PR #1, opened 2022-07-13
Fix pipeline

juspence:
@JimFuller-RedHat I'm going to auto-merge this so the pipeline is less broken, but there are a few items that look like bugs here I wanted to ask about.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -530,8 +530,9 @@ class ProductSerializer(serializers.ModelSerializer):
                 )
         return p
 
-    def get_build_count(self, instance):
-        return instance.builds.count()
+    @staticmethod
+    def get_build_count(instance):
+        return len(instance.builds)
 
     class Meta:
         model = Product
@@ -636,8 +637,9 @@ class ProductVersionSerializer(serializers.ModelSerializer):
                 )
         return p
 
-    def get_build_count(self, instance):
-        return instance.builds.count()
+    @staticmethod
+    def get_build_count(instance):
+        return len(instance.builds)
 
     class Meta:
         model = ProductVersion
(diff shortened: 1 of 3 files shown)


juspence:
GitHub won't let me start a review on lines that didn't change....below is the coverage property defined for any ProductModel subclass in corig/core/models.py:
```python
    @property
    def coverage(self):
        if not self.pnodes.count():
            return 0
        pnode_children = self.pnodes.all()[0].get_children()
        if not pnode_children.count():
            return 0
        has_build = 0
        for pn in pnode_children:
            if not pn.obj.builds:
                # TODO: This looks buggy
                break
            if pn.obj.builds.count():
                has_build += 1
        return round(has_build / pnode_children.count(), 2)
```

We want to report coverage by checking all children of some node, but if any child doesn't have a build, we exit the loop early. Should we use "continue" instead to only skip that one child, and still check others?

Then has_buiild won't get incremented for children that don't have builds, but we'll still count any subsequent children that do have builds, and the coverage report will be more accurate. Or is there a reason to immediately stop processing?

juspence:
Below is the builds property defined for the ProductVariant model in corgi/core/models.py:

```python
    @property
    def builds(self):
        prodcomprelations = (
            ProductComponentRelation.objects.filter(product_ref__contains=self.name)
            .values("build_id")
            .distinct()
        )
        return [pcr["build_id"] for pcr in prodcomprelations]
```

It's overriding the builds property defined for all ProductModel subclasses, but seems to be the same behavior. Ideally it would also return a queryset and have a type hint, to keep the signature identical to the overridden property:
```python
    @property
    def builds(self) -> QuerySet:
        return (
            ProductComponentRelation.objects.filter(product_ref__contains=self.name)
            .values_list("build_id", flat=True)
            .distinct()
        )
```
I didn't make this change since I'm not sure whether it was overridden for a reason and shouldn't go away. I think you just wanted the length of the Queryset / list to report the number of builds. You can do that with my_queryset.count() though so I don't think the custom logic is needed.

JimFuller-RedHat:
will fix/review - fwiw this was a draft MR which I thought I had closed before we did the crossover (lfs/public repo) - in your defense I was in the land of nod and you made the right decision ;)

PR #4, opened 2022-07-14
PSDEVOPS-3745, PSDEVOPS-3755: Misc fixes

juspence:
@RedHatProductSecurity/corgi-devs This has two quick bugfixes for the mentioned tickets, as well as cleanup for some overridden properties that I proposed to Jim yesterday. This uncovered a bug in get_related_errata(), which is really the same as get_product_component_relations() with some extra filtering added, so I deduped both into the same method and fixed the incorrect filtering logic.

I didn't end up adding tests for the Content Security Policy since:
1) Currently, we only really test the API and not the frontent / views
2) We'd need a different test for each view / page
3) The list of allowed resources / external JS libraries we use is relatively static and not likely to change much


mprpic:
LGTM!

mprpic (dismissed):

PR #9, opened 2022-07-18
PSDEVOPS-3708: Enable daily monitoring email now that UMB tasks run automatically

juspence:
@RedHatProductSecurity/corgi-devs Follow-up PR that will send us daily emails about failed tasks, once we turn on UMB tasks that will run periodically.

### DIFF
--- a/corgi/tasks/monitoring.py
+++ b/corgi/tasks/monitoring.py
@@ -1,7 +1,11 @@
 import logging
+from datetime import timedelta
 
+from celery.signals import beat_init
 from django.conf import settings
 from django.core.mail import EmailMessage
+from django.utils import timezone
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat:
apart from wondering if pytz needs to be in our requirements/base ... LGTM

juspence:
> apart from wondering if pytz needs to be in our requirements/base ... LGTM

It's already present as a dependency of Celery and Django. I could add it to the .in file, but it's probably better to switch to django.timezone.utc or similar. Let me test that and then I'll push a fix. Good catch!

EDIT: And django.timezone.utc is just an alias for pytz.utc, so going to merge this once CI passes. Thanks for the review!

PR #10, opened 2022-07-18
PSDEVOPS-3622 - minimal coverage report

JimFuller-RedHat:
Generates minimal coverage report on products (versions, streams) that have at least one build against them, denoting # of builds, # of components and latest build dt . 

**Note** - It is likely that this report will evolve in future epics.


jasinner:
Seems fine to me.

PR #13, opened 2022-07-19
Interal urls in local compose

jasinner:
Load internal url from .env file into celery pods launched by podman-compose.


config/settings/base.py
```diff
@@ -321,6 +321,6 @@
 MANIFEST_HINTS_URL = os.getenv("CORGI_MANIFEST_HINTS_URL")
 
 DISTGIT_DIR = os.getenv("CORGI_DISTGIT_DIR", "/opt/distgit")
-LOOKASIDE_CACHE_BASE_URL = f"https://{os.getenv('CORGI_TEST_PKGS_HOST')}/repo"
+LOOKASIDE_CACHE_BASE_URL = f"https://{os.getenv('CORGI_LOOKASIDE_CACHE_URL')}/repo"
```
JimFuller-RedHat, line 324: makes sense for base settings to not depend on any test machinery

.gitlab-ci.yml
```diff
@@ -16,9 +16,9 @@ before_script:
     - export CORGI_TEST_OSBS_HOST1
     - export CORGI_TEST_OSBS_HOST2
     - export CORGI_TEST_OSBS_HOST3
-    - export CORGI_TEST_PKGS_HOST
     - export CORGI_TEST_REGISTRY_URL
-    # Not used in tests directly, but needed for tests to pass
+  # Not used in tests directly, but needed for tests to pass
+    - export CORGI_LOOKASIDE_CACHE_URL
     - export CORGI_APP_STREAMS_LIFE_CYCLE_URL
```
juspence, line 22: @jasinner This is not quite right - this variable is used directly in the tests so should be part of the group above. I'm fine with the rename though if that's more clear to people.

In general all vars with CORGI_TEST_ are needed in Gitlab CI as well as locally / in OpenShift, and their value is used directly in one of the test modules like test_sca.py's test_download_lookaside_sources below.

config/settings/base.py
juspence: @JimFuller-RedHat It's not a dependency on test machinery, it's a naming convention to indicate this variable is used directly in tests (see test_sca.py's test_download_lookaside_sources) and must be set in Gitlab for CI to pass.

Other vars, which don't have CORGI_TEST_ as a prefix, either aren't used directly in tests (but may still be needed for them to pass) or aren't used in tests at all and aren't present in Gitlab CI variables.

.gitlab-ci.yml
jasinner: I guess the main reason for renaming this env var was that it's used in `config/settings/base.py`. Also the old name didn't give any indication as to what the variable represented. I guess you're right though and it should be put in the group above this comment as it's also used directly in tests.

PR #14, opened 2022-07-19
minor fix to coverage report

JimFuller-RedHat:
should use exists() instead of count() !


jasinner:
fine by me, but didn't wait for CI to pass.

JimFuller-RedHat:
hmmm, @jasinner CI fails now ... though getting a pass when I run full test suite locally ... investigating

PR #15, opened 2022-07-19
PSDEVOPS-3719 - add retry for rhel compose task

JimFuller-RedHat:
added retry logic to app.task for rhel_compose task.


jasinner:
shouldn't hurt to retry.

PR #16, opened 2022-07-19
PSDEVOPS-3772: Use prefork pool (default) so workers respond to probes even when busy

juspence:
@RedHatProductSecurity/corgi-devs In order to add liveness probes, which will detect hung Celery workers and restart them, the worker must be able to respond to Celery management commands like "celery -A config inspect ping".

Using the solo pool means that there is only one Celery process / thread running inside the container, so whenever it's busy executing a task, the worker will not respond to any management commands / probes. This means the worker will appear hung and get restarted by OpenShift constantly (whenever it's executing a task) if we add a liveness probe.

We should use the prefork pool instead, which is almost the same as solo except that a parent process waits to respond to any management commands, as well as a child process which actually runs the tasks. I also fixed the Celery worker hostnames to match SDEngine to make configuring the probe easier for the different worker types. The type of worker (slow vs. fast) is still present in the hostname, just not in the prefix.


mprpic:
Nice, lgtm!

PR #22, opened 2022-07-20
PSDEVOPS-3529: surface up COMPOSE relation 

JimFuller-RedHat:
We ingest rhel-compose data as part of loading product data. This allows us to 'wire in' COMPOSE ProductComponentRelation data to product_streams in REST API.

While I was in the area we also expose ERRATA relations on product_variants.

Specifically this surfaces up in the REST API as follows:

* **api/v1/product_streams** - now have relations enumerating COMPOSE 
* **api/v1/product_variants** - now have relations enumerating ERRATA
* **api/v1/relations** - nascent listing of all relations

**Note** - It maybe eventually we migrate view/serializer approach to directly populating a model attribute which would make then make this filterable/searchable. Alternately we could entertain using tags to encode this information. Decided it was better to wait for more refined user stories defining behaviour.

**Note** - Need to add tests ;)


jasinner:
I tried this out in dev and it looks good. Need to fix black according to CI.

PR #27, opened 2022-07-22
use correct package_type path when downloading container lookaside source

jasinner:
This fixes a bug which is showing up in prod after we added syft support for containers.


JimFuller-RedHat:
LGTM

PR #28, opened 2022-07-22
allow importing products multiple times

jasinner:
Fixes PSDEVOPS-3661 so that you can run `python3 manage.py loadproductdata` multiple times without error. We could potentially add a scheduled task to run this periodically and update product data with changes in ET, product_definitions and rhel_composes.


JimFuller-RedHat:
LGTM

PR #31, opened 2022-07-22
PSDEVOPS-3805: Schedule update of product data periodically

juspence:
@jasinner I went ahead and did this since it was a very quick change, but feel free to tweak if the update should run more than once a day, or if the order should be different. I copied the order the "loadproductdata" command uses, but not sure if that's right.


JimFuller-RedHat:
LGTM

PR #32, opened 2022-07-23
PSDEVOPS-3468: Add minimal rhel modules support

JimFuller-RedHat:
This PR enables ingestion of rhel modules ... for now we only create the root component node and associated RHEL_MODULE component.

Future stories will need to provide more detail and how we might populate component tree underneath RHEL_MODULE (for now its provided in meta data).


jasinner:
LGTM

PR #33, opened 2022-07-25
PSDEVOPS-3529 greed matching

jasinner:
Fixes a bug where components are eagerly matched to all products if they don't have any entry in ProductComposeRelations.

Also add `loadcomposedata` management command for loading builds in composes.


JimFuller-RedHat:
LGTM!

PR #34, opened 2022-07-25
Fix "not registered" error / no decorator when running load_composes

juspence:
@RedHatProductSecurity/corgi-devs Quick fix for an error seen in recent monitoring emails.


JimFuller-RedHat:
nice catch, LGTM!

PR #35, opened 2022-07-26
PSDEVOPS-3814: Try to fix deadlock when creating ProductComponentRelations

juspence:
@RedHatProductSecurity/corgi-devs Review appreciated. Without a traceback it's hard to know what the dependent code / transaction was waiting for. CONTEXT below indicates that inserting fails:

OperationalError('deadlock detected
DETAIL: Process 3580 waits for ShareLock on transaction 834584; blocked by process 3616.
Process 3616 waits for ShareLock on transaction 833789; blocked by process 3580.

HINT: See server log for query details.

CONTEXT: while inserting index tuple (5740,87) in relation "unique_productcomponentrelation"
')

There are only two places in the code where ProductComponentRelations get created, as far as I can tell. I've changed these to use defaults= inside the get_or_create, so that insert is fully atomic.

I then cleaned up logic that tries to access or read from ProductComponentRelations, to avoid evaluating a queryset multiple times or making many calls to the DB when only one is needed. In other words, all reads of PCR data should be atomic as well, so there should be no more race conditions or deadlocks.

I don't see any place where PCRs get updated or deleted. Did I miss these, or are we only inserting this data and never changing it after it's been created?

### DIFF
--- a/corgi/tasks/errata_tool.py
+++ b/corgi/tasks/errata_tool.py
@@ -87,14 +87,18 @@ def load_errata(errata_names: list[str]) -> list[int]:
         else:
             erratum_id = int(erratum_name)
         # Get all the PCR with errata_id
-        relation_build_ids = ProductComponentRelation.objects.filter(
-            external_system_id=erratum_id
-        ).values_list("build_id", flat=True)
+        relation_build_ids = list(
+            ProductComponentRelation.objects.filter(external_system_id=erratum_id).values_list(
+                "build_id", flat=True
+            )
+        )
         # Check is we have software builds for all of them
         if (
+            # Skip loading erratum if we have all its builds in DB already
+            # But handle case / don't skip it when num_build_ids == num_builds == 0
             0
             < len(relation_build_ids)
-            == SoftwareBuild.objects.filter(build_id__in=list(relation_build_ids)).count()
+            == SoftwareBuild.objects.filter(build_id__in=relation_build_ids).count()
         ):
             logger.info("Already processed %s", erratum_id)
             continue
--- a/corgi/tasks/errata_tool.py
+++ b/corgi/tasks/errata_tool.py
@@ -104,16 +104,17 @@ def load_errata(errata_names: list[str]) -> list[int]:
             for build_obj in build_objects:
                 for build_id, errata_components in build_obj.items():
                     build_ids.add(int(build_id))
-                    pcr, created = ProductComponentRelation.objects.get_or_create(
+                    _, created = ProductComponentRelation.objects.get_or_create(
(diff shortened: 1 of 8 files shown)


corgi/tasks/errata_tool.py
```diff
@@ -87,14 +87,17 @@ def load_errata(errata_names: list[str]) -> list[int]:
...
-        relation_build_ids = ProductComponentRelation.objects.filter(
+        relation_build_ids = list(ProductComponentRelation.objects.filter(
             external_system_id=erratum_id
-        ).values_list("build_id", flat=True)
+        ).values_list("build_id", flat=True))
         # Check is we have software builds for all of them
+        num_build_ids = len(relation_build_ids)
+        num_builds = SoftwareBuild.objects.filter(build_id__in=relation_build_ids).count()
         if (
+            # TODO: Why is 0 present in below?
```
juspence: @RedHatProductSecurity/corgi-devs I don't fully understand this logic. Should we just check that "num_build_ids == num_builds", AKA all the build_ids referenced by this erratum are already present (as builds) in our database?

corgi/tasks/rhel_compose.py
```diff
@@ -63,15 +63,12 @@ def save_compose(stream_name, compose_coords) -> None:
...
                             # found the build_id, stop iterating filenames
                             break
-                    relation, created = ProductComponentRelation.objects.get_or_create(
+                    ProductComponentRelation.objects.get_or_create(
                         external_system_id=compose_id,
                         product_ref=stream_name,
                         build_id=build_id,
+                        defaults={"type": ProductComponentRelation.Type.COMPOSE},
                     )
-                    # Don't include type in get_or_create to avoid slow read of field outside index
```
mprpic, line 71: So is the slowness no longer a concern?

mprpic (approved):

corgi/tasks/rhel_compose.py
juspence: The slowness is due to using keyword arguments in get_or_create which aren't indexed. Every kwarg gets used to look up a matching record, so this forces a scan of the whole table and makes things slow.

Putting these values in defaults= means they're only written when a record is created, and not used to look up a matching record. All the remaining kwargs are indexed, so the lookup stays fast by using an index scan.

corgi/tasks/errata_tool.py
juspence: Looking at git history in archived repo, this is to handle the case when "num_build_ids == num_builds == 0", and we need to load the erratum instead of skipping it.

JimFuller-RedHat:
can confirm we are only inserting data and never changing it

PR #38, opened 2022-07-28
nested components materialize taxonomy

jasinner:
This repurposes the sources field of Component to return all root nodes for the component. The previous usage was too similar to the upstream field.

I also updates the upstream field to be 'upstreams' with an 's'. That makes it more consistent with the other arrayfield fields in Component.

I changes the sca task to use software_build.save_component_taxonomy from component.save_component_taxonomy so that it correctly recursed into components identified and linked during the scan. 

Latestly and I guess most importantly I change the software_build.save_component_taxonomy to recurse into descendants. This will probably have a performance impact but it's required to link back from nested components to their ancestors.


juspence:
@jasinner Change looks fine, but while looking it over I realized there's a lot of code I don't understand here. Would you or @JimFuller-RedHat have some time next week to help me understand:

1) What's the difference between roots, upstreams, and sources?
2) If we are using a tree-based design, how can we ever have more than one root / upstream?
3) What's the reason for using MPTT-based trees / Nodes and not e.g. foreign-key links to parent entities (ProductVersion -> Product, or binary / arch Component -> parent source Component) or many-to-many fields?
3) If ProductComponentRelation is a through table, why is build_id a CharField and not a ForeignKey link to the specific build?

JimFuller-RedHat:
1. root = source -  upstream is a child of root indicating upstream project which we also list under source (that should also stop soon) - Jason and I talked recently about streamlining the conceptual model (aka REST API) 

2. A single component can participate in multiple component trees and thusly have multiple roots - remember a component is a component entity ... which is different then ComponentNode (which can only have one root). Any number of component nodes can point to a single component.

3. Covered by adr some time back - this is the way we are doing it. We use trees because we traverse tree relationships - you can model 'trees' with foreign keys but every time your tree changes you prob have to change your code. Also (as evidenced by deptopia and previous efforts) modelling these relationships classically almost always presents performance pitfalls. Lastly we have no foreign key relationships between product and component trees as some day it is likely products will move out into its own offering registry. Another reason to keep pcr unadorned. Note that there will be many more tree style queries in the future (not just today). Please refer to adr and/or happy to answer broad questions offline (and not as part of any specific issue).

#4 We could do that ... though I would rather wait till we get a 2nd build system integrated before we 'pour cement'. We may find other systems have other kinds of ids ... when in doubt (esp at the beginning of dev) I try to resist foreign relationships until one cannot ;)

PR #39, opened 2022-07-29
PSDEVOPS-3841 dont fetch rpm capabilities

jasinner:
Fixes a performance bottleneck in corgi.collectors.brew.Brew::get_rpm_build_data. We where making 1 call to Koji for each RPM in a SRPM build, which was hammering Brew with new connections and also taking up to 25min.


JimFuller-RedHat:
LGTM

PR #42, opened 2022-07-29
PSDEVOPS-3839: Speed up performance

juspence:
Outdated, incorrect: @RedHatProductSecurity/corgi-devs This change should prevent duplicate rows from being inserted into our DB, which will prevent "ModelName.MultipleObjectsReturnedError" when using "ModelName.get". Note that we may have to clean up any existing duplicate rows before the error goes away, but this fix will keep things from getting worse / more duplicates from being added.

Still accurate: It also improves performance slightly, since using get_or_create with only indexed keyword arguments, and moving unindexed keyword arguments to the defaults= dict, speeds up the database operations. The exact reason is that get_or_create first tries to SELECT a row with matching field values / keyword arguments. The defaults= dict is ignored for this step. If no matching row is found, get_or_create then tries to INSERT a row using the provided keyword arguments and the values in the defaults dict. AKA the defaults= dict is not used for the get_ logic, only the _or_create logic.

Since we only provide indexed fields as keyword arguments, the SELECT query for the get_ logic can now use an index scan, instead of being forced to do a whole table scan to match against the unindexed fields. A whole table scan can be very slow for text fields / large tables with lots of data, although Jason's recent MR was a much better improvement :)

### DIFF
--- a/corgi/tasks/errata_tool.py
+++ b/corgi/tasks/errata_tool.py
@@ -144,8 +144,7 @@ def update_variant_repos() -> None:
         pv_channels = []
         for repo in et_variant_data["repos"]:
             repo, _ = Channel.objects.get_or_create(
-                type=Channel.Type.CDN_REPO,
-                name=repo,
+                name=repo, defaults={"type": Channel.Type.CDN_REPO}
             )
             pv_channels.append(repo.name)
             # TODO: investigate whether we need to delete CDN repos that were removed from a
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -222,7 +222,9 @@ CELERY_RESULT_BACKEND = "django-db"
 CELERY_RESULT_BACKEND_ALWAYS_RETRY = True
 CELERY_RESULT_BACKEND_MAX_RETRIES = 2
 CELERY_DEFAULT_RATE_LIMIT = "8/m"
-CELERYD_SOFT_TIME_LIMIT = 300
+# Set a global 15-minute task timeout. Override this on individual tasks by decorating them with:
+# @app.task(soft_time_limit=<TIME_IN_SECONDS>)
+CELERY_TASK_SOFT_TIME_LIMIT = 900
 
 CELERY_WORKER_CONCURRENCY = 1  # defaults to CPU core count, which breaks in OpenShift
 
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -706,9 +706,8 @@ class Brew:
         return component
 
     def get_builds_with_tag(self, brew_tag: str, inherit: bool = False) -> list[int]:
-        brew = self.get_koji_session()
         try:
-            builds = brew.listTagged(brew_tag, inherit=inherit)
+            builds = self.koji_session.listTagged(brew_tag, inherit=inherit)
             return [b["build_id"] for b in builds]
         except koji.GenericError as exc:
             logger.warning("Couldn't find brew builds with tag %s: %s", brew_tag, exc)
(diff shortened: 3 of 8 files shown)


corgi/tasks/errata_tool.py
```diff
@@ -35,6 +35,8 @@ def link_stream_using_brew_tag(brew_tag: str, stream_name: str, inherit: bool =
         product_variant, _ = ProductVariant.objects.get_or_create(name=variant)
         product_stream_node = ProductNode.objects.get(object_id=product_stream.pk)
         product_variant.pnodes.get_or_create(parent=product_stream_node)
+        # TODO: We already have a reference to product_stream
```
jasinner: This is a good idea. We can probably move the product_stream.save* functions to after the variants loop, which should improve the performance of this function quite a bit. Note that this function is only called if passing the 'stream' option to loadproductdata. I left the code in here, because I think it's a robust way of linking brew tags to product variants, but it's very resource intensive as it involves looking up every build in a specified brew tag from Errata Tool.

jasinner (approved):
After using these branch local I am getting errors like this, see comment below.

on slow_fetch_brew_build(981002)
```
  File "/opt/app-root/src/corgi/tasks/brew.py", line 75, in slow_fetch_brew_build
    softwarebuild.save_component_taxonomy()
  File "/opt/app-root/src/corgi/core/models.py", line 194, in save_component_taxonomy
    d.obj.save_component_taxonomy()
  File "/opt/app-root/src/corgi/core/models.py", line 740, in save_component_taxonomy
    self.upstreams = self.get_upstreams()
  File "/opt/app-root/src/corgi/core/models.py", line 910, in get_upstreams
    roots = self.get_roots
  File "/opt/app-root/src/corgi/core/models.py", line 767, in get_roots
    root = cnode.get_root()
  File "/usr/local/lib/python3.9/site-packages/mptt/models.py", line 424, in _fn
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/mptt/models.py", line 732, in get_root
    return self._tree_manager._mptt_filter(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(
corgi.core.models.ComponentNode.MultipleObjectsReturned: get() returned more than one ComponentNode -- it returned 2!
```

jasinner (changes requested):
I think we should remove the commit which changes [corgi/tasks/brew.py::get_or_create functions](https://github.com/RedHatProductSecurity/component-registry/pull/42/commits/b191bea2317d674aeb7c8f46c75c320689f6d70b). I'm getting `ComponentNode.MultipleObjectsReturned` errors trying to call `save_component_taxonomy` or SRPMs persisted with these changes. I'm not sure exactly why, but I think slower performance is better than getting errors ingesting which prevents a materalization of component relationships.

JimFuller-RedHat:
generally echo @jasinner comments and LGTM, though lets hold off on this for today ... would like to get this in (along with some other improvements) this week.

juspence:
> I think we should remove the commit which changes [corgi/tasks/brew.py::get_or_create functions](https://github.com/RedHatProductSecurity/component-registry/pull/42/commits/b191bea2317d674aeb7c8f46c75c320689f6d70b). I'm getting `ComponentNode.MultipleObjectsReturned` errors trying to call `save_component_taxonomy` or SRPMs persisted with these changes. I'm not sure exactly why, but I think slower performance is better than getting errors ingesting which prevents a materalization of component relationships.

So there are a few things here - I had a small oops in the description, where I talked about fixing this issue. I implied that this PR would fix an error we're having on main (so not related to my changes), Component.MultipleObjectsReturned.

But actually the error on main is ComponentNode.MultipleObjectsReturned, which isn't related to and isn't affected by this change. The get_or_create function changes you mention are only for Component and not for ComponentNode, so they aren't the cause of the error you're seeing.

JimFuller-RedHat:
LGTM

jasinner (approved):
Sorry, I verified I'm also getting the ComponentNode.MultipleObjectsReturned on main without your commits here. I'm happy to merge this as is.

PR #43, opened 2022-08-01
Loadproductdata allstreams

jasinner:
This adds a management command which uses product-definitions brew_tags to find and link builds to product streams.

It also fixes a bug in SoftwareBuild.save_product_taxonomy which was calling get_product_variants twice.


JimFuller-RedHat:
LGTM , I will fix CI in my PR

PR #44, opened 2022-08-01
draft - PSDEVOPS-3830 : deptopia like search endpoint

JimFuller-RedHat:
https://issues.redhat.com/browse/PSDEVOPS-3830


corgi/api/views_search.py
```diff
@@ -0,0 +1,129 @@
...
+        # match_root = True
+        # match_dep = True
+        # if request.query_params.get("match_root"):
+        #     match_root = True
+        #     match_dep = False
+        # if request.query_params.get("match_dep"):
+        #     match_root = False
+        #     match_dep = True
+
+        query = Q()
```
jasinner, line 52: If you hit this without any query parameters it tries to load everything.

corgi/api/views_search.py
```diff
@@ -0,0 +1,129 @@
...
+        if name:
+            query |= Q(name=name)
```
jasinner, line 54: I think instead of searching the name field directly here we should emulate the deptopia_cli behaviour which is to add a wildcard to the start of the search term, and anchor it to the end, eg:

```
Q(name__iendswith=name)
```

Compare the results with a search term such as `yaml`

jasinner:
Looks like a good start. I added a comment inline about the default 'name' behaviour and bug if no query params are added.

I think we also need to do some de-duplicating of results. But overall, looks really good.

corgi/api/views_search.py
JimFuller-RedHat: yes, need to add pagination as well

JimFuller-RedHat:
will reraise in next sprint

PR #45, opened 2022-08-02
PSDEVOPS-3839 improve workflow for errata ingestion

jasinner:
This simplifies load Brew builds and associating them with Errata and prevents a deadlock where load_errata called slow_fetch_brew_build synchronously, and prevented the other from progressing due to load_errata being in a DB transaction. 

I removed the 'loadproductdata' mgnt options -stream, -tag, and -allstreams and related code, as it no longer works since simplifying load_errata to only handle a single erratum_name. I think we should create a new type in ProductComponentRelation called BREW_TAG for this type of link instead of using brew_tags to find an link product variants to streams.

We still need to come up with a solution for save_product_taxonomy when ingesting composes. I created PSDEVOPS-3849 as a follow up task for that.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -56,7 +56,8 @@ def slow_fetch_brew_build(build_id: int):
 
     # Once we have the full component tree loaded
     softwarebuild.save_component_taxonomy()
-    # We don't call save_product_taxonomy here to reduce CPU cycles and allow async call of load_errata task
+    # We don't call save_product_taxonomy here to reduce CPU cycles and
+    # allow async call of load_errata task
 
     # for builds with errata tags set ProductComponentRelation
     # get_component_data always calls _extract_advisory_ids to set tags, but list may be empty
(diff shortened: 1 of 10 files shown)


JimFuller-RedHat:
Generally LGTM - lets discuss tmrw - note I believe originally we had save_product_taxonomy() at the 'end' so this is not so radical eg. we are fine with waiting until all components are known before setting their taxonomies.

FWIW all this product taxonomy saving has several 'reasons' - worth repeating - at the beginning of the project I mentioned the relationship between product and component entities could just as easily be a foreign key relationship. That might make this back and forth a bit easier though it also would:

* make it much harder to extrude out products from component registry (into their own 'offering' service) ... which is the current plan of record.
* make it harder to represent aggregate behaviour eg. a product through its entire lifetime may have a component at one point in time, only to be dropped in a future moment in time ... any foreign relationships are tied to the current 'now'. As we 'merge add' we should never lose this historical data. Of course what matters is what 'bits' are actually in the product ... also I suspect there are behaviours we do not understand yet (ex. existence of a few versions of the same component in a product, used or not used)

PR #46, opened 2022-08-02
Minor: remove unused product taxonomy in component object

JimFuller-RedHat:
removes Component save_product_taxonomy which is an older code path

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -203,8 +203,26 @@ def test_product_component_relations():
     srpm_cnode, _ = srpm.cnodes.get_or_create(
         type=ComponentNode.ComponentNodeType.SOURCE, parent=None
     )
-    srpm.save_product_taxonomy()
-    assert ps1.ofuri in srpm.product_streams
+    sb.save_product_taxonomy()
+    c = Component.objects.get(uuid=srpm.uuid)
+    assert ps1.ofuri in c.product_streams
+
+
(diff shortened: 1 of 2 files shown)


tests/test_model.py
```diff
@@ -203,8 +203,26 @@ def test_product_component_relations():
...
+    )
+    p1, ps1, ps2, ps3, pv1a, pv1b = create_product_hierarchy()
+    relate_product_hierarchy(p1, ps1, ps2, ps3, pv1a, pv1b)
+    srpm = ComponentFactory(software_build=sb, type=Component.Type.SRPM)
+    srpm_cnode, _ = srpm.cnodes.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE, parent=None
+    )
+    sb.save_product_taxonomy()
+    c = Component.objects.get(uuid=srpm.uuid)
+    assert ps3.ofuri in c.product_streams
```
mprpic, line 225: Could this test be perhaps rolled into the previous one with a [parametrization](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) for different `ProductComponentRelation.Type`s?

JimFuller-RedHat: It could, I typically wait for a 'third' example before collapsing down/parameterising ... will do so when we get another ProductComponentRelation type (which maybe soon ex. yum repos)

PR #47, opened 2022-08-03
PSDEVOPS-3842 add some safety around Component get_roots

JimFuller-RedHat:
This is the first part of solving PSDEVOPS-3842 - adds safety around calling Component get_roots property. While I was in the area also ensure we return a distinct list of roots.

Note - 2nd PR will address inserts


jasinner (approved):
LGTM

PR #48, opened 2022-08-03
Use hasattr(), not getattr(), when checking exception object in boolean context

juspence:
getattr will return the requests.response object, and it will be evaluated as a boolean. If the response had a 200ish code, it evaluates to True. If the code was 400ish or similar, it evaluates to False.

So any failed request will not enter the if block I changed in the diff, and we skip checking whether the code is retryable or not. Instead we just retry all requests, which is dangerous.


JimFuller-RedHat:
LGTM

PR #51, opened 2022-08-04
PSDEVOPS-3853 use links for results from products

JimFuller-RedHat:
This commit replaces arrays of data with links as per PSDEVOPS-3853:

* manifest -> /api/v1/{product_streams|product_variants}/{uuid}/manifest
* components -> /api/v1/components?ofuri={}
* upstreams -> /api/v1/components?ofuri={}&type=UPSTREAM

which avoids having to return gigantic http response payloads.

While I was in the area I reordered attributes as well as exclude manifest links on Products and Product Versions (as they make no sense). Generally, deactivated anything that slows down processing of the response.

**Note** - there will be more work in REST API area to include/exclude data ... this is the minimal amount of work to ensure endpoints are usable.


jasinner (approved):
LGTM

PR #55, opened 2022-08-04
CORGI-46: Fix miscellaneous items

juspence:
@RedHatProductSecurity/corgi-devs While waiting on another PR, I took a quick look over the code and found:
1) Possible bugs in sca.py that could hide errors or return multiple Components in a .get() call
2) A lot of duplicated code in serializers.py
3) A bug with Django management commands when updating taxonomies for only the specified build IDs
4) A bug in views.py that could lead to 500 errors being raised, whenever model instances don't exist in the DB

### DIFF
--- a/corgi/tasks/sca.py
+++ b/corgi/tasks/sca.py
@@ -11,7 +11,6 @@ from typing import IO, Any, Optional, Tuple
 
 import requests
 from django.conf import settings
-from django.db.models import Q
 from requests import Response
 
 from config.celery import app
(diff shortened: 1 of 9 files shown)


corgi/tasks/sca.py
```diff
@@ -57,32 +56,29 @@ def save_component(component: dict[str, Any], parent: ComponentNode):
...
 
     root_node = root_component.cnodes.first()
     if not root_node:
-        logger.error("Didn't find root component node for %s", root_component.purl)
-        return
+        raise ValueError(f"Didn't find root component node for {root_component.purl}")
     package_type = "rpms"
     if root_component.type == Component.Type.CONTAINER_IMAGE:
         package_type = "containers"
+        # TODO: Below looks like a bug - shouldn't this be unindented one level?
```
jasinner: This should only be run for container images. SRPMs don't have remote-sources. However we also want to scan the distgit sources for containers (like we do for SRPMs), so we do that for both below.

jasinner (approved):
Looks good to me. Minor nitpick: we should remove that TODO comment you added though.

juspence:
> Looks good to me. Minor nitpick: we should remove that TODO comment you added though.

Thanks for the review! I always clean these up and do a final review before merging, they're just there to call out items that need discussion. Speaking of, what do you think about the recursive code in views.py above?

There's one case the existing code doesn't handle, so the return value can be None in some cases, which causes issues elsewhere. Do you know what it should be returning instead? Usually it's a dict, but I don't know what keys should be in it for that case. Or should it raise an error if we reach that condition?

jasinner:
> > Looks good to me. Minor nitpick: we should remove that TODO comment you added though.
> 
> Thanks for the review! I always clean these up and do a final review before merging, they're just there to call out items that need discussion. Speaking of, what do you think about the recursive code in views.py above?
> 
> There's one case the existing code doesn't handle, so the return value can be None in some cases, which causes issues elsewhere. Do you know what it should be returning instead? Usually it's a dict, but I don't know what keys should be in it for that case. Or should it raise an error if we reach that condition?

Probably best to ask @JimFuller-RedHat about views.py. I've not been involved much in that.

jasinner:
After the changes in https://github.com/RedHatProductSecurity/component-registry/pull/61 I'm not sure that we really need to increase to timeout to 1 hour. I think perhaps the original value of 900 might be more appropriate.

juspence:
> After the changes in #61 I'm not sure that we really need to increase to timeout to 1 hour. I think perhaps the original value of 900 might be more appropriate.

I think you were referring to changes in a different PR (https://github.com/RedHatProductSecurity/component-registry/pull/57), but ack - will close that one. Thanks for fixing the slowness!

Last remaining change here is just to return the same dict with a default set of keys in all cases where some required data is missing. Trying to get to this after meetings today.

juspence:
@jasinner Sorry for the delay in responding here - I had to finish the TODOs and clean up some commits after rebasing, since we now provide links instead of object lists in serializers.py.

This should be ready to merge as-is and I don't think there are major changes, but please feel free to look again if you'd like. Or if you're busy that's fine too - if I don't hear anything I'll just merge this tomorrow.

PR #56, opened 2022-08-05
PSDEVOPS-3851 brew tag relations

jasinner:
This adds BREW_TAG type to relations. It also adds several fields to ProductStream model for all the existing methods of ingesting builds and relating them to products in Deptopia/product-definitions.

We need to adjust the existing compose collector to use the new composes field of ProductStream to relate those.


corgi/core/models.py
```diff
@@ -443,6 +443,13 @@ class ProductStream(ProductModel, TimeStampedModel):
 
     cpe = models.CharField(max_length=1000, default="")
 
+    # NOTE brew_tags and yum_repositories values shouldn't be exposed outside of Red Hat
+    brew_tags = models.JSONField(default=dict)
```
JimFuller-RedHat, line 447: It is fine and good to add specific attributes to ProductStream ... mentally I read the ProductModel base
as RedHatProductModel.

corgi/core/models.py
```diff
@@ -561,6 +568,8 @@ class ProductComponentRelation(TimeStampedModel):
     class Type(models.TextChoices):
         ERRATA = "ERRATA"
         COMPOSE = "COMPOSE"
+        BREW_TAG = "BREW_TAG"
```
JimFuller-RedHat, line 571: nice ...

corgi/tasks/brew.py
```diff
@@ -300,3 +307,24 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
...
+
+
+def load_brew_tags() -> None:
+    for ps in ProductStream.objects.all():
+        brew = Brew()
+        for brew_tag, inherit in ps.brew_tags.items():
+            try:
+                builds = brew.get_builds_with_tag(brew_tag, inherit)
+            except GenericError as exc:
+                logger.error("Got error getting builds for %s: %s", brew_tag, exc)
```
JimFuller-RedHat, line 319: Is it intentional to keep running a task and just log the error ? Alternately celery would trap this right ?

JimFuller-RedHat:
apart from minor comments, this is a nice direction and very handy LGTM

corgi/tasks/brew.py
jasinner: Ok, removed the try/catch from here, and letting Celery retry this in case of error.

PR #57, opened 2022-08-05
Fix SCA SoftTimeLimitExceeded, and move it to slow queue to keep fast queue fast

juspence:
@RedHatProductSecurity/corgi-devs We've been seeing the SoftTimeLimitExceeded error a lot recently for the SCA task, so rather than letting it constantly fail, I've increased our Celery task timeout to 1 hour (to match other apps).

I've also moved the SCA task to the slow queue, so that other fast tasks can complete quickly and avoid being blocked. But ingest of new builds from Brew will still be slowed down by this change (potentially up to 4 times slower). Is this the right approach here, or do we need to reopen the "Corgi performance regression" ticket to figure out why save_component_taxonomy takes forever?

Note the SCA task slowness is also seen locally when running tests, and there was a recent regression which bumped the SCA test runtime from 2 minutes to 7 minutes. The slowest tests are:

405s calling test_sca.py::test_software_composition_analysis[1898406
23s calling test_sca.py::test_scan_remote_sources

47s calling test_brew_collector.py::test_fetch_container_build_components

40s calling test_products.py::test_products

15s calling test_errata_data.py::test_update_variant_repos

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -224,7 +224,7 @@ CELERY_RESULT_BACKEND_MAX_RETRIES = 2
 CELERY_DEFAULT_RATE_LIMIT = "8/m"
 # Set a global 15-minute task timeout. Override this on individual tasks by decorating them with:
 # @app.task(soft_time_limit=<TIME_IN_SECONDS>)
-CELERY_TASK_SOFT_TIME_LIMIT = 900
+CELERY_TASK_SOFT_TIME_LIMIT = 3600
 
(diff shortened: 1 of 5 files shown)


JimFuller-RedHat:
LGTM

JimFuller-RedHat:
make sense to me, feel free to merge

juspence:
Jason's changes in #61 should be a better fix for this issue.

PR #58, opened 2022-08-05
CORGI-88: Fix MultipleObjectsReturned by using tree_id as extra lookup kwarg

juspence:
@RedHatProductSecurity/corgi-devs This is definitely not ready to merge, but I wanted to use the code here as a starting point for discussion. After some investigation, it looks like ProductVariants can be linked to multiple ProductStreams in different trees. See CORGI-25 for notes and details.

Given that's the case, I added a tree_id= kwarg to all ProductNode.objects.get_or_create() lookups, which should prevent the MultipleObjectsReturned error we've been seeing. AKA we use the tree_id of the parent node to choose which tree to return objects from. But there are other possible bugs, since ProductModel's saveProductTaxonomy uses helper methods from the ProductNode class.

Those helper methods assume each ProductModel instance has exactly one linked ProductNode, and use product_model_instance.linked_product_nodes.first() as a result. But this isn't true for (at least) a ProductVariant, which could have multiple ProductNodes, one for each ProductStream it's linked to / tree it's a part of. So whenever we call saveProductTaxonomy() for some ProductVariant, we're probably saving the taxonomy for only one tree / ProductStream, and silently ignoring the other.

I'm not sure what the right fix is here. Should we just loop over all linked nodes in every tree, and return the names of each linked object to be saved? I'm afraid that will mess up our tree, since it won't look like a tree anymore (a node will have multiple parents). Thoughts appreciated.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -32,6 +32,11 @@ class ProductNode(MPTTModel, TimeStampedModel):
 
     class MPTTMeta:
         level_attr = "level"
+        # Make get_or_create with object_id, tree_id kwargs atomic so it doesn't insert duplicates
+        constraints = [
+            models.UniqueConstraint(name="unique_productnode", fields=("object_id", "tree_id")),
+        ]
+        indexes = [models.Index(fields=("object_id", "tree_id"))]
 
     # Tree-traversal methods for product-related models below.
     #
@@ -46,6 +51,7 @@ class ProductNode(MPTTModel, TimeStampedModel):
     #
     # - Each method assumes that an object model will always link to a single ProductNode (thus
     #   the use of `pnodes.first()`).
+    #   TODO: Need to revisit, we have ProductVariants which are linked to two separate streams
     #
     # - The `values_list()` query relies on the GenericRelation of each model's pnodes
     #   attribute's related query name.
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -80,6 +80,7 @@ def update_products() -> None:
                 )
                 product_version_node, _ = ProductNode.objects.get_or_create(
                     object_id=product_version.pk,
+                    tree_id=product_node.tree_id,
                     defaults={
                         "parent": product_node,
                         "obj": product_version,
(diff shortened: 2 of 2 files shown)


JimFuller-RedHat:
The right approach is to first file an issue ... and we can discuss ... please close this PR.

juspence:
> The right approach is to first file an issue ... and we can discuss ... please close this PR.

@JimFuller-RedHat I'm a little confused - are you saying this is a different issue than CORGI-25?

juspence:
Per Jim's request, split CORGI-25 into two tickets - one for ComponentNode (which this PR doesn't fix) and one for ProductNode (CORGI-88 that this should fix).

corgi/core/models.py
```diff
@@ -46,6 +51,7 @@ class MPTTMeta:
     #
     # - Each method assumes that an object model will always link to a single ProductNode (thus
     #   the use of `pnodes.first()`).
+    #   TODO: Need to revisit, we have ProductVariants which are linked to two separate streams
```
JimFuller-RedHat, line 54: in this case each child ProductNode from the two streams would just point to the same ProductVariant entity

corgi/core/models.py
```diff
@@ -46,6 +51,7 @@ class MPTTMeta:
     #
     # - Each method assumes that an object model will always link to a single ProductNode (thus
```
JimFuller-RedHat, line 52: those methods probably need to be fixed rather then adulterate the model.

corgi/tasks/prod_defs.py
```diff
@@ -159,8 +161,10 @@ def update_products() -> None:
                                             }
                                         },
                                     )
+                                    # TODO: Why are there two places that create a variant node?
```
JimFuller-RedHat, line 164: feel free to rationalise ... there maybe no good reason why (other then we never got to 'three' to boil it down!).

corgi/core/models.py
juspence: Yes, that's what I was proposing here. I thought we could add a for loop which replaces the .first(), in order to process all the related ProductNodes instead of only the first one.

If we do that, though, two different trees will point to the same ProductVariant, which might cause issues elsewhere if we don't expect a ProductVariant to be in two different trees.

corgi/core/models.py
juspence: I wasn't proposing any model changes, just noting that our tree will no longer look like a tree and trying to find e.g. all linked ProductStreams may return odd results.

juspence:
After discussing with Martin, going to close this for now. We don't think the ProductVariant should be part of two ProductStreams for completely different products, AKA this seems like a bug in how things get linked. Maybe dropping and reloading DB will fix this error.

PR #61, opened 2022-08-08
improve efficiency of software_build.save_component_taxonomy

jasinner:
This reduces the time taken by test_software_composition_analysis to 90 seconds, from 400.

I also changes the behaviour of the app by always returning an emtpy list for upstreams when a component does not have any software_builds. It doesn't make sense for a Component without a software_build to have upstreams. That is because Components without software_builds are upstream.


juspence:
LGTM, thanks for working on this!

juspence (approved):

PR #64, opened 2022-08-09
Simplify Dockerfile

mprpic:
Remove upgrade of all packages since fixing Lows and Moderates is really
not necessary and just adds time to the build process. We can trust Red
Hat's SLAs around addressing meaningful vulnerabilities in the UBI images.

Remove contact address for the image owner, the team name should be
sufficient to identify the image internally. The distribution scope of this
image is private anyway.

Install syft directly from the URL without an extra curl step.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,6 @@ services:
       args:
         - PIP_INDEX_URL=${PIP_INDEX_URL}
         - PIP_REQUIREMENT=./requirements/dev.txt
-        - PRODSEC_EMAIL=${PRODSEC_EMAIL}
         - ROOT_CA_URL=${ROOT_CA_URL}
     image: corgi
     depends_on: ["corgi-db"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
 FROM registry.redhat.io/ubi8/ubi:8.6
 
 ARG PIP_INDEX_URL="https://pypi.org/simple"
-ARG PRODSEC_EMAIL="secalert@redhat.com"
 ENV PYTHONUNBUFFERED=1 \
     PIP_NO_CACHE_DIR=off \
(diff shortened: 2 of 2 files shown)


juspence:
> Remove upgrade of all packages since fixing Lows and Moderates is really not necessary and just adds time to the build process. We can trust Red Hat's SLAs around addressing meaningful vulnerabilities in the UBI images.

Mostly LGTM, but does it really add that much time to the build to check for security updates? "Only rebuilt every six weeks" seems rather infrequent, and even if they're only Moderates and Lows, I'd feel more comfortable running an image that doesn't have any bugs. And I don't imagine we'll be rebuilding the image that often anyway.

docker-compose.yml
```diff
@@ -37,7 +37,6 @@ services:
       args:
         - PIP_INDEX_URL=${PIP_INDEX_URL}
         - PIP_REQUIREMENT=./requirements/dev.txt
-        - PRODSEC_EMAIL=${PRODSEC_EMAIL}
```
juspence, line 40: It's fine to remove the label from this image, but we'll need to set this env var in OpenShift since one of our views has a "contact info" address that uses this value.

juspence (approved):

docker-compose.yml
mprpic: Right, let's set it in the new deployment configs then.

mprpic:
> > Remove upgrade of all packages since fixing Lows and Moderates is really not necessary and just adds time to the build process. We can trust Red Hat's SLAs around addressing meaningful vulnerabilities in the UBI images.
> 
> Mostly LGTM, but does it really add that much time to the build to check for security updates? "Only rebuilt every six weeks" seems rather infrequent, and even if they're only Moderates and Lows, I'd feel more comfortable running an image that doesn't have any bugs. And I don't imagine we'll be rebuilding the image that often anyway.

Discussed this offline. It's OK to rely on image refreshes only.

PR #65, opened 2022-08-10
Fix detect-secrets slowness

juspence:
@mprpic Looks like your recent MR to remove pip-audit also removed the daily scheduled detect-secrets CI job, as well as the "only: refs: schedules" logic for detect-secrets in .gitlab-ci.yml. That "only:" block made it run only once a day, instead of on every PR.

This change should make it faster (only scan changed files instead of the whole tree), or I could just add back the "only:" block and the daily detect-secrets CI job to keep PRs fast.


juspence:
> This change should make it faster (only scan changed files instead of the whole tree)

This works locally, but not in Gitlab CI due to the use of shallow clone (both the "main" and "origin/main" refs don't exist).

So I've added back the "only:" block to make detect-secrets run only as part of scheduled CI jobs.

PR #71, opened 2022-08-15
CORGI-107 remove distgit mount related code before move to MP+

jasinner:
This removes use of the distgit and lookaside NFS shares.


JimFuller-RedHat:
LGTM

jasinner:
I tagged main before this commit as 0.0.1

PR #72, opened 2022-08-15
Fix linting

juspence:
@RedHatProductSecurity/corgi-devs Copied from Adrian's recent MR: https://github.com/RedHatProductSecurity/osidb/pull/32

 - Fixes an issue with pip-compile which would add the index-url to requirements.txt files when compiling deps.
 - Adds the flake8-django plugin which provides extra checks for common Django mistakes
 - Removes unused ignored errors from the flake8 tox config
 - Removes various "type: ignore" and "noqa" annotations
 - Fixes various warnings from flake8, mypy, Pycharm, etc.


tox.ini
```diff
@@ -53,12 +53,7 @@ commands =
 # NOTE: Any ignored errors/warnings specified below are subjective and can be changed based on
 #       common agreement of all developers contributing to this project.
 #
-# W504: line break after binary operator
-# W503: line break before binary operator (used by black formatter)
-# E203: whitespace before ':' (used by black formatter)
-#
 # TODO: move config to pyproject.toml once https://github.com/PyCQA/flake8/issues/234 is resolved
-ignore = W504,W503,E203
```
mprpic, line 61: Can we at least keep the ones that black enforces? Otherwise if someone adds a piece of code in the future that uses e.g. whitespace before a `:` then we'd just be adding these rules back.

requirements/lint.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with python 3.9
 # To update, run:
 #
-#    pip-compile --allow-unsafe --generate-hashes requirements/lint.in
+#    pip-compile --allow-unsafe --generate-hashes --no-emit-index-url requirements/lint.in
```
mprpic, line 5: We never emitted index URLs in the requirements files though so I'm not sure why this is needed. It only works if you define an extra index URL using the `-i` flag, which we don't.

JimFuller-RedHat:
modulo open conversations, LGTM

requirements/lint.txt
Elkasitu: if you compile the dependencies and you have `PIP_INDEX_URL` in your env vars pip-compile will add it by default to the generated requirements.txt

mprpic: Ah, right you are, I guess I never had that env var defined when running pip-compile. Sounds good then!

tox.ini
juspence: I might be a little confused here - do you mean the one black *doesn't* enforce? AKA if black handles something, there's no problem if flake8 enforces it also. So I added back below to match Adrian's MR:
```
extend-ignore = E203
```

According to Adrian's notes, black ignores above so flake8 should as well.

mprpic: The way I read the comments above this line is that for W503 flake8 will complain line break before binary operator  while that is how Black formats the code. Is that no longer true on Black's side? Or does flake8 no longer flag it?

Elkasitu: The key change here is using `extend-ignore` vs `ignore`, the latter overrides *all* ignored codes whereas the former simply appends to the list of things to ignore.

According to the black documentation [1], `W503` is already ignored by default by `flake8`, and black complies with `W504` I think, so `W503` is included in the ignore list and `extend-ignore` appends `E203` to it

[1] https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#id1

juspence: Ack, so I undid my changes here except for switching "ignore" to "extend-ignore".

EDIT: Whoops, just saw Adrian's reply here. Will fix up per that guidance then merge.

mprpic (approved):

PR #73, opened 2022-08-16
load only latest builds in a tag by default

jasinner:
This makes ingestion work similar to deptopia. Latest will move as time goes by, so we'll still need a search function for latest.


JimFuller-RedHat:
LGTM

PR #74, opened 2022-08-16
CORGI-137: add reconciliation and smoke tests

JimFuller-RedHat:
Smoke tests are designed to be run against live environments.

`> python3 scripts/smoke-tests.py {HOST}`

if no host is provided, defaults to localhost:8000.

In addition have added generalised reconciliation script (rhel, openshift, rhacm, satellite & ansible).

`> python3 scripts/reconciliation.py {HOST}`

**Note** - this is by no means comprehensive but a start - we will enhance reconciliation script as we go.

### DIFF
--- a/scripts/reconciliation.py
+++ b/scripts/reconciliation.py
@@ -1,0 +1,114 @@
+# -----------------
+# product reconciliation scripts
+# -----------------
+#
+# Smoke tests are designed to be run against live environments
+#
+# > python3 scripts/reconciliation.py {HOST}
+#
+# if no host is provided defaults to localhost:8000
+#
+# Note- require installation of corgi-bindings:
+
+import argparse
+import json
+import logging
+import sys
+
+import corgi_bindings
+import requests
+import urllib3
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+# setup logging
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
+# get args
+parser = argparse.ArgumentParser()
+parser.add_argument("corgi_url", nargs="?")
+args = parser.parse_args()
+
+logger.info("smoke-test: start")
+
+session = None
+if args.corgi_url is None:
+    session = corgi_bindings.new_session(corgi_server_uri="http://localhost:8000", verify_ssl=False)
+else:
+    session = corgi_bindings.new_session(corgi_server_uri=args.corgi_url, verify_ssl=False)
+
+# check specific streams [corgi ofuri, deptopia id]
+stream_corpus = [
+    ["o:redhat:rhn_satellite:6.7", "93"],
+    ["o:redhat:rhn_satellite:6.8", "50"],
+    ["o:redhat:rhn_satellite:6.9", "72"],
+    ["o:redhat:ansible_automation_platform:1.2", "2085"],
+    ["o:redhat:ansible_automation_platform:2.0", "2030"],
+    ["o:redhat:ansible_automation_platform:2.1", "2154"],
+    ["o:redhat:ansible_automation_platform:2.2", "2340"],
+    ["o:redhat:openshift:4.9.z", "100"],
+    ["o:redhat:rhacm:2.4.z", "1444"],
+    ["o:redhat:rhel:8.6.0", "2302"],
+]
+
+# check if stream exists in product stream
+for stream_ofuri, deptopia_id in stream_corpus:
+    logger.info(f"reconciliation: check product_stream: {stream_ofuri}.")
+
+    response = requests.get(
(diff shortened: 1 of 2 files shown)


scripts/smoke-tests.py
```diff
@@ -0,0 +1,96 @@
...
+# get args
+parser = argparse.ArgumentParser()
+parser.add_argument("host", nargs="?")
+args = parser.parse_args()
+
+logger.info("smoke-test: start")
+
+# simple rest api checks
+logger.info("smoke-test: simple REST-API checks")
+response = requests.get(f"https://{args.host}")
```
jasinner: Can we please remove the scheme from these urls. I had an instance running on another machine I wanted to test which did not have TLS. In other words I want to be able to run it as:

./scripts/smoke-test.py http://<my-host>:8008

scripts/reconciliation.py
```diff
@@ -0,0 +1,127 @@
...
+parser.add_argument("host", nargs="?")
+args = parser.parse_args()
+
+logger.info("smoke-test: start")
+
+session = None
+if args.host is None:
+    session = corgi_bindings.new_session(corgi_server_uri="http://localhost:8000", verify_ssl=False)
+else:
+    session = corgi_bindings.new_session(corgi_server_uri=f"https://{args.host}", verify_ssl=False)
```
jasinner: Also please remove the scheme here.

scripts/reconciliation.py
```diff
@@ -0,0 +1,127 @@
...
+    response.raise_for_status()
+    deptopia_builds = json.loads(response.text)
+
+    logger.info(f"deptopia ps_update_stream {deptopia_builds['ps_update_stream']}")
+
+    session = None
+    if args.host is None:
+        session = corgi_bindings.new_session(corgi_server_uri="http://localhost:8000")
+    else:
+        session = corgi_bindings.new_session(corgi_server_uri=f"https://{args.host}")
```
jasinner: And here.

scripts/reconciliation.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        if single_response.count > 0:
+            logger.info("SUCCESS: %s", deptopia_build["nvr"] + "," + deptopia_build["build_type"])
+            success.append(deptopia_build["nvr"])
+        else:
+            check_name = session.components.retrieve_list(
+                name=deptopia_build["name"], type=component_type
+            )
+            if check_name.count > 0:
+                check_version = session.components.retrieve_list(
```
jasinner: I think we just need to compare NVRs here, not the count of components with the same name and version. Perhaps we need to implement a lookup in Corgi to get all the latest builds in a product stream first though?

jasinner (changes requested):
This is a really good start thank you! If you can just allow a scheme in the host argument, that would be enough for now. We can iron out the NVR comparison later, what you have it good enough for now.

scripts/reconciliation.py
```diff
@@ -0,0 +1,127 @@
...
+    # ["o:redhat:rhn_satellite:6.7", "93"],
+    # ["o:redhat:rhn_satellite:6.8", "50"],
+    # ["o:redhat:rhn_satellite:6.9", "72"],
+    # ["o:redhat:ansible_automation_platform:1.2", "2085"],
+    # ["o:redhat:ansible_automation_platform:2.0", "2030"],
+    # ["o:redhat:ansible_automation_platform:2.1", "2154"],
+    # ["o:redhat:ansible_automation_platform:2.2", "2340"],
+    ["o:redhat:openshift:4.9.z", "100"],
+    ["o:redhat:rhacm:2.4.z", "1444"],
+    ["o:redhat:rhel:8.6.0.z", "2302"],
```
jasinner: This is not the same stream we are ingesting in the UAT load document. In UAT we are loading 8.6.0 (not .z), that is current product id 58 in Deptopia. I think we should be able to accurately compare NVRs between this stream in Corgi, and Deptopia.

jasinner: I verified that the 8.6.0 stream was successful for all builds in Deptopia for the current data in corgi-stage.prodsec. I think we should adjust the test to compare full NVR, not just name and version in order to be more accurate.

scripts/reconciliation.py
```diff
@@ -0,0 +1,127 @@
...
+        session = corgi_bindings.new_session(corgi_server_uri=f"https://{args.host}")
+
+    success = []
+    failure = []
+    different = []
+
+    for deptopia_build in reversed(deptopia_builds["builds"]):
+        component_type = None
+        if deptopia_build["build_type"] == "RPM":
+            component_type = "RPM"
```
jasinner: Shouldn't this be SRPM?

scripts/smoke-tests.py
JimFuller-RedHat: will generalise all ref with scheme

scripts/reconciliation.py
JimFuller-RedHat: Originally was checking nvr then narrowed down for diff ... then changed it for reasons unknown (hehe u would think I would remember it was only Monday!).

scripts/reconciliation.py
JimFuller-RedHat: will fix type and adjust back to nvr

scripts/reconciliation.py
JimFuller-RedHat: corrected to match on 'rpm' and component_type="SRPM"

jasinner (approved):

PR #76, opened 2022-08-16
increase sca task timeout to 30m

jasinner:
After a local run of `loadbrewdata -s rhacm-2.4.z` I got 81 failed tasks, 80 of which where soft time outs from SCA task. That is out of 628 SCA tasks. That's roughly 13% of SCA tasks timing out.


JimFuller-RedHat:
LGTM

PR #77, opened 2022-08-17
CORGI-152 load errata loop

jasinner:
Fixes CORGI-152 by setting key in redis slow_load_errata is fully complete. This prevents an errata with many builds having it's save_product_taxonomy called for every build in that errata.

### DIFF
--- a/corgi/tasks/common.py
+++ b/corgi/tasks/common.py
@@ -65,9 +65,9 @@ def _get_redis():
     return app.broker_connection().default_channel.client
 
 
-def set_task_complete(key: str):
+def set_task_complete(key: str, is_complete: bool = True):
     conn = _get_redis()
-    conn.set(key, int(True))
+    conn.set(key, int(is_complete))
 
 
 def get_task_complete(key: str) -> bool:
--- a/corgi/tasks/common.py
+++ b/corgi/tasks/common.py
@@ -7,6 +7,8 @@ from django_celery_results.models import TaskResult
 from psycopg2.errors import InterfaceError as Psycopg2InterfaceError
 from requests.exceptions import RequestException
 
+from config.celery import app
+
 BACKOFF_KWARGS = {"max_tries": 5, "jitter": None}
 
 # InterfaceError is "connection already closed" or some other connection-level error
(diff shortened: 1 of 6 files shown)


corgi/tasks/common.py
```diff
@@ -57,3 +59,17 @@ def get_last_success_for_task(task_name):
...
     )
+
+
+def _get_redis():
+    return app.broker_connection().default_channel.client
+
+
+def set_task_complete(key: str):
+    conn = _get_redis()
+    conn.set(key, int(True))
```
jasinner: Note: This flag will mean if we ever want to reprocess the task again we'll have to set the associate key back to False or delete all the redis data.

JimFuller-RedHat:
LGTM

corgi/tasks/common.py
juspence: This seems like a really dirty hack. Have you considered using celery-singleton like we do in SDEngine? It also prevents duplicate task runs for the same data, but has configurable lock timeouts and transparently handles all of the Redis logic for you.

JimFuller-RedHat: celery singleton maybe a good route to investigate though not a 'this sprint' thing - it might be useful to step back a little and think what we want from a queue, then incrementalism.

jasinner: I think celery-singleton is a good idea. I posted another PR now which uses it for fetch_brew_build and it seems to work well. I will close this PR, and open another one which splits out the save_prod_tax call to a new task which is a singleton, which should fix CORGI-152 without the need for the direct flag save in Redis proposed in this PR.

corgi/tasks/errata_tool.py
```diff
@@ -24,7 +29,18 @@ def load_et_products() -> None:
     autoretry_for=RETRYABLE_ERRORS,
     retry_kwargs=RETRY_KWARGS,
 )
-def slow_load_errata(erratum_name):
+def load_errata(erratum_name, force_process=False):
```
mprpic, line 32: Won't this make it run on the fast worker now?

corgi/tasks/errata_tool.py
```diff
@@ -79,6 +95,10 @@ def slow_load_errata(erratum_name):
             # once all build's components are ingested we must save product taxonomy
             sb = SoftwareBuild.objects.get(build_id=b)
             sb.save_product_taxonomy()
+        # Note this makes a permanent record in Redis and needs to removed before this task can
+        # be processed again with the same erratum_name. Use force_process true option to clear it,
+        # Or delete the redis persistent storage
+        set_task_complete(f"load_errata:{erratum_name}")
```
mprpic, line 101: I don't love this design. Can't we bail on the fact that the erratum is already loaded in the database? Interacting directly with Redis just introduces another layer of communication between workers and Redis outside of the native Celery stuff.

JimFuller-RedHat: this was closed

PR #79, opened 2022-08-17
CORGI-114: perf improvements saving product taxonomy I

JimFuller-RedHat:
This first PR related to CORGI-114 ensures 

* the list of components save_product_taxonomy works on is distinct
* component get_sources() returns correct information now, as well as avoid multiple objects returned
* added -f (force process) flag to loadbrewdata to force reprocessing of existing softwarebuilds

Note - there will be a couple more PRs related to CORGI-114

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -200,8 +200,8 @@ class SoftwareBuild(TimeStampedModel):
         variant_ids = list(
             ProductComponentRelation.objects.filter(build_id=self.build_id)
             .filter(type=ProductComponentRelation.Type.ERRATA)
-            .values_list("product_ref", flat=True)
             .distinct()
+            .values_list("product_ref", flat=True)
         )
 
         stream_ids = list(
@@ -212,26 +212,30 @@ class SoftwareBuild(TimeStampedModel):
                     ProductComponentRelation.Type.BREW_TAG,
                 ]
             )
-            .values_list("product_ref", flat=True)
             .distinct()
+            .values_list("product_ref", flat=True)
         )
 
         product_details = get_product_details(variant_ids, stream_ids)
 
+        components = set()
         for component in Component.objects.filter(software_build__build_id=self.build_id):
             # This is needed for container image builds which pull in components not
             # built at Red Hat, and therefore not assigned a build_id
             for d in component.cnodes.all().get_descendants(include_self=True):
                 if not d.obj:
                     continue
-                for a in ["products", "product_versions", "product_streams"]:
-                    # Since we're only setting the product details for a specific build id we need
-                    # to ensure we are only updating, not replacing the existing product details.
-                    interim_set = set(getattr(d.obj, a))
-                    interim_set.update(product_details[a])
-                    setattr(d.obj, a, list(interim_set))
-                d.obj.channels = d.obj.get_channels()
-                d.obj.save()
+                components.add(d.obj)
+
+        for component in list(components):
+            for a in ["products", "product_versions", "product_streams"]:
+                # Since we're only setting the product details for a specific build id we need
+                # to ensure we are only updating, not replacing the existing product details.
+                interim_set = set(getattr(component, a))
+                interim_set.update(product_details[a])
+                setattr(component, a, list(interim_set))
+            component.channels = component.get_channels()
+            component.save()
 
         return None
 
@@ -913,17 +917,16 @@ class Component(TimeStampedModel):
     def get_source(self) -> list:
         """return all root nodes"""
         # this uses the mptt get_root function, not the get_root property defined on Component
-        sources = []
+        sources = set()
+
         for cnode in self.cnodes.all():
-            try:
-                source = cnode.get_root()
-                if source:
-                    sources.append(source.purl)
-            except MultipleObjectsReturned:
-                logger.warning(
-                    "Component %s returned multiple objects when returning sources", self.purl
-                )
-        return list(set(sources))
+            if cnode.is_root_node():
+                sources.add(cnode.obj.purl)
+            else:
+                for ancestor in cnode.get_ancestors():
+                    if ancestor.is_root_node():
+                        sources.add(ancestor.obj.purl)
+        return list(sources)
 
     def get_upstreams(self):
         """return upstreams component ancestors in family trees"""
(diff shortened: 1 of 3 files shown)


corgi/core/models.py
```diff
@@ -200,8 +200,8 @@ def save_product_taxonomy(self):
         variant_ids = list(
             ProductComponentRelation.objects.filter(build_id=self.build_id)
             .filter(type=ProductComponentRelation.Type.ERRATA)
-            .values_list("product_ref", flat=True)
             .distinct()
```
juspence, line 203: Is this change needed? I thought ordering of statements in a Django queryset didn't matter: 
```
Finally, note that you can call filter(), order_by(), etc. after the values() call, that means that these two calls are identical:

Blog.objects.values().order_by('id')
Blog.objects.order_by('id').values()
```

Reference: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#values

corgi/core/models.py
```diff
@@ -212,26 +212,30 @@ def save_product_taxonomy(self):
...
+            .values_list("product_ref", flat=True)
         )
 
         product_details = get_product_details(variant_ids, stream_ids)
 
+        components = set()
         for component in Component.objects.filter(software_build__build_id=self.build_id):
             # This is needed for container image builds which pull in components not
             # built at Red Hat, and therefore not assigned a build_id
             for d in component.cnodes.all().get_descendants(include_self=True):
```
juspence: The .all() here will force the queryset to be re-evaluated (make another call to the DB to see if any rows changed), so we'll potentially end up making an extra DB call per component instead of using an already-cached result.

You should be able to avoid this by calling methods on ```component.cnodes``` directly, or else by using ```component.cnodes.get_queryset()```. I don't know it will help that much, but it did jump out at me as I was looking at this.

corgi/core/models.py
```diff
@@ -212,26 +212,30 @@ def save_product_taxonomy(self):
...
-                    # Since we're only setting the product details for a specific build id we need
-                    # to ensure we are only updating, not replacing the existing product details.
-                    interim_set = set(getattr(d.obj, a))
-                    interim_set.update(product_details[a])
-                    setattr(d.obj, a, list(interim_set))
-                d.obj.channels = d.obj.get_channels()
-                d.obj.save()
+                components.add(d.obj)
+
+        for component in list(components):
```
juspence, line 230: You can just iterate through the set directly, there's no need to convert it to a list first. I also wonder if we really need a separate loop just to check distinctness. Can't we do iteration above using e.g.:
```
for component in Component.objects.filter(software_build=self.build_id).distinct("pk"):
    for d in component.cnodes.get_descendants(include_self=True).distinct("pk"):
```

Would also be nice to avoid one-letter variable names (d, a) in favor of something more meaningful (descendant, attr).

EDIT: And after looking further, maybe instead of a nested loop, we should just iterate over ComponentNode directly:
```
for obj in ComponentNode.objects.filter(obj__software_build=self.build_id).distinct("pk").get_descendants(include_self=True).distinct("pk").select_related("obj").values_list("obj", flat=True):
    # Do stuff with component_node's obj
```
Query may need tweaking, but you get the idea - Python is slow and Postgres is fast, so moving logic into the DB should help I think.

corgi/core/models.py
```diff
@@ -913,17 +917,16 @@ def get_provides(self, include_dev=True):
...
-                    sources.append(source.purl)
-            except MultipleObjectsReturned:
-                logger.warning(
-                    "Component %s returned multiple objects when returning sources", self.purl
-                )
-        return list(set(sources))
+            if cnode.is_root_node():
+                sources.add(cnode.obj.purl)
+            else:
+                for ancestor in cnode.get_ancestors():
```
juspence, line 926: Instead of using a for-loop to find the ancestor, can't we just filter for the ancestor with no parent node, AKA the root of the tree? Like ```root = cnode.get_ancestors().get(parent=None)```, then add ```root.obj.purl``` to the set of sources directly?

corgi/core/models.py
```diff
@@ -212,26 +212,30 @@ def save_product_taxonomy(self):
...
-                    # to ensure we are only updating, not replacing the existing product details.
-                    interim_set = set(getattr(d.obj, a))
-                    interim_set.update(product_details[a])
-                    setattr(d.obj, a, list(interim_set))
-                d.obj.channels = d.obj.get_channels()
-                d.obj.save()
+                components.add(d.obj)
+
+        for component in list(components):
+            for a in ["products", "product_versions", "product_streams"]:
```
juspence, line 231: Nit: Iterating over a tuple ```("products", "product_versions", "product_streams")``` is faster than iterating over a list, though it makes very little difference for small collections. I only suggest since this is happening in a nested loop, so there might be some aggregate benefit.

corgi/tasks/brew.py
```diff
@@ -20,9 +20,9 @@
 
 
 @app.task(autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
-def slow_fetch_brew_build(build_id: int, save_product: bool = True):
+def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_process: bool = False):
     logger.info("Fetch brew build called with build id: %s", build_id)
-    if SoftwareBuild.objects.filter(build_id=build_id).count() > 0:
+    if not force_process and SoftwareBuild.objects.filter(build_id=build_id).count() > 0:
```
juspence: Might be better to use .exists() here instead of .count() > 0

corgi/core/models.py
jasinner: I tried querying ComponentNode directly using `(obj__software_build_id=self.build_id)`. While it's a good idea, we'd have to change ComponentNode.obj type to be a GenericRelation from GenericForeignKey as it currently does allow reverse lookups.

corgi/core/models.py
JimFuller-RedHat: it shouldnt but cprofile says otherwise

corgi/tasks/brew.py
JimFuller-RedHat: oh yea, nice catch

corgi/core/models.py
JimFuller-RedHat: thx will try it out!

corgi/core/models.py
JimFuller-RedHat: We are making a distinct list of component entities (not ComponentNodes) and cprofile seemed to indicate otherwise - will give a try.

corgi/core/models.py
JimFuller-RedHat: this was pre-existing code - small speed difference.

corgi/core/models.py
JimFuller-RedHat: thx will try

PR #82, opened 2022-08-18
CORGI-152-split-save-from-load-errata

jasinner:
resolved https://issues.redhat.com/browse/CORGI-152


JimFuller-RedHat:
LGTM though we may need another recapture of test mock

PR #83, opened 2022-08-18
CORGI-114: enable fine grained configuration of postgresql local dev env

JimFuller-RedHat:
This PR enables configuration of postgresql (via adding postgresql.conf). We use the same idiom (w/ pg container) in other projects.

This work was done during analysis of CORGI-114 though is not specific to save_product_taxonomy() eg. provides a good performance bump 'across the board'.

**Note**- I will have a matching PR for infra repo.


jasinner:
LGTM

PR #90, opened 2022-08-23
CORGI-7: Fix misc task logic

juspence:
Quick fixes for some missing decorators / retry logic, and a few small bugs.

### DIFF
--- a/corgi/tasks/rhel_compose.py
+++ b/corgi/tasks/rhel_compose.py
@@ -48,7 +48,12 @@ def save_compose(stream_name) -> None:
                     build_id = rpm_data["build_id"]
                     # found the build_id, stop iterating filenames
                     break
-            # TODO: What if build_id still not found?
+                # if no filenames had RPM data
+                if not build_id:
+                    raise ValueError(
+                        f"When saving compose for {stream_name}'s {compose_id},"
+                        f"no filenames had RPM data for {srpm}"
+                    )
             _, created = ProductComponentRelation.objects.get_or_create(
                 external_system_id=compose_id,
                 product_ref=stream_name,
--- a/corgi/tasks/rhel_compose.py
+++ b/corgi/tasks/rhel_compose.py
@@ -11,7 +11,7 @@ from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 logger = logging.getLogger(__name__)
 
 
-@app.task
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def save_composes() -> None:
     logger.info("Setting up relations for all streams with composes")
     for stream in ProductStream.objects.exclude(composes__exact={}):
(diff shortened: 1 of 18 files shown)


corgi/tasks/sca.py
```diff
@@ -140,15 +139,11 @@ def _get_distgit_sources(source_url: str, package_type: str) -> list[Path]:
...
-        logger.error("Invalid url passed to archive_source %s", e)
-        return None, ""
+    url = urlparse(source_url)
 
-    if url.scheme != "git":
-        logger.error("Cannot download raw source for anything but git protocol")
-        return None, ""
+    # We only support git, git+https, git+ssh
+    if not url.scheme.startswith("git"):
+        raise ValueError("Cannot download raw source for anything but git protocol")
```
juspence, line 146: @jasinner I made this change here before we discussed changing the git-archive logic in standup today. Please let me know if I should drop this, or if only "git" should be explicitly supported, and not tunneled git protocols like git+https.

corgi/tasks/rhel_compose.py
```diff
@@ -28,39 +28,41 @@ def save_compose(stream_name) -> None:
...
+                        filename,
+                    )
+                    rpm_data = brew.koji_session.getRPM(filename)
+                    if not rpm_data:
+                        # Try the next srpm rpm filename
+                        continue
+                    build_id = rpm_data["build_id"]
+                    # found the build_id, stop iterating filenames
+                    break
+            # TODO: What if build_id still not found?
```
juspence: @jasinner This could happen when all the filenames have no rpm_data - in that case, should we raise an error? Right now we pass build_id=None to ProductComponentRelation.objects.get_or_create().

tests/test_sca.py
```diff
@@ -62,7 +62,8 @@ def test_archive_source(
 ):
     target_file, package_name = _archive_source(source_url, package_type)
     mock_git_archive.assert_called_once()
-    assert f"--remote={remote_name}" in mock_git_archive.call_args.args[0]
+    assert remote_name == mock_git_archive.call_args.args[0]
+    # TODO: What is below supposed to be asserting?
```
juspence: @jasinner Any idea what this is supposed to check?

corgi/tasks/rhel_compose.py
jasinner: If filenames have no rpm_data I guess the best we can do is log a warning about it and skip this SRPM.

corgi/tasks/sca.py
jasinner: I think it's fine to add it for now. I'll do a bit of investigation into what protocols are supported using the new git clone method in CORGI-158

tests/test_sca.py
jasinner: Sorry for the mistake, it was probably suppose to check another argument to the function. I'll also fix it during my changes for CORGI-158.

jasinner:
LGTM

PR #94, opened 2022-08-25
CORGI-158 git clone

jasinner:
This switches the SCA task to use `git clone` instead of `git archive`. Unfortunately an update to dist-git meant that we wouldn't use `git archive` anymore.


JimFuller-RedHat:
LGTM

PR #95, opened 2022-08-25
Fix incorrect value for auto_accept

mprpic:
We are accepting messages (confirming we've received them and they
should be removed from the UMB queue) manually if we succeed in
scheduling a Celery task to process the message itself. So auto_accept
should be disabled.

Also, add an explicit setting to auto-settle all accepted messages (this
is the default) to indicate the application considers the delivery to be
complete.

Thanks @juspence for noticing this!


juspence (approved):
LGTM! Although it occurs to me we could re-raise any exception, after we release the message on line 74. Then the manage.py command would exit and the pod would restart.

This would trigger a SignalFX alert, so we'd know how many messages coming out of UMB are failing to get put onto our Redis queues / failing to schedule the slow_fetch_brew_build() task.

GitHub won't let me add suggestions outside of the diff :(

mprpic:
> LGTM! Although it occurs to me we could re-raise any exception, after we release the message on line 74. Then the manage.py command would exit and the pod would restart.
> 
> This would trigger a SignalFX alert, so we'd know how many messages coming out of UMB are failing to get put onto our Redis queues / failing to schedule the slow_fetch_brew_build() task.

I think it would be better to just log it in Splunk as an application-level error and create an alert for those messages specifically (we should figure out how to do this; I'm sure it's more or less the same as SignalFX with some clicky stuff in the web UI). We'll need to do it for other errors anyway since we shouldn't really bring down any pod that encounters an exception.

PR #96, opened 2022-08-26
get first cnode where possible

jasinner:
This improves the performance of the get_upstreams, and get_provides methods called by save_component_taxonomy. Seems to give about a 50% reduction in the time taken by save_component_taxonomy on builds that recently timed out in MP+ stage, eg. build_id=2134940

Unfortunately we could not also add this to get_sources because of CORGI-183


JimFuller-RedHat:
LGTM - note I am working on perf improvements in this area with CORGI-175 (specifically addressing get_sources)

PR #103, opened 2022-08-29
always iterate cnodes after index of generic foreign key

jasinner:
The generic foreign key index added in pull/98 means that iterating of cnodes is fast, so let's just revert to the original behaviour of doing that. This makes the code a bit more future proof in case we start linking multiple component nodes to a component for types other than RPM.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -222,9 +222,9 @@ class SoftwareBuild(TimeStampedModel):
         """it is only possible to update ('materialize') component taxonomy when all
         components (from a build) have loaded"""
         for component in Component.objects.filter(software_build__build_id=self.build_id):
-            cnode = component.cnodes.get_queryset().first()
-            for d in cnode.get_descendants(include_self=True):
-                d.obj.save_component_taxonomy()
+            for cnode in component.cnodes.get_queryset():
+                for d in cnode.get_descendants(include_self=True):
+                    d.obj.save_component_taxonomy()
         return None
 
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat:
LGTM

corgi/core/models.py
```diff
@@ -923,11 +923,11 @@ def get_provides(self, include_dev=True):
...
         if include_dev:
             type_list.append(ComponentNode.ComponentNodeType.PROVIDES_DEV)
-        first_parent = self.cnodes.get_queryset().first()
-        if first_parent.get_descendant_count() == 0:
-            return []
-        for descendant in first_parent.get_descendants().filter(type__in=type_list):
-            provides.add(descendant.purl)
+        for cnode in self.cnodes.get_queryset():
+            if cnode.get_descendant_count() == 0:
+                break
```
juspence: Should this be "continue" instead of "break"? Even if one cnode doesn't have any descendants, we still want to find descendants with PROVIDES type for other cnodes, right?

And if that's the case, you can just remove this check - if you try to do "for item in empty_list", the for loop will just be skipped. Since there's no logic below the for-loop, it just continues to check the next cnode.

LGTM once this is fixed.

jasinner: Good catch, it should be continue so I changed it. I left the check of get_descendant_count because it avoids a DB call, see https://django-mptt.readthedocs.io/en/latest/models.html#get-descendant-count

PR #105, opened 2022-08-29
CORGI-166: Fix deadlock when creating ComponentNode

juspence:
@RedHatProductSecurity/corgi-devs Currently we're missing an index and a unique constraint, so when we call ComponentNode.get_or_create:
1) Django searches for an existing ComponentNode that matches the keyword arguments given (only type= and parent= before this MR). This is slow since there's no cnode index for type and parent (we have to do a full table scan through all the existing cnodes, which slows down more the more cnodes we have).
2) Django creates a new ComponentNode if no match is found in step 1 above, using all the keyword arguments given above, plus any values in the "defaults" dict. This can deadlock or insert multiple rows since there's no unique constraint for type and parent.

To fix 2), I added purl= to the list of keyword arguments, so that each cnode we create really is unique. Some cnodes may have the same purl, but they will not be under the same parent with the same type. That allows adding a UniqueConstraint on all of (type, parent, purl) together. Then adding an index on the same (type, parent, purl) fields together fixes 1) above.


JimFuller-RedHat:
LGTM

PR #106, opened 2022-08-30
CORGI-166: Fix missing constraints / bad logic for parent=None case

juspence:
Quick fix for yesterday's MR - Postgres treats any row with a NULL value in any column as being distinct, per the SQL spec. For example, (1, 2, NULL) is considered distinct from (1, 2, NULL).

This change tweaks our constraint for (type, parent, purl) so that all three are compared when parent isn't NULL, and only (type, purl) are compared when parent is NULL. This is needed to prevent inserting duplicate data.


juspence:
@jasinner I think the SCA change is redundant, but adding it just to make code consistent in all places (in case that helps the DB query planner figure out which index to use). Review appreciated.

JimFuller-RedHat:
LGTM

PR #108, opened 2022-08-31
CORGI-184 cleanup sca scratch

jasinner:
This changes the folder structure used to by the `slow_software_composition_analysis` task. When a build is being processed it's source code is checked out to SCA_SCRATCH_DIR/<build_id>/  (default to /tmp). Because the `slow_software_composition_analysis` task is a singleton this should ensure we only have a single task processing files in that directory at any time.

If the source code has lookaside sources those are checked out to SCA_SCRATCH_DIR/lookaside/<build_id> that ensures that a scan of the folder SCA_SCRATCH_DIR does not also include the lookaside sources. Instead we scan those independently as an archive.

After a scan of both possible source locations is complete both are deleted to avoid an ever growing SCA_SCRATCH_DIR.
- /tmp/<build_id>/
- /tmp/lookaside/<build_id>/

This also adjusts the `slow_software_composition_anslysis` task so that it only calls `save_*_taxonomy` functions if new components are detected by a scan. It also returns the number of new components detected as a task result.


jasinner:
Converting to draft while I debug failing tasks in dev. Getting some errors like this:

```
  File "/opt/app-root/src/corgi/tasks/sca.py", line 102, in slow_software_composition_analysis
    shutil.rmtree(source.parents[0])
  File "/usr/lib64/python3.9/shutil.py", line 717, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/usr/lib64/python3.9/shutil.py", line 715, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/lookaside/1820621'
```

jasinner:
Setting this ready to merge now. The error was caused by some builds having multiple files in the lookaside cache, I've changed the code to accommodate that now.

JimFuller-RedHat:
LGTM

jasinner:
Going to merge this now, to get some feedback from running it in stage. @mprpic @juspence please review anyway.

tox.ini
```diff
@@ -98,6 +98,6 @@ deps = -r requirements/lint.txt
 allowlist_externals = bash
 # Check only files in the current branch which have changed, compared to the main branch, for secrets
 # Scan all files for secrets if the first form fails, since Gitlab CI uses shallow clone and does not have a "main" ref
-commands = /usr/bin/bash -c 'detect-secrets-hook --baseline .secrets.baseline $(\
-    git diff --name-only "origin/main..HEAD" || \
+commands = /usr/bin/bash -c 'detect-secrets-hook --baseline .secrets.baseline --exclude-files tests/data/2018747/sources \
+    $(git diff --name-only "origin/main..HEAD" || \
```
juspence, line 102: This seems to be redundant with above - we ignore this file using both a command-line argument, and an entry in the .secrets.baseline file.

I also don't understand why the .secrets.baseline file needs to use the "detect_secrets.filters.regex.should_exclude_file" filter. Can't we just ignore line 1 of the file (which contains the hash / hex high-entropy string), like we do for other files with commit identifiers?

corgi/tasks/sca.py
```diff
@@ -18,6 +17,7 @@
 from corgi.core.models import Component, ComponentNode, SoftwareBuild
 from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
+LOOKASIDE_SCRATCH_SUBDIR = "lookaside"
```
juspence, line 20: Any reason to make this a constant instead of a Django setting (that we can control with an env var) like before? I don't have a preference either way, just curious if there's some advantage.

corgi/tasks/sca.py
```diff
@@ -69,83 +69,67 @@ def save_component(component: dict[str, Any], parent: ComponentNode):
...
-    software_build.save_component_taxonomy()
-    software_build.save_product_taxonomy()
+    # clean up source code so that we don't have to deal with reuse and an ever growing disk
+    for source in distgit_sources:
+        rm_target = source
+        if source.is_file():
+            rm_target = source.parents[0]
+        # Ignore errors because the dir might already be deleted due to multiples sources
+        # being in the same directory
+        shutil.rmtree(rm_target, ignore_errors=True)
```
juspence, line 103: This seems a little risky, but maybe I don't understand. is there ever a chance that we'd clean up sources for another task?

Or if not, can't we just remove the top-level /tmp/lookaside/build_id/ and /tmp/build_id/ folders directly, without needing to iterate?

tests/test_sca.py
```diff
@@ -38,60 +37,61 @@ def test_parse_components():
...
+            _clone_source(f"{source_url}#{commit}", build_id)
         mock_subprocess.assert_not_called()
 
 
 @patch("subprocess.check_call")
 @patch("corgi.tasks.sca._download_lookaside_sources")
 def test_get_distgit_sources(mock_check_call, mock_download_lookaside_sources):
-    expected_path = "tests/data/rpms/test"
+    shutil.rmtree("/tmp/1")
+    expected_path = "/tmp/1"
```
juspence, line 88: Minor nit: Instead of hardcoding these, could we move them into a variable like test_build_id = 1?

tests/test_sca.py
```diff
@@ -167,66 +171,81 @@ def test_download_lookaside_sources(
...
-            type=Component.Type.SRPM, software_build=sb, name=package_name
+            type=Component.Type.SRPM,
+            software_build=sb,
         )
     root_component.cnodes.get_or_create(
         type=ComponentNode.ComponentNodeType.SOURCE, parent=None, purl=root_component.purl
     )
     assert not Component.objects.filter(purl=expected_purl).exists()
+
+    expected_url = f"https://{os.getenv('CORGI_LOOKASIDE_CACHE_URL')}/repo/{download_path}"
```
juspence, line 242: Maybe I'm confused, but didn't we get rid of this setting? Or if I'm thinking of a different setting, let's keep all the related options in one place (either as constants in sca.py, or as settings in base.py).

tests/test_sca.py
```diff
@@ -138,15 +138,15 @@ def test_download_lookaside_sources(
...
     downloaded_sources = _download_lookaside_sources(
-        distgit_source_archive, package_name, package_type
+        distgit_source_archive, 1, package_type, package_name
     )
     if expected_filename:
-        full_expected_filename = (
-            f"tests/data/lookaside/{package_type}/{package_name}/{expected_filename}/"
-        )
+        expected_hash_prefix = expected_path.split("/")[1][:6]
+        full_expected_filename = f"/tmp/lookaside/1/{expected_hash_prefix}-{expected_filename}/"
```
juspence, line 147: Let's import the lookaside dir constant from sca.py (or make it a Django setting and import that) instead of hardcoding here.

tox.ini
juspence: Fixed this on main since I had to update the .secrets.baseline file anyway.

PR #109, opened 2022-08-31
CORGI-183: Disable saving comp tax

JimFuller-RedHat:
We decided to disable persisting component taxonomy information at ingestion time and dynamically generate when the REST API is called for any specific component. This has a positive impact on ingestion performance though I have taken a minimal approach (eg. we can consider ripping out deprecated function and removing from model related attributes at some future point).

Also tweaked reconciliation script (used to compare components against deptopia).


jasinner (approved):
I pushed one unrelated commit to reduce the verbosity of the brew task logging.

PR #113, opened 2022-09-01
CORGI-156: Handle sourceless builds

mprpic:
This puts in some safeguards around processing builds that may be missing source URLs. It also removes processing of build types that we can't analyzed (yet), e.g. "maven".

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -2,6 +2,7 @@ import itertools
 import json
 import logging
 import re
+from datetime import datetime
 from types import SimpleNamespace
 from typing import Any, Tuple
 from urllib.parse import urlparse
@@ -692,10 +693,18 @@ class Brew:
 
         # Determine build source
         if not build.get("source"):
-            # Sometimes there is no source URL on the build but it can be found in the task
+            # Sometimes there is no source URL on the build, but it can be found in the task
             # request info instead.
             logger.info("Fetching source from task info for %s", build_id)
-            build["source"] = self.get_source_of_build(build)
+            try:
+                build["source"] = self.get_source_of_build(build)
+            except BrewBuildSourceNotFound as exc:
+                # Some older builds do not specify source URLs; the below date was chosen based
+                # on some initial analysis of source-less builds in Brew.
+                if datetime.fromtimestamp(build["completion_ts"]) < datetime(2015, 1, 1):
+                    return {}
+                else:
+                    raise exc
 
         # Add list of Brew tags for this build
         tags = self.koji_session.listTags(build_id)
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -670,13 +670,24 @@ class Brew:
 
         # Determine build type
         build_type_info = self.koji_session.getBuildType(build)
+        build_type = next(
+            type_ for type_ in build_type_info.keys() if type_ in self.SUPPORTED_BUILD_TYPES
+        )
         if not any(type_ in self.SUPPORTED_BUILD_TYPES for type_ in build_type_info.keys()):
             raise BrewBuildTypeNotSupported(
                 f"Build {build_id} type is not supported: {build_type_info}"
             )
-        build_type = next(
-            type_ for type_ in build_type_info.keys() if type_ in self.SUPPORTED_BUILD_TYPES
-        )
+        # TODO: refactor Brew.CONTAINER_BUILD_TYPE to be a generic IMAGE_TYPE with image types
+        #  identified in a separate attribute on the build itself.
+        if build_type == self.CONTAINER_BUILD_TYPE:
+            # If this is an "image" type build, it may be building a container image, ISO image,
+            # or other types of images. Check the content generator name to determine where this
+            # image was built, which indicates what type of image it is.
+            if build["cg_name"] != "atomic-reactor":
+                # Container images are built in OSBS, which uses atomic-reactor to build them.
+                raise BrewBuildTypeNotSupported(
+                    f"Image build {build_id} is not supported: {build['cg_name']} content generator used"
+                )
         build["type"] = build_type
 
         # Determine build source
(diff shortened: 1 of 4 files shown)


tests/test_brew_collector.py
```diff
@@ -116,8 +117,9 @@ def test_get_component_data(build_id, build_source, build_ns, build_name, licens
     c = Brew().get_component_data(build_id)
     if build_type == "module":
         assert list(c.keys()) == ["type", "namespace", "meta", "analysis_meta", "build_meta"]
-    elif build_type == "maven":
-        assert list(c.keys()) == ["type", "namespace", "meta", "build_meta"]
+    # TODO: uncomment when Maven build type is supported
+    # elif build_type == "maven":
+    #     assert list(c.keys()) == ["type", "namespace", "meta", "build_meta"]
```
juspence, line 122: I'd prefer using ```with pytest.raises(BrewBuildTypeNotSupported)``` here.

corgi/tasks/brew.py
```diff
@@ -29,17 +29,21 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     else:
         logger.info("Fetching brew build with build_id: %s", build_id)
 
-    build = Brew().get_component_data(build_id)
-    build_meta = build["build_meta"]["build_info"]
+    component = Brew().get_component_data(build_id)
+    if not component:
+        logger.info("No data fetched for build %s from Brew, exiting...", build_id)
+        return
```
juspence, line 35: I'd rather make this an uncaught exception, so we can see how may builds of an unsupported type we keep trying to ingest. Or at least error-level logging if we want to silence this, so it's easier to find in logs.

corgi/collectors/brew.py
```diff
@@ -697,14 +722,10 @@ def get_component_data(self, build_id: int) -> dict:
...
             component = self.get_rpm_build_data(build_id)
-        elif build_type == self.MAVEN_BUILD_TYPE:
-            component = self.get_maven_build_data(build, build_type_info)
         elif build_type == self.MODULE_BUILD_TYPE:
             component = self.get_module_build_data(build)
-        elif build_type == self.WIN_BUILD_TYPE:
-            component = {}
         else:
-            component = {}
+            return {}
```
juspence: Can we do ```raise BrewBuildTypeNotSupported()``` in this case?

mprpic: Technically we should never even get to this else branch but then linting complaints that `component` may be undefined. I guess I can change it to raise an exception, to make it more explicit what is happening.

corgi/tasks/brew.py
mprpic: This doesn't catch unsupported builds but rather builds that don't have a source URL that are too old, we just kind of ignore them since we can't do anything about them. I'll add an error log whenever we detect an old build though.

tests/test_brew_collector.py
mprpic: We're not calling this though (and no unsupported build types are present in test parameters), and whenever support is added, it won't raise the exception.

corgi/collectors/brew.py
```diff
@@ -666,21 +671,45 @@ def get_component_data(self, build_id: int) -> dict:
...
         if not any(type_ in self.SUPPORTED_BUILD_TYPES for type_ in build_type_info.keys()):
             raise BrewBuildTypeNotSupported(
                 f"Build {build_id} type is not supported: {build_type_info}"
             )
-        build_type = next(
-            type_ for type_ in build_type_info.keys() if type_ in self.SUPPORTED_BUILD_TYPES
-        )
+        # TODO: refactor Brew.CONTAINER_BUILD_TYPE to be a generic IMAGE_TYPE with image types
+        #  identified in a separate attribute on the build itself.
+        if build_type == self.CONTAINER_BUILD_TYPE:
```
JimFuller-RedHat, line 683: nice !

JimFuller-RedHat:
LGTM

tests/test_brew_collector.py
juspence: I meant uncommenting the Maven-type test data defined in build_corpus above, then explicitly testing we raise the right exception for unsupported types. But it's fine to leave it commented I guess.

corgi/tasks/brew.py
juspence: Ah, I see now. I was looking at an old version of the diff earlier. Error logging sounds fine then.

corgi/collectors/brew.py
```diff
@@ -666,21 +671,45 @@ def get_component_data(self, build_id: int) -> dict:
...
+                build["source"] = self.get_source_of_build(build)
+            except BrewBuildSourceNotFound as exc:
+                # Some older builds do not specify source URLs; the below date was chosen based
+                # on some initial analysis of source-less builds in Brew.
+                if datetime.fromtimestamp(build["completion_ts"]) < datetime(2015, 1, 1):
+                    logger.error(
+                        f"Build {build_id} has no associated source URL but is too old "
+                        f"to process; returning an empty component."
+                    )
+                    return {}
```
juspence, line 710: Just saw this change (was looking at old diff before). Are we really unable to do anything without a source URL? I guess we can't do SCA without something to download, but we still can get tags and errata_tags. Does build_data below depend on having a source_url to analyze?

If not, could we still get all other data besides the source URL, then return the built component dict's data (just missing a source URL) instead of an empty dict? It seems more useful to try and return something. But I'm not really sure how the source URL is used, how we want to indicate "this build has no source", and what the dependencies are. So it's fine to leave it like this if you think that's better.

juspence (approved):

corgi/collectors/brew.py
mprpic: We could certainly improve this to do exactly what you say (record a subset of the data), but I haven't found a single build that we can analyze right now that wouldn't have a source URL, so we'd be adding code that would never run. We can certainly file it as an RFE for the backlog if we come across a legit source-less rpm/container/module build. At the latest, we can add this if we ever support analyzing ISOs.

tests/test_brew_collector.py
mprpic: I'll leave it as is for now.

PR #114, opened 2022-09-01
get module version from brew_info instead of modulemd

jasinner:
This fixes the issue described in CORGI-199.

We still need to link the module RPMs as children of the RHEL_MODULE component. There doesn't seem to be enough information in brew alone to create the link.


JimFuller-RedHat:
LGTM

PR #115, opened 2022-09-01
CORGI-195: Fix pnode ordering since it's no longer guaranteed

juspence:
Quick fix for issue in CI tests.


JimFuller-RedHat:
this PR was committed without running checks

PR #118, opened 2022-09-02
CORGI-201: Use password authentication even if a valid Kerb ticket exists

mprpic:
This saves us an extra request to the KDC to validate the ticket against
the DB, which inevitably fails because the DB server is not registered as a
valid principal in the KDC.


juspence:
LGTM, thanks for catching this!

juspence (approved):

JimFuller-RedHat:
LGTM

PR #121, opened 2022-09-06
Account for unknown build types

mprpic:
This is a regression from #113. If we encounter a build type that isn't
supported, the next() call ends with StopIteration and an exception is
emitted. Add a default that isn't included in the supported build type list
so we raise the correct exception, `BrewBuildTypeNotSupported`.


JimFuller-RedHat:
LGTM

PR #122, opened 2022-09-06
CORGI-210: Add skiplist for certain internal-only and test components

mprpic:


corgi/collectors/brew.py
```diff
@@ -57,6 +57,11 @@ class Brew:
         MODULE_BUILD_TYPE,
     )
 
+    COMPONENT_NAME_SKIPLIST = (
```
JimFuller-RedHat, line 60: nitpick - **component_excludes** might be a simpler name.

Have you considered defining as an env var ? That approach might come useful for potential external users who may have a different exclude list.

mprpic: Sure, changed to "exclude" and made it load a list from an env var as well.

PR #123, opened 2022-09-06
Remove bandit and radon from linting

mprpic:
The results these two tools report have not proven to be valuable enough to
maintain them as dependencies and include them in our automated CI runs.


JimFuller-RedHat:
sure - these are useful when first starting a project ... their usefulness declines over time ... happy to have them removed.

PR #125, opened 2022-09-07
CORGI-213: fix indexes

JimFuller-RedHat:
turns out we needed index for purl on ComponentNode

once applied, the following psql query should show correct indexes being used.

```
EXPLAIN ANALYZE SELECT "core_componentnode"."id",
       "core_componentnode"."parent_id",
       "core_componentnode"."lft",
       "core_componentnode"."rght",
       "core_componentnode"."tree_id"
  FROM "core_componentnode"
 WHERE "core_componentnode"."purl" = 'pkg:rpm/redhat/curl@7.61.1-18.el8_4.2?arch=aarch64';
```


mprpic (approved):
LGTM, we should split out all the other indexes into individual ones as well.

PR #126, opened 2022-09-07
related to CORGI-213: add a few more indexes

JimFuller-RedHat:
minor REST API perf improvements


corgi/core/models.py
```diff
@@ -172,6 +172,7 @@ class Meta:
             ),
         ]
         indexes = [
+            models.Index(fields=("type", "parent", "purl")),
```
mprpic, line 175: What's the benefit of the tuple of fields over the individual ones? Do we do a lot of look-ups on all three fields?

JimFuller-RedHat: on further thought it felt safer to add this one back - the only negative if it turns out it is unneeded is we use up a little extra storage space ... will remove these kind of indexes once I absolutely confirm such tuples indexes are not needed.

juspence: FYI this was added as part of CORGI-166. In order for Django's get_or_create() function to be safely atomic, and not insert duplicate rows, you have to enforce uniqueness on all the keyword arguments that appear in the get_or_create() call at the database level.

So I added the unique constraint, and in order for that uniqueness check / get_or_create() to stay performant, indexing the same fields that appear in the constraint is also needed.

PR #127, opened 2022-09-08
Remove unused code

mprpic:


JimFuller-RedHat:
makes sense - we can always retrieve back from git history

PR #133, opened 2022-09-09
related to CORGI-213: enable force loading when loading from loadcomposedata

JimFuller-RedHat:
Enables force loading of builds when invoking loadcomposedata command.

`./manage.py loadcomposedata -f -s rhel-8.7.0 `

### DIFF
--- a/corgi/tasks/management/commands/loadcomposedata.py
+++ b/corgi/tasks/management/commands/loadcomposedata.py
@@ -33,6 +33,12 @@ class Command(BaseCommand):
             action="store_true",
             help="Schedule build for ingestion inline (not in celery)",
         )
+        parser.add_argument(
+            "-f",
+            "--force",
+            action="store_true",
+            help="Force ingestion even if the Compose exists",
+        )
 
     def handle(self, *args, **options) -> None:
         if options["compose_names"]:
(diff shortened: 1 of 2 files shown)


corgi/tasks/management/commands/loadcomposedata.py
```diff
@@ -33,6 +33,12 @@ def add_arguments(self, parser: CommandParser) -> None:
             action="store_true",
             help="Schedule build for ingestion inline (not in celery)",
         )
+        parser.add_argument(
+            "-f",
+            "--force",
+            action="store_true",
+            help="force ingestion, even if it exists",
```
mprpic:
```suggestion
            help="Force ingestion even if the Compose exists",
```

corgi/tasks/management/commands/loadcomposedata.py
```diff
@@ -60,6 +70,6 @@ def handle(self, *args, **options) -> None:
         for build_id in build_ids:
             build_id = int(build_id)
             if options["inline"]:
-                slow_fetch_brew_build(build_id)
+                slow_fetch_brew_build(build_id, force_process=force_process)
```
mprpic: You can just pass `options["force"]` here (and below), it'll already be True or False based on the input.

mprpic (approved):

PR #135, opened 2022-09-12
CORGI-221 purl url parameter broken when it contains reserved characters

jasinner:
reserved characters in purl params get correctly encoded and looked up from db in components list view


JimFuller-RedHat:
LGTM

JimFuller-RedHat (approved):
LGTM

PR #139, opened 2022-09-12
CORGI-215 - return array of components when using /api/v1/components?ofuri={}

JimFuller-RedHat:
Invoking

` /api/v1/components?ofuri={}`

should return an array of {link,purl} objects.

**NOTE**- The only other open question is if we can do boolean || with django rest api params (eg. type=SRPM&type=CONTAINER_IMAGE or type=SRPM||CONTAINER_IMAGE)


mprpic (approved):
Looks fine as a quick solution, but longer term we should refactor this to a simple include/exclude_fields filter that could replicate the behavior without needing the extra ComponentListSerializer.

PR #129, opened 2022-09-08
CORGI-200 compose modules

jasinner:
This builds the component tree to link RHEL_MODULE component with it's RPM component children when ingesting composes.

This also fixes CORGI-211

PR #140, opened 2022-09-12
CORGI-208: Improve purl generator with more correct namespace/type values

mprpic:
We should consider submitting a proposal to include `rpmmod` as a supported purl type. I filed CORGI-224 to further improve the namespace vs type fields.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -190,8 +190,7 @@ class ComponentNode(MPTTModel, TimeStampedModel):
         return self.desc.name
 
     def save(self, *args, **kwargs):
-        if not self.purl:
-            self.purl = self.obj.purl
+        self.purl = self.obj.purl
         super().save(*args, **kwargs)
 
 
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -808,7 +808,7 @@ class Component(TimeStampedModel):
                 type="container",
                 namespace=self.Namespace.REDHAT.lower(),
                 name=self.name,
-                version=f"{self.version}",
+                version=f"{self.version}-{self.release}",
                 qualifiers={
                     "arch": self.arch,
                     "digest": digest,
@@ -818,17 +818,23 @@ class Component(TimeStampedModel):
         elif self.type == Component.Type.UPSTREAM:
             # Upstream components should default to the "generic" purl type and not use any
             # namespaces. In the future, we may extend this to map to upstream-defined purls.
+            version = self.version
+            if self.release:
+                version = f"{version}-{self.release}"
             return PackageURL(
                 type="generic",
                 name=self.name,
-                version=f"{self.version}",
+                version=version,
             )
         else:
-            # unknown
+            # Unknown
+            version = self.version
+            if self.release:
+                version = f"{version}-{self.release}"
             return PackageURL(
                 type=self.type,
                 name=self.name,
-                version=f"{self.version}",
+                version=version,
             )
 
     def save_component_taxonomy(self):
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -768,20 +768,34 @@ class Component(TimeStampedModel):
         return str(self.name)
 
     def get_purl(self):
-        if self.type in (Component.Type.RPM, Component.Type.SRPM, Component.Type.RHEL_MODULE):
-            evr = ""
+        if self.type in (Component.Type.RPM, Component.Type.SRPM):
             qualifiers = {
                 "arch": self.arch,
             }
             if self.epoch:
-                evr = f"{self.epoch}:"
                 qualifiers["epoch"] = str(self.epoch)
-            evr = f"{evr}{self.version}-{self.release}"
             return PackageURL(
-                type=self.type,
+                type="rpm",  # Purl defines no type specific to SRPMs
                 namespace=self.Namespace.REDHAT.lower(),
                 name=self.name,
-                version=evr,
+                version=f"{self.version}-{self.release}",
+                qualifiers=qualifiers,
+            )
+        elif self.type == Component.Type.RHEL_MODULE:
+            # Break down RHEL module version into its specific parts:
+            # NSVC = Name, Stream, Version, Context
+            version, _, context = self.release.partition(".")
+            qualifiers = {
+                "stream": self.version,
+                "version": version,
+                "context": context,
+            }
+            return PackageURL(
+                # Purl does not define a type for RHEL/CentOS modules; TODO: propose rpmmod
+                type="rpmmod",
+                namespace=self.Namespace.REDHAT.lower(),
+                name=self.name,
+                version=f"{self.version}-{self.release}",
                 qualifiers=qualifiers,
             )
         elif self.type == Component.Type.CONTAINER_IMAGE:
@@ -803,15 +817,12 @@ class Component(TimeStampedModel):
                 subpath=None,
             )
         elif self.type == Component.Type.UPSTREAM:
-            ns = self.meta_attr.get("url", "")
-            if "//" in ns:
-                ns = ns.split("//")[1]
-                # TODO: below does not use ns??
+            # Upstream components should default to the "generic" purl type and not use any
+            # namespaces. In the future, we may extend this to map to upstream-defined purls.
             return PackageURL(
-                type="upstream",
+                type="generic",
                 name=self.name,
                 version=f"{self.version}",
-                subpath=None,
             )
         else:
             # unknown
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -766,7 +766,7 @@ class Component(TimeStampedModel):
         """return name"""
         return str(self.name)
 
-    def get_purl(self):
+    def get_purl(self) -> PackageURL:
         if self.type in (Component.Type.RPM, Component.Type.SRPM):
             qualifiers = {
                 "arch": self.arch,
@@ -856,7 +856,7 @@ class Component(TimeStampedModel):
         self.nvr = self.get_nvr()
         self.nevra = self.get_nevra()
         purl = self.get_purl()
-        self.purl = purl.to_string() if purl else ""
+        self.purl = purl.to_string()
         super().save(*args, **kwargs)
 
     @property


JimFuller-RedHat:
thx for this - now reading this PR a few thoughts come to mind:

* We would have to regenerate all purls for all components after this is applied.
* Unclear if we conflate SRPM with RPM if that does not introduce potential for collision. Also we would keep SRPM component type ... and hide that fact in purl
* Package url is barely a spec eg. it is an adhoc community standard ... if we have something non standard I am fine with it in the short term (and propose changes to purl community)

Are we proposing anything for handling epoch ?

My pref is to keep proposed enhancements with rhel_module and upstream->generic changes and consider holding off on conflating srpm with rpm ... that in particular seems slightly risky. Maybe I am being paranoid.

mprpic:
> thx for this - now reading this PR a few thoughts come to mind:
> 
>     * We would have to regenerate all purls for all components after this is applied.
> 

Ah true, but that could be a simple data migration. I'll add one. We can consider squashing all migrations into one (and then deleting any data migrations) before our 1.0.0 release.

>     * Unclear if we conflate SRPM with RPM if that does not introduce potential for collision. Also we would keep SRPM component type ... and hide that fact in purl
> 

There should still be uniqueness based on the arch type, for which SRPMs should use `src`. Revisiting the Type and Namespace values can be solved in that follow-up CORGI-224 issue and would also de-risk changing the purl package type (you could still filter on the Type=SRPM field).

The point of not treating SRPM as a unique type is that SRPMs are RPMs just like any other ones. They just happen to contain sources instead of binaries. The only way to detect if an RPM is a source RPM is to check the file's header for some special value. So proposing adding an `srpm` type to the purl spec would (I assume) not be accepted.

>     * Package url is barely a spec eg. it is an adhoc community standard ... if we have something non standard I am fine with it in the short term (and propose changes to purl community)
> 
> 
> Are we proposing anything for handling epoch ?
> 

In the purl spec?

> My pref is to keep proposed enhancements with rhel_module and upstream->generic changes and consider holding off on conflating srpm with rpm ... that in particular seems slightly risky. Maybe I am being paranoid.

@jasinner wdyt?

JimFuller-RedHat:
thx for explanation - and +1 to your set of commits (we can always ?type=SRPM to facet on these ) ... wld be good to get @jasinner thoughts as well

mprpic:
Actually, I'll take the lazy route out and once this is deployed just run this in Python shell before we migrate the data from stage to prod:
```
for c in Components.objects.all():
    c.save()
for c in ComponentNode.objects.all():
    c.save()
```
That refreshes the nevra, nvr, and purl values and bumps the last changed timestamp (which is fine).

JimFuller-RedHat:
sounds like a plan, maybe we take a paranoid backup before applying that change ;)

jasinner:
> Actually, I'll take the lazy route out and once this is deployed just run this in Python shell before we migrate the data from stage to prod:
> 
> ```
> for c in Components.objects.all():
>     c.save()
> for c in ComponentNode.objects.all():
>     c.save()
> ```
> 
> That refreshes the nevra, nvr, and purl values and bumps the last changed timestamp (which is fine).

ComponentNode.save() only updates the purl is the purl is None. I think we'll have to do something more manual like:

```
for c in ComponentNode.objects.all():
   c.purl = c.obj.purl
```

mprpic:
> ComponentNode.save() only updates the purl is the purl is None. I think we'll have to do something more manual like:
> 

@jasinner Why does it do that actually? Is there any harm in removing that IF condition?

juspence:
Sorry for delayed reply here - I didn't see this MR since no one got tagged / requested as a reviewer. If we're going to regenerate purls, we need to make sure the ComponentNodes have their purl updated as well. Martin (in gchat) proposed running the below code in addition to https://github.com/RedHatProductSecurity/component-registry/pull/140#issuecomment-1244196863 above:
```
for c in Component.objects.all():
    c.save()
for c in ComponentNode.objects.all():
    c.save()
```

I don't think this will be enough, since if I remember right, the purl is only copied from the Component to the Node the first time the Node is created. If the node's purl is already set, calling my_node.save() again won't update the purl (I think).

We could just filter for all Components with Type=RHEL_MODULE and set the purl manually, though:
```
for c in Component.objects.filter(type=Component.Type.RHEL_MODULE):
    c.save()  # Updates the purl automatically using get_purl
    for node in c.cnodes.all():
        # Update the purl manually, since it only changes automatically when unset
        node.purl = c.purl
        node.save()
```

Shouldn't be a blocker or a big deal, just something to be aware of.

EDIT: I'm slow today, +1 to Jason's comment above. The harm would be an additional query / data read (to look up the linked component) every time a node has its .save() method called. Not a big deal, but I can see it hurting performance in some cases. 

Since the purl isn't really supposed to change, I'd be in favor of keeping the current code. We can just run something like above to manually fix the node purls if we ever change the purl format again.

jasinner:
This seems to be what we need, waiting for it to complete before spot checking the data:

```
for c in m.Component.objects.all():
     c.save()

 for n in m.ComponentNode.objects.all():
     n.purl = m.Component.objects.get(n.object_id).purl
     n.save()
```

mprpic:
Pushed a new commit that removes the condition on refreshing the ComponentNode.purl on save; now it really should be just:
```python
for c in Component.objects.all():
    c.save()
for c in ComponentNode.objects.all():
    c.save()
```

jasinner:
There is a corrupt data probably created before [this commit](https://github.com/RedHatProductSecurity/component-registry/commit/37f70af40b62b90c074ce91d2d0a7ac4249ae4df).

We'll have to clean it up before we can re-generate all the ComponentNode purls, maybe like this:

```
import corgi.core.models as m
from django.db import IntegrityError, DatabaseError
from corgi.tasks.brew import slow_fetch_brew_build

for c in m.Component.objects.all():
    c.save()


def fix_root(node):
    print(f"finding roots for '{node.purl}'")
    for root in node.obj.get_roots:
        print(f"deleting root node '{root.purl}' and re-fetching")
        slow_fetch_brew_build(root.obj.software_build.build_id, force_process=True)
        root.obj.delete()


for n in m.ComponentNode.objects.all():
    if not n.obj:
        print(f"node with null obj found '{n.purl}', deleting")
        n.delete()
        continue
    try:
        n.save()
    except IntegrityError:
        print(f"got integrity error for '{n.purl}'")
        if n.obj.software_build:
            build_id = n.obj.software_build.build_id
            print(f"deleting software build for '{build_id}' and re-fetching")
            slow_fetch_brew_build.delay(build_id, force_process=True)
        else:
            fix_root(n)
    except DatabaseError as e:
        fix_root(n)

```

corgi/core/models.py
```diff
@@ -190,8 +190,7 @@ def desc(self):
         return self.desc.name
 
     def save(self, *args, **kwargs):
-        if not self.purl:
-            self.purl = self.obj.purl
+        self.purl = self.obj.purl
```
jasinner, line 193: I found that sometimes self.obj is None which causes this to fail. Let's add a if self.obj check here.

mprpic: That would indicate that there are dangling nodes without actual objects that they should be pointing to. I think in general we lack some sort of deletion mechanism for nodes when their objects get deleted. I think we always assumed that we'd be appending data only and never deleting so it would be interesting to know how a ComponentNode came to be orphaned.

jasinner: You are right, never mind it. I could be caused by the data cleanup loop. It should never happen in normal operations.

JimFuller-RedHat: for completeness ... 

root cause is likely due to change near models.py#170  via

https://github.com/RedHatProductSecurity/component-registry/pull/129/commits/37f70af40b62b90c074ce91d2d0a7ac4249ae4df

the change itself is perfectly valid - though the existence of these nodes before the change would have needed a custom migration to clean up.

jasinner:
> import corgi.core.models as m
> from django.db import IntegrityError, DatabaseError
> from corgi.tasks.brew import slow_fetch_brew_build
> 
> for c in m.Component.objects.all():
>     c.save()
> 
> 
> def fix_root(node):
>     print(f"finding roots for '{node.purl}'")
>     for root in node.obj.get_roots:
>         print(f"deleting root node '{root.purl}' and re-fetching")
>         slow_fetch_brew_build(root.obj.software_build.build_id, force_process=True)
>         root.obj.delete()
> 
> 
> for n in m.ComponentNode.objects.all():
>     if not n.obj:
>         print(f"node with null obj found '{n.purl}', deleting")
>         n.delete()
>         continue
>     try:
>         n.save()
>     except IntegrityError:
>         print(f"got integrity error for '{n.purl}'")
>         if n.obj.software_build:
>             build_id = n.obj.software_build.build_id
>             print(f"deleting software build for '{build_id}' and re-fetching")
>             slow_fetch_brew_build.delay(build_id, force_process=True)
>         else:
>             fix_root(n)
>     except DatabaseError as e:
>         fix_root(n)

This completed successfully for me in dev and it seems like all the purls have been updated successfully. I lost the logs, but it looks like 63 builds where re-fetched.

corgi/core/models.py
```diff
@@ -808,7 +808,7 @@ def get_purl(self):
                 type="container",
                 namespace=self.Namespace.REDHAT.lower(),
                 name=self.name,
-                version=f"{self.version}",
+                version=f"{self.version}-{self.release}",
```
jasinner, line 811: I think we should only add '-' if release is not None or empty.

jasinner: Not related to this PR, but NVR, and NEVRA field for container container components has the same training '-' problem that this would create.

mprpic: @jasinner Are there container images that *don't* have a release field?

juspence (approved):

corgi/core/models.py
mprpic: There don't seem to be any:
```
In [3]: Component.objects.filter(type=Component.Type.CONTAINER_IMAGE, release="").count()
Out[3]: 0
```
NVR and NEVRA do have the problem for other component types though, but I guess I'll leave that for a follow-up PR.

corgi/core/models.py
mprpic: Resolved by #145.

juspence: #145 doesn't fix nodes without a linked component (obj field). It only prevents duplicate nodes from being inserted when the parent field is NULL / None.

I guess ComponentNodes could be orphaned if we're doing manual cleanup. e.g. I think the other day we ran some code to delete and refetch a SoftwareBuild. Due to the ForeignKey between SoftwareBuild and Component (on_delete=models.CASCADE), the Component would also be deleted:
```
# on_delete=models.CASCADE means that deleting a SoftwareBuild also deletes all linked Components
software_build = models.ForeignKey(SoftwareBuild, on_delete=models.CASCADE, null=True)
```

But GenericForeignKey doesn't cascade deletions by default, unless there's a reverse relation that links a ComponentNode to its parent Component. We have this set up between Products and ProductNodes:
```
# related_query_name= with any value means that deleting a Product also deletes all linked ProductNodes
pnodes = GenericRelation(ProductNode, related_query_name="product")
```

But not between Components and ComponentNodes:
```
# no related_query_name=, so ComponentNodes are left dangling when the linked Component is deleted
cnodes = GenericRelation(ComponentNode)
```

https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericRelation.related_query_name says at the very end:
> Note also, that if you delete an object that has a [GenericRelation](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericRelation), any objects which have a [GenericForeignKey](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey) pointing at it will be deleted as well. In the example above, this means that if a Bookmark object were deleted, any TaggedItem objects pointing at it would be deleted at the same time.

> Unlike [ForeignKey](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey), [GenericForeignKey](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey) does not accept an [on_delete](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ForeignKey.on_delete) argument to customize this behavior; if desired, you can avoid the cascade-deletion by not using [GenericRelation](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericRelation), and alternate behavior can be provided via the [pre_delete](https://docs.djangoproject.com/en/3.2/ref/signals/#django.db.models.signals.pre_delete) signal.

I recall reading somewhere else that (in addition to adding a GenericRelation) a related_query_name= is also required to cascade deletions. I don't think we need to change this behavior as part of this PR, but mentioning for awareness.

jasinner (approved):

PR #141, opened 2022-09-12
Remove dev-only endpoints

mprpic:
Coverage has been replaced by a management command, and relations is not a standalone entity worth exposing on its own endpoint. Relations are exposed on components themselves.


JimFuller-RedHat (approved):
looks exactly as discussed ... thx for cleaning up my mess!

PR #142, opened 2022-09-12
CORGI-225: Add to user guide documentation

mprpic:


juspence (approved):
I don't really get the change to serializers.py (using software_build.name instead of component.name). No objection necessarily, just confused.

We also need to update the link on the homepage to point to user_guide.md instead of tutorial.md - other than that, LGTM.

mprpic:
> I don't really get the change to serializers.py (using software_build.name instead of component.name). No objection necessarily, just confused.

Ah, I meant to submit that as a separate change. It fixes the URL for downloading RPMs from Brew, turns out the path uses the build's name not the actual RPM name.

> We also need to update the link on the homepage to point to user_guide.md instead of tutorial.md - other than that, LGTM.

I don't actually see a link to the tutorial anywhere on the home page. We only point to the general docs at https://github.com/RedHatProductSecurity/component-registry/tree/main/docs.

mprpic:
Re-submitted the download URL fix as https://github.com/RedHatProductSecurity/component-registry/pull/143

PR #145, opened 2022-09-13
CORGI-166: Fix DB constraint that allows duplicate data

juspence:
@RedHatProductSecurity/corgi-devs Looks like we have a DB constraint that got changed by mistake. We want all nodes to be unique on (type, parent, purl), and have to handle a special case when parent=None.

Postgres treats any row with a NULL value as being unique (correct behavior according to SQL spec) so inserting data like below would normally succeed:
```
(type=A, parent=None, purl=B)
(type=A, parent=None, purl=B)
```

To handle this case, we have a unique constraint on only type= and purl=, and this constraint should get used only when parent=None. So constraint should look at only below fields and block duplicate insertion:
```
(type=A, purl=B)
(type=A, purl=B)
```

The null parent field got added to this constraint by mistake, so this MR adds a test to confirm the correct behavior, then removes the parent= field from the constraint when it's null.


mprpic:
> The null parent field got added to this constrant by mistake, so this MR adds a test to confirm the correct behavior, then removes the parent= field from the constraint when it's null.

Why was it added in the first place though? @jasinner, do you remember the motivation behind adding https://github.com/RedHatProductSecurity/component-registry/commit/37f70af40b62b90c074ce91d2d0a7ac4249ae4df ? Is there a legitimate use case for having multiple root nodes that point to the same Component model?

The constraint fix here doesn't address the root cause of the duplicates though, that happens because the container image purl is not unique enough in that it doesn't include the release string. I'll push a fix for that into the #140 PR.

*edit*: purl uniqueness fix here: https://github.com/RedHatProductSecurity/component-registry/pull/140/commits/9f5a492e2b5f77bc78fefa73848b3e35f3e540b7

juspence:
> The constraint fix here doesn't address the root cause of the duplicates though, that happens because the container image purl is not unique enough in that it doesn't include the release string. I'll push a fix for that into the #140 PR.

Hmm...When I first added this change, I ran ```ComponentNode.objects.count()``` and compared that to ```ComponentNode.objects.distinct("type", "parent", "purl").count()```. Both reported the same number of nodes, so I assumed that we didn't have any duplicate nodes with the same type / parent / purl.

But I guess I forgot to filter for parent=None, then check total counts vs. distinct counts for only type + purl. So container images which are also root nodes (parent=None) should be the only case where we might have duplicates.

Or maybe everything in our DB was distinct when I ran that check, but we hadn't loaded enough data yet to hit this issue with duplicate container image purls for different components. Thanks for adding the fix to #140.

If we want to allow two root nodes with the same type + purl (for the same component), we should just remove the constraint entirely. Having a constraint where some values can be null, means that Postgres won't actually enforce the constraint. Any row with a NULL value will always be considered unique, even if all the other fields are identical to some other row.

Should we also add a unique constraint  for the Component model's purl field, whenever it's not an empty string? Currently we don't enforce uniqueness on our purls, so creating two different components with the same purl will succeed. But then creating nodes for one component might fail. The other component would definitely have nodes with the same purl, and maybe the same type + parent as well.

jasinner:
I changed it to fix CORGI-211. But looking at the logs in the description of that ticket and looking at the [commit](https://github.com/RedHatProductSecurity/component-registry/commit/9f5a492e2b5f77bc78fefa73848b3e35f3e540b7) Martin referenced to fix duplicate container image purls I think perhaps that issue was caused by the issue Martin fixed, not [my commit](https://github.com/RedHatProductSecurity/component-registry/pull/129/commits/37f70af40b62b90c074ce91d2d0a7ac4249ae4df) which broke the index. 

I'm OK with merging this, and trying to find the real root cause of CORGI-211, if it's not already fixed by Martin's commit.

jasinner (approved):

PR #147, opened 2022-09-14
Corgi 163 errata info

jasinner:
Part of fixining CORGI-163. Make sure channels are always up to date.

This also adds a et_product_version field to ProductStreams model which is pull from product_definitions errata_info field


corgi/collectors/errata_tool.py
```diff
@@ -158,41 +158,38 @@ def get_erratum_components(self, erratum_id: str):
...
 
             for variant in repo_rels["variants"]:
                 variant_name = variant["name"]
-                if variant_name not in variants_with_repos:
-                    logger.warning(
+                try:
+                    variant_obj = CollectorErrataProductVariant.objects.get(name=variant_name)
+                    variant_obj.repos.append(repo_attrs["name"])
+                except CollectorErrataProductVariant.DoesNotExist:
+                    logger.debug(
```
mprpic, line 177: Do we want this to be a debug statement only? I think it indicates some faulty data so should probably be reported as an error or a warning.

corgi/collectors/errata_tool.py
```diff
@@ -158,41 +158,38 @@ def get_erratum_components(self, erratum_id: str):
...
             repo_attrs = repo["attributes"]
             repo_rels = repo["relationships"]
 
             for variant in repo_rels["variants"]:
                 variant_name = variant["name"]
-                if variant_name not in variants_with_repos:
-                    logger.warning(
+                try:
+                    variant_obj = CollectorErrataProductVariant.objects.get(name=variant_name)
+                    variant_obj.repos.append(repo_attrs["name"])
```
mprpic: This can probably be moved outside of the try-except since it doesn't fall within the scope of the caught exception.

Also, if this runs twice, will it append the same set of repo names to the queried variant?

mprpic (approved):
LGTM!

One final note: if we ever want this to display the actual repo names (called content sets) served by Pulp, we'll have to reconcile the variants to those content sets. SDEngine does this to figure out the externally-facing repo names, but I don't think we have to worry about this just yet. Just figured I'd mention it.

corgi/collectors/errata_tool.py
jasinner: I will update it to warning.

corgi/collectors/errata_tool.py
jasinner: Ok, fixed this up in the latest version of the branch.

jasinner:
I made a small change to update_variant_repos tasks so that updated ProductVerison's associated ProductStreams also have their channels array updated by save_product_taxonomy call.

PR #149, opened 2022-09-15
CORGI-208: Exclude self as a source for itself

mprpic:
Looking at e.g.:

/api/v1/components?purl=pkg%3Asrpm%2Fredhat%2F389-ds-base%401.2.11.15-97.el6_10%3Farch%3Dsrc

It's an SRPM component but in "sources" it lists a link to itself:

```
"sources": [
    {
        "link": "/api/v1/components?purl=pkg%3Asrpm/redhat/389-ds-base%401.2.11.15-97.el6_10%3Farch%3Dsrc",
        "purl": "pkg:srpm/redhat/389-ds-base@1.2.11.15-97.el6_10?arch=src"
    }
],
```

The "sources" relationship is defined as listing components that a
component embeds or is a source for, so it listing itself is
feels incorrect.


corgi/core/models.py
```diff
@@ -978,7 +978,7 @@ def get_source(self) -> list:
         return list(
             ComponentNode.objects.get_queryset()  # type: ignore
             .filter(purl=self.purl)
-            .get_ancestors(include_self=True)
+            .get_ancestors(include_self=False)
```
juspence, line 1018: I think include_self=False is the default, so up to you whether you just remove the kwarg, or leave it here to be more explicit.

tests/test_brew_collector.py
```diff
@@ -611,13 +611,18 @@ def test_fetch_rpm_build(mock_sca):
     ]:
         assert package in provides
     assert len(srpm.get_upstreams()) == 1
```
juspence, line 613: get_upstreams() also uses get_ancestors(include_self=True). I wonder if that should be changed as well (but doesn't have to be in this PR necessarily).

juspence (approved):

corgi/core/models.py
mprpic: I'd prefer to keep it explicit.

tests/test_brew_collector.py
mprpic: That one looks a bit more tricky because it inherits that behavior from `get_roots()` and I'm a bit wary of changing that one. I'll file an issue to investigate this a bit more in-depth.

mprpic:
Test is failing because it's assuming values after the fix from #140, I'll just wait for that one to be merged and then rebase here.

PR #150, opened 2022-09-16
Do not fail a task when processing an unsupported build

mprpic:
It adds too much noise to the error emails and failed task counts and we can do nothing about it. If we need to pull a listing of builds that were being processed and are unsupported, we can always extract it from the logs.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -29,7 +29,12 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     else:
         logger.info("Fetching brew build with build_id: %s", build_id)
 
-    component = Brew().get_component_data(build_id)
+    try:
+        component = Brew().get_component_data(build_id)
+    except BrewBuildTypeNotSupported as exc:
+        logger.warning(str(exc))
+        return
+
     if not component:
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat (approved):
LGTM

corgi/tasks/brew.py
```diff
@@ -29,7 +29,12 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     else:
         logger.info("Fetching brew build with build_id: %s", build_id)
 
-    component = Brew().get_component_data(build_id)
+    try:
+        component = Brew().get_component_data(build_id)
+    except BrewBuildTypeNotSupported as exc:
+        logger.warning(str(exc))
```
juspence, line 35: I'd generally prefer logger.error but other than that nit, LGTM.

juspence (approved):

corgi/tasks/brew.py
mprpic: Why?

juspence: When logging an exception that got raised, higher log level feels more natural to me. But it's just a personal preference really.

PR #155, opened 2022-09-22
CORGI-163 pulp modules

jasinner:
This add setup_pulp_relations task to `loadproductdata` (and celery beat) and a new `loadpulpdata` management command. Now when you call `loadproductdata` management command CDN_REPO relations are setup for all rhel_modules and SRPMS in a pulp repository. In addition Collector models are created using the module metadata in pulp. 

When loading the builds in a stream using the `loadpulpdata` all builds from that stream in the relations table are processed using both `slow_fetch_brew_build`, and `fetch_modular_build` commands. The 'fetch_modular_build` command is required to unpack the Collector model relationships and populate the rhel_module RPM children relationships stored in the collector models in the core models.


corgi/core/models.py
```diff
@@ -255,7 +255,12 @@ def save_product_taxonomy(self):
         """update ('materialize') product taxonomy on all build components"""
         variant_ids = list(
             ProductComponentRelation.objects.filter(build_id=self.build_id)
-            .filter(type=ProductComponentRelation.Type.ERRATA)
+            .filter(
+                type__in=[
+                    ProductComponentRelation.Type.ERRATA,
+                    ProductComponentRelation.Type.CDN_REPO,
+                ]
```
juspence: Small nit: Since this list isn't being changed anywhere, it would be better to make it a tuple:
```suggestion
                type__in=(
                    ProductComponentRelation.Type.ERRATA,
                    ProductComponentRelation.Type.CDN_REPO,
                )
```

docs/developer_guide.md
```diff
@@ -62,6 +62,9 @@ export CORGI_DOMAIN
 export CORGI_EMAIL_HOST
 export CORGI_SERVER_EMAIL
 export CORGI_UMB_BROKER_URL
+export CORGI_PULP_URL
+export CORGI_PULP_USERNAME
+export CORGI_PULP_PASSWORD
```
juspence, line 66: Note to self: I'll need to update my venv once this gets merged.

corgi/collectors/models.py
```diff
@@ -43,21 +43,45 @@ def __str__(self):
...
     def __str__(self):
         return f"{self.nvra}"
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                name="unique_collectorrpms",
+                fields=("nvra", "srpm"),
+            ),
+        ]
```
juspence, line 71: I think this isn't needed since we already require NVRA to be unique by itself.

corgi/collectors/models.py
```diff
@@ -43,21 +43,45 @@ def __str__(self):
...
+
+    def __str__(self):
+        return f"{self.name}"
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                name="unique_collectorrpmrepos",
+                fields=("name", "content_set"),
+            ),
```
juspence: As above - since name is always unique, forcing name + content_set to be unique together is redundant.

corgi/collectors/brew.py
```diff
@@ -768,3 +770,110 @@ def brew_rpm_lookup(self, rpms) -> tuple:
...
+    def sans_epoch(self, srpm) -> str:
+        name, version, release = self.split_nvr(srpm)
+        version_parts = version.split(":")
+        if len(version_parts) > 1:
+            srpm = f"{name}-{version_parts[1]}-{release}"
+        return srpm
+
+    def module_key_to_nvr(self, module_key) -> str:
+        module_parts = module_key.split(":")
+        return f"{module_parts[0]}-{module_parts[1]}-{module_parts[2]}.{module_parts[3]}"
```
juspence: Could you add a header comment explaining why these are needed? I understand what the first one does, but I don't quite get the second one. Probably because I don't know what a module_key is or what module_parts it contains. A special kind of NVR used by modular RPMs?

corgi/tasks/pulp.py
```diff
@@ -0,0 +1,59 @@
...
+    ProductStream,
+    ProductVariant,
+)
+from corgi.tasks.brew import fetch_modular_builds
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS, _create_relations
+
+logger = logging.getLogger(__name__)
+
+
+def setup_pulp_relations() -> None:
```
juspence, line 33: This needs an ```@app.task()``` decorator if we want to run it automatically every day. Which I think we should, so that running loadproductdata manually behaves identically to what celery-beat runs automatically.

juspence (commented):
Still about half to go - lots of very good work here. I'll finish the rest tomorrow since I don't want to rush and miss something, plus I need to finish a few other things before I step out today.

Unrelated to the PR, I saw you requested me as a reviewer but I still didn't get an email notification. I guess GitHub only sends those if you tag @juspence in a comment....which is frustrating.

corgi/tasks/pulp.py
jasinner: That's a very good point, thanks. I added setup_pulp_relations to the beat tasks as well.

corgi/collectors/brew.py
jasinner: Will do.

corgi/collectors/models.py
jasinner: This was a suggestion from @JimFuller-RedHat 
I reached out to him for help because using get_or_create with an nvra for CollectorRPM model manager was throwing duplicate key errors, see [this comment](https://issues.redhat.com/browse/CORGI-163?focusedCommentId=20988378&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20988378) on the CORGI-163 task. I confirmed this suggestion of a UniqueConstraint fixed that issue.

corgi/core/models.py
jasinner: Ok, done. I also adjusted the stread_ids loop below to use a tuple.

jasinner:
Created a [follow up task](https://issues.redhat.com/browse/CORGI-257) for keeping modules in CDN_REPOs up to date: 

Created a MR  for ops repo. (component-registry-ops/-/merge_requests/55)

corgi/collectors/pulp.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def get_rpm_data(self, repo: str) -> Generator[str, None, None]:
+        rpms_by_srpm = self._get_rpm_data(repo)
+        yield from Brew().lookup_build_ids(rpms_by_srpm)
+
+    def _get_rpm_data(self, repo) -> defaultdict:
+        rpms_by_srpm = defaultdict(list)
+        page = 0
+        while True:
+            RPM_CRITERIA["criteria"]["skip"] = PAGE_SIZE * page
```
mprpic: You may want to modify a local copy of the global RPM_CRITERIA otherwise it'll persist across different runs of fetching this data.

corgi/collectors/migrations/0006_auto_20220922_1834.py
```diff
@@ -0,0 +1,25 @@
...
+                fields=("nvra", "srpm"), name="unique_collectorrpms"
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="collectorrpmrepository",
+            constraint=models.UniqueConstraint(
+                fields=("name", "content_set"), name="unique_collectorrpmrepos"
+            ),
+        ),
+    ]
```
mprpic: Maybe all three of these migrations can be squashed to just one? Or is there a reason to keep them separate?

corgi/collectors/models.py
juspence: Hmm, I don't get how a redundant UniqueConstraint would fix that issue. Maybe get_or_create() wasn't behaving atomically because we didn't have an index for both its keyword arguments (nvra and srpm), so it was trying to insert duplicate rows and failing on the unique nvra requirement.

In that case you should be able to just explicitly add an index on both nvra and srpm, instead of relying on UniqueConstraint (which implicitly creates an index on all the fields in the constraint IIUC). As far as I'm aware, though, if you have a unique constraint / index on a single field, get_or_create() should be atomic - you don't need the constraint / index on all fields.

Either way I'd like to have a better idea of why this bug happens before we merge this code. I don't want us to assume it's fixed, only to see it happen again later. For example, if it was triggered by some kind of race condition which just happened to not occur after the constraint changes, but which is still present in the code.

corgi/tasks/pulp.py
```diff
@@ -0,0 +1,63 @@
...
+def slow_setup_pulp_module_relations(channel, variant):
+    module_build_ids = Pulp().get_module_data(channel)
+    no_of_relations = _create_relations(
+        module_build_ids, channel, variant, ProductComponentRelation.Type.CDN_REPO
+    )
+    if no_of_relations > 0:
+        logger.info("Created %s new relations for rhel_modules in %s", no_of_relations, channel)
+
+
+def get_builds_by_cdn_repo(stream_names: list[str]):
```
juspence: It looks like this is only called when the management command is run manually. If that's the case, it might make more sense to move this into the loadpulpdata.py module. Or if it really is a Celery task, we should add the app.task() decorator here as well.

corgi/tasks/errata_tool.py
```diff
@@ -130,8 +135,18 @@ def update_variant_repos() -> None:
...
+                try:
+                    rpm_repo = CollectorRPMRepository.objects.get(name=repo)
+                except CollectorRPMRepository.DoesNotExist:
+                    logger.debug("Not creating Channel for inactive repo %s", repo)
+                    continue
+                repo, created = Channel.objects.update_or_create(
+                    name=repo,
+                    defaults={
+                        "type": Channel.Type.CDN_REPO,
+                        "meta_attr": {"content_set": rpm_repo.content_set},
```
juspence, line 145: This looks like a bug. rpm_repo.content_set can (I assume) change from time to time, but because it's inside the defaults= dict, the content_set will only be stored when the Channel object is first created.

Whenever we try to update the Channel object, we'll only update the keyword arguments (name, which I assume won't change) and everything else in the defaults dict will stay unchanged.

We also try to avoid creating a Channel for inactive repos, but it doesn't look like we remove the Channel if an active repo becomes inactive. I know nothing about Pulp - whether that's important or even possible. But it might be a case we want to handle. Should we add an active boolean to the Channel model, always create Channels even for inactive repos, then update that boolean using update_or_create()?

corgi/tasks/errata_tool.py
```diff
@@ -117,6 +119,9 @@ def update_variant_repos() -> None:
     CDN repos are saved as Channels and are linked to Variants as their children. Multiple Variants
     can link to the same CDN repo.
     """
+    logger.info("Getting active repositories from Pulp")
+    Pulp().get_active_repositories()
```
juspence: I wonder if calling this method here would make the update_variant_repos() task time out. Maybe it would be better to:
1) Instead of calling update_variant_repos(), call get_active_repositories() first as a Celery task
2) At the end of that Celery task, call update_variant_repos() next as a separate Celery task

corgi/tasks/common.py
```diff
@@ -57,3 +59,19 @@ def get_last_success_for_task(task_name):
     return (
         last_success - timedelta(minutes=30) if last_success else timezone.now() - timedelta(days=3)
     )
+
+
+def _create_relations(build_ids, external_system_id, product_ref, relation_type) -> int:
+    no_of_relations = 0
+    for build_id in build_ids:
+        if not build_id:
+            continue
```
juspence: Not clear on what this is doing - won't we always have a build_id?

corgi/tasks/brew.py
```diff
@@ -382,3 +421,12 @@ def load_brew_tags() -> None:
...
             logger.info("Saving %s new builds for %s", no_of_created, brew_tag)
+
+
+def fetch_modular_builds(relations_query):
+    build_ids: set[int] = set()
+    for build_id in relations_query:
+        build_id_int = int(build_id)
+        build_ids.add(build_id_int)
+        fetch_modular_build.delay(build_id_int)
+    return list(build_ids)
```
juspence: Not related to this MR specifically, but this could be cleaned up:
```suggestion
def fetch_modular_builds(relations_query):
    for build_id in relations_query:
        fetch_modular_build.delay(int(build_id))
    return relations_query
```
I thought at first this code had a bug, where we might call fetch_modular_build() multiple times if the same build_id appeared multiple times in relations query.

But actually, all the code I see in rhel_compose.py that calls fetch_modular_build() already uses ```.values_list("build_id", flat=True).distinct()```, so there's no need to convert the build_ids to a set.

We already get a distinct queryset of the build IDs, so we can just convert that to a list and iterate through it, instead of building a set which would be slower and take more memory. We could also save some more effort if whatever uses the list of build_ids we return here doesn't need to modify that list. In other words, either return a tuple, or iterate on and return the original queryset directly without modifying it.

corgi/tasks/pulp.py
```diff
@@ -0,0 +1,63 @@
...
+        for pv in ProductVariant.objects.filter(name__in=ps.product_variants):
+            relations_query = (
+                ProductComponentRelation.objects.filter(
+                    product_ref=pv.name,
+                    type=ProductComponentRelation.Type.CDN_REPO,
+                )
+                .values_list("build_id", flat=True)
+                .distinct()
+            )
+            build_ids.update(fetch_modular_builds(relations_query))
```
juspence: This looks like a bug - fetch_modular_builds() returns a list of ints, but the other build IDs here might be strings. I would recommend changing fetch_modular_builds() to just return the queryset of string build_ids, then update the fetch_modular_build() Celery task to take a string build ID as an argument, then convert the build_id to an int only in that task.

corgi/tasks/rhel_compose.py
```diff
@@ -33,65 +28,14 @@ def save_compose(stream_name) -> None:
             compose_url, variants
         )
         for key in "srpms", "rhel_modules":
-            no_of_relations += _create_relations(compose_data, key, compose_id, stream_name)
+            if key not in compose_data:
+                continue
```
juspence, line 34: Should we raise an exception or log an error in case some data we expect is missing? When does this happen and why?

corgi/collectors/models.py
JimFuller-RedHat: Other then the general - 'sometimes ORMs does surprising things with sql' (thankfully rare with django)... I do not have any good explanation - as this is a collector model I think the risk is small - having no free cycles to dig into a root cause ... might be worth raising a task onto the backlog if/when this does arise in the future.

juspence: Rereading CORGI-163, just the unique=True constraint on nvra should be enough to prevent duplicates from being inserted. There's really no need for a separate UniqueConstraint on both nvra + srpm together.

I assume when you saw the errors with duplicate keys, you added both the unique=True condition on nvra and the UniqueConstraint on (nvra, srpm) as mentioned in Jim's comment. They're both doing the same job and only one of them is needed - let's either force nvra to be unique by itself, or force nvra + srpm to be unique together, but not apply uniqueness checks to both nvra alone and nvra + srpm.

corgi/collectors/migrations/0006_auto_20220922_1834.py
jasinner: OK Done.

corgi/collectors/pulp.py
jasinner: Ok, will do, thanks.

corgi/tasks/pulp.py
jasinner: OK, will do.

corgi/tasks/errata_tool.py
jasinner: That's not how I understand update_or_create. Defaults are used to update the fetched object, which is what we want in this case. See https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create it says "The defaults is a dictionary of (field, value) pairs used to update the object.". If the conent_set is updated we want to set that on the named repo. I don't think this will ever happen, but if it does the code here should update the conent_set field using the value passed in the defaults dict.

Inactive in this case mean it has RPMs shipped to it. I don't think there is a case when a Repo would have RPMs removed from it, so it won't become inactive.

corgi/tasks/errata_tool.py
jasinner: It's a good idea. I added a new task to pulp called update_cdn_repo_channels() as the new entrypoint for both calling 'get_active_repositories' as a task and also calling the errata task update_variant_repos()

corgi/tasks/common.py
jasinner: I found that in some cases the collectors.pulp.Pulp().get_module_data or collectors.brew.Brew.persist_modules() generators where returning None values, so this is to a void trying to persist those as relations which causes a DB Error.

corgi/tasks/pulp.py
jasinner: fetch_modular_builds is also called by loadcomposedata module. We're still left with the cast to int there, but I think that's better than the alternative, which is to change the build_id parameter of slow_fetch_brew_build also to a str.

I have updated this as you suggested though.

corgi/tasks/rhel_compose.py
jasinner: It's expected in some cases that "rhel_modules" key does not exist in compose_data dict returned by RhelCompose.fetch_compose_data, that's all this is checking for. Most composes don't have modules, in fact I think only the appstream compose in RHEL 8 has modules. But that may change in the future.

corgi/tasks/brew.py
jasinner: I updated this as you suggested and return the original queryset as well.

corgi/collectors/models.py
jasinner: I removed unique=True from the nvra field and kept the uniqueconstraint on nvra + srpm. I also have to updated the CollectorRPM.get_or_create() call in Brew().persist_modules to include srpm in the fields instead of in default values.

corgi/tasks/pulp.py
jasinner: I also had to move setup_pulp_relations() call to the end of upload_variant_repos. That will mean it's called daily which is OK for now, but once we complete CORGI-257 we can probably just call it manually after updateproductdata once when we populate the database from scratch. In CORGI-257 we'll query the cdn_repo on demand and create collector models and entry in the relations table each time we process an erratum.

corgi/tasks/brew.py
jasinner: I actually made another improvement here. What was happening with the original code here was the fetch_modular_build was being called with all build_ids and then only slow_fetch_brew_build was being called, which meant that modular builds where 'half baked' until the whole stream was processed. I improved it by calling slow_fetch_brew_brew after fetch_modular_build for each build_id. That also means I don't need to return anything from fetch_modular_builds function.

corgi/tasks/pulp.py
jasinner: Now I just cast the build_id to an int when calling slow_fetch_brew_ build from fetch_modular_builds.

corgi/tasks/errata_tool.py
juspence: Whoops, you're right - I was misremembering the behavior as the other way around (only update kwargs). And re: inactive, sounds good - thanks for explaining.

corgi/tasks/rhel_compose.py
juspence: Maybe a comment here to explain that would be good, otherwise sounds good to me.

corgi/tasks/common.py
juspence: Noting that here might be good - thanks for explaining.

corgi/tasks/pulp.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    if no_of_relations > 0:
+        logger.info("Created %s new relations for rhel_modules in %s", no_of_relations, channel)
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def update_cdn_repo_channels() -> int:
+    logger.info("Getting active repositories from Pulp")
+    no_of_created_repos = Pulp().get_active_repositories()
+    logger.info("Created %s new active CDN repositories")
```
juspence: Looks like this is missing the format-string argument (and sorry for not catching it earlier):
```suggestion
    logger.info("Created %s new active CDN repositories", no_of_created_repos)
```

corgi/collectors/models.py
juspence: We should remove the UniqueConstraint here and rely on name= alone to be unique, to match the behavior in Pulp().get_active_repositories():
```
_, created = CollectorRPMRepository.objects.update_or_create(
                name=repo["id"], defaults={"content_set": content_set}
            )
```
That lookup means that name= should be enough on its own to uniquely identify the repo object, and we then want to update its content_set, so I don't think name and content_set need to appear together in a UniqueConstraint.

corgi/tasks/management/commands/loadpulpdata.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+        parser.add_argument(
+            "-f",
+            "--force",
+            action="store_true",
+            help="Force ingestion even if the Build exists",
+        )
+
+    def handle(self, *args, **options) -> None:
+        if options["stream_name"]:
```
juspence, line 26: To avoid build_ids being undefined when stream_name isn't given:
```suggestion
        build_ids = None
        if options["stream_name"]:
```

corgi/tasks/management/commands/loadpulpdata.py
```diff
@@ -0,0 +1,68 @@
...
+                f"Fetching component data for builds: {', '.join(map(str, build_ids))}"
+            )
+        )
+
+        for build_id in build_ids:
+            build_id = int(build_id)
+            if options["inline"]:
+                slow_fetch_brew_build(build_id, force_process=options["force"])
+            else:
+                slow_fetch_brew_build.delay(build_id, force_process=options["force"])
```
juspence: This passes a string build_id from get_builds_by_cdn_repo() / fetch_modular_builds(), but slow_fetch_brew_build() expects an int build_id.
```suggestion
            if options["inline"]:
                slow_fetch_brew_build(int(build_id), force_process=options["force"])
            else:
                slow_fetch_brew_build.delay(int(build_id), force_process=options["force"])
```

corgi/collectors/brew.py
```diff
@@ -760,11 +762,123 @@ def get_builds_with_tag(
...
+        }
+        nested_builds: set[int] = set()
+        rpm_components: list[dict] = []
+        for rpm in rhel_module.collectorrpm_set.all():
+            srpm_build_id = rpm.srpm.build_id
+            nested_builds.add(srpm_build_id)
+            name, version, release = Brew.split_nvr(rpm.nvra)
+            release_split = release.rsplit(".", 1)
+            if len(release_split) == 2:
+                arch = release_split[1]
```
juspence, line 869: To avoid arch being undefined when len(release_split) != 2:
```suggestion
            if len(release_split) == 2:
                arch = release_split[1]
            else:
                arch = None  # Or should we default to an empty "" string, or "noarch"?
```

corgi/collectors/pulp.py
```diff
@@ -0,0 +1,108 @@
...
+    def get_rpm_data(self, repo: str) -> Generator[str, None, None]:
+        rpms_by_srpm = self._get_rpm_data(repo)
+        yield from Brew().lookup_build_ids(rpms_by_srpm)
+
+    def _get_rpm_data(self, repo) -> defaultdict:
+        rpms_by_srpm = defaultdict(list)
+        page = 0
+        while True:
+            rpm_criteria = copy.deepcopy(RPM_CRITERIA)
+            rpm_criteria["criteria"]["skip"] = PAGE_SIZE * page
```
juspence: To avoid deep-copying a dict on every loop iteration:
```suggestion
        page = 0
        rpm_criteria = copy.deepcopy(RPM_CRITERIA)
        while True:
            rpm_criteria["criteria"]["skip"] = PAGE_SIZE * page
```

corgi/collectors/pulp.py
```diff
@@ -0,0 +1,108 @@
...
+            page += 1
+            rpm_data = self._get_unit_data(repo, rpm_criteria)
+            for entry in rpm_data:
+                entry_meta = entry.get("metadata")
+                filename = entry_meta.get("filename")
+                source_rpm = entry_meta.get("sourcerpm")
+                rpms_by_srpm[source_rpm.removesuffix(".rpm")].append(filename)
```
juspence: Are we using .get() here because these keys might not be present in the entry dict? If yes, we need to handle the case when entry_meta or source_rpm are None. With the current code, ```entry_meta.get("filename or sourcerpm")``` and ```source_rpm.removesuffix(".rpm")``` will fail when if either of those is None.

Or if not, and we expect those values to always be present:
```suggestion
            for entry in rpm_data:
                filename = entry["metadata"]["filename"]
                source_rpm = entry["metadata"]["sourcerpm"].removesuffix(".rpm")
                rpms_by_srpm[source_rpm].append(filename)
```

corgi/collectors/pulp.py
```diff
@@ -0,0 +1,108 @@
...
+        rpms_by_module = {}
+        for entry in module_data:
+            entry_meta = entry.get("metadata")
+            name = entry_meta.get("name")
+            # eg. build_id 1500134
+            stream = entry_meta.get("stream").replace("-", "_")
+            version = entry_meta.get("version")
+            context = entry_meta.get("context")
+            module_key = f"{name}-{stream}-{version}.{context}"
+            rpms_by_module[module_key] = entry_meta.get("artifacts")
```
juspence: See other comment - let's either handle the case when entry_meta / name / stream / version / context / artifacts is missing, or if they're always present, just directly access dict["key"] instead of using dict.get("key").

juspence (approved):

corgi/collectors/brew.py
```diff
@@ -760,11 +762,123 @@ def get_builds_with_tag(
...
+                    rpm_data = self.koji_session.getRPM(filename)
+                    if not rpm_data:
+                        # Try the next srpm rpm filename
+                        continue
+                    build_id = rpm_data["build_id"]
+                    # found the build_id, stop iterating filenames
+                    break
+                # if no filenames had RPM data
+                if not build_id:
+                    logger.warning("Unable to find build_id for %s when saving %s", srpm, build_id)
```
juspence: @jasinner This is probably the case you mentioned about generators returning None instead of a build_id. Maybe we should raise an exception here instead of logging a warning. Or we could ```continue``` to the next (srpm, call) in find_build_id_calls instead of returning None.

corgi/collectors/models.py
jasinner: sorry I missed this one on the last pass. Good idea, I will change as you suggest.

corgi/tasks/pulp.py
jasinner: ok, thanks.

corgi/tasks/management/commands/loadpulpdata.py
jasinner: sorry, it looks like I had some local changes which I failed to commit last night, will do it soon.

corgi/tasks/management/commands/loadpulpdata.py
jasinner: These lines are removed in my uncommitted changes.

corgi/collectors/brew.py
jasinner: defaulted arch to empty "", as the brew task does as well.

corgi/collectors/pulp.py
jasinner: good idea!

corgi/collectors/pulp.py
jasinner: I don't expect them to be None, so using your suggestion.

corgi/collectors/pulp.py
jasinner: ok

corgi/collectors/brew.py
jasinner: yeah, let's continue.

JimFuller-RedHat:
LGTM

PR #156, opened 2022-09-26
Minor: minor perf test with new indexes in stage

JimFuller-RedHat:
This week I will trickle a few optimisations I have collected over the past few weeks ... culminating in a custom django migration. This specific commit has very minor code change which will take advantage of some test indexes currently set on stage which we want to measure there first.


jasinner:
LGTM

jasinner (approved):

PR #157, opened 2022-09-27
CORGI-223:  re-enable get-upstreams on REST API as perf has improved

JimFuller-RedHat:
Verified that get_upstreams() no longer suffers from perf degradation - this is due to indexes being added and [work elsewhere](https://gitlab.corp.redhat.com/product-security/dev/component-registry-ops/-/merge_requests/56) improving django-mptt performance.


mprpic (approved):
:+1:

PR #158, opened 2022-09-27
CORGI-256: Fix scanner warnings

juspence:
@RedHatProductSecurity/corgi-devs Please review. This resolves the most important outstanding issues from the SonarQube scan that Judy Kelly performed.

You can run a SonarQube scan locally in a container using the following steps:
```
podman pull docker.io/library/sonarqube
podman run -d --name sonarqube -p 9000:9000 -p 9092:9092 sonarqube
podman logs -f sonarqube
# Wait for a line like
# 2021.09.10 11:15:32 INFO  app[][o.s.a.SchedulerImpl] SonarQube is up
# Browse to localhost:9000
# Login with username "admin" and password "admin"
# Set up a manual project, use “component-registry” as name
# Choose to analyze locally using Python on Linux
# Follow the directions to download and install the scanner
# https://docs.sonarqube.org/latest/analysis/scan/sonarscanner/
# https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-4.7.0.2747-linux.zip
# A command is displayed, the important bit is the authentication token
# From the top directory of the component-registry repository, run:
sonar-scanner \
  -Dsonar.projectKey=component-registry \
  -Dsonar.sources=. \
  -Dsonar.host.url=http://localhost:9000 \
  -Dsonar.login=pre_generated_token
  -Dsonar.python.version=3
# The page will automatically refresh when the scan completes
```

You'll need to add "CORGI_TEST_PULP_URL" to your venv after this is merged - see the Gitlab CI variable for the value to use.
I'm going to try to add SonarQube directly into our Gitlab CI pipeline (in a second PR), but in the meantime the fixes here are ready for review.

PR #159, opened 2022-09-28
CORGI-256: Add sonarqube integration

juspence:
@RedHatProductSecurity/corgi-devs Please review this MR to add SonarQube scanning to the CI pipeline. #158 must be merged first to fix the outstanding issues, and once that's done we can merge this MR to turn on the scanning.

The internal SonarQube server has the results, under the URL "dashboard?id=component-registry&branch=add-sonarqube-integration", but doesn't report the coverage correctly even though this works locally. Fixing this would require running the sonarqube job after the test job that generates the coverage.xml file (either in a new stage or as a dependent job in the same stage).

This slows down the pipeline quite a bit and we don't really use the coverage report anyway, so going to ignore this for now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,9 @@ include:
   - project: 'product-security/dev/component-registry-ops'
     ref: "$CORGI_OPS_BRANCH"
     file: '/templates/gitlab/ansible-run.yml'
+  - project: 'product-security/dev/component-registry-ops'
+    ref: "$CORGI_OPS_BRANCH"
+    file: '/templates/gitlab/sonarqube.yml'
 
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat:
LGTM, question - are we running this with every commit to stage or maybe we could consider a weekly ?

juspence:
> LGTM, question - are we running this with every commit to stage or maybe we could consider a weekly ?

It takes about 30s - 1m to return results, so not much longer than our existing linters. And it seems more useful to prevent security issues from being introduced, i.e. every commit, instead of a weekly check that might create failures for us to then go clean up.

mprpic (approved):

PR #162, opened 2022-10-01
skip sca of kernel

jasinner:
I'm very confident we don't need to scan the kernel. It's all source code, there are no embedded components I am aware of. It's eating up lots of resources in stage, see CORGI-270.


jasinner:
@JimFuller-RedHat

JimFuller-RedHat:
sounds reasonable, LGTM

PR #170, opened 2022-10-03
CORGI-261 - custom django migration with indexes specifically for opt…

JimFuller-RedHat:
set of indexes for optimising pg

* django mptt specific (specifically partial index helps things along)
* pcr index 

**Note** - before this PR gets merged we will need to remove test indexes on stage


corgi/core/migrations/0037_custom_indexes.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0036_productcomponentrelation_core_produc_product_c88392_idx"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "CREATE INDEX core_componentnode_tree_parent_lft_idx ON public.core_componentnode USING btree(tree_id, parent_id, lft)"  # noqa: E501
```
mprpic: Any reason this can't be broken onto two lines to not need to linter exclusion comment?

JimFuller-RedHat: sure np

juspence: If we create indexes like this (using direct SQL statements), there's very little in the models code that tells us this index exists. So it could get accidentally removed / changed in the future with another direct SQL statement, for example if someone runs something in stage for debugging. I'd prefer creating the index in models.py like below:
```
        indexes = [
            models.Index(fields=("tree_id", "parent_id", "lft")),
            models.Index(fields=("type", "parent", "purl")),
```
Testing locally now but I think this should work the same, and be more maintainable.

JimFuller-RedHat: generally agree - though these in particular (right now) are something we probably want to isolate apart from the models ... as they are the result of pure performance analysis ... once we have run for a bit we can port back into the model.

juspence: This doesn't actually work anyway since MPTT generates the field names dynamically, so Django can't create indexes on these fields before the app registry is fully loaded.

mprpic: Resolving this after some discussion on chat: these raw indexes touch fields that are dynamically generated by the models so the in-Django model indexes won't work (unfortunately). Keeping this as is.

juspence (approved):

PR #171, opened 2022-10-03
Revert enabled django-debug-toolbar and debug=true on stage

mprpic:
These are debugging tools that should *never* run in any other environment but a local one.


juspence (approved):
LGTM. Not sure if more is needed anywhere else, but at least this much is fine to start with.
EDIT: Reviewed recent PRs, this is everything. Thanks for fixing!

PR #173, opened 2022-10-03
CORGI-229 component api filter for products is too inclusive.

jasinner:
This ensures that when we filter components by products in the api we only get exact matches and not matches for substrings. For example a query for the product_stream rhel-8.6.0 was including results for rhel-8.6.0, and rhel-8.6.0.z.


JimFuller-RedHat:
LGTM

PR #179, opened 2022-10-10
Fix ofuri queries and format for product variant

jasinner:
The `fetch_unprocessed_cdn_relations` weekly scheduled task failed over the weekend. This attempts to fix it for future runs.

Also fixes 2 other bugs related to ofuri:

- CORGI-292
- CORGI-293


jasinner:
In order to update all the ProductVariant object instances ofuri we need to run the following:
```
import corgi.core.models as m

for pv in m.ProductVariant.objects.all():
     pv.save()
```

JimFuller-RedHat:
we may have to contemplate updating product taxonomies on variants (to populate missing product_variants)

PR #180, opened 2022-10-10
CORGI-289: Load composes periodically

juspence:
@RedHatProductSecurity/corgi-devs This change should cause us to automatically load all compose builds every day, if they don't already exist. I don't know how much processing this involves, after the existing data is loaded and we only have to ingest new builds.

Maybe it's better to only check for / load new builds once a week? Or maybe we'd prefer to keep our data more current (no more than 24 hours delay / staleness) and it's better to run the task daily.

### DIFF
--- a/corgi/tasks/monitoring.py
+++ b/corgi/tasks/monitoring.py
@@ -57,8 +57,9 @@ def setup_periodic_tasks(sender, **kwargs):
     # task at all: https://github.com/celery/django-celery-beat/issues/221
     upsert_cron_task("errata_tool", "load_et_products", hour=0, minute=0)
     upsert_cron_task("prod_defs", "update_products", hour=1, minute=0)
-    upsert_cron_task("pulp", "update_cdn_repo_channels", hour=1, minute=30)
+    upsert_cron_task("pulp", "update_cdn_repo_channels", hour=2, minute=0)
     upsert_cron_task("rhel_compose", "save_composes", hour=3, minute=0)
+    upsert_cron_task("rhel_compose", "get_builds", hour=4, minute=0)
     upsert_cron_task("monitoring", "email_failed_tasks", hour=10, minute=45)
 
     # Automatic task result expiration is currently disabled
--- a/corgi/tasks/rhel_compose.py
+++ b/corgi/tasks/rhel_compose.py
@@ -38,36 +38,16 @@ def save_compose(stream_name) -> None:
     logger.info("Created %s new relations for stream %s", no_of_relations, stream_name)
 
 
-def get_builds_by_compose(compose_names):
-    relations_query = (
-        ProductComponentRelation.objects.filter(
-            external_system_id__in=compose_names,
-            type=ProductComponentRelation.Type.COMPOSE,
-        )
-        .values_list("build_id", flat=True)
-        .distinct()
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def get_builds(compose_names: list[str] = None, stream_name: str = None) -> None:
+    """Get compose build IDs, optionally for only a particular stream or set of composes"""
+    relations_query = ProductComponentRelation.objects.filter(
+        type=ProductComponentRelation.Type.COMPOSE
     )
-    fetch_modular_builds(relations_query)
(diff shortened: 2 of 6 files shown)


corgi/tasks/monitoring.py
```diff
@@ -57,8 +57,9 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
     # task at all: https://github.com/celery/django-celery-beat/issues/221
     upsert_cron_task("errata_tool", "load_et_products", hour=0, minute=0)
     upsert_cron_task("prod_defs", "update_products", hour=1, minute=0)
-    upsert_cron_task("pulp", "update_cdn_repo_channels", hour=1, minute=30)
-    upsert_cron_task("rhel_compose", "save_composes", hour=3, minute=0)
+    upsert_cron_task("pulp", "update_cdn_repo_channels", hour=2, minute=0)
+    upsert_cron_task("rhel_compose", "get_builds", hour=3, minute=0)
```
jasinner: I think we should move this to after the save_composes call so that we can process any new builds which were discovered by that task and saved as relations?

corgi/tasks/monitoring.py
```diff
@@ -57,8 +57,9 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
...
+    upsert_cron_task("rhel_compose", "save_composes", hour=4, minute=0)
```
jasinner: I think we should probably update this task to look for composes which are not already in the relations table I think we store the compose name in external_system_id field there. Once a compose is released it doesn't change, so we only need to save new composes.

We could do this as a follow up task if you'd prefer to get this change in first.

corgi/tasks/rhel_compose.py
```diff
@@ -38,36 +38,16 @@ def save_compose(stream_name) -> None:
...
-def get_all_builds():
-    relations_query = (
-        ProductComponentRelation.objects.filter(
-            type=ProductComponentRelation.Type.COMPOSE,
-        )
-        .values_list("build_id", flat=True)
-        .distinct()
-    )
-    fetch_modular_builds(relations_query)
+    relations_build_ids = relations_query.values_list("build_id", flat=True).distinct()
```
jasinner, line 52: Everywhere I see the use of distinct I think we should add an order_by call as well. I've found it's necessary for distinct to work correctly with Django and Postgres.

Perhaps we should add an updated filter here as well and only update composes build which might have changed since the last run? See for example the slow_fetch_unprocess_cdn_relatiions task.

```
        created_during = timezone.now() - timedelta(days=8)
        relations_query = relations_query.filter(created_at__gte=created_during)
```

jasinner (approved):
I added some ideas for improvement. But I'm happy for this to be merged as is if you prefer. All the things I suggested are optional to improve the efficience of these 2 tasks.

corgi/tasks/monitoring.py
juspence: Thanks, will fix! Agree we should process as many builds as possible - I made my best guess at the dependencies here but wasn't sure what the right order was.

corgi/tasks/monitoring.py
juspence: Yeah, I'll file a follow-up task for this. I want this change to just be "automate what we already run manually", and then we can make improvements to the code after it runs for a while, if nothing shows up in the daily monitoring email.

EDIT: Filed CORGI-296 for this feature.

corgi/tasks/rhel_compose.py
juspence: Hmm, that's actually a really big gotcha I wasn't aware of. I've tested e.g. ```.values_list("some_field", flat=True).distinct()``` and it seemed to work OK. But that was for very little data, so maybe a non-trivial number of records in the DB runs into that ordering bug.

I'll file a follow-up bug - can you send me a link to any docs you know of describing this behavior? Here or on Gchat or in the ticket is fine - or if you haven't seen mentioned in docs that's fine too.

EDIT: Here's an example when you use .distinct("some_field") to only consider distinctness of certain fields:
https://stackoverflow.com/questions/21621879/django-distinct-and-order-by
But I don't see anything about how .distinct() with no fields specified behaves. I filed CORGI-295 for this issue.

PR #182, opened 2022-10-12
CORGI-276: Add copyright_text, license_concluded logic and fix Django warning

juspence:
@jasinner Quick MR to allow OpenLCS to beta-test the scanning feature in stage. We won't turn this on in production until we've enabled authentication and had a more thorough requirements discussion (e.g. the endpoint name and design will likely change).

Going to let this sit overight, add a quick test tomorrow morning and then merge. I've tested manually.


JimFuller-RedHat:
Why are we not using the natural extension points to the logical data model and use tags to represent most/all of this data ? 

From this list:

- "copyright_text",
- "license_concluded",
- "license_concluded_list",
- "license_declared",
- "license_declared_list",
- "openlcs_scan_url",
- "openlcs_scan_version",

copyright_text is the only one that might be germane to the component model ... the rest look like the result of processing from external systems and likely to change (maybe in both structure and values). One reason to isolate 'things that change together' is to avoid disturbing the generic data model constantly which generates churn for all applications/ux downstream.

I might be missing something but I always see downsides adding a bunch of textfields to a model (with no constraints, etc) ... if tags are not up to the task (or to limited) we could entertain adding a generic meta model (like osidb) and give it an 'openlcs' type with whatever structure we deem fit (and could change independently of main component model).

The decisions are always the same with databases eg. when should we apply 3NF (eg. represent with a different table), use extensibility (simple  = tags, more complex = generic meta) or use a external system id. Of these three I think we already have things covered with tagging ... or we can improve tagging to accommodate.

In summary - small efforts now to help keep the data model extensible (and loosely coupled) will pay off big later on.

juspence:
> Why are we not using the natural extension points to the logical data model and use tags to represent most/all of this data ?

After a discussion on Gchat, our options were either "change tags to a textfield" or keep them limited in length. We prefer limiting the tag length so that people don't start storing full text documents in them. Anything too complicated to fit in a tag should probably get its own model field, or if not we can always increase the tag length or make them a text field in the future.

We also think that the license data will only change rarely once scanned, so is worth storing in its own model field. As you note below, this doesn't have to be on the Component model - we could have a dedicated model for OpenLCS metadata.
 
> From this list:
> 
>     * "copyright_text",
>     * "license_concluded",
>     * "license_concluded_list",

The above are model properties since they're reported in the SPDX manifest for each component, and are unlikely to change once they've been uploaded / set once.

>     * "license_declared",
>     * "license_declared_list",

These two already exist (and aren't related to OpenLCS), they've just been renamed to distinguish them from the license scan results that get uploaded.

>     * "openlcs_scan_url",
>     * "openlcs_scan_version",

CORGI-276 didn't mention these, but there were additional requirements requested in the meeting and in the original CORGI-56 story.

> copyright_text is the only one that might be germane to the component model ... the rest look like the result of processing from external systems and likely to change (maybe in both structure and values). One reason to isolate 'things that change together' is to avoid disturbing the generic data model constantly which generates churn for all applications/ux downstream.
> 
> I might be missing something but I always see downsides adding a bunch of textfields to a model (with no constraints, etc) ... if tags are not up to the task (or to limited) we could entertain adding a generic meta model (like osidb) and give it an 'openlcs' type with whatever structure we deem fit (and could change independently of main component model).

This is designed to be a simple beta test - we fully expect to change the implementation in the future (endpoint name and structure / logic, model field types and constraints, etc.). What that looks like will be the outcome of a more thorough requirements discussion, after the OpenLCS team has beta-tested for a bit.

> The decisions are always the same with databases eg. when should we apply 3NF (eg. represent with a different table), use extensibility (simple = tags, more complex = generic meta) or use a external system id. Of these three I think we already have things covered with tagging ... or we can improve tagging to accommodate.
> 
> In summary - small efforts now to help keep the data model extensible (and loosely coupled) will pay off big later on.
> 
> Note - my comments are about the logical model ... we can choose to display this in the conceptual model (aka REST API) however way we want.
> 
> Lastly - if I am not mistaken I believe openlcs can use tagging feature today if they want to test things.

The tagging feature isn't usable because there are no endpoints that allow writing the tags. Writing a generic tagging endpoint that works correctly for all cases was a little more complicated than adding a dedicated, temporary endpoint and fields just for the OpenLCS testing. The olcs_test endpoint is also easier to deprecate or change in the future, which will be likely given current priorities. Again, this is all temporary just to unblock the OpenLCS team. They know it will change in the future and I'm not proposing this as a final implementation.

JimFuller-RedHat:
thx for explanation - agree we do not want to save other ppl's documents (eg. document URIs are fine)

corgi/api/views.py
```diff
@@ -400,6 +401,47 @@ def list(self, request, *args, **kwargs):
...
+        openlcs_scan_url = request.data.get("openlcs_scan_url")
+        openlcs_scan_version = request.data.get("openlcs_scan_version")
+        if (
+            not copyright_text
+            and not license_concluded
+            and not openlcs_scan_url
+            and not openlcs_scan_version
+        ):
+            # At least one of above is required, else Bad Request
+            return Response(status=400)
```
mprpic, line 428: I think this should be a 200 actually. It's essentially a PUT that changes nothing, but that is not an invalid operation, simply a noop one. It's analogous to a PUT that submits the text that is already in a particular field.

mprpic (approved):
LGTM!

corgi/api/views.py
juspence: I'll probably keep this unchanged for now given time constraints, and since I already told the OpenLCS team at least one of these fields is required. I also think it's good for a beta-test, to explicitly help the user realize "none of the fields you submitted were recognized", e.g. if there was a typo in the JSON data / field name they submitted.

We can definitely change it for the final implementation though.

PR #183, opened 2022-10-12
CORGI-222 allow null value for cg_name field image build

jasinner:
This fixes CORGI-222 where 3 image builds from ansible_automation_platform-1.2 where not being fetched from brew.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -683,10 +683,18 @@ class Brew:
         #  identified in a separate attribute on the build itself.
         if build_type == self.CONTAINER_BUILD_TYPE:
             # If this is an "image" type build, it may be building a container image, ISO image,
-            # or other types of images. Check the content generator name to determine where this
-            # image was built, which indicates what type of image it is.
-            if build["cg_name"] != "atomic-reactor":
+            # or other types of images.
+            build_extra = build.get("extra")
+            if build["cg_name"] == "atomic-reactor":
+                # Check the content generator name to determine where this
+                # image was built, which indicates what type of image it is.
                 # Container images are built in OSBS, which uses atomic-reactor to build them.
+                pass
+            elif build_extra and build_extra.get("submitter") == "osbs":
+                # Some builds such as 903565 have the cg_name field set to None
+                # In that case check the extra/submitter field for osbs value
+                pass
+            else:
                 raise BrewBuildTypeNotSupported(
                     f"Image build {build_id} is not supported: "
                     f"{build['cg_name']} content generator used"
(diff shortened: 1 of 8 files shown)


scripts/reconciliation.py
```diff
@@ -42,26 +42,26 @@
...
-    ["o:redhat:rhacm:2.7", "2395"],  # GA 1.0
-    ["o:redhat:rhel:8.4.0.z", "35"],  # GA 1.0
-    ["o:redhat:rhel:8.6.0.z", "2302"],  # GA 1.0
-    ["o:redhat:rhel:8.7.0", "2301"],  # GA 1.0
-    ["o:redhat:rhel:9.0.0.z", "2307"],  # GA 1.0
-    ["o:redhat:rhel:9.1.0", "2308"],  # GA 1.0
-    ["o:redhat:rhn_satellite:6.7", "93"],  # GA 1.0
-    ["o:redhat:rhn_satellite:6.8", "50"],  # GA 1.0
-    ["o:redhat:rhn_satellite:6.9", "72"],  # GA 1.0
+    # ["o:redhat:ansible_automation_platform:2.0", "2030"],  # GA 1.0
```
JimFuller-RedHat, line 45: I do not think you meant to commit this change

corgi/collectors/brew.py
```diff
@@ -684,7 +684,10 @@ def get_component_data(self, build_id: int) -> dict:
             # If this is an "image" type build, it may be building a container image, ISO image,
             # or other types of images. Check the content generator name to determine where this
             # image was built, which indicates what type of image it is.
-            if build["cg_name"] != "atomic-reactor":
+            # Some builds such as 903565 have the cg_name field set to None
+            if build["cg_name"] is None or build["cg_name"] == "atomic-reactor":
```
JimFuller-RedHat, line 688: question: is an iso image a kind of component we want to represent someday ?

mprpic: @jasinner I'm not sure this is actually a good idea. Some old stuff might have not set the content generator value but will probably not be a container image build. Maybe we should refactor this to check if the build has a `buildContainer` task? That would be a good indicator whether its a container image build or some random ISO image build. Wdyt?

jasinner: I change the check from `build["cg_name"] == None` to `build["extra"]["submitter"] == "osbs"` instead. @mprpic can you resolve this conversation if you think that is OK.

jasinner: I had a quick look at some of the iso images in brew, but it's wasn't immediately clear how we could ingest them based solely on the data in Brew, so I'm going to punt that to a future (yet to be specified) epic.

scripts/reconciliation.py
jasinner: Thanks, I removed changes to the reconciliation script from this PR.

JimFuller-RedHat:
LGTM

PR #184, opened 2022-10-13
CORGI-136/CORGI-30 add latest mode and fix links corgi-269/corgi-292

JimFuller-RedHat:
This PR provides a 'latest' mode that narrows down components by ofuri to be those root components which were the last built version (and actually shipped) of components. 

For example to see 'latest' for ansible_automation_platform_2.1:
```
http://localhost:8008/api/v1/components?ofuri=o%3Aredhat%3Aansible_automation_platform%3A2.1&view=summary&mode=latest
```
or 'latest' for rhel 8.7.0:
```
http://localhost:8008/api/v1/components?ofuri=o%3Aredhat%3Arhel%3A8.7.0.z&view=summary&mode=latest
```

this concept of 'latest' only makes sense in terms of a specific product/offering eg. the latest built root component specifically for this offering/product.

* surfaced up completion_time on SoftwareBuild which is needed to determine latest component 
* while in the area - fixed SoftwareBuild .components to be true foreign key relationship (which already existed)
* fixed broken link generation (CORGI-260 and addresses dupe CORGI-292)
* added a `view` url param which can be used to switch to a 'summary' view (instead of just specifically for when using ofuri)
* added a `mode` url param (as we might have other modes in the future)
* product_stream component link now generates using view=summary&mode=latest

Once this PR is applied we need to refresh SoftwareBuilds to populate completion_time column:
```
 for sb in SoftwareBuild.objects.all():
    sb.save()
```

**Note** - there will be a separate PR addressing when a product (actually) uses multiple builds of the same component ... maybe best illustrated with an example eg. taking rhel which includes python-pygments  for both python27 and python36

* python27:2.7/python-pygments	2.2.0-22.module+el8.5.0+10788+**a4cea9e0**
* python36:3.6/python-pygments	2.2.0-22.module+el8.5.0+10789+**e4939b94**

which should probably be treated as 2 root components ...as well as some minor perf optimisations.


corgi/core/models.py
```diff
@@ -302,6 +306,14 @@ def save_product_taxonomy(self):
 
         return None
 
+    def save(self, *args, **kwargs):
+        if "completion_time" in self.meta_attr:
+            dt = datetime.datetime.strptime(
+                self.meta_attr["completion_time"].split(".")[0], "%Y-%m-%d %H:%M:%S"
+            )
+            self.completion_time = make_aware(dt, timezone=TZINFO)
```
mprpic: If we're making completion_time a proper field, let's set it in the collector itself, just like we set e.g. name, tags, or type.

You can also parse the date with Django's utils:

```python
from django.utils import timezone, dateparse
make_aware(dateparse.parse_datetime(completion_time))
```

`make_aware` uses the timezone that is already configured in settings, which is UTC. Re-defining it in an extra TZINFO constant is not necessary.

corgi/core/migrations/0039_auto_20221013_0748.py
```diff
@@ -0,0 +1,33 @@
...
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="components",
+                to="core.softwarebuild",
+            ),
+        ),
+        migrations.AddIndex(
+            model_name="softwarebuild",
+            index=models.Index(fields=["completion_time"], name="core_softwa_complet_c38fc7_idx"),
+        ),
+    ]
```
mprpic, line 45: It would be nice to add a data migration here to populate all the completion dates for all builds instead of relying on save().

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
     filterset_class = ComponentFilter
     lookup_url_kwarg = "uuid"
 
+    def get_queryset(self):
+        mode = self.request.query_params.get("mode")
+        if mode == "latest":
```
mprpic: Do we expect any other modes to ever exist? I think this would be fine as `show_latest=true|false` boolean flag, no?

corgi/core/models.py
```diff
@@ -225,15 +228,16 @@ class Type(models.TextChoices):
     type = models.CharField(choices=Type.choices, max_length=20)
     name = models.TextField()  # Arbitrary identifier for a build
     source = models.TextField()  # Source code reference for build
+    completion_time = models.DateTimeField(null=True)  # meta_attr["completion_time"]
```
mprpic, line 228: I think this should be a non-nullable field. Each build must have a completed time, otherwise it's not completed yet and we don't want to collect it.

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
+            #      multiple root components with the same component name yet different versions.
+            ofuri = self.request.query_params.get("ofuri")
+            if ofuri:
+                component_types = Q(type=Component.Type.SRPM) | Q(
+                    type=Component.Type.CONTAINER_IMAGE
+                )
+                results = (
+                    Component.objects.filter(
+                        component_types,
+                        product_streams__overlap=[ofuri],
```
mprpic: Should this not be `__contains`?

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
+                        software_build__completion_time__isnull=False,
+                    )
+                    .annotate(
+                        latest=Subquery(
+                            Component.objects.filter(
+                                component_types,
+                                name=OuterRef("name"),
+                                product_streams__overlap=[ofuri],
+                                software_build__completion_time__isnull=False,
```
mprpic: Are all the filters besides `name=OuterRef("name")` actually necessary? I think the parent query already pairs it down to the components that we need.

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
+                            )
+                            .order_by("software_build__completion_time")
+                            .values("uuid")[:1]
+                        )
+                    )
+                    .filter(
+                        uuid=F("latest"),
```
mprpic: I haven't tested this query, but does this actually work for lists of components? If e.g. ofuri:rhel-8 ships emacs-1, emacs-2, emacs-3, openssl-4, openssl-5, will this spit out [emacs-3, openssl-5]? The way I'm reading this query suggest it would only spit out "openssl-5".

It would be nice to have a test to go along with this.

corgi/core/models.py
JimFuller-RedHat: thx will do!

corgi/core/migrations/0039_auto_20221013_0748.py
JimFuller-RedHat: sure will add it

corgi/api/views.py
JimFuller-RedHat: yes (in discussion with Marek) but equally happy to constrain to show_latest for now

mprpic:
In the future, can you please split out changes like this into multiple commits? It's a lot easier to review and you even listed them out in the PR description in logical groups that would have made for good individual commits:

> * surfaced up completion_time on SoftwareBuild which is needed to determine latest component
> * while in the area - fixed SoftwareBuild .components to be true foreign key relationship (which already existed)
> * fixed broken link generation (CORGI-260 and addresses dupe CORGI-292)
> * added a view url param which can be used to switch to a 'summary' view (instead of just specifically for when using ofuri)
> * added a mode url param (as we might have other modes in the future)
> * product_stream component link now generates using view=summary&mode=latest

corgi/core/models.py
JimFuller-RedHat: that was my instinct as well, though there are software builds without completion time ;( ... unsure of what the right answer should be

mprpic: Got an example?

corgi/api/views.py
JimFuller-RedHat: when we do not include these the ordered list includes outside of original parent query ... which then results in a wrong answer ... let me retest again to verify this is the case (possible I was confused)

corgi/api/views.py
JimFuller-RedHat: yes it works as you describe ([emacs-3, openssl-5]) - can do a demo tmrw to show you behaviour and yes will add a test ;)

corgi/core/models.py
JimFuller-RedHat: hmmm looping through SoftwareBuild - all have completion time... weird ... maybe the err I was seeing was converting it into datatime ... one moment

corgi/api/views.py
mprpic: Also, it just occurred to me that a more readable query might start by finding the distinct builds that ship latest components and then filter Components on that, something like:

```python
Component.objects.filter(
    software_build__in=SoftwareBuild.objects.filter(completion_time__isnull=False)
   .order_by("components__name", "-completion_time")
   .distinct("components__name")
).distinct()
```

(I left out the other ofuri and type filters.)

corgi/api/views.py
jasinner: I think overlap is better, as I found when investigating [CORGI-229](https://github.com/RedHatProductSecurity/component-registry/pull/173/files) that __contains was returning matches for arrayfield value substrings. For example a query for `o:redhat:rhel:8.6.0` was return matches for both `o:redhat:rhel:8.6.0` and `o:redhat:rhel:8.6.0.z`.

corgi/api/views.py
jasinner: I think we should default to showing only the latest for ofuri param on components, and have a boolean to `show_all`.

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
+
     def list(self, request, *args, **kwargs):
         # purl are stored with each segment url encoded as per the specification. The purl query
         # param here is url decoded, to ensure special characters such as '@' and '?'
-        # are not interpreted as part of the request.
-        ofuri = request.query_params.get("ofuri")
-        if ofuri:
+        # are not interpreted  as part of the request.
+        view = request.query_params.get("view")
+        if view == "summary":
```
jasinner, line 408: I'm not convinced we need a mode for this. Can we leave mode out for now until we have a better use case or requirement for it?

corgi/api/views.py
JimFuller-RedHat: ok - after chatting with @jasinner  - we will bake behaviour into ?ofuri={} usage and not implement any additional url param switches eg. 'latest' only makes sense in context of a specific offering/product ... and we are always free to facet on ofuri with ?product_streams=ofuri (if we do not want latest mode).

Note - 'latest' should be used with manifest selection as well

corgi/api/views.py
JimFuller-RedHat: this is 'view' (not 'mode' which we decided to remove) .... this simply switches between summary and full listing (only components for now, but we may provide summary serialisations of other entities in the future) ... before we baked this in when using ?ofuri={} ... I do not think we need to bake it in just for ?ofuri={} and seemed generically separate (and useful).

jasinner: Ok, now that I tried it out and heard your explanation it seems good to keep to keep the view.

corgi/api/views.py
```diff
@@ -380,12 +381,49 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
+    def get_queryset(self):
+        mode = self.request.query_params.get("mode")
+        if mode == "latest":
+            # 'latest' mode only relevent in terms of a specific offering/product
+            # TODO - subject of a near-future PR will handle instances of when we have
+            #      multiple root components with the same component name yet different versions.
+            ofuri = self.request.query_params.get("ofuri")
+            if ofuri:
+                component_types = Q(type=Component.Type.SRPM) | Q(
+                    type=Component.Type.CONTAINER_IMAGE
```
jasinner: I think this should be as follows to eliminate arch specific containers, and also include RHEL_MODULE types.

```
root_components = (
    Q(type=Component.Type.SRPM)
    | Q(type=Component.Type.CONTAINER_IMAGE, arch="noarch")
    | Q(type=Component.Type.RHEL_MODULE)
)
```
There is probably a better way to do this using ComponentNode.parent == None, but I can't think of it right now.


Perhaps this component_types query could actually be a distinct view, and added to latest by default?

jasinner (changes requested):

corgi/api/views.py
JimFuller-RedHat: sounds good to me

corgi/api/views.py
JimFuller-RedHat: removed

JimFuller-RedHat:
* baked in ofuri behaviour (only really makes sense with product_stream)
* removed setting on software_build.save(), added data migration and setting of completion_time in brew task
* enhanced tests 
* updated openapi (was complaining about StatusViewSet which has not queryset)

jasinner (approved):
Looks good to me, thanks.

corgi/api/views.py
mprpic: That might be true for string matches, but this is an ArrayField lookup.

tests/test_api.py
```diff
@@ -526,12 +527,47 @@ def test_api_component_400(client, api_path):
 
 
 def test_product_components_ofuri(client, api_path):
-    openssl = ComponentFactory(name="openssl")
-    curl = ComponentFactory(name="curl")
+    """test 'latest' filter on components"""
```
mprpic, line 530: So we're testing that this returns a certain number of components, can we also verify that it returns the right one please?

corgi/tasks/brew.py
```diff
@@ -52,6 +53,17 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     build_meta["corgi_ingest_start_dt"] = dateformat.format(timezone.now(), "Y-m-d H:i:s")
     build_meta["corgi_ingest_status"] = "INPROGRESS"
 
+    completion_dt = None
+    if "completion_time" in build_meta:
+        completion_dt = make_aware(
+            dateparse.parse_datetime(build_meta["completion_time"].split(".")[0])  # type:ignore
```
mprpic, line 59: Why is this being split on the `.`?

JimFuller-RedHat:
re 'why not latest on products, product_versions, etc' ... this PR is focused on product_streams ... also in chat this morning with @jasinner was unclear if this is needed at this time eg. manifests will need to use product_stream 'latest' filter - and I do not think we are ever contemplating generating manifests for product, product_version, etc. (what would they represent in terms of delivered bits to end users ?)

corgi/core/models.py
```diff
@@ -774,7 +803,9 @@ class Namespace(models.TextChoices):
     upstreams = fields.ArrayField(models.CharField(max_length=1024), default=list)
 
     cnodes = GenericRelation(ComponentNode)
-    software_build = models.ForeignKey(SoftwareBuild, on_delete=models.CASCADE, null=True)
+    software_build = models.ForeignKey(
+        SoftwareBuild, on_delete=models.CASCADE, null=True, related_name="components"
+    )
```
juspence, line 808: This change broke the components property:
```
    @property
    def components(self):
        return self.component_set.values_list("purl", flat=True)
```

PR #186, opened 2022-10-14
Manifest should use .get_latest_components now

JimFuller-RedHat:
Now that we have a 'latest' filter which returns the latest built top level components (SRPM, noarch CONTAINER_IMAGE and RHEL_MODULE) ... we should employ this when generating product stream manifest.

Tagging @juspence specifically ... there is a failing test which we need to figure out ... though running out of time this week to diagnose ... raising this PR for discussion.

Note - we also may consider removing from REST API manifest links from product, product_version and product_variants

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from random import choice, randint
 
 import factory
+from django.utils import timezone
 
 from corgi.core import models
 
@@ -26,6 +27,7 @@ class SoftwareBuildFactory(factory.django.DjangoModelFactory):
     build_id = factory.sequence(lambda n: n + 100)
     name = factory.Faker("word")
     tag = factory.RelatedFactory(SoftwareBuildTagFactory, factory_related_name="software_build")
+    completion_time = timezone.now()
 
 
 class ProductTagFactory(TagFactory):
--- a/tests/test_manifests.py
+++ b/tests/test_manifests.py
@@ -30,14 +30,24 @@ def test_product_manifest_properties():
     pvnode = ProductNode.objects.create(parent=pnode, obj=version, object_id=version.pk)
     psnode = ProductNode.objects.create(parent=pvnode, obj=stream, object_id=stream.pk)
     _ = ProductNode.objects.create(parent=psnode, obj=variant, object_id=variant.pk)
-    # This generates and saves the product_variants, and product_streams property of `variant`
-    variant.save_product_taxonomy()
+    # This generates and saves the product_variants property of stream
+    stream.save_product_taxonomy()
+    assert variant.name in stream.product_variants
 
     build = SoftwareBuildFactory(build_id=1)
-    component = ComponentFactory(software_build=build)
+    build.save()
+    component = ComponentFactory(
+        software_build=build,
+        type="SRPM",
+        product_variants=[variant.ofuri],
+        product_streams=[stream.ofuri],
+    )
+    component.save()
     _, _ = component.cnodes.get_or_create(
         type=ComponentNode.ComponentNodeType.SOURCE, parent=None, purl=component.purl
     )
+    # The product_ref here is a variant name but below we use it's parent stream
+    # to generate the manifest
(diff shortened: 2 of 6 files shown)


tests/test_manifests.py
```diff
@@ -25,14 +25,22 @@ def test_product_manifest_properties():
     product = ProductFactory(product_variants=["1"])
     version = ProductVersionFactory(product_variants=["1"])
     stream = ProductStreamFactory(product_variants=["1"])
+    stream.save()
```
juspence: You don't need to call .save() here, the factory creates the model instance for you.

I don't know why the test fails, but you've changed how the manifest is generated so you probably also need to change what the test is checking for in the various assertions.  You can run the tests with -vvvv to get debug output that shows the full diff between what was asserted and was actually generated. The error on Gitlab is:
```
        # One component linked to this product
        num_components = len(variant.components)
        assert num_components == 1
        # Manifest contains info for all components + the product itself
>       assert len(manifest["packages"]) == 2
E       AssertionError: assert 1 == 2
```

"len(variant.components)" probably needs to change to "len(variant.get_latest_components)" to match the manifest template. If we're generating the manifest using get_latest_components, I would check to make sure that filter is including the test component in its output.

Once you fix the above len() statement, I bet "assert num_components == 1" will start failing too, AKA we return the test component in the full list of components but not in the filtered list. In other words, I think this is probably a bug in the filter (or a bug in how we generate the test data), and not a bug in the manifest template itself.

tests/factories.py
```diff
@@ -26,6 +26,7 @@ class Meta:
     build_id = factory.sequence(lambda n: n + 100)
     name = factory.Faker("word")
     tag = factory.RelatedFactory(SoftwareBuildTagFactory, factory_related_name="software_build")
+    completion_time = datetime.now()
```
juspence: This looks like it's generating a bunch of warnings in the test output. I would use Django's timezone.now() function to automatically generate TZ-aware datetimes using whatever timezone is configured in settings.

tests/test_manifests.py
```diff
@@ -25,14 +25,22 @@ def test_product_manifest_properties():
...
     pvnode = ProductNode.objects.create(parent=pnode, obj=version, object_id=version.pk)
     psnode = ProductNode.objects.create(parent=pvnode, obj=stream, object_id=stream.pk)
     _ = ProductNode.objects.create(parent=psnode, obj=variant, object_id=variant.pk)
 
     build = SoftwareBuildFactory(build_id=1)
-    component = ComponentFactory(software_build=build)
+    build.save()
+    component = ComponentFactory(
+        software_build=build,
+        type="SRPM",
```
juspence, line 41: Does your latest filter code exclude SRPMs when listing components? You might try removing the type="SRPM" line you added here and seeing if that makes the test pass.

In general, I'm happy to help review or debug things, but this doesn't really look like a manifesting bug to me. It's kind of hard to say for sure though, since I never reviewed (never got tagged) on the MR that implemented the latest filter.

In future you might tag me as a reviewer on other MRs too. There's no need to wait for approval from me before merging, but tagging me sends me an email notification, so will keep me aware of MRs submitted and new code added, which helps me give you a better answer when debugging instead of just guessing :)

juspence: I got a notification on this MR. I meant that I never got tagged / never got to review your previous MR that added the "latest" filter, so I'm not sure what it's doing and why the data it returns isn't correct.

tests/factories.py
jasinner: I fixed this.

tests/test_manifests.py
jasinner: The problem was were were using variant, not stream. All fixed up now.

jasinner:
I took ownership of this branch/PR because I want to work on CORGI-238 and CORGI-239 and it requires the changes here. @JimFuller-RedHat it should be ready to take out of draft and review now.

JimFuller-RedHat:
LGTM

PR #188, opened 2022-10-17
CORGI-309: Fix monitoring emails fail to send when too many tasks have failed

juspence:
@RedHatProductSecurity/corgi-devs Fix for the missing monitoring email in stage since 10/14. When the number of failed tasks is very large, iterating over the queryset fails.

To join all the TaskResults into a string email body, we should use a queryset iterator to only load some (chunk_size) rows into memory at one time, instead of loading all the rows into memory when the queryset is evaluated. We should also use .values_list() to only return the fields / columns we're interested in, instead of loading all the fields for a particular record and constructing a model instance in memory.

This doesn't handle the case when the email body itself is too big to be processed by the SMTP server. That's a little trickier to implement, so I think not worth spending time on. We can't just send one email per e.g. every 1000 rows. Some errors may have a very long traceback that puts us over the size limit, while other errors might have a smaller traceback that allows sending 1000 rows at a time.

We could pick some very small value (e.g. only 500 rows per email), or we could build the email body first, then split it apart into e.g. 10MB chunks. But that seems like a lot of processing for an issue that should only happen rarely. I also don't like the idea of always splitting the email since it may not be needed, and it's easier to search for text in one combined message (e.g. show me all "SoftTimeLimitExceeded" errors).

### DIFF
--- a/corgi/tasks/monitoring.py
+++ b/corgi/tasks/monitoring.py
@@ -79,10 +79,14 @@ def email_failed_tasks():
         return
 
     failed_tasks_threshold = get_last_success_for_task("corgi.tasks.monitoring.email_failed_tasks")
-    failed_tasks = TaskResult.objects.filter(
-        status__exact="FAILURE",
-        date_done__gte=failed_tasks_threshold,
-    ).order_by("task_name", "date_done")
+    failed_tasks = (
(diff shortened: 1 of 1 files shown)


jasinner (approved):
LGTM, thanks.

PR #189, opened 2022-10-18
CORGI-154 fix builds property of productmodel

jasinner:
I removed both the `errata`, and `upstream` properties of ProductModel which were using `get_product_component_relations` function but where not being called by anything. That allowed me to consolidate the `get_product_component_relations` function into the `builds` function, and therefore simplify it.

While I was simplifying functions I also fixed the `errata` function of the Component model so that it didn't check the meta_attr components field unnecessarily. I removed both `get_product_variants`, and `get_product_streams` functions from Component which were a legacy of when we had a `save_product_taxonomy` function directly on the Component Model. Today component product taxonomies are updated by calling SoftwareBuild's `save_product_taxonomy` function.

Also added 2 new tests, and refactored some tests to make them work after the changes to ProductModel `builds` , and easier to understand.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -347,34 +347,30 @@ class ProductModel(models.Model):
         # Else self.builds is an empty QuerySet
         return Component.objects.none()
 
-    @staticmethod
-    def get_product_component_relations(
-        variant_ids: list[str], sort_field="external_system_id", only_errata=False
-    ) -> QuerySet["ProductComponentRelation"]:
-        """Returns unique productcomponentrelations with at least 1 matching variant_id,
-        ordered by sort_field. if only_errata is True, only return PCRs with Type=ERRATA
-        """
-        if variant_ids:
-            query = Q()
-            for variant_id in variant_ids:
-                query |= Q(product_ref__contains=variant_id)
-            result = ProductComponentRelation.objects.filter(query).distinct().order_by(sort_field)
-            if only_errata:
-                result = result.filter(type=ProductComponentRelation.Type.ERRATA)
-            return result
-        # Else self.product_variants is an empty QuerySet
-        return ProductComponentRelation.objects.none()
-
     @property
     def builds(self) -> QuerySet:
-        """Return unique build IDs for errata with variant_ids matching self.product_variants"""
-        return (
-            self.get_product_component_relations(
-                self.product_variants, "build_id", only_errata=True
+        """Returns unique productcomponentrelations with at least 1 matching variant or stream,
+        ordered by build_id.
+        """
+        product_refs = [self.name]
+        if isinstance(self, ProductStream):
+            # we also want to include child product_variants of this product stream
+            product_refs.extend(self.product_variants)
+        elif isinstance(self, ProductVersion) or isinstance(self, Product):
+            # we don't store product or product versions in the relations table therefore
+            # we only want to include child product_streams and product_variants in the query
+            product_refs = self.product_streams + self.product_variants
+        # else it was a product variant, only look up by self.name
+
+        if product_refs:
+            return (
+                ProductComponentRelation.objects.filter(product_ref__in=product_refs)
+                .order_by("build_id")
+                .values_list("build_id", flat=True)
+                .distinct()
             )
-            .values_list("build_id", flat=True)
-            .distinct()
-        )
+        # Else no product_variants or product_streams
+        return ProductComponentRelation.objects.none()
 
     @property
     def components(self) -> QuerySet["Component"]:
@@ -383,33 +379,9 @@ class ProductModel(models.Model):
         return self.get_related_components(self.builds, "purl")
 
     @property
-    def errata(self) -> QuerySet:
-        """Return unique errata IDs with variant_ids matching self.product_variants"""
-        return (
-            self.get_product_component_relations(self.product_variants, only_errata=True)
-            .values_list("external_system_id", flat=True)
-            .distinct()
-        )
-
-    @property
     def manifest(self) -> str:
         """Return an SPDX-style manifest in JSON format."""
         return ProductManifestFile(self).render_content()
-
-    @property
-    def upstream(self) -> QuerySet:
-        """Return unique upstream values for components with build_ids matching self.builds"""
-        # TODO: Need to add @extend_schema_field to fix drf-spectacular warning
-        #  'unable to resolve type hint for function "upstream".
-        #  Consider using a type hint or @extend_schema_field. Defaulting to string.'
-        #  list or Iterable[str] work for drf-spectacular, but break mypy
-        #  Union[anything, QuerySet] or QuerySet work for mypy, but not drf-spectacular
-        # for RPMS its sibling, SRPM its child
-        return (
-            self.get_related_components(self.builds, "upstreams")
-            .values_list("upstreams", flat=True)
-            .distinct()
-        )
 
     @property
     def coverage(self) -> int:
@@ -525,6 +497,7 @@ class ProductStream(ProductModel, TimeStampedModel):
     active = models.BooleanField(default=False)
     et_product_versions = fields.ArrayField(models.CharField(max_length=200), default=list)
 
+    # redefined from parent class
     product_streams = None  # type: ignore
     pnodes = GenericRelation(ProductNode, related_query_name="product_stream")
 
@@ -592,6 +565,7 @@ class ProductVariant(ProductModel, TimeStampedModel):
 
     cpe = models.CharField(max_length=1000, default="")
 
+    # redefined from parent class
     product_variants = None  # type: ignore
     pnodes = GenericRelation(ProductNode, related_query_name="product_variant")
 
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat:
LGTM

corgi/core/models.py
```diff
@@ -592,7 +561,7 @@ class ProductVariant(ProductModel, TimeStampedModel):
 
     cpe = models.CharField(max_length=1000, default="")
 
-    product_variants = None  # type: ignore
+    product_variants = []  # type: ignore
```
mprpic: I guess we can now remove the ignored type hint check that would have caught this in the first place :-)

mprpic (approved):

corgi/core/models.py
juspence: This behavior was copied from the original code, and isn't technically a bug, but I'm fine with changing it. Originally we had code in different models that was almost identical:
```
# No inheritance in original code
class ProductStream():
    products = ArrayField()
    product_versions = ArrayField()
    # product_streams field wasn't originally defined on ProductStream model
    product_variants = ArrayField()

# No inheritance in original code
class ProductVariant():
    products = ArrayField()
    product_versions = ArrayField()
    product_streams = ArrayField()
    # product_variants field wasn't originally defined on ProductVariant model
```

When I factored out that common code to the ProductModel class, I defined all four properties on the parent ProductModel class to avoid duplication and to make the schema generation work. To keep the behavior identical to the old code, I added e.g. ```product_variants = None``` so that the ProductVariant model still had the same properties as before, instead of getting a new product_variants property (inherited from the parent class).

In this case, setting ```product_variants = None``` means the field isn't defined on the model, even though it's defined in the parent class. So the field is "not present", which is different from "present but always None / not iterable". Maybe that's not the behavior we want, and we'd rather force all ProductModel subclasses to have the same properties.

In that case, the product_variants field here could just be removed (inherit the field definition from the parent class), or we could redefine the product_variants field on the ProductVariant model to make it read-only (force it to always be an empty ArrayField).

Setting it to an empty list isn't quite what we want, I think. It's not "forcing the ArrayField to be initialized to an empty list" as I understand, but rather "replacing an ArrayField with a Python list" which won't define the Django methods for e.g. lookups.

corgi/core/models.py
```diff
@@ -347,70 +347,39 @@ def get_related_components(build_ids: QuerySet, sort_field: str) -> QuerySet["Co
...
+        """Returns unique productcomponentrelations with at least 1 matching variant or stream,
+        ordered by build_id.
+        """
+        product_refs = []
+        if isinstance(self, ProductVariant):
+            product_refs.append(self.name)
+        elif isinstance(self, ProductStream):
+            product_refs.append(self.name)
+            # we also want to include child product_variants for each product stream in the query
+            product_refs.extend(self.product_variants)
```
juspence: I'm not clear why we're using self.name for a ProductStream or a ProductVariant, but not for a Product or a ProductVersion.

Assuming the product_streams / product_variants property is always present as an ArrayField, even on the ProductStream / ProductVariant model:
```suggestion
        # for all ProductModel subclasses, use the list of child streams and child variants
        product_refs = self.product_streams + self.product_variants
        # for streams, "child streams" is an empty list so add the stream's name
        # for variants, "child variants" is an empty list so add the variant's name
        if isinstance(self, ProductStream) or isinstance(self, ProductVariant):
            product_refs += [self.name]
```

Or maybe:
```suggestion
        # for all ProductModel subclasses, look up based on current name, child streams, and child variants
        # for ProductStream and ProductVariant classes, one of below will be an empty list
        product_refs = [self.name] + self.product_streams + self.product_variants
```

corgi/core/models.py
```diff
@@ -347,70 +347,39 @@ def get_related_components(build_ids: QuerySet, sort_field: str) -> QuerySet["Co
...
+        if product_refs:
+            return (
+                ProductComponentRelation.objects.filter(product_ref__in=product_refs)
```
juspence, line 367: This does an exact lookup, where the product_ref stored in the PCR table has to exactly match one of the variant names in the product_refs list / self.product_variants field.

The old get_pcr method (which I refactored slightly without changing logic) did a lookup to check if each product_ref contained a variant name (AKA check if the variant name was a substring of the product_ref). Is that substring check unneeded, AKA do we always store exactly the variant name as the product_ref?

corgi/core/models.py
jasinner: I'll switch it to ArrayField, and test removing the mytest skip flag as well. Thanks for explaining the history of the changes here.

juspence: You're welcome - if you want it to be an ArrayField, you can just remove the line completely. It should get inherited from the parent ProductModel class.

corgi/core/models.py
jasinner: Great feedback, thank you. I'll go with the one-liner here.

corgi/core/models.py
jasinner: I think it's better to use exact matches here to make relations very concrete. I think if we're going to do this kind of inclusive matching the rules should probably be defined in product_definitions or somewhere the IR / Secure Engineering teams can modify the logic themselves.

juspence: Ack, I don't necessarily have a preference either way. I'm just not sure why the old behavior was written that way - if you think it's OK to change it, I'm fine with this as-is.

juspence (approved):

corgi/core/models.py
jasinner: Switching to ArrayField would have required a migration, so I'm going to change it back to None, and just be careful about what we include in the ProductModel.builds properities so that we don't try to concatenate None to a list.

corgi/core/models.py
jasinner: On section thought and closer inspection we don't want to include product_streams list when the type is ProductVariant, because that is too general. I'm leaving this closer to the original code with a small tweak to avoid concatenating None and list.

jasinner: I added some tests for Product and ProductVersion nested builds.

corgi/core/models.py
```diff
@@ -990,15 +944,12 @@ def errata(self) -> list[str]:
         """Return errata that contain component."""
```
juspence, line 946: I don't see anything that uses this errata property on the component model either. Did I miss something, or is this also unused? Fine to leave if you'd rather not change this, just a heads-up you might also be able to remove this one.

corgi/core/models.py
```diff
@@ -347,70 +347,40 @@ def get_related_components(build_ids: QuerySet, sort_field: str) -> QuerySet["Co
...
+        ordered by build_id.
+        """
+        product_refs = [self.name]
+        if isinstance(self, ProductStream):
+            # we also want to include child product_variants for each product stream in the query
+            product_refs.extend(self.product_variants)
+        elif not isinstance(self, ProductVariant):
+            # If it's a product variant the product_variants list will be None
+            product_refs = self.product_streams + self.product_variants
+
```
juspence: Code above looks almost right, but we don't look up by self.name when self is a ProductVersion or Product. I can't tell if that's a typo or not.

Algorithm for what you have above:
1. Start with only self.name in the lookup.
2. If it's a stream, just add the child variants to the lookup.
3. Else if it's not a stream and not a variant (so must be either a product or product version), add both the child streams and child variants to the lookup. TODO: You overwrite ```product_refs =``` instead of appending ```product_refs +=```, so we lose self.name in this case. Is that a bug?
5. Else it must have been a variant, and none of the if-blocks will be entered so the lookup will only contain the variant's name.

Assuming 3. was on purpose, below might be a bit easier to follow since it avoids the negative if case and documents the else case with a comment:
```suggestion
        if isinstance(self, ProductStream):
            # we also want to include child product_variants for each product stream in the query
            product_refs.extend(self.product_variants)
        elif isinstance(self, ProductVersion) or isinstance(self, Product):
            # we also want to include child product_streams and product_variants in the query
            product_refs = self.product_streams + self.product_variants
        # else it was a product variant, only look up by self.name
```

Or if 3. was a bug, just add the += on the second-to-last line.

juspence (approved):

corgi/core/models.py
juspence: To clarify a little more (probably overexplaining but want to be clear): it was set to None to avoid inheriting the property from the parent class. This was copied from the original code. We didn't define the "product_variants" field on the ProductVariant model, since it didn't make sense at the time.

In other words, this code defines what fields exist on the ProductVariant class. We aren't changing the value stored in the field or setting the "product_variants" field on any particular model instance. "None" here doesn't store a value inside the field, and you aren't trying to concatenate a None value to a list of values. "None" completely removes the field from the model definition. Then if you tried to add self.product_streams and self.product_variants from some ProductVariant instance, you would be trying to access a nonexistent product_variants property, which raises the below error:
```
In [1]: from corgi.core.models import *

In [2]: ProductVariant.objects.last()
Out[2]: <ProductVariant: Supplementary-9.0.0.Z.MAIN.EUS>

In [3]: ProductVariant.objects.last().product_streams
Out[3]: ['rhel-9.0.0.z']

In [4]: ProductVariant.objects.last().product_streams2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [4], in <module>
----> 1 ProductVariant.objects.last().product_streams2

AttributeError: 'ProductVariant' object has no attribute 'product_streams2'
```

So removing "None" adds back the "product_variant" field the exact same way it's defined in the parent class. And changing what properties are defined on a model will always require a migration. We could make this change if we wanted to make the product_streams field on the ProductStream model default to a one-item list of [self.name], for example. Then all of our ProductModel subclasses would be consistent and have the same properties.

corgi/core/models.py
jasinner: It seems to me it's used by the api ComponentSerializer class, that's why I left it in.

corgi/core/models.py
jasinner: good idea, I'll use this.

PR #191, opened 2022-10-19
CORGI-74 Add modules and latest to user guide

jasinner:
Supports CORGI-205 by filtering out inactive product_streams by default in the product_streams list endpoint. All product streams can be included by adding `view=all` to the endpoint parameters.

Also adds an example of getting components from a product stream to the user docs.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -290,11 +290,14 @@ class ProductVersionViewSet(ProductDataViewSet):
 class ProductStreamViewSetSet(ProductDataViewSet):
     """View for api/v1/product_streams"""
 
-    queryset = ProductStream.objects.get_queryset()
+    queryset = ProductStream.objects.filter(active=True)
     serializer_class = ProductStreamSerializer
 
     def list(self, request, *args, **kwargs):
         req = self.request
+        active = request.query_params.get("active")
+        if active == "all":
+            self.queryset = ProductStream.objects.filter()
         ofuri = req.query_params.get("ofuri")
         if not ofuri:
             return super().list(request)
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -291,8 +291,7 @@ curl -s 'https://{CORGI_HOST}/api/v1/components?type=NPM&name=is-svg&version=2.1
 ```
 
 This query returns a list of results include the component count. The component data can be found in the results field.
-The sources field lists all the components which embed this component, at the time of writing we are yet to implement
-latest filtering, so it's useful to process the results on the client side to get a clearer picture of the packages included:
+The sources field lists all versions of all components which embed this component, it's useful to process the results on the client side to get a clearer picture of the packages included:
 
 ```bash
 $ curl -L -s 'https://{CORGI_HOST}/api/v1/components?purl=pkg:npm/is-svg@2.1.0' | jq '.sources[] | .purl' | awk -F@ '{print $1}' | cut -c2- | sort | uniq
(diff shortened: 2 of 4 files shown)


JimFuller-RedHat:
apart from flake8 CI failure, LGTM - one proposal - consider changing view=active to active={} ... view= seems more a generic switch to do with serialisation rather then data selection.

mprpic (approved):

tests/factories.py
```diff
@@ -72,6 +72,7 @@ class Meta:
     description = factory.Faker("word")
     cpe = "cpe:/o:redhat:enterprise_linux:9"
     tag = factory.RelatedFactory(ProductStreamTagFactory, factory_related_name="product_stream")
+    active = True
```
juspence, line 75: Should we default this to a random value, so we can accurately test latest vs. non-latest streams without hardcoding the behavior? Or maybe it should be latest=False to test the non-latest behavior by default, and force factory users to opt-in if they want to work with a latest stream.

corgi/api/views.py
```diff
@@ -290,11 +290,15 @@ def taxonomy(self, request, uuid=None):
...
-    queryset = ProductStream.objects.get_queryset()
+    queryset = ProductStream.objects.filter(active=True)
     serializer_class = ProductStreamSerializer
 
     def list(self, request, *args, **kwargs):
         req = self.request
+        active = request.query_params.get("active")
+        if active == "all":
+            self.queryset = ProductStream.objects.filter()
+            return super().list(request)
```
juspence: I think the return statement should be removed, otherwise ?active=all&ofuri=value won't work together.

docs/user_guide.md
```diff
@@ -601,6 +600,418 @@ curl -s 'https://{CORGI_HOST}/api/v1/components?nvr=bare-metal-event-relay-opera
...
+"pkg:rpm/redhat/galera@25.3.26-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-backup-debuginfo@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-debuginfo@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-server-debuginfo@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-server-utils-debuginfo@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-embedded-devel@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/mariadb-gssapi-server-debuginfo@10.3.17-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+"pkg:rpm/redhat/Judy-debugsource@1.0.5-18.module%2Bel8%2B2765%2Bcfa4f87b?arch=i686"
+"pkg:rpm/redhat/galera-debuginfo@25.3.26-1.module%2Bel8.1.0%2B3974%2B90eded84?arch=i686"
+```
```
juspence, line 588: Doc updates LGTM, but the full output isn't needed I think. You can just give a few lines as a sample and then there's less scrolling involved for the user.

juspence (approved):

tests/factories.py
jasinner: Well I set it to active=True because the existing test expect that. I don't expect this flag to be used much outside of the active url parameter on products though. I think the rest of the filtering we will push out to clients aka service layer.

corgi/api/views.py
jasinner: good catch. Fixed.

PR #192, opened 2022-10-20
Latest view perf

JimFuller-RedHat:
This PR finalises the work revolving around 'latest view' as well as cleaning up various indexes.

Fixed date sorting typo for latest filter so it actually returns the latest (instead of the 'earlist'). Added a start date param that is used to define the latest dt ... useful for generating a manifest at some previous point in time (this was requested by Riccardo at some point). In addition fixed links from products where they would have been broken.

For performance added:
* generic ofuri index on products
* specific set of 'latest view' partial indexes

Also added get_latest_components() to ProductVariant (in addition to ProductStream) though we still need to be able to generate latest view on a 'productless' component(s) ... that will be subject of some future PR.

While I was in the area:
* changed data migration for completion time to use an iterator to be more efficient (as it triggered pod killer doing migration in stage and we want to avoid that when we push to prod)
* replaced counts with estimates on status view (eg. /api/v1/status) eg. hitting status page will take 50ms vs explicitly counting (well known fun with [count perf in postgres](http://wiki.postgresql.org/wiki/Slow_Counting)) 

**Note** - once this is merged I will ensure all test indexes are deleted on stage.


corgi/api/serializers.py
```diff
@@ -60,6 +60,23 @@ def get_model_ofuri_link(
...
+    related_type=None,
+    view=None,
+) -> str:
+    """Generic method to get an ofuri link for an arbitrary Model subclass."""
+    link = f"{CORGI_API_URL}/{model_name}?product_streams={product_stream}"
+    if model_name == "components":
+        link = f"{link}"
+    if related_type:
+        link += f"&type={related_type}"
+    if view:
```
mprpic: What is this used for? I don't see it used anywhere.

corgi/api/serializers.py
```diff
@@ -60,6 +60,23 @@ def get_model_ofuri_link(
...
+def get_model_product_stream_link(
+    model_name: str,
+    product_stream: str,
+    related_type=None,
+    view=None,
+) -> str:
+    """Generic method to get an ofuri link for an arbitrary Model subclass."""
+    link = f"{CORGI_API_URL}/{model_name}?product_streams={product_stream}"
+    if model_name == "components":
+        link = f"{link}"
```
mprpic: This IF branch can be removed since it's redundant, the link is returned by itself at the end anyway.

tests/test_api.py
```diff
@@ -1,8 +1,8 @@
-from datetime import datetime
 from urllib.parse import quote
 
 import pytest
 from django.conf import settings
+from django.utils.datetime_safe import datetime
```
mprpic: This is a [deprecated](https://docs.djangoproject.com/en/4.0/releases/4.0/#id2) function, why are we using it here?

corgi/core/models.py
```diff
@@ -595,6 +603,38 @@ def get_ofuri(self) -> str:
             product_stream = f"{product_stream_node.obj.ofuri}:{self.name.lower()}"
         return product_stream
 
+    def get_latest_components(self, start_dt=now()):
```
mprpic: This is the same method that already exists on the ProductModel, which this class even inherits from, why are we duplicating it here?

corgi/core/models.py
```diff
@@ -515,13 +519,16 @@ def manifest(self) -> str:
...
+        """Return root components from latest builds (or from a supplied start datetime)."""
         root_components = (
             Q(type=Component.Type.SRPM)
             | Q(type=Component.Type.CONTAINER_IMAGE, arch="noarch")
             | Q(type=Component.Type.RHEL_MODULE)
         )
+        dt_range = Q()
+        if start_dt:
+            dt_range &= Q(software_build__completion_time__lte=start_dt)
```
mprpic: This seems counterintuitive: start time filters components that have been built *before* the specified datetime? I would expect it to return everything built *after* (`gte`) that datetime? Or am I reading this wrong?

corgi/core/migrations/0041_custom_indexes.py
```diff
@@ -0,0 +1,31 @@
...
+             USING btree (type, name) WHERE (((type)::text = 'SRPM'::text) OR \
+             (((arch)::text = 'noarch'::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_idx ON public.core_component \
+            USING btree (uuid, software_build_id, type, name, product_streams) \
+            WHERE (((type)::text = 'SRPM'::text) OR (((arch)::text = 'noarch'::text) \
+            AND ((type)::text = 'CONTAINER_IMAGE'::text)) OR ((type)::text = 'RHEL_MODULE'::text))"
+        ),
```
mprpic: Can all of these not be added on the model itself? Why are we using raw SQL here?

corgi/core/migrations/0040_auto_20221020_1057.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.2.15 on 2022-10-20 10:57
```
mprpic, line 1: This migration will conflict with the existing 0040 that's already on main. Please rebase and refresh your migrations.

corgi/api/views.py
```diff
@@ -114,6 +113,19 @@ def list(self, request):
...
+            )
+            pcr_count = cursor.fetchone()
+            cursor.execute(
+                "SELECT reltuples AS estimate FROM pg_class WHERE relname = 'core_component';"
+            )
+            component_count = cursor.fetchone()
+            cursor.execute(
+                "SELECT reltuples AS estimate FROM pg_class WHERE relname = 'core_softwarebuild';"
+            )
+            sb_count = cursor.fetchone()
```
mprpic, line 132: Can you add a comment in the code please explaining why these are done using estimates / raw SQL?

corgi/api/serializers.py
JimFuller-RedHat: get_model_product_stream_link() is used to create upstream link now ... I just emulated status quo eg. will narrow down to get_upstream_link

corgi/api/serializers.py
JimFuller-RedHat: sure, above comment will remove configurability

tests/test_api.py
JimFuller-RedHat: will fix (I let pycharm do the picking)

corgi/core/models.py
JimFuller-RedHat: unsure what you mean ... get_latest_components() should exist on both ProductStream and ProductVariant .. it is not currently on ProductModel.

corgi/core/models.py
JimFuller-RedHat: yes, it is confusing ... Ricardo used start ... it could just as easily be date_before and we could contemplate adding a date_after ... will ping you/Jason for further thoughts

corgi/core/migrations/0041_custom_indexes.py
JimFuller-RedHat: these are custom partial indexes which I believe in django can kind of be represented with constraints but (for me) representing what is a perf only index in that way obscures intent... prefer to isolate perf only indexes as raw sql.

corgi/core/migrations/0040_auto_20221020_1057.py
JimFuller-RedHat: oh did something slip in already ? will update

corgi/core/models.py
mprpic: Ah, I must have misread the diff, it is indeed on both ProductStream and ProductVariant. Still though, why do we have to duplicate it? Can't we move it to ProductModule or make a separate mixin?

corgi/core/models.py
mprpic: Just exclude it from this PR, it can be added later once we figure out exactly what we need here. I don't think this is urgent to get in, is it?

corgi/core/migrations/0041_custom_indexes.py
mprpic: I think these indexes are obscured if they only exist in a migration though. How is a future developer supposed to find out about them unless they want to read through 40+ migration files? At the very least I'd add a call out to this file on the affected models in models.py.

corgi/core/models.py
JimFuller-RedHat: hmmm .... I added this to address some aspect of manifesting on 'other things' but things have moved on ...  'other things' in question are less product/offering based and more on 'productless' things like generating a manifest on a container (ex. adding a /manifest to a component eg. /api/v1/component/{uuid}/manifest). I will remove this from this PR.

corgi/core/models.py
JimFuller-RedHat: not urgent, will remove

corgi/core/models.py
```diff
@@ -515,13 +519,16 @@ def manifest(self) -> str:
         """Return an SPDX-style manifest in JSON format."""
         return ProductManifestFile(self).render_content()
 
-    def get_latest_components(self):
-        """Return root components from latest builds."""
+    def get_latest_components(self, start_dt=now()):
```
mprpic: This `now()` call gets evaluated once on start time and then it'll stay the same until you restart. I don't think that's what we want. Mutable arguments are usually passed as:

```python
def get_latest_components(self, start_dt=None):
    if start_dt is None:
        start_dt = now()
```

JimFuller-RedHat: ug ... thx for that ! (mutters: too many dynamic languages ...)

corgi/core/migrations/0041_custom_indexes.py
JimFuller-RedHat: admittedly depending on grep CREATE INDEX is not very explicit ... thx for highlighting - if it is ok with you I will add comments calling out (and will backfill comments for last custom index in the same manner )

corgi/api/serializers.py
JimFuller-RedHat: replaced with get_upstream_link()

corgi/api/serializers.py
JimFuller-RedHat: simplified

corgi/core/models.py
JimFuller-RedHat: removed start_dt

corgi/core/migrations/0040_auto_20221020_1057.py
JimFuller-RedHat: I do not see 0040 on main ... I did conflate the migrations into one

mprpic (approved):
Looks good now!

corgi/core/migrations/0040_auto_20221020_1057.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        migrations.AddIndex(
+            model_name="productvariant",
+            index=models.Index(fields=["ofuri"], name="core_produc_ofuri_ccf4bd_idx"),
+        ),
+        migrations.AddIndex(
+            model_name="productversion",
+            index=models.Index(fields=["ofuri"], name="core_produc_ofuri_c62cf2_idx"),
+        ),
+        migrations.RunSQL(
```
jasinner, line 29: Do you think we'd be able to adjust the query in get_latest_components to filter by component nodes which don't have parents instead of this type filter which is likely to change?

I just worry that if a new root type starts being used, we'll have to adjust and rebuild this index as well.

corgi/core/migrations/0040_auto_20221020_1057.py
```diff
@@ -0,0 +1,47 @@
...
+            "CREATE INDEX core_compon_latest_name_type_idx ON public.core_component \
+             USING btree (name, type) WHERE (((type)::text = 'SRPM'::text) OR \
+             (((arch)::text = 'noarch'::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
+        ),
+        migrations.RunSQL(
```
jasinner, line 35: Why is there more than one index being added here?

jasinner (approved):
Ok to merge as is. I hope we can make it more resilient to the addition of future root types though.

corgi/core/migrations/0040_auto_20221020_1057.py
JimFuller-RedHat: using django-mptt the answer is probably no eg. structurally we will be better placed to have more generic indexes if we use something like ltree in the future.

corgi/core/migrations/0040_auto_20221020_1057.py
JimFuller-RedHat: to aid subquery ... note that these partial queries are very cheap in terms of storage and ingestion performance eg. around 12MB ... the bigger the table the more useful having foward/backward type queries

PR #193, opened 2022-10-20
Refactor build completion time parsing to not require type:ignore

mprpic:
Minor change but submitting it separately because my purl type/namespace PR is growing to quite a large set of changes.

@RedHatProductSecurity/corgi-devs, please review.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -53,15 +53,16 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     build_meta["corgi_ingest_start_dt"] = dateformat.format(timezone.now(), "Y-m-d H:i:s")
     build_meta["corgi_ingest_status"] = "INPROGRESS"
 
-    completion_dt = None
-    if "completion_time" in build_meta:
-        completion_dt = make_aware(
-            dateparse.parse_datetime(build_meta["completion_time"].split(".")[0])  # type:ignore
-        )
+    completion_time = build_meta.get("completion_time", "")
+    if completion_time:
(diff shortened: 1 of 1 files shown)


corgi/tasks/brew.py
```diff
@@ -53,15 +53,15 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
     build_meta["corgi_ingest_start_dt"] = dateformat.format(timezone.now(), "Y-m-d H:i:s")
     build_meta["corgi_ingest_status"] = "INPROGRESS"
 
-    completion_dt = None
     if "completion_time" in build_meta:
-        completion_dt = make_aware(
-            dateparse.parse_datetime(build_meta["completion_time"].split(".")[0])  # type:ignore
-        )
+        dt = dateparse.parse_datetime(build_meta["completion_time"].split(".")[0])
```
juspence: General LGTM (and I had almost the same change locally, heh). But maybe we should do below instead:
```
    completion_time = build_meta.get("completion_time", "")
    if completion_time:
        dt = dateparse.parse_datetime(completion_time.split(".")[0])
```
That handles the case where the key is present in the build_meta dictionary but the value is set to None, so calling .split() would fail.

juspence (approved):

corgi/tasks/brew.py
JimFuller-RedHat: FWIW failure was intentional ... we do not want to ingest something without a completion_time ... granted it could be made more explicit

mprpic: Sure, fixed!

PR #194, opened 2022-10-20
CORGI-23: Add yum repos collector

juspence:
@RedHatProductSecurity/corgi-devs Basic PoC for a Yum repo collector that mostly reuses the Pulp / CDN repo collector code. We need this to collect RPMs and modules from community repos, as well as for Red Hat products that don't use Errata Tool and Pulp.

Note that for everything else, we should rely on the Pulp repo collector (AKA Yum repos that are managed by and searchable in Pulp). However, some entries in product-definitions define a yum_repositories line that is actually just a list of Pulp / CDN repo URLs.

To avoid creating two channels for the same repo, we now save the Pulp repo's relative url on ingest, then try to find a Channel with matching relative url when processing a Yum repo. If it turns out the Yum repo is really a Pulp repo and has an existing CDN_REPO channel, we'll reuse it, else we'll create a new CDN_REPO Channel for the Yum repo.

Once the channel is created, we store relations (stream names, channel / repo names, and RPM / module build IDs) for all components that are shipped to that repo. The existing Pulp collector code will (weekly) find all unprocessed CDN_REPO relations, load the module / RPM data, and create the actual Components.

Reusing this code is why I didn't create a separate YUM_REPO relation type, but it's still possible to distinguish between Pulp and Yum CDN_REPOs. Pulp-managed repos are always linked to ProductVariants, while unmanaged / "Yum" repos are linked to ProductStreams. AKA streams are what's listed in product-definitions and what define the yum_repositories line.

Tests are added now and should be in a final state, unless changes are needed as part of review.


corgi/collectors/yum.py
```diff
@@ -0,0 +1,149 @@
...
+                modules_lines.append(output[start_line_no:line_no])
+                start_line_no = line_no + 1
+
+        return [self.parse_module(lines) for lines in modules_lines]
+
+    def find_modules_from_yum_repos(
+        self, repos: list[str]
+    ) -> list[dict[str, dict[str, list[str]]]]:
+        """List and collect info for all RPM modules in particular Yum repos."""
+        command = ["dnf", "--quiet", "module", "list"]
```
mprpic: DNF has a pretty decent Python API, I wonder if it would be easier to use that instead of the raw `dnf` commands that don't output nice machine-readable text:

https://dnf.readthedocs.io/en/latest/api_queries.html

corgi/tasks/common.py
```diff
@@ -73,3 +75,23 @@ def _create_relations(build_ids, external_system_id, product_ref, relation_type)
...
+) -> tuple[subprocess.CompletedProcess, list[str]]:
+    """Simple wrapper function to securely run an external command using the subprocess module
+    Raises a CalledProcessError by default if the command returned a non-zero exit code"""
+    logger.info(f"Running external command: {command}")
+    result = subprocess.run(command, *args, **kwargs, capture_output=True, check=True, text=True)
+    logger.info(f"Command completed with result: {result}")
+    if result.stderr:
+        # Treat warnings like errors even if the command had a successful exit code
+        raise ValueError(f"Command {command} failed: {result.stderr}")
+    output = result.stdout.split("\n")
```
mprpic: Can this just strip the trailing/leading whitespace instead of trimming lines below?

```python
output = result.stdout.strip().splitlines()
```

corgi/collectors/yum.py
juspence: I probably should, thanks for the link! It's my first time working with repoquery / dnf in general, so Jason suggested copying the Deptopia logic. And obviously that's written in go so doesn't directly interface with the DNF Python library. I'll try to get to this after finishing tests.

mprpic: Don't feel like you have to spend tons of time on it though. If the current way works then it's fine as is. Only if it's trivial to use DNF through its Python API it might be worth it to refactor this.

juspence: Quick look doesn't provide any obvious way to list only modules in a repo, so going to leave as-is for now. If I missed the obvious and there is some way to do this, please feel free to file a backlog ticket.

jasinner:
LGTM from a cursory review. Could probably use some tests. Especially testing components and their product relationships in the openshift-4.x streams.

We need to exclude community projects from being ingested. Probably a simple way for now could be to only ingest from the repo if a channel already exists for it. Although I think a more robust way would be to add egress filtering to block all network traffic attempting to reach resources outside the RH firewall from the MP+ project. If we go for the later, we won't have to change the code at when we run a community version of Corgi.

corgi/tasks/yum.py
```diff
@@ -0,0 +1,63 @@
...
+    srpm_relations = _create_relations(
+        srpm_build_ids, channel.name, stream, ProductComponentRelation.Type.CDN_REPO
+    )
+    if srpm_relations > 0:
+        logger.info(f"Created {srpm_relations} new relations for SRPMs in {channel.name}")
+
+    logger.info(f"Querying Yum repository {repo} for modules")
+    module_build_ids = yum.get_modules_from_yum_repos((repo,))
+    module_relations = _create_relations(
+        module_build_ids, channel.name, stream, ProductComponentRelation.Type.CDN_REPO
```
jasinner: This will cause issues with SoftwareBuild.save_product_taxonomy(). In that function we expect CDN_REPO types to be associated with variants, not streams. Perhaps it's better to use a new type YUM_REPO and link it with a stream instead, so that we know to expect a stream name always for YUM_REPOS.

jasinner: An alternative, more invasive fix is to use ofuri in product_ref column of relations table. When reading relations back from the table we could then parse the ofuri to determine the product type it refers to.

juspence:
> LGTM from a cursory review. Could probably use some tests. Especially testing components and their product relationships in the openshift-4.x streams.

I'm a little confused. Do you mean testing what modules / RPMs get created? That only depends on the build IDs that the collector returns, which is being tested already.


> We need to exclude community projects from being ingested. Probably a simple way for now could be to only ingest from the repo if a channel already exists for it. Although I think a more robust way would be to add egress filtering to block all network traffic attempting to reach resources outside the RH firewall from the MP+ project. If we go for the later, we won't have to change the code at when we run a community version of Corgi.

I'd rather add a feature flag that makes it clear whether community-only products are being ingested or not. But I'm also confused about this - do we have any logic that might ingest community stuff? I thought at the moment we only ingested data for whatever's in prod-defs (and that there are no community products there).

corgi/tasks/yum.py
juspence: Ack, in that case I'll switch the type of the created relations to YUM_REPO, then copy the Pulp task code that finds unprocessed CDN relations to the Yum task module and make it look for unprocessed YUM relations instead.

jasinner:
> > LGTM from a cursory review. Could probably use some tests. Especially testing components and their product relationships in the openshift-4.x streams.
> 
> I'm a little confused. Do you mean testing what modules / RPMs get created? That only depends on the build IDs that the collector returns, which is being tested already.
> 
Fair enough, no need tests for existing code.

> > We need to exclude community projects from being ingested. Probably a simple way for now could be to only ingest from the repo if a channel already exists for it. Although I think a more robust way would be to add egress filtering to block all network traffic attempting to reach resources outside the RH firewall from the MP+ project. If we go for the later, we won't have to change the code at when we run a community version of Corgi.
> 
> I'd rather add a feature flag that makes it clear whether community-only products are being ingested or not. But I'm also confused about this - do we have any logic that might ingest community stuff? I thought at the moment we only ingested data for whatever's in prod-defs (and that there are no community products there).

Currently prod-defs contains community projects. There is some logic in Deptopia to filter it out based on business_unit field of ps_product, which we could also use. It also filters openstack-rdo into a distinct instance. I'm not sure what the logic is behind that, seems like it is also a community project, probably best to check with Deptopia folks such as Sam.

juspence:
Replying here: Jason approved the MR in Gchat, the filtering requested above was added in a separate MR. This wasn't merged previously because community product data had already been loaded and needed to be removed first. Now that we've reloaded stage DB data, this can get merged as soon as the pipeline passes.

PR #196, opened 2022-10-24
CORGI-321: Add filter to prevent ingesting community products

juspence:
@RedHatProductSecurity/corgi-devs Please review. This filter was requested by Jason so we don't accidentally load a bunch of community-only components / builds. After checking the data in prod-defs, it looks like only two entries are marked as community. See the Jira ticket for details.


mprpic (approved):
LGTM!

PR #197, opened 2022-10-24
CORGI-224: Refactor purl type and namespace

mprpic:
This is quite a large because it touches so many parts of the code base.
To update the default list of choices, we have to add a new attribute,
which also means an update to all the places that fill out our data
(mainly in the collectors) as well as the API layer.

The list of component types now aligns more closely to the purl spec
(with the exception of the "rpmmod" type that we have to request a new
type for). This means we'll produce purls that can be distributed
outside of Red Hat and potentially reused by other systems that also use
purl for component identification.


mprpic:
@RedHatProductSecurity/corgi-devs This is kind of ready for review (the brew collector tests are still failing and need to be updated). But feedback is welcome. I don't love that so much had to be updated to make a simple change to the list of component types, it indicates that we've spread out the logic in collectors way too much. This will also obviously require a full DB reload so I'm not even bothering with writing a data migration, but we've already discussed that last week.

corgi/api/serializers.py
```diff
@@ -347,7 +352,9 @@ def get_components(instance: Product) -> str:
 
     @staticmethod
     def get_upstreams(instance: Product) -> str:
-        return get_model_ofuri_link("components", instance.ofuri, related_type="UPSTREAM")
+        return get_model_ofuri_link(
```
juspence: My main branch may be out of date, but in my copy of the code there's a "get_upstream_link()" method that adds a "&type=UPSTREAM" parameter to the link it returns. I guess that should be changed to "&namespace=UPSTREAM" as well.

corgi/collectors/brew.py
```diff
@@ -421,6 +418,12 @@ def _extract_remote_sources(self, go_stdlib_version, remote_sources):
             )
             for pkg_type in remote_source.pkg_managers:
                 if pkg_type in ("npm", "pip", "yarn"):
+                    # Convert Cachito-reported package type to Corgi component type.
+                    pkg_type = {
+                        "npm": Component.Type.NPM,
+                        "pip": Component.Type.PYPI,
+                        "yarn": Component.Type.NPM,
+                    }.get(pkg_type)
```
juspence: Super minor nit but I'd prefer if this were a constant, and that we avoid defining the three types twice. Below only a suggestion:
```
# At top of method, or top of module - wherever's convenient
CACHITO_PKG_TYPE_MAPPING = {
    "npm": Component.Type.NPM,
    "pip": Component.Type.PYPI,
    "yarn": Component.Type.NPM
}
...
# many lines omitted
...
# get value from dict, else if not found keep current value unchanged
pkg_type = CACHITO_PKG_TYPE_MAPPING.get(pkg_type, pkg_type):
```

corgi/core/models.py
```diff
@@ -218,7 +218,6 @@ class SoftwareBuild(TimeStampedModel):
 
     class Type(models.TextChoices):
         BREW = "BREW"  # Red Hat Brew build system
-        PNC = "PNC"  # Red Hat PNC (Project Newcastle) build system
```
juspence, line 226: Any reason why this is getting removed? Are we only trying to track Red Hat (Brew) vs. community (Koji) builds, and we don't care about which Red Hat build system gets used as long as the build eventually ends up in Brew?

If that's the case, maybe we should change the enum names to RED_HAT and COMMUNITY...in another MR.

tests/test_sca.py
```diff
@@ -281,6 +281,6 @@ def test_slow_software_composition_analysis(
             type=Component.Type.CONTAINER_IMAGE, arch="noarch", software_build=sb
         )
     else:
-        root_component = Component.objects.get(type=Component.Type.SRPM, software_build=sb)
+        root_component = Component.objects.root_components().get(software_build=sb)
```
juspence: Should we also assert that type is RPM and arch is "src"? AKA assert that the component properties were set to the correct value, and not some other type that the root_components() queryset can return.

corgi/core/models.py
```diff
@@ -718,18 +717,31 @@ def get_product_details(variant_ids: list[str], stream_ids: list[str]) -> dict[s
...
+        GEM = "GEM"  # Rubygem packages
+        GENERIC = "GENERIC"  # Fallback if no other type can be used
+        GITHUB = "GITHUB"
         GOLANG = "GOLANG"
         MAVEN = "MAVEN"
         NPM = "NPM"
-        RHEL_MODULE = "RHEL_MODULE"
-        RPM = "RPM"
-        SRPM = "SRPM"
+        RHEL_MODULE = "RPMMOD"  # Not an actual purl type, propose addition in CORGI-226
```
juspence: Hmm, while reviewing some earlier code I mixed up the RHEL_MODULE enum name vs. the RPMMOD enum value. I get why we want to say Type.CONTAINER_IMAGE since just "OCI" is a bit cryptic. But can we make this enum name / value match like the others do?

Either RHEL_MODULE or RPMMOD is fine. Or if RPMMOD seems confusing, maybe we can call it RPMMODULE? I guess that's technically more correct than RHEL_MODULE since Fedora has modules too.

corgi/tasks/brew.py
```diff
@@ -332,6 +331,7 @@ def save_container(softwarebuild, build_data) -> ComponentNode:
         for module in build_data["meta"]["upstream_go_modules"]:
             new_upstream, created = Component.objects.get_or_create(
                 type=Component.Type.GOLANG,
+                namespace=Component.Namespace.UPSTREAM,
```
juspence: I think somewhere else you added "go-mod" vs "go-package" to a field in the Component's meta_attr. Should we also track that here as well?

tests/test_model.py
```diff
@@ -304,7 +308,7 @@ def test_get_upstream():
     assert container_nested.get_upstreams() == [container_source.purl]
 
     container_o_source = ComponentFactory(
-        name="contain_upstream_other", type=Component.Type.UPSTREAM
+        name="contain_upstream_other", type=Component.Namespace.UPSTREAM
```
juspence: This still has ```type=``` instead of ```namespace=```.

tests/test_model.py
```diff
@@ -288,7 +290,9 @@ def test_get_upstream():
     )
     assert container_rpm.get_upstreams() == []
 
-    container_source = ComponentFactory(name="container_upstream", type=Component.Type.UPSTREAM)
+    container_source = ComponentFactory(
+        name="container_upstream", type=Component.Namespace.UPSTREAM
```
juspence: This still has ```type=``` instead of ```namespace=```.

tests/test_api.py
```diff
@@ -336,16 +343,23 @@ def test_re_purl_filter(client, api_path):
...
 
 def test_nvr_nevra_filter(client, api_path):
-    c1 = ComponentFactory(type=Component.Type.RPM, meta_attr={"epoch": "1"}, name="autotrace-devel")
+    c1 = ComponentFactory(
+        type=Component.Type.RPM,
+        meta_attr={"epoch": "1"},
+        name="autotrace-devel",
+        version="3.2.1",
+        release="1.0.1e",
+        arch="noarch",
```
juspence, line 352: Does arch="noarch" make sense for RPMs? I thought that was only for CONTAINER_IMAGEs. Should it be arch="src" instead?

juspence (approved):
LGTM once tests are fixed.

corgi/core/models.py
mprpic: No, we can track the individual systems under their names, so Brew and Koji, but we have no code that pulls anything out of PNC so it's unused. Also, all PNC builds are imported into Brew so it's questionable if we will ever have PNC builds represented in our DB.

tests/test_api.py
mprpic: It does make sense; most Python packages for example are not arch-specific so use noarch. Check out e.g. Brew build 1399682.

corgi/core/models.py
mprpic: The value should match w/e is the equivalent purl type, the name is really arbitrary and should match w/e we prefer to see the value called throughout the code base. I suppose if we do support Fedora modules down the line it would be better to call it RPMMOD. I just hope that that is clear enough for everyone who suddenly has to context switch from thinking about RHEL modules to RPM modules. @jasinner, thoughts?

tests/test_model.py
mprpic: Good eye, fixed!

tests/test_model.py
mprpic: Fixed!

tests/test_sca.py
mprpic: Ah yes, that was supposed to call `.srpms()` not `.root_components()`. Fixed!

corgi/collectors/brew.py
mprpic: I also combined the listing with some of the types in `_extract_bundled_provides()`, see 3109367.

corgi/api/serializers.py
mprpic: Yep, my branch wasn't rebased on master. Fixed now!

mprpic:
I fixed up all the tests except these three:
```
       3 failed
         - tests/test_brew_collector.py:349 test_extract_remote_sources
         - tests/test_brew_collector.py:373 test_extract_multiple_remote_sources
         - tests/test_brew_collector.py:696 test_fetch_container_build_rpms
```
The first two assert some random counts from the associated file but I have no idea what those counts refer to, the last one asserts that a method is *not* called but I have no idea why it shouldn't be called. @jasinner @JimFuller-RedHat, could you help me figure out these three? In the code they're all marked with a `TODO FAIL` comment.

corgi/tasks/brew.py
mprpic: The `upstream_go_modules` dict is actually set in the collector in:

```python
            # Legacy OSBS builds such as 1890187 copy source code into dist-git but specify where
            # the source code came from using the 'go' stanza in container.yaml
            # ref: https://osbs.readthedocs.io/en/osbs_ocp3/users.html#go
            go_modules = []
            if "go" in build_info["extra"]["image"]:
                for module in build_info["extra"]["image"]["go"]["modules"]:
                    if "module" in module:
                        go_modules.append(module["module"])
            if go_modules:
                component["meta"]["upstream_go_modules"] = go_modules
```

That's a different code path from the one that sets up proper Go components (which will have their `meta` object stored since it goes through the other path that checks `sources` (`...defaults={"meta_attr": source["meta"]...`). So for these modules we don't have much information other than the name. @jasinner, do we need to store anything more in the above `go_modules` list?

corgi/core/models.py
jasinner: I would probably keep RPMMOD and change the enum name to the same "RPMMOD" as well.

corgi/tasks/brew.py
jasinner: Unfortunately those legacy OSBS builds are lacking in upstream metadata. The go stanza is really all we have to go on. There was actually a whole project [retrodep](https://github.com/release-engineering/retrodep) around trying to reverse engineer the commit in use. I seem to recall that project wrote the information into the container image. There builds are all pretty old though, so I don't think we need to extract that.

jasinner:
> I fixed up all the tests except these three:
> 
> ```
>        3 failed
>          - tests/test_brew_collector.py:349 test_extract_remote_sources
>          - tests/test_brew_collector.py:373 test_extract_multiple_remote_sources
>          - tests/test_brew_collector.py:696 test_fetch_container_build_rpms
> ```
> 
> The first two assert some random counts from the associated file but I have no idea what those counts refer to, the last one asserts that a method is _not_ called but I have no idea why it shouldn't be called. @jasinner @JimFuller-RedHat, could you help me figure out these three? In the code they're all marked with a `TODO FAIL` comment.

Sure, I'll push a new commit to fix these up later in my day today.

corgi/tasks/sca.py
```diff
@@ -69,19 +69,14 @@ def save_component(component: dict[str, Any], parent: ComponentNode):
...
-            software_build=software_build,
-            type=Component.Type.CONTAINER_IMAGE,
-            arch="noarch",
-        )
+
+    # Get root component for this build; fail the task if it does not exist.
+    # TODO: ditch type:ignore when https://github.com/typeddjango/django-stubs/pull/1025 is released
+    root_component = (
+        Component.objects.filter(software_build=software_build)
+        .root_components()  # type:ignore
```
jasinner, line 77: Nice, thanks for splitting this into a distinct filter!

corgi/core/models.py
mprpic: Ack, I'll fix that up first thing tomorrow morning.

corgi/core/models.py
```diff
@@ -793,6 +806,8 @@ class Namespace(models.TextChoices):
     data_score = models.IntegerField(default=0)
     data_report = fields.ArrayField(models.CharField(max_length=200), default=list)
 
+    objects = ComponentQuerySet.as_manager()
```
jasinner, line 809: Why did you add objects as a specific declared field of Component? How does this change the behavior of the regular (built-in) Django model manager?

corgi/core/models.py
```diff
@@ -863,45 +873,39 @@ def get_purl(self) -> PackageURL:
                     digest = self.meta_attr["digests"].get(digest_fmt)
                     if digest:
                         break
-            return PackageURL(
-                type="container",
-                namespace=self.Namespace.REDHAT.lower(),
+            purl_data = dict(
```
jasinner, line 905: CONTAINER_IMAGE types should have their version set to the digest instead of `f"{self.version}-{self.release}`. According to https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci

jasinner (changes requested):
Thanks so much for doing this PR. It includes many changes, but it's a worthwhile change to improve the key output of the application, the purls! Let's just make one adjustment to the OCI purl version that would be great.

I'll also push a commit to fix up the brew collector tests.

corgi/core/models.py
jasinner: The value of `f"{self.version}-{self.release}` can be included as a tag qualifier.

corgi/core/models.py
jasinner: Nevermind, I see you are using this to add `srpms()`, and `root_components` filter methods.

jasinner:
> > I fixed up all the tests except these three:
> > ```
> >        3 failed
> >          - tests/test_brew_collector.py:349 test_extract_remote_sources
> >          - tests/test_brew_collector.py:373 test_extract_multiple_remote_sources
> >          - tests/test_brew_collector.py:696 test_fetch_container_build_rpms
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The first two assert some random counts from the associated file but I have no idea what those counts refer to, the last one asserts that a method is _not_ called but I have no idea why it shouldn't be called. @jasinner @JimFuller-RedHat, could you help me figure out these three? In the code they're all marked with a `TODO FAIL` comment.
> 
> Sure, I'll push a new commit to fix these up later in my day today.

I didn't finish this yet sorry. Still need to fix `tests/test_brew_collector.py:test_extract_multiple_remote_sources`

corgi/core/models.py
jasinner: I fixed this up

corgi/core/models.py
mprpic: Renamed!

mprpic:
> I didn't finish this yet sorry. Still need to fix tests/test_brew_collector.py:test_extract_multiple_remote_sources

@jasinner I fixed up the last failing test, it was an easy one to update. Thanks for fixing the others!

The last thing that remains is to refresh the migration to apply the new values. I did notice though that you moved the custom indexes from 0040 into the migration here. I think those indexes need to be updated but I'm not sure what specific queries they were targeting in the first place. @JimFuller-RedHat, can you help us understand how those indexes need to change now that we have different values for what is an SRPM and container image?

Specifically these indexes:
```
        migrations.RunSQL(
            "CREATE INDEX core_compon_latest_name_type_idx ON public.core_component \
             USING btree (name, type) WHERE (((type)::text = 'SRPM'::text) OR \
             (((arch)::text = 'noarch'::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
             OR ((type)::text = 'RHEL_MODULE'::text))"
        ),
        migrations.RunSQL(
            "CREATE INDEX core_compon_latest_type_name_idx ON public.core_component \
             USING btree (type, name) WHERE (((type)::text = 'SRPM'::text) OR \
             (((arch)::text = 'noarch'::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
             OR ((type)::text = 'RHEL_MODULE'::text))"
        ),
        migrations.RunSQL(
            "CREATE INDEX core_compon_latest_idx ON public.core_component \
            USING btree (uuid, software_build_id, type, name, product_streams) \
            WHERE (((type)::text = 'SRPM'::text) OR (((arch)::text = 'noarch'::text) \
            AND ((type)::text = 'CONTAINER_IMAGE'::text)) OR ((type)::text = 'RHEL_MODULE'::text))"
        ),
```

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -43,4 +43,25 @@ class Migration(migrations.Migration):
             name="type",
             field=models.CharField(choices=[("BREW", "Brew"), ("KOJI", "Koji")], max_length=20),
         ),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_name_type_idx"),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_type_name_idx"),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_idx"),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_name_type_idx ON public.core_component \
+             USING btree (name, type) WHERE (((type)::text = 'SRPM'::text) OR \
```
juspence:
```suggestion
             USING btree (name, type) WHERE ((((type)::text = 'RPM'::text) AND ((arch)::text = 'src')) OR \
```

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -43,4 +43,25 @@ class Migration(migrations.Migration):
...
+             (((arch)::text = ''::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_type_name_idx ON public.core_component \
+             USING btree (type, name) WHERE (((type)::text = 'SRPM'::text) OR \
```
juspence:
```suggestion
             USING btree (type, name) WHERE ((((type)::text = 'RPM'::text) AND ((arch)::text = 'src')) OR \
```

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -43,4 +43,25 @@ class Migration(migrations.Migration):
...
+             (((arch)::text = ''::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_idx ON public.core_component \
+            USING btree (uuid, software_build_id, type, name, product_streams) \
+            WHERE (((type)::text = 'SRPM'::text) OR (((arch)::text = ''::text) \
```
juspence:
```suggestion
            WHERE ((((type)::text = 'RPM'::text) AND ((arch)::text = 'src')) OR (((arch)::text = ''::text) \
```

juspence (commented):
I think below suggestions will fix the index. But maybe we need to add ```, arch,``` to the USING clause now that we filter on arch = '' for container images and arch = 'src' for SRPMs.

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -0,0 +1,67 @@
...
+            field=models.CharField(choices=[("BREW", "Brew"), ("KOJI", "Koji")], max_length=20),
+        ),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_name_type_idx"),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_type_name_idx"),
+        migrations.RunSQL("DROP INDEX IF EXISTS core_compon_latest_idx"),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_name_type_idx ON public.core_component \
+             USING btree (name, type) WHERE (((type)::text = 'SRPM'::text) OR \
+             (((arch)::text = ''::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
```
juspence:
```suggestion
             OR ((type)::text = 'RPMMOD'::text))"
```

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -0,0 +1,67 @@
...
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_type_name_idx ON public.core_component \
+             USING btree (type, name) WHERE (((type)::text = 'SRPM'::text) OR \
+             (((arch)::text = ''::text) AND ((type)::text = 'CONTAINER_IMAGE'::text)) \
+             OR ((type)::text = 'RHEL_MODULE'::text))"
```
juspence:
```suggestion
             OR ((type)::text = 'RPMMOD'::text))"
```

corgi/core/migrations/0041_auto_20221024_1844.py
```diff
@@ -0,0 +1,67 @@
...
+        ),
+        migrations.RunSQL(
+            "CREATE INDEX core_compon_latest_idx ON public.core_component \
+            USING btree (uuid, software_build_id, type, name, product_streams) \
+            WHERE (((type)::text = 'SRPM'::text) OR (((arch)::text = ''::text) \
+            AND ((type)::text = 'CONTAINER_IMAGE'::text)) OR ((type)::text = 'RHEL_MODULE'::text))"
```
juspence:
```suggestion
            AND ((type)::text = 'CONTAINER_IMAGE'::text)) OR ((type)::text = 'RPMMOD'::text))"
```

jasinner:
I adjusted the indexes to use django indexes in the models. I would still appreciate a look over from @JimFuller-RedHat 

@mprpic FYI I also forced push to this branch to remove hashes from the component_data.json file which were causing secrets tox target to fail locally.

corgi/core/migrations/0041_auto_20221024_1844.py
mprpic: Jason and I discussed this over chat, he pushed a new commit that turns these into model indexes. See latest commit.

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
...
+                    | Q(Q(type="OCI") & Q(arch=""))
+                ),
+            ),
+            models.Index(
+                fields=("name", "type"),
+                name="compon_latest_type_name_idx",
+                condition=Q(
+                    Q(Q(type="RPM") & Q(arch="src"))
+                    | Q(type="RPMMOD")
+                    | Q(Q(type="OCI") & Q(arch=""))
```
juspence:
```suggestion
                    Q(type=Component.Type.RPM, arch="src")
                    | Q(type=Component.Type.RPMMOD)
                    | Q(type=Component.Type.CONTAINER_IMAGE, arch="")
```

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
             models.Index(fields=["product_streams"]),
             models.Index(fields=["product_variants"]),
             models.Index(fields=("type", "product_streams")),
+            models.Index(
+                fields=("type", "name"),
+                name="compon_latest_name_type_idx",
+                condition=Q(
+                    Q(Q(type="RPM") & Q(arch="src"))
+                    | Q(type="RPMMOD")
+                    | Q(Q(type="OCI") & Q(arch=""))
```
juspence:
```suggestion
                    Q(type=Component.Type.RPM, arch="src")
                    | Q(type=Component.Type.RPMMOD)
                    | Q(type=Component.Type.CONTAINER_IMAGE, arch="")
```

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
...
+                ),
+            ),
+            models.Index(
+                fields=("uuid", "software_build_id", "type", "name", "product_streams"),
+                name="compon_latest_idx",
+                condition=Q(
+                    Q(Q(type="RPM") & Q(arch="src"))
+                    | Q(type="RPMMOD")
+                    | Q(Q(type="OCI") & Q(arch=""))
```
juspence:
```suggestion
                    Q(type=Component.Type.RPM, arch="src")
                    | Q(type=Component.Type.RPMMOD)
                    | Q(type=Component.Type.CONTAINER_IMAGE, arch="")
```

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
             models.Index(fields=["product_streams"]),
             models.Index(fields=["product_variants"]),
             models.Index(fields=("type", "product_streams")),
+            models.Index(
+                fields=("type", "name"),
```
juspence: Not sure if below is needed, but we do want to filter on it...
```suggestion
                fields=("type", "name", "arch"),
```

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
...
+                name="compon_latest_name_type_idx",
+                condition=Q(
+                    Q(Q(type="RPM") & Q(arch="src"))
+                    | Q(type="RPMMOD")
+                    | Q(Q(type="OCI") & Q(arch=""))
+                ),
+            ),
+            models.Index(
+                fields=("name", "type"),
```
juspence:
```suggestion
                fields=("name", "type", "arch"),
```

corgi/core/models.py
```diff
@@ -819,27 +825,52 @@ class Meta:
...
+                name="compon_latest_type_name_idx",
+                condition=Q(
+                    Q(Q(type="RPM") & Q(arch="src"))
+                    | Q(type="RPMMOD")
+                    | Q(Q(type="OCI") & Q(arch=""))
+                ),
+            ),
+            models.Index(
+                fields=("uuid", "software_build_id", "type", "name", "product_streams"),
```
juspence:
```suggestion
                fields=("uuid", "software_build_id", "type", "name", "arch", "product_streams"),
```

juspence (approved):
Still LGTM, index suggestions are minor but it would be nice to use those enum names we spent so much effort on :)

jasinner:
I couldn't use the Component.Type in Meta because it's not defined yet. I did add arch to fields in the indexes though. Good catch. @JimFuller-RedHat Please review the indexes.

I also adjusted the purl name to use last part of the repository_url path. Also added repository_url as a qualifier. See https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci

jasinner (approved):

jasinner:
We should probably fix this [bug related to rhel_modules](https://issues.redhat.com/browse/CORGI-316) before we do our data refresh:

JimFuller-RedHat:
indexes look good and glad to see we can represent partial indexes normally (instead as a constraint!) using condition (which I did not know).

jasinner:
When testing this branch locally I noticed an issue with CONTAINER_IMAGE arch = "". It's very difficult to search for an empty string using the API. Therefore I reverted the commit which switched arch for IMAGE INDEX containers from "" to "noarch".

corgi/collectors/brew.py
```diff
@@ -499,14 +515,19 @@ def _extract_image_components(
         child_component["rpm_components"] = arch_specific_rpms
         return noarch_rpms_by_id, child_component
 
+    def _get_name_label(self, docker_config):
+        config = docker_config.get("config", {})
+        labels = config.get("Labels", {})
+        return labels.get("name")
```
jasinner: Just a reminder for myself. This needs a default value such as "".

jasinner:
I adjust a few commits to improve repository_name for container purls.

mprpic:
The last three commits adjust the purls for rpmmod to be in line with the purl spec proposal. I also removed arch from the qualifier since our RHEL module builds don't provide it (only the included individual RPMs are built for specific arches).

PR #198, opened 2022-10-25
CORGI-308: Add manifest for components

juspence:
@RedHatProductSecurity/corgi-devs Please review. This adds a generic manifest that can be used for any component (including UBI) to list all its PROVIDES and PROVIDES_DEV relationships.

Based on information in the user guide, we translate Corgi's PROVIDES relationship into SPDX's "CONTAINS" relationship (e.g. for the CONTAINER_IMAGE being manifested, which contains embedded RPMs). We translate Corgi's PROVIDES_DEV relationship into SPDX's "BUILD_DEPENDENCY_OF" relationship (e.g. for the RPM being manifested, which is used to build the RPMs in the provides list).

Note that the ticket mentions UBI, but during a meeting we were asked to implement this manifesting capability for any component. If we end up using a library to generate the manifest, the logic to generate it may change a bit but the relationships, model properties, and manifest structure should all stay the same.

Going to add sample data to the Jira ticket in a bit. A future MR can also tweak the product manifest, e.g. to reference a component manifest's SPDX ID instead of including that information directly.


corgi/core/models.py
```diff
@@ -1020,18 +1024,24 @@ def get_channels(self):
...
         type_list = [ComponentNode.ComponentNodeType.PROVIDES]
         if include_dev:
             type_list.append(ComponentNode.ComponentNodeType.PROVIDES_DEV)
-        for cnode in self.cnodes.get_queryset():
-            if cnode.get_descendant_count() == 0:
-                continue
-            for descendant in cnode.get_descendants().filter(type__in=type_list):
-                provides.add(descendant.purl)
-        return list(provides)
+        return self.cnodes.get_queryset().get_descendants().filter(type__in=type_list)
```
jasinner, line 1033: I'm gladly surprised that we don't have to iterate over the cnodes. I guess this must return the same results because you didn't have to alter the tests.

I'd be tempted to double check that a component with multiple cnodes (like a container with rpms and other types) still returns the same list of provides as before.

jasinner (approved):
Looks great. Thank you. 

We could probably consolidate the 2 templates into sub-templates in order to reuse them, but let's do it in a following PR.

JimFuller-RedHat (approved):
LGTM

corgi/core/models.py
juspence: I ran the below code which seems to still work:
```
component_qs = Component.objects.get_queryset().iterator(chunk_size=500)

for component in component_qs:
     node_qs = component.cnodes.get_queryset()
     node_count = node_qs.count()

     if node_count < 2:
         continue

     descendant_count = node_qs.get_descendants().count()

     if descendant_count < 2:
         continue

     print("Descendants:", descendant_count)
     print("Nodes:", node_count)

     for node in node_qs:
         print(node.get_descendants().count())

     break
```
I got the following output, AKA calling get_descendants() on the whole queryset or on each individual node in the queryset gives the same result either way:
```
Descendants: 28
Nodes: 4
7
7
7
7
```

PR #204, opened 2022-10-27
CORGI-238 product manifest includes provided relationships

jasinner:
Include all the provided relationships for each root level component in the product manifests.

- Validates product and component manifest against [SPDX 2.2.2 schema](https://raw.githubusercontent.com/spdx/spdx-spec/development/v2.2.2/schemas/spdx-schema.json)
- Utilized django templates blocks and includes to avoid repeating ourselves (DRY)

@RedHatProductSecurity/corgi-devs

### DIFF
--- a/corgi/web/templates/base_manifest.json
+++ b/corgi/web/templates/base_manifest.json
@@ -1,3 +1,4 @@
+{# Used as the base of all manifest.json templates #}
 {
     "creationInfo": {
         "created": "{% now 'Y-m-d' %}T{% now 'H:i:00' %}Z",
--- a/corgi/web/templates/base_manifest.json
+++ b/corgi/web/templates/base_manifest.json
@@ -1,0 +1,28 @@
+{
+    "creationInfo": {
+        "created": "{% now 'Y-m-d' %}T{% now 'H:i:00' %}Z",
+        "creators": [
+            "Organization: Red Hat Product Security (secalert@redhat.com)"
+        ],
+        "licenseListVersion": "3.8"
+    },
+    "dataLicense": "CC-BY-4.0",
+    "documentDescribes": [
+        "SPDXRef-{{obj.uuid}}"
+    ],
+    "documentNamespace": "https://access.redhat.com/security/data/manifest/spdx/{{obj.name}}-{{obj.uuid}}",
+    "name": "{{obj.name}}",
+    "packages": [
+      {% block packages %}{% endblock %}
+    ],
+    "relationships": [
+       {% block relationships %}{% endblock %}
+       {
+          "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# Document describes component being manifested #}
+          "relationshipType": "DESCRIBES",
+          "spdxElementId": "SPDXRef-DOCUMENT"
+       }
+    ],
+    "SPDXID": "SPDXRef-DOCUMENT",
+    "spdxVersion": "SPDX-2.2"
+}
--- a/corgi/web/templates/base_manifest.json
+++ b/corgi/web/templates/base_manifest.json
@@ -18,7 +18,7 @@
     "relationships": [
        {% block relationships %}{% endblock %}
        {
-          "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# Document describes component being manifested #}
+          "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# Document describes stream or component being manifested #}
           "relationshipType": "DESCRIBES",
           "spdxElementId": "SPDXRef-DOCUMENT"
        }
--- a/corgi/web/templates/provided_relationships_manifest.json
+++ b/corgi/web/templates/provided_relationships_manifest.json
@@ -1,0 +1,11 @@
+{% for node in obj.get_provides_nodes %}
+    {
+      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# subcomponent is built from, or contained in, component #}
+      "relationshipType": {% if node.type == "PROVIDES_DEV" %}"DEV_DEPENDENCY_OF"{% else %}"CONTAINED_BY"{% endif %},
+      "spdxElementId": "SPDXRef-{{node.obj.uuid}}"
+    },
+    {
+      "relatedSpdxElement": "NONE",{# subcomponent is a leaf node #}
+      "relationshipType": "CONTAINS",
+      "spdxElementId": "SPDXRef-{{node.obj.uuid}}"
+    },{% endfor %}
--- a/tests/test_manifests.py
+++ b/tests/test_manifests.py
@@ -30,7 +30,7 @@ pytestmark = pytest.mark.unit
 def test_product_manifest_properties():
     """Test that all models inheriting from ProductModel have a .manifest property
     And that it generates valid JSON."""
-    component, stream = setup_products_and_components()
+    component, stream, provided, dev_provided = setup_products_and_components()
 
     manifest = json.loads(stream.manifest)
 
@@ -45,15 +45,19 @@ def test_product_manifest_properties():
         assert False
 
     # Test will fail with JSONDecodeError if above isn't valid
-    # Eventually, we should also check the actual manifest content is valid SPDX data
-    # Then most of below can go away
 
     # One component linked to this product
     num_components = len(stream.get_latest_components())
     assert num_components == 1
 
-    # Manifest contains info for all components + the product itself
-    assert len(manifest["packages"]) == 2
+    num_provided = 0
+    for component in stream.get_latest_components():
+        num_provided += component.get_provides_nodes().count()
+
+    assert num_provided == 2
+
+    # Manifest contains info for all components, their provides, and the product itself
+    assert len(manifest["packages"]) == num_components + num_provided + 1
 
     # Last "component" is actually the product
     component_data = manifest["packages"][0]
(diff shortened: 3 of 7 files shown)


JimFuller-RedHat:
might be worth isolating this work into an SPDXv# directory ... otherwise LGTM ... its great to see maturation of manifesting in general

corgi/web/templates/provided_relationships_manifest.json
```diff
@@ -0,0 +1,11 @@
+{% for node in obj.get_provides_nodes %}
+    {
+      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# subcomponent is built from, or contained in, component #}
+      "relationshipType": {% if node.type == "PROVIDES_DEV" %}"DEV_DEPENDENCY_OF"{% else %}"CONTAINED_BY"{% endif %},
+      "spdxElementId": "SPDXRef-{{node.obj.uuid}}"
```
juspence, line 5: This is backwards - obj.uuid / component being manifested is a DEPENDENCY_OF node.obj.uuid / subcomponent in the PROVIDES list, not the other way around. I think there's no inverse DEV_DEPENDENCY_OF relationship, so let's flip this to match the other relationships in the document.

Those start with the higher-level element identified in spdxElementId (e.g. DOCUMENT, or object being manifested), and relate that using relationshipType to a lower-level element identified in relatedSpdxEelement (e.g. object being manifested, or subcomponent in a PROVIDES listing).

I'm also not sure DEV_DEPENDENCY_OF is the right relationship. According to docs/user_guide.md:
```
- `provides`: an inverse relationship to the `sources` list, a list of components that this component provides,
   either as components built from this component, or components embedded in this component.
```
So based on above, I used BUILD_DEPENDENCY_OF as the relationship since the parent component is required to build the subcomponent in the PROVIDES list.

```suggestion
      "relatedSpdxElement": "SPDXRef-{{node.obj.uuid}}",{# subcomponent is built from, or contained in, component #}
      "relationshipType": {% if node.type == "PROVIDES_DEV" %}"BUILD_DEPENDENCY_OF"{% else %}"CONTAINS"{% endif %},
      "spdxElementId": "SPDXRef-{{obj.uuid}}"
```

corgi/web/templates/base_manifest.json
```diff
@@ -0,0 +1,26 @@
...
+        ],
+        "licenseListVersion": "3.8"
+    },
+    "dataLicense": "CC-BY-4.0",
+    "documentDescribes": [
+        "SPDXRef-{{obj.uuid}}"
+    ],
+    "documentNamespace": "https://access.redhat.com/security/data/manifest/spdx/{{obj.name}}-{{obj.uuid}}",
+    "name": "{{obj.name}}",
+    {% block packages %}{% endblock %}
```
juspence: Minor nit: I see we have e.g. the relationships key below in this document, and then the included template only iterates over the relationships. But we have the packages key in the included template - can we move it up to this document instead?

tests/test_manifests.py
```diff
@@ -22,53 +28,35 @@
...
 
+    # From https://raw.githubusercontent.com/spdx/spdx-spec/development/
+    # v2.2.2/schemas/spdx-schema.json
+    with open("tests/data/spdx-22-spec.json", "r") as spec_file:
+        schema = json.load(spec_file)
+    try:
+        jsonschema.validate(manifest, schema)
+    except jsonschema.exceptions.ValidationError as e:
+        print(e)
+        assert False
```
juspence:
```suggestion
    jsonschema.validate(manifest, schema)
```

tests/test_manifests.py
```diff
@@ -97,16 +85,140 @@ def test_product_manifest_properties():
...
+    assert manifest["relationships"][0] == provided_contained_by_component
+    assert manifest["relationships"][1] == provided_contains_nothing
+    assert manifest["relationships"][2] == dev_provided_dependency_of_component
+    assert manifest["relationships"][3] == dev_provided_contains_nothing
+    assert manifest["relationships"][-2] == component_is_package_of_product
     assert manifest["relationships"][-1] == document_describes_product
+
+
+def test_component_manifest_properties():
+    """Test that all models inheriting from ProductModel have a .manifest property
```
juspence:
```suggestion
    """Test that all Components have a .manifest property
```

juspence (approved):
LGTM I guess. But it was really confusing and hard to follow how each template is included where. Do we really expect to reuse or extend these templates often enough that deduplicating them like this is useful?

I thought one manifest for components and one for products would be enough for most future needs, so I didn't expect these to change much. Maybe there's some useful information we can pull out (e.g. the top of the manifest with basic SPDX metadata), and the rest can be a giant component / product manifest with some duplication.

Or maybe we can add a {# header comment #} at the top of each template explaining where it's included from and what it does / how it's used.

jasinner:
> LGTM I guess. But it was really confusing and hard to follow how each template is included where. Do we really expect to reuse or extend these templates often enough that deduplicating them like this is useful?
> 
> I thought one manifest for components and one for products would be enough for most future needs, so I didn't expect these to change much. Maybe there's some useful information we can pull out (e.g. the top of the manifest with basic SPDX metadata), and the rest can be a giant component / product manifest with some duplication.
> 
> Or maybe we can add a {# header comment #} at the top of each template explaining where it's included from and what it does / how it's used.

I think avoiding duplication in these templates makes working with them easier. Although I do agree that it's now harder to get an overview of the entire manifest. I will add some headers to each template to make that a bit easier.

corgi/web/templates/base_manifest.json
jasinner: I don't think we can move the packages blocks up to the base_manifest. Both product_manifest.json and component_manifest.json have different packages blocks. The product_manifest one relates each component to it's subcomponents whereas the component one only has a single layer of nesting, so only subcomponents in the one component we are describing.

tests/test_manifests.py
jasinner: Ok, I'll change this to just throw the error instead of an assertion/failure.

corgi/web/templates/provided_relationships_manifest.json
jasinner: The specification and documentation is not very clear here, but I think that relatedSpdxElement is on the right side of the relationship when it's all in one line. I think that because following [RDF example](https://spdx.github.io/spdx-spec/v2.2.2/relationships-between-SPDX-elements/#111-relationship-field) which embeds a relationship inside a file element:

```
<File rdf:about="#SPDXRef-45">
  <relationship>
    <Relationship>
      <relatedSpdxElement>
        <File rdf:about="http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82..."/>
      </relatedSpdxElement>
      <relationshipType>http://spdx.org/rdf/terms#relationshipType_contains
      </relationshipType>
    </Relationship>
  </relationship>
```

I read this as `SPDXRef-45 CONTAINS spdx-tools-v1.2`. Notice the `spdx-tools-v1.2` has the 'relatedSpdxElement' label here, the same as the JSON element above.

Given that, I think I have the relationship around the right way. The [spec ](https://spdx.github.io/spdx-spec/v2.2.2/relationships-between-SPDX-elements/#111-relationship-field) expands on the 'DEV_DEPENCY_OF relationship by saying "Is to be used when SPDXRef-A is a development dependency of SPDXRef-B." 

Here's an example of an actualy PROVIDES_DEV relationship from one of our containers [acm-governance-policy-addon-controller-container](https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2157159) produced with the existing code.

```
    {
      relatedSpdxElement: SPDXRef-38049df5-0dd3-4859-a450-3ab63e2a4b65,
      relationshipType: DEV_DEPENDENCY_OF,
      spdxElementId: SPDXRef-d291152d-7a4e-49d0-9249-a49a7f686961
    },
```

Expanding the SpdxIds:
```
$ cat manifest.1 | jq '.packages[] | select(.SPDXID=="SPDXRef-38049df5-0dd3-4859-a450-3ab63e2a4b65") | .externalRefs'
[
  {
    "referenceCategory": "PACKAGE_MANAGER",
    "referenceLocator": "pkg:oci/search-api-container@sha256:cee5135056464f3ea5d08cd12a8eb49b48de29b9f82ca62a2a4aced9e9823bba?tag=v2.6.1-1",
    "referenceType": "purl"
  }
]
cat manifest.1 | jq '.packages[] | select(.SPDXID=="SPDXRef-d291152d-7a4e-49d0-9249-a49a7f686961") | .externalRefs'
[
  {
    "referenceCategory": "PACKAGE_MANAGER",
    "referenceLocator": "pkg:npm/whatwg-encoding@1.0.5",
    "referenceType": "purl"
  }
]
```

It reads: `npm/whatwg-encoding DEV_DEPENDENCY_OF search-api-container` which is correct.
```
    {
      relatedSpdxElement: search-api-container,
      relationshipType: DEV_DEPENDENCY_OF,
      spdxElementId: npm/whatwg-encoding
    },
```

Just another [example](https://github.com/spdx/tools-java/blob/92e1c5d29eacb0081139b0c05e5e6270b231788c/testResources/SPDXJSONExample-v2.2.spdx.json#L252-L256) to show the relatedSpdxElement should be on the right hand side:
```
{
    "spdxElementId" : "SPDXRef-DOCUMENT",
    "relatedSpdxElement" : "SPDXRef-Package",
    "relationshipType" : "DESCRIBES"
  }
```



You can read more about the PROVIDES_DEV relationship we get from [OSBS here](https://github.com/containerbuildsystem/cachito/#feature-definitions).

tests/test_manifests.py
jasinner: Also changed.

corgi/web/templates/base_manifest.json
juspence: Sorry, I should have added an example. I'm looking at the logic for the packages block in this template:
```
{% block packages %}{% endblock %}    
```

Which is different from the relationships block:
```
    "relationships": [
       {% block relationships %}{% endblock %}
    ],
```

Let's add the key (not the values in the array) to this template so they're both consistent:
```suggestion
    "packages": [
       {% block packages %}{% endblock %}
    ],
```

corgi/web/templates/provided_relationships_manifest.json
juspence: tl;dr - you're right, I am very confused. There may be a bug in our docs / relationship types, or maybe I'm still confused. Maybe we should use BUILD_DEPENDENCY_OF, otherwise code you have here LGTM.

> The specification and documentation is not very clear here, but I think that relatedSpdxElement is on the right side of the relationship when it's all in one line. 
> I read this as SPDXRef-45 CONTAINS spdx-tools-v1.2. Notice the spdx-tools-v1.2 has the 'relatedSpdxElement' label here, the same as the JSON element above.

Yep, I'm reading this the same way. I think my confusion was the difference between PROVIDES (contains / contained_by) and PROVIDES_DEV. For the first, I originally said ```container CONTAINS component```, so the provided object is "below" ithe container in the hierarchy. From the container, you get the component.

But for PROVIDES_DEV, we'd say ```component is a DEPENDENCY_OF container``` that's needed to build the container, so the provided object is "above" the container in the hierarchy. From the component, you get the container.

After thinking through this a bit too long, what you have (using CONTAINED_BY instead of CONTAINS and flipping the ordering from my original version) seems right. The provided component is always on the left side, and RELATES_TO the component being manifested on the right side, either as a DEPENDENCY_OF or something CONTAINED_BY something else. Thanks for spending the time to generate those examples, and for catching my mistake :)

The info in docs/user_guide.md may also need to be updated:
```
`provides`: ....a list of components that this component provides,
   ....as components built from this component
```
(Below is considering only PROVIDES_DEV)
For type=rpm, this relationship makes sense since the SRPM "provides" the RPMs. AKA the RPMs are built from the SRPM, and the SRPM is a BUILD_DEPENDENCY_OF the RPMs.

But for type=container_image, isn't it the other way around? The container "provides" the RPMs, but the container is built from the RPMs, so each RPM should be a BUILD_DEPENDENCY_OF the container. The RPMs aren't built from the container, as the docs imply.

Do we need another relationship type to distinguish this? Otherwise we'll have one relationship type that can be interpreted two different ways, depending on if you're talking about RPMs or containers.

I'd also like to use BUILD_DEPENDENCY_OF as the relationship in the SPDX manifest, since I think that's more specific. My understanding is that all the devDependencies we get from OSBS are being used to build something else, for example SRPMs building RPMs or RPMs building containers.

I assume OSBS just doesn't distinguish between build / dev dependencies the way SPDX does. But maybe we don't know for sure that every devDependency / PROVIDES component is always for building, and never for development / testing. In that case using DEV_DEPENDENCY_OF is fine, but let's update the user guide to make this clear (in a separate bug ticket - I can file if you think it's needed).

jasinner: I think the user_guide for "provides" is OK. The relationship type itself could be split out into more types, but it seems like overkill for the complexity it will create.

OSBS doesn't distinguish between build and dev dependencies the way SPDX does.

corgi/web/templates/base_manifest.json
jasinner: Ok, I'll add the packages array declaration to the base template and include the packages block inside it.

PR #206, opened 2022-10-28
CORGI-283: Tweak Content-Security-Policy to simplify ESS compliance

juspence:
...is what I would like this MR to do, but it's not complete yet. "unsafe-inline" in our CSP config means that we aren't actually protected from Cross-Site Scripting attacks. But it seems unlikely this is exploitable.

An attacker who includes JavaScript code in a Brew build license string or Erratum description could theoretically run that JavaScript code when it's displayed in Corgi. However, we can probably rely on the Django-REST-Framework serializer to prevent this kind of code execution, and I'll note that in the ESS assessment.

I was sort of hoping to fully rule out this issue by writing stronger CSP directives, but these break Django-REST-Framework because it requires inline scripts and CSS styles. There are open PRs that fix this, but they haven't gotten much attention:
https://github.com/encode/django-rest-framework/pull/7016

@RedHatProductSecurity/corgi-devs Please review. The last commit is optional and could be excluded if we like the logo even less than the console spam it fixes.


jasinner:
Sorry for the late reply here. Can you please provide a link for the ESS with more details of the requirement you are trying to achieve with these changes? I think overall the changes look fine, I just want to understand the justification for the changes better.

corgi/web/templates/base.html
```diff
@@ -10,65 +10,45 @@
     <link rel="stylesheet" type="text/css" crossorigin="anonymous"
           href="https://cdnjs.cloudflare.com/ajax/libs/patternfly/3.59.5/css/patternfly-additions.min.css"
           integrity="sha512-EYt56ZZC96V80F1+H7qwhi3NUIzR7onpd+RsVWHUxQfZSiIWmsQFD+6J/bS6OsAGKgvRftxtkxB16efK4mDFaA==">
+    <link rel="stylesheet" type="text/css" href="{% static 'base.css' %}">
+    <link rel="icon" href="{% static 'favicon.ico' %}">
     <title>Red Hat Component Registry | {% block title %}{% endblock %}</title>
+    <script defer src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js"
+            integrity="sha512-bLT0Qm9VnAYZDflyKcBaQ2gg0hSYNQrJ8RilYldYQ1FxQYoCLtUjuuRuZo+fjqhx/qtq/1itJ0C2ejDxltZVFg=="
```
jasinner: nit: I'm curious how these hashes are obtained or did you calculate them yourself?

jasinner:
Suggestion: I'm definitely not an expert on Content Security Policy, but I visited the home page using your branch in my dev environment, and grabbed the `Content-Security-Policy` HTTP Response header value which I plugged into [this tool](https://csp-evaluator.withgoogle.com/).

I gave some suggestion in how it could be further improved, see [this screenshot](http://file.bne.redhat.com/~jshepher/Screenshot%20from%202022-11-01%2013-40-01.png).

jasinner (approved):
This is overall an improvement to the code, so I'm going to approve it. Please fix failing black check on CI and rebase on latest main.

corgi/web/templates/base.html
juspence: They were generated using https://www.srihash.org/

juspence:
> Sorry for the late reply here. Can you please provide a link for the ESS with more details of the requirement you are trying to achieve with these changes? I think overall the changes look fine, I just want to understand the justification for the changes better.

Yes, I'll add details of the requirement to the Jira ticket since I don't want to paste internal links here.

> Suggestion: I'm definitely not an expert on Content Security Policy, but I visited the home page using your branch in my dev environment, and grabbed the `Content-Security-Policy` HTTP Response header value which I plugged into [this tool](https://csp-evaluator.withgoogle.com/).
> 
> I gave some suggestion in how it could be further improved, see [this screenshot](http://file.bne.redhat.com/~jshepher/Screenshot%20from%202022-11-01%2013-40-01.png).

Thank you! I'll add the "object-src: 'none'" directive and do another review in a bit. Note that it seems like our templates here were copied from our other tools, and we use Bootstrap / jQuery / PatternFly across all of these to make things look prettier. They come with some built-in CSS and JS niceness we use, if I'm understanding the template right.

So removing the cdnjs.cloudflare.com line is technically possible, but it will make things ugly and might even break things in other places. I also don't think removing them greatly enhances security, since we have SRI hashing in place to prevent some kind of resource substitution / attacker running their own code.

Removing the 'unsafe-inline' directive is not possible yet, as noted in the MR description and the comments I added in the third commit.

> This is overall an improvement to the code, so I'm going to approve it. Please fix failing black check on CI and rebase on latest main.

Yep, I just worked on this Friday while blocked on other tickets, so didn't look at pipeline before pushing (or yesterday while handling CORGI-295). I may let this sit for a while since other more urgent tasks are on my plate now, but will fix before merging.

JimFuller-RedHat:
Please consider adding to PR description rationale for inclusion of drf_spectacular_sidecar (as replacement for drf_spectacular) ... otherwise LGTM and as this PR has been around for a while best to get in as soon as. Nice work!

JimFuller-RedHat (approved):
Please consider adding to PR description rationale for inclusion of drf_spectacular_sidecar (as replacement for drf_spectacular) ... otherwise LGTM and as this PR has been around for a while best to get in as soon as. Nice work!

juspence:
> Please consider adding to PR description rationale for inclusion of drf_spectacular_sidecar (as replacement for drf_spectacular) ... otherwise LGTM and as this PR has been around for a while best to get in as soon as. Nice work!

Thanks for reviewing! Unfortunately this still isn't done - I need to make more changes based on Jason's feedback above. I'll tag and request review again once this is ready.

drf_spectacular_sidecar isn't replacing drf_spectacular. As noted in the third commit message, we want to avoid 'unsafe-inline' in our Content-Security-Policy, which drf-spectacular without the sidecar requires.

juspence:
@RedHatProductSecurity/corgi-devs OK, round two. Sorry for the churn here - Jason's response above had some good feedback which led to some changes:

1) I merged some PRs into Django-REST-Framework, then upgraded the version we use. Now DRF no longer requires "'unsafe-inline'" in a Content-Security-Policy.
2) We now whitelist external resources (hosted on a CDN) by hash instead of by hostname. This avoids some known attacks when an HTML injection bug is present on any of our pages.
3) I added sandbox directives and tightened up our "default-src" directive, to better follow CSP best practices.

The reason to avoid 'unsafe-inline' is twofold. The first reason is that I don't think I can convince the ESS auditors that unsafe-inline is actually safe, even though DRF's serializers probably prevent injection bugs that would lead to a CSP bypass / code execution.

The second reason is that after reading more on this topic, I'm not 100% sure myself that our configuration is safe. At the end of the day, a stronger CSP is better for us and easier to explain in an audit.

Have a great weekend!

config/settings/base.py
```diff
@@ -77,30 +81,59 @@
 X_FRAME_OPTIONS = "DENY"  # Header: X-Frame-Options DENY
 
 # Content Security Policy
+# Use CSP_STYLE_SRC_ELEM / exclude inline styles once browser support is better
 CSP_STYLE_SRC = (
     "'self'",
-    "'unsafe-inline'",
```
mprpic, line 82: Do we have to remove this? I find inline CSS to be much more readable and maintainable for one-off changes.

mprpic (approved):

config/settings/base.py
juspence: I'm almost certain the answer is "yes". For other ESS audits, it was enough to say "we have a CSP to prevent XSS attacks". For this audit, we're being asked to show specific evidence of what our CSP contains and how we follow best practices. As noted above, I don't think I can justify "unsafe-inline" as "actually it's safe in the configuration we use" to a less-technical auditor.

I also don't think I should. If we allow "unsafe-inline" only for CSS, an attacker who finds an HTML injection bug could, for example, inject form elements with inline styles that set opacity:0 to make them invisible. The form target can then be set to attacker-site.com to exfiltrate data.

In general we mostly use classes from the Patternfly stylesheets, our custom stylesheet only has four classes in it, and our templates don't tend to change much. FInally, those classes are only used on one element each, so making changes won't have breaking side-effects in other places. So I don't think this should be too much additional maintenance burden.

PR #207, opened 2022-10-31
Fix missing softwarebuild on rpmmod

jasinner:
Includes commits from https://github.com/RedHatProductSecurity/component-registry/pull/197

I removed 'namespace' attribute from Component.*_or_create calls as it's not part of the uniquness constraint for that model.


jasinner:
I will incorporate these changes into #197 instead

PR #208, opened 2022-11-01
CORGI-295: Fix distinct duplicates

juspence:
@RedHatProductSecurity/corgi-devs This MR fixes a bug where calling distinct() can sometimes return duplicates, depending on if the queryset we call .distinct() on previously had any ordering defined. This ordering can be applied in several places:

1) The model meta options themselves
2) A higher-level queryset, like ```a = Model.objects.order_by("b")``` followed by ```c = a.values_list("d").distinct()```
3) The django-mptt library, which tries to always return MPTTModel nodes in "tree order" (based on tree ID and left child's node ID).

Hoping we can discuss this in a planning or standup meeting before merge. There are quite a few changes needed to handle the above three cases, but this avoids us having to always write ```.order_by().distinct()``` every time we interact with a queryset. If we'd prefer that approach instead, please let me know and I can switch this.

I'm also not sure this handles all the ways django-mptt might return an ordered queryset - review appreciated.


jasinner:
The performance of the product manifest is amazing! Great work.

corgi/core/models.py
```diff
@@ -533,6 +530,8 @@ def get_latest_components(self):
                         root_components,
                         name=OuterRef("name"),
                         product_streams__overlap=(self.ofuri,),
+                        # TODO: What is overlap for??
```
jasinner: nit: This was previously `contains`, however I found that contains not only matched against list entries, but also substrings of list entries returning incorrect results for searches such as "rhel-8" which is a substring of many product streams such as "rhel-8.6.0.z"

corgi/core/models.py
```diff
@@ -533,6 +530,8 @@ def get_latest_components(self):
...
+                        # TODO: Exclude software_build__isnull=True???
```
jasinner: nit: Not a bad idea. Although I don't think there will be any of these as the root components types should always have software_build set. There is no DB constraint for that however so if it was true, it would be a bug in the way those components were created.

jasinner (approved):
Great work, happy to merge this!

If this does have any impact on MPTT TreeManager default ordering I can't see it. I downloaded the data from stage and tried navigating the API Web UI using these code changes and everything was ligtening fast, even the manifest/ and component?ofuri endpoints with limits set to high value like 5000 for rhel-8.6.0.z

corgi/core/models.py
juspence: Thanks for the heads-up! I just tested locally but couldn't reproduce this running code like below:
```
c = Component.objects.last()
c.products = ["abc", "def"]
c.save()

# Working as expected
Component.objects.filter(products__contains=["abc"]).values_list("products", flat=True)
<ComponentQuerySet [['abc', 'def']]>

# No results, AKA it doesn't seem to do a substring match for me
Component.objects.filter(products__contains=["ab"]).values_list("products", flat=True)
<ComponentQuerySet []>

# Throws an error, as expected
Component.objects.filter(products__contains="abc").values_list("products", flat=True)
Component.objects.filter(products__contains="ab").values_list("products", flat=True)
```

The reason I ask is that I switched some code in a view's get_queryset function to use a __contains lookup. Previously we were only returning components which had exactly ```["some value"]``` in their taxonomy fields, which seemed like a bug.

I'm going to leave the __contains lookup, since that seems to be identical to the __overlap lookup when there's only one value in the list on the right side (see https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/fields/#contains).

I also don't see __overlap mentioned in the list of standard Django lookups (https://docs.djangoproject.com/en/3.2/ref/models/querysets/#std-fieldlookup-contains), only the Postgres-specific extensions (above). But please let me know if you'd prefer __overlap, or if the original version was working as intended.

jasinner: If contains and overlap behave the same way, I guess using the standard contains function is better.

PR #209, opened 2022-11-02
CORGI-126 maven purl not built with namespace

jasinner:
This fixes the bug described in CORGI-126. The namespace attribute of Component has only 2 types, Redhat, and Upstream, so I didn't use it to persist the Maven namespace. Instead I store it in `meta_attr["group_id"]` for the maven type, and check that property when building the PURL. This means we don't have to change the type of namespace now, although we might want to revisit this decision later as it will make it more difficult to search for Maven components by their namespace. You will however be able to search the purl attribute for the namespace.

While in the area, I also expanded the types saved during a SCA scan to include GEM, and CARGO. I also upgraded Syft to the latest version and reduced the time taken by the test_sca.py test by removing some test data which wasn't used.


jasinner:

corgi/collectors/syft.py
```diff
@@ -59,6 +60,10 @@ def parse_components(cls, syft_json):
                     pkg_type = Component.Type.MAVEN
                 elif artifact["type"] == "rpm":
                     pkg_type = Component.Type.RPM
+                elif artifact["type"] == "gem":
+                    pkg_type = Component.Type.GEM
+                elif artifact["type"] == "rust-crate":
+                    pkg_type = Component.Type.CARGO
```
juspence: In corgi.core.models, we have a constant mapping of cachito types to Corgi types. I think we should reuse the same pattern here, then just do:
```
# Look up artifact["type"] in the mapping, and if it doesn't exist keep the value unchanged
pkg_type = SYFT_CORGI_TYPE_MAPPING.get(artifact["type"], artifact["type"])
```

tests/test_sca.py
```diff
@@ -33,7 +33,16 @@ def test_parse_components():
...
-    assert results[0]["analysis_meta"] == {"source": "syft", "version": "0.48.1"}
+    assert results[0]["analysis_meta"] == {"source": "syft-0.60.1"}
+
+
+def test_parse_maven_components():
+    with open("tests/data/hawkular-metrics-schema-installer-syft.json", "r") as maven_test_data:
+        results = Syft.parse_components(maven_test_data.read())
+    assert len(results) > 0
+    group_ids = [r["meta"].get("group_id") for r in results if r["type"] == Component.Type.MAVEN]
+    assert "" not in group_ids
```
juspence: I think you're trying to assert that .get() always returns a value, right? But by default .get() returns None if a key isn't found. So either change the first line to ```.get("group_id", "")``` or the second line to ```assert None not in group_ids```

tests/test_sca.py
```diff
@@ -275,12 +295,7 @@ def test_slow_software_composition_analysis(
...
-    if is_container:
-        root_component = Component.objects.get(
-            type=Component.Type.CONTAINER_IMAGE, arch="noarch", software_build=sb
-        )
-    else:
-        root_component = Component.objects.srpms().get(software_build=sb)
-        root_component.get_provides_purls()
+    expected_component = Component.objects.get(purl=expected_purl)
+    assert expected_component
+    assert expected_component.meta_attr["source"] == "syft-0.60.1"
```
juspence: I think the old version was a more useful test, since it asserted that a component with the right type (either a container image or an SRPM) got created. In the new code we assert the component exists, but don't check it has the right type.

I think you should just be able to ```assert root_component.meta_attr["source"] == "syft-0.60.1"``` after the if / else block, since root_component will be defined either way. There's also no need to ```assert expected_component``` since the .get() call should raise a Component.DoesNotExist exception if it's missing.

juspence (approved):
Overall LGTM, comments are only suggestions. I'm not so familiar with the SCA code though, so might have missed something.

corgi/collectors/syft.py
jasinner: Good idea, I will update it.

tests/test_sca.py
jasinner: You are correct, updated.

tests/test_sca.py
jasinner: Ah, it looked weird that we were using `get` to test, but you are right it was useful because if it doesn't exist it will thrown an DoesNotExist error.

PR #210, opened 2022-11-02
CORGI-271 - remove deptopia reconcilation script

JimFuller-RedHat:
Removing deptopia reconciliation script as it is being refactored into ops.


jasinner (approved):
OK, good idea.

PR #211, opened 2022-11-03
CORGI-245 inconsistent nvr values

jasinner:
This updates the `get_nvr`, and `get_nvra` functions of Components to avoid trailing `-`. There was also a problem with trailing `-` in the RPM purls in the UPSTREAM namespace.

I'm not providing any data migration here, because we plan to do a full import again as part of CORGI-317.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -840,49 +840,11 @@ class Component(TimeStampedModel):
 
     def get_purl(self) -> PackageURL:
         if self.type == Component.Type.RPM:
-            qualifiers = {
-                "arch": self.arch,
-            }
-            if self.epoch:
-                qualifiers["epoch"] = str(self.epoch)
-            purl_data = dict(
-                name=self.name,
-                version=f"{self.version}-{self.release}",
-                qualifiers=qualifiers,
-            )
+            purl_data = self._build_rpm_purl()
         elif self.type == Component.Type.RPMMOD:
-            # Break down RHEL module version into its specific parts:
-            # NSVC = Name, Stream, Version, Context
-            version, _, context = self.release.partition(".")
-            stream = self.version
-            purl_data = dict(
-                name=self.name,
-                version=f"{stream}:{version}:{context}",
-            )
+            purl_data = self._build_module_purl()
         elif self.type == Component.Type.CONTAINER_IMAGE:
-            digest = ""
-            if self.meta_attr.get("digests"):
-                for digest_fmt in CONTAINER_DIGEST_FORMATS:
-                    digest = self.meta_attr["digests"].get(digest_fmt)
-                    if digest:
-                        break
-            purl_name = self.name
-            name_from_label = self.meta_attr.get("name_from_label")
-            if name_from_label:
-                purl_name = name_from_label
-            qualifiers = {
-                "tag": f"{self.version}-{self.release}",
-            }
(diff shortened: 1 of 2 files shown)


corgi/core/models.py
```diff
@@ -847,7 +847,7 @@ def get_purl(self) -> PackageURL:
                 qualifiers["epoch"] = str(self.epoch)
             purl_data = dict(
                 name=self.name,
-                version=f"{self.version}-{self.release}",
+                version=f"{self.version}{f'-{self.release}' if self.release else ''}",
```
mprpic: f-string nesting is a bit iffy in my opinion. It's not used anywhere in the Python docs and feels a bit less readable then just separating it out into an if statement before the dict is created. Do you prefer this one-line form?

mprpic (approved):
Minor formatting question/comment, otherwise LGTM!

corgi/core/models.py
```diff
@@ -909,12 +909,13 @@ def is_srpm(self):
...
     def get_nvr(self) -> str:
-        return f"{self.name}-{self.version}-{self.release}"
+        return f"{self.name}-{self.version}{f'-{self.release}' if self.release else ''}"
 
     def get_nevra(self) -> str:
         return (
             f"{self.name}{f':{self.epoch}' if self.epoch else ''}"
-            f"-{self.version}-{self.release}.{self.arch}"
+            f"-{self.version}{f'-{self.release}' if self.release else ''}"
+            f"{f'.{self.arch}' if self.arch else ''}"
```
juspence: +1 to Martin's comment above. Maybe this is a little more readable:
```suggestion
        epoch = f":{self.epoch}" if self.epoch else ""
        release = f"-{self.release}" if self.release else ""
        arch = f".{self.arch}" if self.arch else ""

        return (
            f"{self.name}{epoch}"
            f"-{self.version}{release}"
            f"{arch}"
)
```

juspence (approved):
LGTM except for the "oom" file. Seems like this got committed on accident?

juspence:
Actually I was going through emails and remembered one more thing:
https://github.com/RedHatProductSecurity/component-registry/pull/197#discussion_r1005089777
added f"{self.version}-{self.release}" as a container image tag qualifier.

Do we also need to handle an empty self.release there? Or maybe we don't, if we expect container image components to always have a self.release defined.

jasinner:
> Actually I was going through emails and remembered one more thing: [#197 (comment)](https://github.com/RedHatProductSecurity/component-registry/pull/197#discussion_r1005089777) added f"{self.version}-{self.release}" as a container image tag qualifier.
> 
> Do we also need to handle an empty self.release there? Or maybe we don't, if we expect container image components to always have a self.release defined.

I didn't find any containers without a release set in the stage data but I added a similar check for `release ==  ""` for the tag qualifier of container tag just in case.

corgi/core/models.py
jasinner: Ok, I removed nested f-stings.

jasinner:
> LGTM except for the "oom" file. Seems like this got committed on accident?

Thanks, I removed the rouge file.

corgi/core/models.py
jasinner: Ok, I removed nested f-stings. I saw it already existed in the code so I just reused it as it looked succinct.

PR #213, opened 2022-11-07
CORGI-334: Fix returning a datetime when a timedelta is expected

juspence:
@mprpic Quick review please. I made all the related code pass a datetime (instead of a timedelta) to fetch_unprocessed_relations, since other code that uses get_last_success_for_task expects it to return a timezone.

I didn't want to change get_last_success_for_task from what's implemented in SDEngine. But I could tweak get_last_success_for_task and its callers instead, if you think returning a timedelta there would be less invasive or more sensible.


mprpic (approved):
LGTM! I think returning a datetime is fine.

PR #214, opened 2022-11-08
CORGI-31 go package scan

jasinner:
This add a go_list collector which gets called from sca task after a Syft scan.

The `go list` command returns a stream of json objects, we I had to process it as a stream using the splitstream library. 

The 'go list` command returns go packages from go modules, and the standard library. In order to add versions to go standard library packages we have to lookup which go version was used to compile them. This is possible today with container images. For RPMs, I filed a follow task (CORGI-338).

The `go list` command only works on go source code. Modern go projects all use a go.mod file, so I look for that as an indicator of where to perform the `go list` command. Deptopia also checks for Gopkg.lock files, but I think those projects are less command these days so I didn't add it.

I verified this by importing a view packages mentioned in CORGI-249 detect the crypto/tls standard library after the scan.


jasinner:
CI is failing. Locally it works by setting GOCACHE env var to `/tmp`. I tried to do that in CI env as well, but the setting doesn't seem to have the desired affected. I'll fix it before merging.

corgi/collectors/go_list.py
```diff
@@ -0,0 +1,91 @@
...
+                            results.append(result)
+                    else:
+                        logger.debug("Did not find go.mod in %s", target_path)
+            else:
+                for result in cls.invoke_process_popen_poll_live(GO_LIST_COMMAND, target_path):
+                    results.append(result)
+        return results
+
+    @classmethod
+    def find_go_dir(cls, extract_dir):
```
mprpic:
```suggestion
    def find_go_dir(cls, extract_dir: str) -> Path:
```

corgi/collectors/go_list.py
```diff
@@ -0,0 +1,91 @@
...
+        cls, go_list_pipe: Optional[IO[bytes]]
+    ) -> Generator[dict[str, Any], None, None]:
+        # use of splitstream here as `go list` output is actually a stream of json objects,
+        # not a fully formed valid json document
+        for jsonstr in splitfile(go_list_pipe, format="json"):
+            artifact = json.loads(jsonstr)
+            typed_component: dict[str, Any] = {
+                "type": Component.Type.GOLANG,
+                "meta": {
+                    "name": artifact["ImportPath"],
```
mprpic, line 83: In the Brew collector, we also record the individual go package type: `"go_component_type": "go-package",`. Can we add that here as well?

requirements/rpms.txt
```diff
@@ -1,6 +1,7 @@
 # To compile C bindings in certain Python dependencies
 gcc-8.5.0
 git-2.31.1
+golang-bin-1.17.12-1.module+el8.6.0+16014+a372c00b
```
mprpic: Just curious, is this installing the `-bin` package specifically for a reason?

mprpic (approved):
LGTM!

requirements/rpms.txt
jasinner: It's a smaller package and include the `go` package which we need. See https://developers.redhat.com/blog/2017/03/17/basics-of-go-in-fedora#how_to_install_fedora_golang_

That's what deplist installs as well.

corgi/collectors/go_list.py
jasinner: Sure, I'll add that.

corgi/collectors/go_list.py
jasinner: Updated, thanks.

requirements/rpms.txt
mprpic: I think both `dnf install golang` and `dnf install golang-bin` results in the same package set. `golang-bin` is just a subpackage of `golang`, which it needs anyway.

jasinner: Ok, I'll install the golang package.

PR #216, opened 2022-11-09
CORGI-317: Fix some bugs and clean up code

juspence:
@jasinner @mprpic I have the taxonomy changes mostly finished locally, but I'm still working on getting the tests to pass and fixing bugs.

In the meantime, this MR is a set of changes that can stand alone but will help lay groundwork for the other changes. Splitting this out since the rest of the changes keep growing.


corgi/core/models.py
```diff
@@ -986,6 +970,34 @@ def get_roots(self) -> list[ComponentNode]:
...
 
+    @property
+    def download_url(self) -> str:
+        """Report a URL for RPMs and container images that can be used in manifests"""
+        if self.software_build and self.software_build.type == SoftwareBuild.Type.BREW:
+            # RPM ex:
+            # /vol/rhel-7/packages/emacs/24.3/23.el7/ppc64le/emacs-common-24.3-23.el7.ppc64le.rpm
+            if self.type == Component.Type.RPM:
+                return (
+                    f"{settings.BREW_DOWNLOAD_ROOT_URL}/vol/"
```
jasinner: BREW_DOWNLOAD_ROOT_URL is not available outside the RH firewall. Therefore we wouldn't want to use it in manifests.

jasinner (changes requested):
Looks good me. I like that your able to move some of the functions of Product up to the ProductModel with a more generic query.

I think we shouldn't include the BREW_DOWNLOAD_URL in the manifests though. I think previously we were setting the download location to a generic url on the customer portal which is probably more appropriate. Although most users will get the software via CDN (for rpms), or container repository.

requirements/rpms.txt
```diff
@@ -12,8 +12,8 @@ openldap-devel-2.4.46
 # Dependency for SSL support in the python-qpid-proton Python package
 openssl-1.1.1k
 openssl-devel-1.1.1k
-python39-3.9.13-1.module+el8.7.0+15656+ffd4a257
```
jasinner, line 15: nit: Martin and I discussed this yesterday when some old RPMs where removed from the UBI-8 RPM repos that we should probably not set the version as at, and just use for example `python-39` here. Otherwise we might miss some security updates, and also the builds will break again when these RPMs are purged. It shouldn't have any compatibility consequences.

corgi/core/models.py
```diff
@@ -104,54 +104,48 @@ class Meta(NodeModel.Meta):
...
-            .get_family()
-            .filter(content_type=ContentType.objects.get_for_model(Product))
-            .values_list("product__name", flat=True)
-            .distinct()
+    def get_node_pks_for_type(qs: QuerySet["ProductNode"], target_model: str) -> QuerySet:
+        """For a given ProductNode queryset, find all nodes with the given type
+        and return the primary keys of their related objects"""
+        # "product_version" -> "ProductVersion"
+        mapping_model = target_model.title().replace("_", "")
+        return qs.filter(level=MODEL_NODE_LEVEL_MAPPING[mapping_model]).values_list(
```
jasinner, line 117: I like the use of level here.

corgi/core/models.py
```diff
@@ -495,21 +500,16 @@ def manifest(self) -> str:
...
+                ROOT_COMPONENTS_CONDITION,
                 product_streams__overlap=(self.ofuri,),
             )
             .annotate(
-                latest=Subquery(
+                latest=models.Subquery(
                     Component.objects.filter(
-                        root_components,
-                        name=OuterRef("name"),
+                        ROOT_COMPONENTS_CONDITION,
```
jasinner, line 512: nit: I wonder if we can also level the ComponentNode(level=0) trick here as ComponentTypes at the root level are likely to change in future.

corgi/web/templates/product_manifest.json
```diff
@@ -9,7 +9,7 @@
     {% endfor %}
     {
         "copyrightText": "NOASSERTION",
-        "downloadLocation": {% if obj.channels %}"{{obj.channels.0}}"{% else %}"NOASSERTION"{% endif %},
```
jasinner, line 12: A better download location is probably [package browser](https://access.redhat.com/downloads/content/package-browser) for (S)RPMs and Modules, and [catalog](https://catalog.redhat.com/software/containers/search) for Containers.

tests/test_model.py
```diff
@@ -78,14 +78,22 @@ def test_cpes():
...
-    assert ":" not in no_epoch.nevra
-    no_arch = ComponentFactory(arch="")
-    assert not no_arch.nevra.endswith("-")
+    for component_type in Component.Type.values:
+        no_release = ComponentFactory(
+            name=component_type, release="", version="1", type=component_type
+        )
+        assert not no_release.nvr.endswith("-")
+
+        if component_type == "OCI":
```
jasinner: nit: I have a different fix for this in https://github.com/RedHatProductSecurity/component-registry/pull/215

requirements/rpms.txt
juspence: Ack, I'll tweak this. Originally Judy Kelly recommended we pin the versions we use: https://docs.google.com/document/d/1n-4CU_Mim8Fauy0iFzelLLy4QaGy747VF5cMU60txDU/edit#heading=h.a2f0mozgq7rz

It's not worth it though since those package versions can get removed. I thought they'd be stable for the lifetime of a particular UBI8 version, and only backported security updates (with the same version and a newer release) would be added.

corgi/core/models.py
juspence: We could! I wasn't sure about the logic for ROOT_COMPONENTS, but I can definitely switch it to report only components with level=0 / at the root of a tree.

corgi/web/templates/product_manifest.json
juspence: Thank you! I was mostly just guessing when I decided to use channels[0] originally (didn't even really know what a channel was back then). I saw while tweaking other code that we had a nice download_url property, and didn't realize it was internal. I'll switch this to use above.

tests/test_model.py
juspence: Ack, and thank you! I'll fix the assert to match your version during a rebase, but I think I'll also keep the for loop I added. Right now we only test one type of component randomly, but we have different code for different purl types. So if one of those codepaths is broken, the test will sometimes pass and sometimes fail randomly.

Switching to a for loop improves coverage by testing every code path for every type, and means that tests will pass or fail consistently if something is broken, instead of randomly like they do today.

corgi/core/models.py
juspence: Details below - will fix.

corgi/core/models.py
juspence: Actually, we could do this but it would require some more invasive changes. Right now we filter on Component, so we can't check for level=0. We'd need to switch to filtering ComponentNodes with level=0, then look up (and dedupe) their linked Component objects. So I'll leave this for another MR.

corgi/api/serializers.py
```diff
@@ -142,11 +143,20 @@ def get_product_data_list_by_ofuri(
...
-        {"type": pcr.type, "external_system_id": pcr.external_system_id} for pcr in related_pcrs
+    relations = [  # type: ignore
+        # Django's values_list() returns a QuerySet of Optional[str]
+        # even though the fields aren't nullable - this is a known bug
+        # https://github.com/typeddjango/django-stubs/issues/444
+        # We can upgrade djangorestframework-stubs to 1.7.0 to fix
+        # But there are other bugs in our code we must also fix
+        {"type": pcr_type, "external_system_id": pcr_id}
+        for (pcr_type, pcr_id) in related_pcrs
+        if pcr_type is not None and pcr_id is not None
```
mprpic: Can this be move to a filter/exclude on the ProductComponentRelation query?

corgi/core/constants.py
```diff
@@ -9,3 +11,18 @@
...
+    "ProductStream": 2,
+    "ProductVariant": 3,
+    "Channel": 4,
+}
+
+# Filter on "root components": SRPMs, modules, or index container images
+SRPM_CONDITION = Q(type="RPM", arch="src")
+MODULE_CONDITION = Q(type="RPMMOD")
+INDEX_CONTAINER_CONDITION = Q(type="OCI", arch="noarch")
+ROOT_COMPONENTS_CONDITION = SRPM_CONDITION | MODULE_CONDITION | INDEX_CONTAINER_CONDITION
```
mprpic, line 28: Is there a specific reason to define these in a global context (in a separate `constants.py` file) instead on the queryset class where they are used? It makes it easier to see what a certain method does if the values are defined right above it :slightly_smiling_face:

corgi/core/models.py
```diff
@@ -104,54 +110,48 @@ class Meta(NodeModel.Meta):
...
 
     @staticmethod
-    def get_product_versions(node_model):
-        return list(
-            node_model.pnodes.first()
+    def get_node_names_for_type(product_model: "ProductModel", target_model: str) -> QuerySet:
+        """For a given ProductModel / Channel, find all related nodes with the given type
+        and return the names of their related objects"""
+        # "product_version" -> "ProductVersion"
+        mapping_model = target_model.title().replace("_", "")
```
mprpic, line 126: Mmm, this feels somewhat hacky and surprising. Can we not pass the model class itself here?

corgi/core/models.py
```diff
@@ -1090,7 +1103,8 @@ def get_upstreams(self):
         for root in roots:
             # For SRPM/RPMS, and noarch containers, these are the cnodes we want.
             source_children = tuple(
-                c for c in root.get_children().filter(type=ComponentNode.ComponentNodeType.SOURCE)
+                c
+                for c in root.get_descendants().filter(type=ComponentNode.ComponentNodeType.SOURCE)
```
mprpic: Is the `c for c` part necessary here?

mprpic (approved):
Just minor nitpicks mostly. LGTM!

corgi/api/serializers.py
juspence: Unfortunately not...mypy is not very smart here so .exclude(field__isnull=True) doesn't help. The newer version is much better, since it actually finds more bugs in our code. But fixing all of them is too much effort for this one MR, so I just hacked around it like this. I guess I could switch to ```# type: ignore``` though. Please feel free to leave a comment or thumbs up here if you'd prefer that way, otherwise I'll just leave it as is.

corgi/core/constants.py
juspence: There are like a million places where the ROOT_COMPONENTS_CONDITION is used, so "right above it" doesn't really work. I just reused constants.py since it already existed, but I could move it to the top of models.py instead. Please feel free to leave a comment or thumbs up here if you'd prefer that way, otherwise I'll just leave it as is.

I also thought we might end up using this same logic in other files (e.g. some task or collector that filters only root components), so I figured defining in and importing from constants.py was best.

corgi/core/models.py
juspence: I used string keys in the mapping dict since it was defined in constants.py. Using class names as the dict keys would have caused a circular import.

EDIT: There was also some existing code which used "product_stream__name", so to make that generic we also need the underscored model name. Let me look and see if there's a less ugly way to do something similar.

EDIT2: I can fix this as part of the next MR. Right now we need to go from field_name to ModelName (make into Title Case then strip underscore), and the other way around is tricky. The next MR renames some of these properties we're using for lookups (underscore will disappear because reasons), so I'll switch it to use str(ClassName).lower() in that MR.

corgi/core/models.py
juspence: Probably not, good catch!

corgi/api/serializers.py
mprpic: Either way is fine. In general I don't like bending code to make linters happy so will always choose the type:ignore, but that's really just my preference :-)

corgi/core/constants.py
mprpic: Ah, I grepped for SRPM_CONDITION only and saw it used once. This is fine then, carry on.

corgi/core/models.py
mprpic: Ack!

corgi/core/models.py
```diff
@@ -986,6 +969,35 @@ def get_roots(self) -> list[ComponentNode]:
     def build_meta(self):
         return self.software_build.meta_attr
 
+    @property
+    def download_url(self) -> str:
+        """Report a URL for RPMs and container images that can be used in manifests"""
+        # RPM ex:
+        # /downloads/content/aspell-devel/0.60.8-8.el9/aarch64/fd431d51/package
+        # We can't build URLs like above because the fd431d51 signing key is required,
+        # but isn't available in the meta_attr / other properties (CORGI-342). Get it with:
```
jasinner, line 978: Awesome, thanks for investigating this.

corgi/core/models.py
```diff
@@ -986,6 +969,35 @@ def get_roots(self) -> list[ComponentNode]:
...
+        # Image ex:
+        # /software/containers/container-native-virtualization/hco-bundle-registry/5ccae1925a13467289f2475b
+        # /software/containers/openshift/ose-local-storage-diskmaker/
+        # 5d9347b8dd19c70159f2f6e4?architecture=s390x&tag=v4.4.0-202007171809.p0
+        # We can't build URLs like above because the hash is required,
+        # but doesn't match any hash in the meta_attr / other properties.
+        # We could use repository_url instead, if we decide to store it on the model.
+        # Currently it's only in the meta_attr and should not be used (CORGI-341).
+
+        # TODO: self.filename, AKA the filename of the image archive a container is included in,
```
jasinner, line 992: Nice detective work, thanks.

jasinner (approved):

PR #219, opened 2022-11-11
expose productstream active param on api schema

jasinner:
I discovered this while testing the ops repo deptopia-reconcilication scripts. The bindings generated against the 1.1.0 schema failed with an error about the productstreams list endpoint not having an active parameter.


jasinner:
The component list endpoint was also missing a couple of parameters, `ofuri`, `summary`, and `purl`.

PR #220, opened 2022-11-14
CORGI-346 fix unprocessed relations loading

jasinner:
This reduces the number of builds we try to process in a day while still staying below our max memory for Redis of 1,200mb. I've also adjust the number of slow workers to 10 to include the throughput.

When redis was crashing last week with a memory limit of 500mb I saw that most redis keys where 500k keys in the slow list. We need to avoid trying to load so many tasks in one hit. I calculated if we load less than 150,000 builds a day, and process 10 builds a minute we should be able to avoid evicting messages from redis.

If we do end up evicting messages from redis SoftwareBuild won't be created and we can try to process them again the next day. After this is merged and we try to reload all builds we should keep an eye on the number of keys in the Redis slow list to make sure it's not not getting too close to the limit.


jasinner:

corgi/tasks/brew.py
```diff
@@ -483,46 +484,36 @@ def fetch_modular_builds(relations_query: QuerySet, force_process: bool = False)
...
+            # build_id defaults to "" and int() will fail in this case
+            continue
+        if not SoftwareBuild.objects.filter(build_id=int(build_id)).exists():
+            logger.info("Processing CDN relation build with id: %s", build_id)
+            slow_fetch_modular_build.delay(build_id, force_process=force_process)
+            processed_builds += 1
     return processed_builds
 
 
+#
```
juspence: Committed by mistake, or some changes didn't make it into your commit?

corgi/core/constants.py
```diff
@@ -26,3 +26,8 @@
 MODULE_CONDITION = Q(type="RPMMOD")
 INDEX_CONTAINER_CONDITION = Q(type="OCI", arch="noarch")
 ROOT_COMPONENTS_CONDITION = SRPM_CONDITION | MODULE_CONDITION | INDEX_CONTAINER_CONDITION
+
+# The ratio of build types to total in the relation table
+CDN_RELATIONS_RATIO = 0.8
+BREW_RELATIONS_RATIO = 0.15
+YUM_RELATIONS_RATIO = 0.05
```
juspence, line 33: Do we think these ratios might change a lot once we load more data? And maybe we should add a test that these add up to 1.0, in case one is changed in the future without changing the others, or a new type gets added.

corgi/tasks/rhel_compose.py
```diff
@@ -49,5 +51,12 @@ def get_builds(compose_names: list[str] = None, stream_name: str = None) -> None
...
 
-    relations_build_ids = relations_query.values_list("build_id", flat=True).distinct()
-    fetch_modular_builds(relations_build_ids)
+    processed_builds = 0
+    for build_id in relations_query.values_list("build_id", flat=True).distinct().iterator():
+        if not build_id:
+            continue
+        if not SoftwareBuild.objects.filter(build_id=int(build_id)).exists():
+            logger.info("Processing Compose relation build with id: %s", build_id)
+            slow_fetch_modular_build.delay(build_id, force_process=force_process)
```
juspence, line 60: Should we move this logic into the fetch_modular_builds function?

corgi/tasks/monitoring.py
```diff
@@ -64,7 +63,8 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
     upsert_cron_task("brew", "fetch_unprocessed_brew_tag_relations", hour=6, minute=0)
     upsert_cron_task("yum", "load_yum_repositories", hour=7, minute=0)
     upsert_cron_task("yum", "fetch_unprocessed_yum_relations", hour=8, minute=0)
-    upsert_cron_task("monitoring", "email_failed_tasks", hour=10, minute=45)
+    upsert_cron_task("pulp", "fetch_unprocessed_cdn_relations", hour=9, minute=0)
+    upsert_cron_task("monitoring", "email_failed_tasks", hour=11, minute=45)
```
juspence, line 67: What would happen if we made only this change to process CDN relations daily, without any of the other logic, and switched created_since to 1? AKA process only relations created in the last day, every 1 day. Would that reduce the number of tasks being processed enough to avoid these errors?

juspence (commented):
Thanks for all the effort you've spent debugging these Celery / Redis failures. I'm not sure I like this change, though, since it feels very fragile. If we ever change the Redis memory limits, or add more Celery workers, the MAX_BUILDS_TO_PROCESS value will also need to be changed. The different PCR type ratios also might change once we load more / different data.

I'd rather raise the Redis memory limit to the maximum (3GB per pod, I think?) and see if we can process 1 day's worth of CDN relations without restarting. We could also request a quota increase and raise the number of slow Celery workers.

Finally, we could just get rid of the Redis eviction policy and allow Redis to restart. All tasks / keys in the Redis queue get saved to a PVC, so if Redis restarts we shouldn't lose any data. I think by default Celery retries connections to the broker if it goes down, so this also shouldn't cause task failures, but I haven't looked at any recent monitoring email to confirm this. Even if it does cause failures, letting some task explicitly fail due to a broken connection seems less scary to me.

Evicting keys might make a currently-running task fail, but might also make two copies of the same singleton task run at the same time (if a locking key added by celery-singleton gets removed). Evicting keys might also cause some scheduled task to never run in the first place. In that last case, we'd never see it in our monitoring email and likely wouldn't even know it happened until we investigated why some data was missing.

So explicit task failures we can debug, due to broken connections / Redis restarts, seem better than silent failures we don't even know are happening. But you've done the most debugging on these issues, so I'm happy to go with whatever solution you think is best - probably there are no good ones here.

corgi/tasks/monitoring.py
juspence: Discussed in a meeting: No, this won't help. We create a bunch of CDN relations all at once, so whether we look at all relations created in the past day or past week, we still have way too many to process.

corgi/tasks/brew.py
jasinner: removed

corgi/tasks/rhel_compose.py
jasinner: I don't think that's a good idea. We want to avoid create a task (and adding it to Redis) if the build has already been processed.

corgi/core/constants.py
jasinner: I don't think these will change much. I took a guess at these before calculating that actual ratios and my guess was pretty accurate. That indicates to me they have probably been in this ratio for a long time.

jasinner:
Looking at the relations table, we have almost 200,000 unique builds we need to process in total. Given we need to process both `slow_fetch_brew_build`, and `slow_software_composition_analysis` tasks on each build that doubles the number of tasks to 400,000. If we leave room for an extra 100,000 tasks, our capacity should only need to be 500,000, which we know we can fit in 500mb. If already know we have trouble at 500mb, so I suggest we go a little higher to 1200mb If we leave 20% overhead to the Redis pod, it's pod limit should be set to 1.5gb. That is less than the 3gb suggested by Justin, but should give us plenty of room to fit all the tasks from the relations table, while still leaving an extra 1.5gb in the OpenShift project to allocate to slow workers. 

Increasing slow workers increases our throughput. I calculate we can process 144,000 tasks a day with 10 workers. That is an average of 1 task a minute. If we leave room for 24,000 ah-hoc tasks in a day, we should set our MAX_BUILDS_TO_PROCESS setting here to 120,000. If we want to disable this feature in future, just set it to 0 in the djnago config or associated environment variable.

Unfortunately restarting Redis does not help, the tasks are read from the persistent storage and all loaded into memory again on restart. To be able to use a restart to clear the memory we'd have to remove the persistent storage volume altogether. I think with this setup it's unlikely we'll have any keys (tasks) evicted, but if they are they should `slow_*` tasks backed up on the slow queue, which have the oldest created time, and never been accessed. They should be picked up again on the next load from the relations table, because we order by created_date when fetching them.

We should be able to get through the entire relations table in 5 days with this setup.

corgi/tasks/rhel_compose.py
juspence: I was referring to the helper function ```corgi.tasks.brew.fetch_modular_builds```, not the Celery task ```corgi.tasks.brew.slow_fetch_modular_build```. The logic you added above to ```corgi.tasks.rhel_compose.get_builds``` is the same as in ```corgi.tasks.brew.fetch_unprocessed_relations```.

I thought there might be other places in the code that needed to avoid scheduling Celery tasks for builds that already exist. I thought we should combine that logic in one place (fetch_modular_builds), then reuse it for any code that processes data in the PCR table. But after looking further, there aren't any other places that need this logic, so what you have here is fine.

PR #222, opened 2022-11-15
CORGI-317: Use foreign keys in taxonomies

juspence:
@jasinner @mprpic Please review when you have time and are not busy. This still needs much more work, including a bugfix for the two failing tests and manual testing locally. But I wanted to discuss the approach and see if there was a better way to implement these changes, before I spend more time testing.

I'll clean up the TODO comments, do a final review, and test locally before merging, but there should be no further changes here. I'm going to make the ProductComponentRelation table use foreign keys in another MR since this one is already huge.

In general, this should change how we link models together, and present only minor user-facing changes (mostly the order of properties in the schema). Most of the work is in simplifying the models.py code, with a small amount of cleanup in the tasks code, templates, and tests. I also included some bugfixes I noticed as I was going through each file.

In particular, there should be no changes to the available API endpoints, arguments they take, or field names they return. Currently there are some differences in the autogenerated openapi.yml schema file. I'm working now to debug these and I expect them to be gone in the final version.

---

The approach here is to use ManyToManyFields to link Channels to ProductVariants. This is because one ProductVariant can obviously have many Channels, but two different ProductVariants may ship to the same Channel. Because of this, we also have to use ManyToManyFields to link Channels to all other ProductModel subclasses (Products, ProductVersions, and ProductStreams).

Similar logic applies to Components, so they're also linked to ProductModel subclasses using ManyToManyFields.

The ProductModel subclasses themselves are a little special. Because we know that one ProductVariant always has only one parent ProductStream, I chose to link them together using a ForeignKey. However, this may be a bad choice since it complicates the code a bit. Whenever we call save_product_taxonomy() (or certain generic serializer methods), we have to check which ProductModel subclass we're operating on, then do one of two things:

If we have a forward relationship (e.g. foreign key from ProductVariant to ProductStream), we can assign a model instance to the foreign key field and access its properties directly:
```
my_pv.productstreams = my_ps
print(my_pv.productstreams.name)
```

If we have a reverse relationship (e.g. related_set from ProductStream to ProductVariant), we have to use .add() or .set() to link the models together, and we have to iterate over each related model instance to access properties:
```
my_ps.productvariants.add(my_pv)
for pv in my_ps.productvariants:
    print(pv.name)
```

This could be simplified if we just use a ManyToManyField for the ProductModel taxonomy, like we already do for the Channel and Component taxonomies. Then it won't matter which side of the relationship we're on, since we'll always use a manager and iteration to access the related objects (the second form given above), even if there's only one linked object.

ManyToManyFields would also keep our code a bit more flexible, if we ever think that e.g. we might have to link one ProductVariant to multiple ProductStreams in the future (to support layered products??). However, ManyToManyFields alone won't be enough to fully support ProductModels that participate in multiple trees / taxonomies, since our current code has many baked-in assumptions.

I think I slightly prefer the current approach, since the stricter taxonomy is more accurate relative to the relationships we actually use today. It also simplifies some boilerplate and lets us avoid calling save_product_taxonomy() in some cases.  Using a ForeignKey should also be more performant since they're automatically indexed, but ManyToManyFields use a through table (which requires an extra JOIN). I don't have any strong preferences, though - any thoughts you have are welcome.


jasinner:
I only looked over the changes in the core/models class and I'm generally in favour of this approach. I agree it simplifies the code, and will hopefully make it easier to maintain.

Thanks for explaining the ForeignKey vs ManyToMany field approaches for the ProductModels. I'm happy to go with whatever approach you think is best. I doubt the current Product relationships will change much.

jasinner (approved):

corgi/core/models.py
```diff
@@ -327,13 +328,19 @@ class ProductModel(TimeStampedModel):
...
+    # Override below on the Product and ProductVersion models
+    productversions = models.ForeignKey(
+        "ProductVersion", on_delete=models.CASCADE, related_name="%(class)ss"
+    )
+
+    # below fields are added implicitly to all ProductModel subclasses
+    # using Django reverse relations, but mypy needs an explicit type hint
+    productstreams: Union[models.ForeignKey, models.Manager["ProductStream"]]
+    productvariants: Union[models.ForeignKey, models.Manager["ProductVariant"]]
+    channels: models.Manager["Channel"]
```
mprpic: I'm not a huge fan of these abstracted fields, it makes it harder to read the child models and they still contain tons of comments referring to these same fields. We could just define them there instead and keep this abstract model for the fields that are the same in all models?

corgi/core/models.py
```diff
@@ -710,7 +781,48 @@ def root_components(self) -> models.QuerySet["Component"]:
         return self.filter(ROOT_COMPONENTS_CONDITION)
 
 
-class Component(TimeStampedModel):
+class ComponentTaxonomyMixin(models.Model):
+    """Add component taxonomy fields and methods to save them for Components."""
+
+    # Separate class not really needed, but easier to add in other places
+    # if we add different types of component classes in the future
```
mprpic: Which means we can just do that abstraction in the future as well, no? :slightly_smiling_face:

mprpic (commented):
Overall approach LGTM. There are way too many changes here for a thorough review so as long as tests pass and you've tried it locally, I'd be fine letting this run on stage and we can catch any bugs there.

I would like us to use a bit less abstractions in general, so that the code is more readable and easier to understand without having to hop from one class to another to build a full mental model of something.

This was a massive effort, good work!

corgi/core/models.py
juspence: Yep, I pulled it out thinking I was going to reuse it, then realized I was mistaken. So I added the comment instead of putting it back - will fix.

corgi/core/models.py
juspence: Yeah....a lot of this was "thinking on my feet", where I saw some duplication I tried to get rid of, instead of needing to change the same thing in 5 different places. Originally this was cleaner and more useful / reusable, but then mypy started complaining and I had to hack around that. I'll just put this back and stop trying to be clever.

But one caveat is we do need *something* defined on the ProductModel base class, so that mypy knows these properties exist and our other methods can stay generic (just handle a ProductModel). So the type hints will need to stay even after the fields themselves get moved to below,

Another caveat is that this gets complicated due to the ForeignKey relationship. We define e.g. a productstreams field on the ProductVariant model, and all ProductModel subclasses will have this productstreams field.

But the type is different depending on which direction of the relationship you're looking at, like I describe in the MR description. Following the ForeignKey link "up" from variants to streams gives you one parent object directly. Whereas following the ForeignKey link "down" from products or versions to streams gives you a set of many related child objects that you have to iterate over.

Using a ManyToManyField would simplify that approach and make the relationships look identical from both ends. That also makes this taxonomy identical to the others. Maybe there's reasons not to do so, though (other than the ones I outline above) - would appreciate any thoughts you have.

mprpic: > But one caveat is we do need something defined on the ProductModel base class, so that mypy knows these properties exist and our other methods can stay generic (just handle a ProductModel). So the type hints will need to stay even after the fields themselves get moved to below,
> 

That's odd. I don't see other abstract models having to define their child fields just to keep mypy happy. Looking at e.g. the Tag model, even if I comment out the `tagged_model` abstract method, mypy doesn't complain about `Tag` not having the `tagged_model` field defined.

> Another caveat is that this gets complicated due to the ForeignKey relationship. We define e.g. a productstreams field on the ProductVariant model, and all ProductModel subclasses will have this productstreams field.
> 
> But the type is different depending on which direction of the relationship you're looking at, like I describe in the MR description. Following the ForeignKey link "up" from variants to streams gives you one parent object directly. Whereas following the ForeignKey link "down" from products or versions to streams gives you a set of many related child objects that you have to iterate over.
> 
> Using a ManyToManyField would simplify that approach and make the relationships look identical from both ends. That also makes this taxonomy identical to the others. Maybe there's reasons not to do so, though (other than the ones I outline above) - would appreciate any thoughts you have.

I think using ForeignKey is fine. It makes it explicit as to how you can link the different levels together. I think it's also fine to just have different sets of fields on each of the product levels without going through all the trouble to make it reusable for all child models. These will hardly ever change and I also don't see us adding yet another level to the product taxonomy.

juspence: > That's odd. I don't see other abstract models having to define their child fields just to keep mypy happy. Looking at e.g. the Tag model, even if I comment out the `tagged_model` abstract method, mypy doesn't complain about `Tag` not having the `tagged_model` field defined.

I think above is because nothing tries to use the generic Tag base class directly. So abstract methods in base classes have to be implemented by subclasses, and method signatures / return types on a subclass can't be more specific than the base class. But if you don't have any methods which take a generic Tag argument, mypy won't care if some Tag subclasses have different properties. The only reason to complain would be if they don't implement a required property, but you commented out the abstract method definition.

The ProductModel errors are a side effect of using ForeignKeys I think. We can't use an abstract property in the base class for e.g. the productvariants field. IIRC at least PyCharm / maybe other tools complain that subclasses aren't implementing all the required properties.

That productvariants field is there (reverse relation from the othermodel field on the ProductVariant model) but tools don't seem to detect it when it's part of a reverse relation. They only see it when it's explicitly defined as a ForeignKey (the forward side of the relation). I looked at defining all four taxonomy fields in the base class, but I think this interefered with the reverse relation's property getting set on the related model.

Thanks for the thoughts about taxonomy approaches - I'll keep the ForeignKeys for now and see if I can make mypy happy with a simpler / more readable product taxonomy.

juspence: Fixed - the newer version uses abstract properties on the ProductModel base class, then overrides these on each subclass to return either a manager or a class instance.

There are still type hints to help mypy figure things out, but the actual logic is closer to where it's being used and has less abstraction. @mprpic Feel free to resolve this thread if it looks good, or please leave a comment if you think it can be improved further.

juspence:
There is one user-facing API change I just noticed. The API endpoint for Channels used to return a list of only related model ofuris, like below:
```
            "products": [
                "o:redhat:3amp"
            ],
            "product_versions": [
                "o:redhat:3amp:2"
            ],
            "product_streams": [
                "o:redhat:3amp:2"
            ],
            "product_variants": [
                "o:redhat:3amp:2:8base-3scale-2.12"
            ]
```

Now we return a list of dicts with the ofuri, name, and a link:
```
"products": [
                {
                    "name": "3amp",
                    "link": "http://localhost:8008/api/v1/products?ofuri=o:redhat:3amp",
                    "ofuri": "o:redhat:3amp"
                }
            ],
            "product_versions": [
                {
                    "name": "3amp-2",
                    "link": "http://localhost:8008/api/v1/product_versions?ofuri=o:redhat:3amp:2",
                    "ofuri": "o:redhat:3amp:2"
                }
            ],
            "product_streams": [
                {
                    "name": "3amp-2",
                    "link": "http://localhost:8008/api/v1/product_streams?ofuri=o%3Aredhat%3A3amp:2",
                    "ofuri": "o:redhat:3amp:2"
                }
            ],
            "product_variants": [
                {
                    "name": "8Base-3SCALE-2.12",
                    "link": "http://localhost:8008/api/v1/product_variants?ofuri=o:redhat:3amp:2:8base-3scale-2.12",
                    "ofuri": "o:redhat:3amp:2:8base-3scale-2.12"
                }
```

There's also a similar change for ProductModel API endpoints. They now return a list of {name, link, uuid} dicts for each Channel (since there's no ofuri property on the Channel model) instead of just a list of Channel names.

I think the newer forms are more useful and would prefer keeping them...we'd just need to make sure our users are aware of the changes when we release this to prod.

corgi/core/models.py
mprpic: Did you forget to push the actual changes? I don't see any new commits on my local branch.

mprpic: Latest version looks good!

juspence:
@jasinner @mprpic FYI, I cleaned up all the TODO comments + a few other minor things. Tests still pass, so merging this. I'll start on CORGI-352 after PTO + email catchup + some other priorities (it never stops...)

Thanks for the advice and reviews!

PR #223, opened 2022-11-16
Add Celery debug pages

juspence:
@jasinner @mprpic  I remembered we added this Celery rate limit setting about 5 minutes before I was supposed to go to sleep. Oh well :smile: 

Jason mentioned "about 8 tasks per minute" is how fast things ran while he was testing...which is exactly what this Celery task rate limit was set to. The comment that "both Brew and SCA tasks take about the same amount of time to finish" also made me suspicious, which eventually jogged my memory about this setting. So thanks for that!

Feel free to click the merge button here, push new commits, or just close if we don't want to do this. I won't touch this branch again. We could merge this as-is, or deploy to stage without merging and see if it actually makes a difference. Although I'd be very suprised if it didn't help. We could also keep the setting instead of completely removing it, and just set it to some higher value.

You can also set or override the rate limit on individual tasks using the task decorator, like we do in other tools:
```
# AKA 36/m across all 6 slow workers
# AKA 3/5s
@app.task(rate_limit="6/m")
```

EDIT: Jason's other MR has the rate limit fix, so I tweaked this one to be just the debug pages. They're useful for me when loading data locally and don't hurt anything, so I might merge this if there's no strong opinions. Will let sit for another day or so.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -113,6 +113,7 @@ INSTALLED_APPS = [
     "django.contrib.sessions",
     "django.contrib.messages",
     "django.contrib.staticfiles",
+    "django.contrib.humanize",
     "django.contrib.postgres",
     "django_celery_results",
     "django_celery_beat",
--- a/corgi/web/templates/app_status.html
+++ b/corgi/web/templates/app_status.html
@@ -1,0 +1,8 @@
+{% extends 'base.html' %}
+
+{# Only task-related view are Celery-dependent; modify below title if this is used anywhere else. #}
+{% block title %}Tasks{% endblock %}
+
+{% block body %}
+<h2 style="margin-left: 1em;">{{ msg }}</h2>
+{% endblock %}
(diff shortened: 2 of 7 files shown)


jasinner:
Thanks for this. I removed the rate limiting in this PR: https://github.com/RedHatProductSecurity/component-registry/pull/224

I had a look at the celery pages but didn't see much value in it for now. With the existing data that is there. Perhaps we could modify it to be more useful though, thanks for sharing!

juspence:
> Thanks for this. I removed the rate limiting in this PR: #224
> 
> I had a look at the celery pages but didn't see much value in it for now. With the existing data that is there. Perhaps we could modify it to be more useful though, thanks for sharing!

Do you mind if I merge these anyway? I find them useful when loading data locally - viewing these pages is faster and easier for me than running the same Redis / Celery inspect commands manually.

It would be nice to make these a permanent part of the app, so I don't have to rebase this commit onto my branches every time I want to use it. I could also add logic to enable these only in stage (or even only in dev, like Django debug toolbar), if you'd rather not expose this info in prod.

jasinner (commented):
LGTM, merging.

PR #225, opened 2022-11-16
Fix email handling for admins

juspence:
@jasinner or @mprpic Quick review please. This change makes Corgi send an email to our mailing list any time there's a 500-level error. 400-level errors are treated as warnings and don't trigger any email.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -171,19 +171,30 @@ LOGGING = {
             "datefmt": "%d/%b/%Y:%H:%M:%S %z",  # Matches the one used by gunicorn
         },
     },
+    "filters": {
+        "require_debug_false": {
+            "()": "django.utils.log.RequireDebugFalse",
+        },
+    },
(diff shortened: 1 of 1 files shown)


jasinner (approved):
LGTM. I'm not opposed to it, but won't this email admins everytime an ERROR level message is logged. Not just 500-level HTTP response?

juspence:
> LGTM. I'm not opposed to it, but won't this email admins everytime an ERROR level message is logged. Not just 500-level HTTP response?

Hmm, I think you're right. I added it to all the logging targets like "root", "django", and "corgi", but the docs used "django.request" in their example. I think if the volume isn't too high, we'd want to know about errors no matter where they came from.

But if it ends up spamming us, I'll open a follow-up PR to get rid of the mail_admins handler in the "root", "django", and "corgi" loggers, and add a new line for "django.requests" like in the docs. That should only send those emails for request-related errors.

PR #229, opened 2022-11-17
Draft: Fix eventlet workers

juspence:
@jasinner This is the draft change I mentioned in my email. Summarizing:

The fast queue shouldn't have tasks like fetch_brew_build or fetch_modular_build, since those spawn many copies. If the queue fills up with many instances of those tasks, the daily load tasks and our daily monitoring task will be delayed and not run on time.

Adding a separate eventlet_* queue and eventlet workers lets us move I/O-bound tasks into this queue. These I/O-bound tasks can be processed quickly by eventlet. They get their own dedicated queue so that the fast tasks aren't delayed and so that these asynchronous / greenthread-based workers don't get blocked by some CPU-bound task.

More detailed info is on each commit message.


jasinner:
Incorporated into https://github.com/RedHatProductSecurity/component-registry/pull/231

PR #230, opened 2022-11-18
CORGI-332 improve logging for splunk

jasinner:
@RedHatProductSecurity/corgi-devs 

This uses key=value pairs separated by commas for logging as per the [splunk best-practices guide](https://dev.splunk.com/enterprise/docs/developapps/addsupport/logging/loggingbestpractices/#Use-clear-key-value-pairs). In order to group log messages together more easily I've switched all tasks to use the celery task logger which includes the task name, and id in log messages. I've also added a thread attribute to help tie logs from collector and other classes together to the task which called it. Finally I override the task_success message to include a task_runtime key which we can use to gather metrics on task runtimes.

### DIFF
--- a/config/celery.py
+++ b/config/celery.py
@@ -1,6 +1,7 @@
 import logging
 
 from celery import Celery  # type: ignore[attr-defined]
+from celery.app import trace
 
 logger = logging.getLogger(__name__)
 
@@ -12,3 +13,7 @@ app.config_from_object("django.conf:settings", namespace="CELERY")
 # So all tasks in any corgi.tasks submodule are automatically discovered
 # And no config changes are needed when a new submodule is added
 app.autodiscover_tasks()
+
+trace.LOG_SUCCESS = """\
+%(return_value)s\" task_name=%(name)s, task_id=%(id)s, task_runtime=%(runtime)ss\
+"""
--- a/config/celery.py
+++ b/config/celery.py
@@ -14,6 +14,9 @@ app.config_from_object("django.conf:settings", namespace="CELERY")
 # And no config changes are needed when a new submodule is added
 app.autodiscover_tasks()
 
-trace.LOG_SUCCESS = """\
-%(return_value)s\" task_name=%(name)s, task_id=%(id)s, task_runtime=%(runtime)ss\
+LOG_FORMAT = """\
+%(return_value)s\", task_name=%(name)s, task_id=%(id)s, task_runtime=%(runtime)ss\
 """
+
+trace.LOG_SUCCESS = LOG_FORMAT
+trace.LOG_FAILURE = LOG_FORMAT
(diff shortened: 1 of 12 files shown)


config/celery.py
```diff
@@ -12,3 +13,7 @@
 # So all tasks in any corgi.tasks submodule are automatically discovered
 # And no config changes are needed when a new submodule is added
 app.autodiscover_tasks()
+
+trace.LOG_SUCCESS = """\
+%(return_value)s\" task_name=%(name)s, task_id=%(id)s, task_runtime=%(runtime)ss\
```
juspence: Logging the runtime is a great idea, thanks for doing the research here!

Should we also log the same info on trace.LOG_FAILURE? And do you think it's valuable to log the task args and task kwargs as well? I'm thinking of something like below:

```
LOG_FORMAT = """\
%(return_value)s\", task_name=%(name)s, task_id=%(id)s, task_args=%(args)s, task_kwargs=%(kwargs)s, task_runtime=%(runtime)ss\
"""

trace.LOG_SUCCESS = LOG_FORMAT
trace.LOG_FAILURE = LOG_FORMAT
```

config/settings/base.py
```diff
@@ -163,13 +164,15 @@
...
+LOG_FORMAT_END = f'level=%(levelname)s, app=corgi, env={get_env()}, msg="%(message)s"'
+
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
     "formatters": {
         "default": {
-            "format": "[%(asctime)s] [%(name)s:%(lineno)d] %(levelname)s: %(message)s",
-            "datefmt": "%d/%b/%Y:%H:%M:%S %z",  # Matches the one used by gunicorn
+            "format": f"{LOG_FORMAT_START}, {LOG_FORMAT_END}",
```
juspence, line 175: For this Django logging config, the new version almost has the same information as the old one, just formatted differently and with thread / app / env added. But it's missing name and lineno which I think are useful.

config/settings/base.py
```diff
@@ -278,6 +281,11 @@
     ],
 )
 
+CELERY_WORKER_LOG_FORMAT = f"{LOG_FORMAT_START}, {LOG_FORMAT_END}"
+CELERY_WORKER_TASK_LOG_FORMAT = (
+    f"{LOG_FORMAT_START}, task_name=%(task_name)s, task_id=%(task_id)s, {LOG_FORMAT_END}"
+)
+
```
juspence, line 290: For this Celery logging config, should we also log the processName here? Celery does this by default, but it looks like we won't anymore after this change.

juspence (approved):
LGTM! Questions are minor so this is fine to merge I think.

config/settings/base.py
jasinner: Let's add it back later.

config/settings/base.py
jasinner: I don't find it useful. It usually says something like MainThread all the time.

config/celery.py
jasinner: It made the line too long for flake8 and I don't want to risk splitting the link in the logs.

PR #231, opened 2022-11-18
CORGI-351 sca eventlet

jasinner:
Adjust the slow worker to also use eventlet. I also incorporated the feedback from https://github.com/RedHatProductSecurity/component-registry/pull/229

- remove dnspython
- move tasks with many instances to slow queue
- increase concurrency of slow queue

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -92,10 +92,10 @@ services:
 
   corgi-celery-fast:
     deploy:
-      replicas: 2
+      replicas: 1
       resources:
         limits:
-          memory: 256M  # Keep in sync with OpenShift mem limits to catch OOM problems
+          memory: 128M  # Keep in sync with OpenShift mem limits to catch OOM problems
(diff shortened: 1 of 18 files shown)


docker-compose.yml
```diff
@@ -92,10 +92,10 @@ services:
 
   corgi-celery-fast:
     deploy:
-      replicas: 2
+      replicas: 1
```
juspence, line 95: @jasinner Assuming this change matches what's in the ops repo, I'd prefer three workers here so that we get one per AWS Availability Zone (data center). From the MP+ High-Availability guidelines:

"Run at least 2 pod replicas (3 replicas for C1 applications) using a replication controller (Deployment/DeploymentConfig/StatefulSet)."

I don't want to paste an internal link here, but you can search for "MP+ high availability" on The Source.

EDIT: Nvm, just saw it's set to three in the ops repo. Thank you!

PR #232, opened 2022-11-18
Only log django.request errors to emails, not all errors at root

juspence:
@mprpic Would you mind giving this a quick sanity check? Or please let me know if you think we should just turn off email logging. I was trying to copy SDEngine's behavior, where we get emails for 500-level errors (which happen only rarely). But the spam is a little frustrating.

CORGI-354 is the specific bug that made me want a traceback, and made me realize we weren't getting these emails.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -190,11 +190,13 @@ LOGGING = {
         },
     },
     "root": {
-        "handlers": ["console", "mail_admins"],
+        "handlers": ["console"],
         "level": "WARNING",
     },
     "loggers": {
         "django": {"handlers": ["console"], "level": "WARNING"},
+        # Mail errors only, but set level=WARNING here to pass warnings up to parent loggers
+        "django.request": {"handlers": ["mail_admins"], "level": "WARNING"},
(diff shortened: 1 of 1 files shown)


mprpic:
Doesn't reporting of 500-level errors work out of the box? I don't see anything in SDEngine that would be specific to mailing out tracebacks on HTTP errors. What's the purpose of the mail_admins logger? We're already running with `disable_existing_loggers=False` which would use the admin logger since it's configured by default.

juspence:
> Doesn't reporting of 500-level errors work out of the box? I don't see anything in SDEngine that would be specific to mailing out tracebacks on HTTP errors. What's the purpose of the mail_admins logger? We're already running with `disable_existing_loggers=False` which would use the admin logger since it's configured by default.

It should, but we didn't get any emails when I first triggered CORGI-354. The emails only started coming in after I set up this logging (and we got spammed because I set it up on the wrong / too many log handlers).

EDIT: I'm going to merge this since I'll be on PTO and don't want to leave the spam on while I'm out. Jason reviewed the initial MR which turned on this logging: https://github.com/RedHatProductSecurity/component-registry/pull/225

Feel free to disable this entirely if we still get spammed after this is merged.

PR #235, opened 2022-11-21
adjust env var not to clash with existing matcher

jasinner:
looks like 'env' is an existing matcher key for events, let's try using something else so we can select the key in a search.


JimFuller-RedHat (commented):
LGTM

PR #240, opened 2022-11-23
Improve analysis meta

jasinner:
Supports CORGI-313

I actually set out to do CORGI-343, and while searching the stage data I found that I couldn't find the analysis_meta / source data on any components except for those found via SCA analysis. That lead me to fix that issue by writing the old analysis_meta field, which tells us how a component for first discovered, so that it now directly writes to the collector 'meta.source' field. From there, its a little easier to persist it into Component.meta_attr. We save everything that wasn't popped off and put in Component fields into meta_attr.

I was trying to test getting bundled deps from a specfile on the podman package when I discovered some errors parsing the 'go list' output. I also found these errors in the stage TaskResult data, so I put some checks around the 'go list' output parsing to skip those results with errors. 

Finally there was some code already using 'meta.source' attribute of collector data and I discovered it had a bug which meant that it wasn't actually returning any results. I fix that, and now added the 'filename' attribute to the ComponentSerializer. I think this attribute is only being set on some components of type 'upstream' but we can probably also use the field for other types too, particularly the 'rpm' type.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -325,7 +325,7 @@
         "filename": "tests/test_sca.py",
         "hashed_secret": "75272b8978226ca0852fd26fb75ed8c2bd4f4680",
         "is_verified": true,
-        "line_number": 150,
+        "line_number": 155,
         "is_secret": false
       },
       {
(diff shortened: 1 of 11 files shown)


jasinner:
put in draft till I can fix the tests (mypy,schema)

JimFuller-RedHat:
small question - now that we are surfacing up 'source' from looser metadata should we consider constraining these values, otherwise LGTM

jasinner:
I don't think it's necessary constrain source (aka the old analysis/meta) at this point. I think it's only really useful for debugging, and not something we want to expose on the API.

JimFuller-RedHat:
Yes, if this is not really a part of core and only used for debugging it could just stay in meta attr, which I've noted I have not followed myself previously (looking at score related attributes which can 'go away').

PR #241, opened 2022-11-24
dont expire singleton locks, clear then on startup

jasinner:
See CORGI-386 for details of why singleton lock expiry was set.

I'm also splitting out sca tasks to a dedicated 'cpu' queue with prefork worker. When testing locally I see the syft process sometimes uses up alot of CPU.

The fast queue was also changed to prefork and concurrency reduced because we almost never have many fast tasks to process.


JimFuller-RedHat:
LGTM and we would like to get this in as soon as we can.

PR #242, opened 2022-11-24
CORGI-345 - remove rpmmod from latest filter

JimFuller-RedHat:
As all of RPM children should have their SRPM parents listed we do not need RPMMOD in latest filter.


jasinner (approved):
LGTM, thanks

PR #254, opened 2022-11-28
Fix celery slow queue length

juspence:
@RedHatProductSecurity/corgi-devs This should address some issues with our slow queue due to processing duplicate tasks in our queue / relations in the PCR table. I've added more detailed notes to the commit message for the big changes. Quick sanity check appreciated, or please let me know if this should sit until Jason's back and can review himself next week. Thanks!

### DIFF
--- a/corgi/tasks/yum.py
+++ b/corgi/tasks/yum.py
@@ -3,11 +3,17 @@ from urllib.parse import urlparse
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
+from django.utils import timezone
 
 from config.celery import app
 from corgi.collectors.yum import Yum
 from corgi.core.models import Channel, ProductComponentRelation, ProductStream
-from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS, _create_relations
+from corgi.tasks.common import (
+    RETRY_KWARGS,
+    RETRYABLE_ERRORS,
+    _create_relations,
+    get_last_success_for_task,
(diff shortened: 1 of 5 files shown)


corgi/tasks/yum.py
```diff
@@ -19,9 +25,15 @@
     retry_kwargs=RETRY_KWARGS,
     soft_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
 )
-def fetch_unprocessed_yum_relations(force_process: bool = False) -> int:
+def fetch_unprocessed_yum_relations(force_process: bool = False, created_since: int = 0) -> int:
```
mprpic: Perhaps call it `days_created_since` so it's immediately clear that it expects an int value for number of days? Or change it to expect the timedelta value as you did in `fetch_unprocessed_relations()`.

juspence: I renamed it to "days_created_since" so we can easily set it to 0 and get the "look up last success" behavior, else explicitly force processing a certain number of days.

But I think when I force-pushed it cleared your review. Or maybe you didn't approve yet? I'll let this sit a bit while doing some paperwork - will merge tomorrow if no further feedback from you / Jim I guess. Thanks for reviewing!

mprpic (approved):
LGTM!

JimFuller-RedHat (approved):
nice catch ... I think I follow and as long as force_process continues to work as before (so we can always override)... this is good to go in now and we can explain to @jasinner when he gets back

PR #255, opened 2022-11-29
CORGI-365: latest filter tweaks

JimFuller-RedHat:
consolidate query based on performance analysis


JimFuller-RedHat:
this PR effectively reverts back to my original change, which we reviewed last week ... I did a series of perf tests which prove out this was the best approach (for now)

PR #256, opened 2022-11-30
CORGI-374: Fix (some of our) daily errors

juspence:
@RedHatProductSecurity/corgi-devs Initial PR which fixes timeouts for Errata Tool tasks, and maybe fixes IntegrityErrors when creating ComponentNodes.

I'm not sure if the second part will really fix the issue, so split this off here for review + discussion. I'll have another PR after lunch with more fixes for other errors.


mprpic (approved):
Ah, nice, I was looking at that exact get_or_create on cnodes as part of CORGI-348. I guess this PR addresses that issue as well then.

juspence:
> Ah, nice, I was looking at that exact get_or_create on cnodes as part of CORGI-348. I guess this PR addresses that issue as well then.

Oops. I just saw that error in today's email and finally had a revelation about the cause - didn't realize we had a separate ticket. I *think* this should fix it, but not 100% sure. Maybe we can leave that ticket open a few more days and see if the IntegrityError keeps happening.

Thanks for reviewing! I'll review your PR as soon as I get back from lunch.

PR #257, opened 2022-12-02
CORGI-365: add heuristic with release 'el# (ex. .el8, .el9) to ensure correct latest filter

JimFuller-RedHat:


jasinner:
Have you seen https://fedoraproject.org/wiki/Archive:Tools/RPM/VersionComparison ? It gives more details about how RPM versions are compared. For now all the versions we have in the latest filter should use these rules for comparison. I think your alorgithm looks pretty good. As it really only has to work for builds with completion_time out of order. I expect it will break in the case an epoch is included in a later build, or if the version part only consists of numbers, and release has no 'el[0-9]' section. But those are not situations we have encountered yet,and are unlikely to encouter, I think.

JimFuller-RedHat:
thx for the link - yes I did see that when researching ... current heuristic is evidence based eg. opted for this as its fast versus being 100% comprehensive and as you mentioned we have not encountered.

jasinner (approved):

PR #258, opened 2022-12-03
CORGI-374, CORGI-376: Fix more errors from daily monitoring email

juspence:
@RedHatProductSecurity/corgi-devs As title says, plus a possible fix for the high CPU usage issue that our slow workers keep hitting. Those slow workers use the eventlet pool and should only process I/O bound tasks.

But they're still configured to pull CPU-bound tasks from the fast queue. If any of the greenthreads inside the eventlet worker pick up a CPU-bound task, they won't be able to yield as often (not doing any I/O) so end up blocking the worker. Then the other 19 greenthreads on the eventlet worker will get starved of CPU time / never execute, which is probably why our tasks look like they're hanging.

Even if that's not *the* issue, it at least simplifies our queueing setup. Now slow workers only process slow tasks, fast workers only process fast tasks, and CPU workers only process SCA tasks (which are the only tasks in the CPU queue). So our setup becomes a little easier to reason about and debug.

Not going to merge this to avoid destabilizing things over the weekend, but opening now so Jason can review in his morning. Hope you enjoyed your time off!

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -362,4 +362,7 @@ APP_STREAMS_LIFE_CYCLE_URL = os.getenv("CORGI_APP_STREAMS_LIFE_CYCLE_URL", "")
 MANIFEST_HINTS_URL = os.getenv("CORGI_MANIFEST_HINTS_URL")
 
 LOOKASIDE_CACHE_BASE_URL = f"https://{os.getenv('CORGI_LOOKASIDE_CACHE_URL')}/repo"
+
+# Set to False to disable software composition analysis tasks.
+SCA_ENABLED = strtobool(os.getenv("CORGI_SCA_ENABLED", "true"))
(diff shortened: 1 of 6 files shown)


jasinner (approved):
LGTM, sorry I missed reviewing this yesterday.

juspence:
> LGTM, sorry I missed reviewing this yesterday.

No worries! I know catch-up after PTO is crazy. I see you already fixed the slow-workers issue (and thank you for that!) and that was really the only reason I wanted to get the queuing fix in quickly,

juspence:
@jasinner Thanks for reviewing! Some of my changes to try and avoid SoftTimeLimitExceeded errors in the SCA task broke the mock-patching in the tests. For now I've ripped out those changes to make the pipeline pass again.

I'll either add back those fixes in a separate MR or (hopefully) find a better root cause and fix for those failures.

I also added a test confirming _clone_source failures raise the errors we expect. Please feel free to merge when you're done testing the Celery workers in stage, or I can merge in my morning.

PR #260, opened 2022-12-05
fix load unprocessed relations mgnt command to work with all relation…

jasinner:
…s, not only latest


jasinner:
Fixes https://github.com/RedHatProductSecurity/component-registry/pull/259

PR #261, opened 2022-12-05
minor: perf improvements

JimFuller-RedHat:
* turns out drf gets a little bit faster if you give it hints about what is read_only or not (we could expand on this once we grok 'write' scenario)
* .exclude is handled internally different then a straight filter

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -166,7 +166,7 @@ class TagSerializer(serializers.Serializer):
 class SoftwareBuildSerializer(serializers.ModelSerializer):
     tags = TagSerializer(many=True, read_only=True)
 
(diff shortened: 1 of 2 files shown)


juspence (approved):
Yep, that .exclude() vs .filter() behavior is annoyingly tricky. Especially when doing lookups on fields across a related model (related_model__field=xyz...)

Nice find with read_only=True making things faster! There's probably more fields we can add that param to as well - I think it's talking about read-only API endpoints (e.g. we don't support PUT / POST / PATCH on anything), not read-only fields in our DB. But no need to cram that all into this MR.

Jason's doing some debugging right now in stage, so please check with him before merging. Otherwise LGTM!

PR #263, opened 2022-12-06
Fix a small amount of inline styling that was missed and the /docs API endpoint

juspence:
There's a related issue with the /docs API returning an error - it works locally but not in stage. I see this line in the traceback:
```
File "/usr/local/lib/python3.9/site-packages/drf_spectacular/views.py", line 117, in _get_sidecar_url
    return static(f'drf_spectacular_sidecar/{package}')
```

Which seems to cause this error:
```
File "/usr/local/lib/python3.9/site-packages/django/contrib/staticfiles/storage.py", line 417, in stored_name
    raise ValueError("Missing staticfiles manifest entry for '%s'" % clean_name)

Exception Type: ValueError at /api/v1/schema/docs
Exception Value: Missing staticfiles manifest entry for 'drf_spectacular_sidecar/swagger-ui-dist'
```

Perhaps it's because we don't redirect 'drf_spectacular_sidecar/swagger-ui-dist', with no trailing slash, to 'drf_spectacular_sidecar/swagger-ui-dist/' with a trailing slash.

This is controlled by the APPEND_SLASH setting:
```
APPEND_SLASH = False  # Default: True
# If True, and request URL doesn't match any patterns in URLconf
# and request URL doesn’t end in slash
# HTTP redirect is issued to same URL with slash appended
# Note that the redirect may cause any data submitted in a POST request to be lost
# With False, our URLconf controls. "api/path/" won't match, should give a 404
# Only "api/path" should succeed
```

I'm still not sure why it works locally...might deploy an experimental fix from this branch for testing. I'll tag when this is ready for review.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -345,13 +345,17 @@ class Brew:
             # Legacy OSBS builds such as 1890187 copy source code into dist-git but specify where
             # the source code came from using the 'go' stanza in container.yaml
             # ref: https://osbs.readthedocs.io/en/osbs_ocp3/users.html#go
-            go_modules = []
-            if "go" in build_info["extra"]["image"]:
-                for module in build_info["extra"]["image"]["go"]["modules"]:
-                    if "module" in module:
-                        go_modules.append(module["module"])
-            if go_modules:
-                component["meta"]["upstream_go_modules"] = go_modules
+            # Handle case when "go" key is present but value is None
+            go = build_info["extra"]["image"].get("go", {})
+
+            # AND handle case when "modules" key is present but value is None
+            if go and go.get("modules", []):
+                go_modules = tuple(
+                    module["module"] for module in go["modules"] if module.get("module")
+                )
+                if go_modules:
+                    # Tuple above can be empty if .get("module") name is always None / an empty str
+                    component["meta"]["upstream_go_modules"] = go_modules
 
             # builds such as 1911112 have all their info in typeinfo as they use remote_sources map
             # in remote_source json, and tar download urls by cachito url
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -694,7 +694,7 @@ class Brew:
             "stream": modulemd["data"]["stream"],
             "context": modulemd["data"]["context"],
             "components": modulemd["data"].get("components", []),
-            "rpms": modulemd["data"]["xmd"]["mbs"]["rpms"],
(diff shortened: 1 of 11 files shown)


juspence:
Setting DEBUG = False locally is enough to reproduce this. It's  issue # 847 in the drf-spectacular repo - will open a PR there tomorrow.

juspence:
@RedHatProductSecurity/corgi-devs This bugfix is ready for review. It turns out drf-spectacular didn't fully support the Content-Security-Policy changes I made, but I didn't realize / it wasn't being tested.

DRF-Spectacular has a sidecar to serve static files locally, which avoids two issues with a strict Content-Security-Policy: inline styles / scripts, and files that are served from a CDN but don't support hashing because they're dynamically generated.

However, the sidecar didn't handle static files correctly when a storage backend like "CompressedManifestStaticFilesStorage" was being used. That backend is also only used in stage and production because we've set DEBUG = True locally.

I opened a PR in the drf-spectacular repo that fixes the issue. Until a new version that includes this fix is released, we can just refer to the specific commit with the fix in our requirements.txt file. If we don't merge this, the /docs endpoint will be unusable and we'll keep getting error emails whenever someone visits it.

juspence:
Added a fix for two monitoring email errors and CORGI-384.

corgi/collectors/brew.py
```diff
@@ -345,13 +345,17 @@ def get_container_build_data(self, build_id: int, build_info: dict) -> dict:
...
             # the source code came from using the 'go' stanza in container.yaml
             # ref: https://osbs.readthedocs.io/en/osbs_ocp3/users.html#go
-            go_modules = []
-            if "go" in build_info["extra"]["image"]:
-                for module in build_info["extra"]["image"]["go"]["modules"]:
-                    if "module" in module:
-                        go_modules.append(module["module"])
-            if go_modules:
-                component["meta"]["upstream_go_modules"] = go_modules
+            # Handle case when "go" key is present but value is None
```
JimFuller-RedHat, line 348: perhaps I missed ... how is this related to inlinie styling and /docs endpoint ?

juspence: > perhaps I missed ... how is this related to inlinie styling and /docs endpoint ?

See above note:
> Added a fix for two monitoring email errors and CORGI-384.

I pushed those fixes here since the diff was small, this hadn't been reviewed yet, and it didn't make sense to me to have two "miscellaneous small bugfix" MRs open at the same time.

PR #264, opened 2022-12-06
switch slow workers to prefork

jasinner:
CORGI-380 and CORGI-376

Try to avoid stuck slow workers by reverting them to the default 'prefork' pool, and also reducing concurrency to the application wide default of 5.


juspence (approved):
LGTM, thank you for picking this one up!

PR #266, opened 2022-12-09
Community mode

jasinner:
This creates a new mode "community" for community projects. 

It removes all internal URLs from config to enable trying the project outside of the RH firewall. 

The openstack-rdo product streams doesn't have any builds because they exist in the [Centos Koji instance.](https://cbs.centos.org/koji/taginfo?tagID=2273)


config/community_product_definitions.json
```diff
@@ -0,0 +1,499 @@
...
+        "fedora-37",
+        "fedora-all",
+        "fedora-rawhide",
+        "epel-7",
+        "epel-8",
+        "epel-all",
+        "openstack-rdo"
+      ],
+      "business_unit": "Community",
+      "errata_product_tags": []
```
mprpic: Are we trying to keep parity of this data to our internal version? I guess we could strip all these attributes that don't really matter for the community version.

config/utils.py
```diff
@@ -7,7 +7,7 @@ def get_env():
 
 
 def running_dev():
-    return get_env() == "dev"
+    return get_env() in ("dev", "community")
```
mprpic: Don't we want this to be a separate `running_community`?

config/community_product_definitions.json
jasinner: I really just wanted a copy of the data without having to rely on the dashboard service, so that we didn't have any reliance on internal services. 

I did remove all the unused attributes now.

jasinner:
@RedHatProductSecurity/corgi-devs I updated the tests so they can be run without any internal URLs and fix the the few suggestions Martin had.

This should be ready for review. It works well for Fedora. We'll have to figure out a solution for openstack-rdo.

jasinner:
I removed VCR from the last 2 tests which were using them, so removed pytest-recording dependency from the project as well. This also allowed to discontinue the use of git-lfs, so all files are now directly part of the repository. Git LFS can be removed as a dev dependency.

jasinner:
Now with SCA enabled

corgi/collectors/prod_defs.py
```diff
@@ -9,7 +12,7 @@ class ProdDefs:
     GSSAPI_AUTH = HTTPSPNEGOAuth()
 
     @classmethod
-    def get_product_definitions(cls) -> dict:
+    def get_enterprise_product_definitions(cls) -> dict:
```
JimFuller-RedHat: nitpick: not a fan of 'enterprise' denotation ... alternately (if community is a mode) then we could route between func impl based on mode instead of incurring different codepaths - note I make this suggestion not really understanding potential for violence in the codebase.

Dockerfile-community
```diff
@@ -0,0 +1,32 @@
+FROM registry.redhat.io/ubi8/ubi:8.6
```
JimFuller-RedHat, line 1: alternately this might be from quay ? I do not know the right answer as to pref in community.

JimFuller-RedHat:
LGTM

config/settings/community.py
```diff
@@ -0,0 +1,26 @@
...
+#    image: sdengine
+#    command: python3 -m smtpd -n -c DebuggingServer localhost:1025
+#    ports:
+#      - "1025:1025"
+
+ALLOWED_HOSTS = ["*"]
+
+UMB_BREW_MONITOR_ENABLED = strtobool("false")
+
+COMMUNITY_MODE_ENABLED = strtobool("true")
```
juspence: You can just set these to False / True directly, instead of using strtobool on a constant string.

corgi/collectors/prod_defs.py
```diff
@@ -18,17 +18,40 @@ def get_product_definitions(cls) -> dict:
...
+
+    @classmethod
+    def load_product_definitions(cls) -> list:
+        try:
+            data = cls.get_product_definitions_service()
+        except ConnectionError as e:
+            if settings.COMMUNITY_MODE_ENABLED:
+                data = cls.get_community_product_definitions()
+            else:
+                raise e
```
juspence, line 48: This looks a little odd. Maybe it's better to check the setting before we make a request:
```suggestion
        if settings.COMMUNITY_MODE_ENABLED:
            data = cls.get_community_product_definitions()
        else:  # Enterprise mode
            data = cls.get_product_definitions_service()
```

config/settings/base.py
```diff
@@ -360,16 +360,13 @@
...
-COMMUNITY_PRODUCTS_ENABLED = strtobool(os.getenv("CORGI_COMMUNITY_PRODUCTS_ENABLED", "false"))
+COMMUNITY_MODE_ENABLED = strtobool(os.getenv("CORGI_COMMUNITY_MODE_ENABLED", "false"))
 
 # Brew
 BREW_URL = os.getenv("CORGI_BREW_URL")
 BREW_WEB_URL = os.getenv("CORGI_BREW_WEB_URL")
 BREW_DOWNLOAD_ROOT_URL = os.getenv("CORGI_BREW_DOWNLOAD_ROOT_URL")
 
-# RHEL Compose
-RHEL_COMPOSE_BASE_URL = os.getenv("CORGI_TEST_DOWNLOAD_URL")
```
juspence, line 371: Instead of removing this, should we rename it to "COMPOSE_BASE_URL"? We could use the internal / Red Hat value when not running in community mode, and default it to the public / Fedora value "
https://kojipkgs.fedoraproject.org/compose/" otherwise.

juspence (commented):
Looks great so far, below are all minor. I still need to review the rest...trying to have this done soon, sorry for the delay.

config/settings/base.py
jasinner: The setting is not used anywhere. I think because the compose data has a fully qualified link in prod_defs.

corgi/collectors/prod_defs.py
jasinner: I designed it that way originally. However that will put a burden on us to keep the local product definitions up to date. I would rather have it this way so that if we can access the internal dashboard we prefer to use it instead of the local copy. The local copy does allow the codebase to standalone outside of the RH firewall, but I think it's better to only use it where dashboard is not available.

juspence: Thanks for explaining! Your approach here sounds good in that case.

EDIT: Actually, I just realized - this code is running in stage / prod as well as locally, right? So if there's some kind of outage or network glitch, stage might fall back to using the community product definitions with the current code. Maybe that's not an issue - if so, please feel free to mark this resolved.

tests/test_brew_collector.py
```diff
@@ -666,32 +660,32 @@ def test_save_component():
...
     noarch_rpms = noarch_rpms_by_id.values()
     assert len(noarch_rpms) == len(NOARCH_RPM_IDS)
     assert set(noarch_rpms_by_id.keys()) == set(NOARCH_RPM_IDS)
 
 
-@pytest.mark.vcr(match_on=["method", "scheme", "host", "port", "path", "body"])
-@patch("corgi.tasks.sca.cpu_software_composition_analysis.delay")
-def test_fetch_rpm_build(mock_sca):
+@patch("corgi.tasks.brew.Brew")
+@patch("corgi.tasks.brew.cpu_software_composition_analysis.delay")
```
juspence: Is this a typo?
```suggestion
@patch("corgi.tasks.sca.cpu_software_composition_analysis.delay")
```

config/settings/base.py
```diff
@@ -398,7 +395,7 @@
 # Manifest hints url
 MANIFEST_HINTS_URL = os.getenv("CORGI_MANIFEST_HINTS_URL")
 
-LOOKASIDE_CACHE_BASE_URL = f"https://{os.getenv('CORGI_LOOKASIDE_CACHE_URL')}/repo"
+LOOKASIDE_CACHE_BASE_URL = os.getenv("CORGI_LOOKASIDE_CACHE_URL")
```
juspence, line 398: The Pulp and Compose URL / env var changes look OK. The old values are only used in tests, or aren't used at all, and the test still pass.

But this URL is used in the sca.py task:
```
# https://<host>/repo/rpms/containernetworking-plugins/v0.8.6.tar.gz/md5/
# 85eddf3d872418c1c9d990ab8562cc20/v0.8.6.tar.gz
lookaside_download_url = (
    f"{settings.LOOKASIDE_CACHE_BASE_URL}/{package_type}/{package_name}/{lookaside_path}"
)
```

I think we need to hardcode the ```https://``` protocol and ```/repo``` path into the lookaside_download_url variable directly, if we remove them from the setting. Or we could keep the code the same, and just update our local environment variables and the Gitlab CI variable to include the protocol and path bits.

Otherwise we will probably have SCA task failures / 404 Not Found errors when trying to download the source for some build. Maybe I'm misunderstanding how this is used - please feel free to mark this resolved if so.

EDIT: I see you updated the Gitlab CI variable already to make this work.

docs/developer_guide.md
```diff
@@ -61,7 +66,6 @@ export CORGI_DOMAIN
 export CORGI_EMAIL_HOST
 export CORGI_SERVER_EMAIL
 export CORGI_UMB_BROKER_URL
-export CORGI_PULP_URL
```
juspence, line 64: You only removed the CORGI_TEST_PULP_URL, right? If the regular CORGI_PULP_URL is still needed, let's keep it in the documentation.

docs/developer_guide.md
```diff
@@ -72,16 +76,16 @@ environment's `venv/bin/activate` script.
...
+```
+
+
+If doing enterprise development, be sure to set this to relevant value from CI environment.
+If it's not set the local product-definitions.json file in the config directory will be used.
+```bash
 CORGI_PRODSEC_DASHBOARD_URL=<value>
-CORGI_PULP_URL=<value>
-CORGI_PULP_USERNAME=<value>
-CORGI_PULP_PASSWORD=<value>
```
juspence, line 84: As above.

docs/developer_guide.md
```diff
@@ -72,16 +76,16 @@ environment's `venv/bin/activate` script.
 In order for the environment variables to be passed into the celery pods started by podman-compose you'll also have to
 add them to a .env file e.g.:
 ```bash
-CORGI_APP_STREAMS_LIFE_CYCLE_URL=<value>
-CORGI_BREW_DOWNLOAD_ROOT_URL=<value>
-CORGI_BREW_URL=<value>
-CORGI_ERRATA_TOOL_URL=<value>
-CORGI_LOOKASIDE_CACHE_URL=<value>
-CORGI_MANIFEST_HINTS_URL=<value>
```
juspence, line 80: I think we should leave the APP_STREAMS, ERRATA_TOOL, and MANIFEST_HINTS URLs in the documentation, if they're still used in enterprise mode. We can just note that they're not needed for community mode.

Or if they're not used at all, then removing them here is fine.

corgi/tasks/management/commands/loadproductdata.py
```diff
@@ -1,56 +1,81 @@
...
+
+            # TODO load_brew_tags, can't find
+        else:  # Enterprise mode
+            self.do_load_et_products()
+            self.do_update_products()
+            self.do_update_cdn_repo_channels()
+            self.do_save_composes()
+            self.do_load_brew_tags()
+            self.do_load_yum_repositories()
+            self.delay_setup_pulp_relations()
```
juspence: I think in enterprise mode, load_yum_repostories should be called with .delay and run after setup_pulp_relations. The Yum repo tasks / collector works better when Channels are already set up, so should always run after the Pulp repo tasks / collector:
```suggestion
            self.delay_setup_pulp_relations()
            self.delay_load_yum_repositories()
```

In community mode we obviously won't have real channels...I don't know what the rest of our data will actually look like in that case, but I think the Yum repo data should be OK. The collector will create a fake Channel based on the Yum repo's URL, if a Channel with a matching URL doesn't already exist.

corgi/tasks/management/commands/loadproductdata.py
```diff
@@ -1,56 +1,81 @@
...
 class Command(BaseCommand):
 
     help = "Fetch product data from Product Definitions."
 
     def handle(self, *args, **options):
+        if settings.COMMUNITY_MODE_ENABLED:
+            self.do_update_products()
+            self.do_load_yum_repositories()
+
+            # TODO load_brew_tags, can't find
```
juspence: I lost track of where you got to on this - did you find a solution for the community Brew tags, or did we decide not to load them / they don't exist?

juspence (approved):
Looks good! Below are minor - feel free to merge this when you think it's ready. We can open follow-up PRs if needed to fix any issues.

I see a bunch of LFS files that changed. The diff in GitHub only says "Git LFS file not shown" - did you mean to change these, or should they have been completely removed instead?

I think we might be able to run "community mode" without needing a separate settings module or docker-compose.yml file. I only mention because I think you said you would have liked to keep this even simpler (or maybe I'm misremembering?)

If so, we can add public defaults to internal env vars in docker-compose. The Pypi value will only be used if the env var is unset:
${PIP_INDEX_URL:-"https://pypi.org/simple"}

And document which env vars should be set / unset for community mode:
UMB_BREW_MONITOR_ENABLED turned off
COMMUNITY_MODE turned on

I think that should remove the need for a separate docker-compose and settings file. The Dockerfile might still need some tweaking. I could do some or all of this in a follow-up MR (or push a commit here) if you're interested.

Dockerfile-community
jasinner: UBI license is not restrictive and it can be used in open source projects [without requiring a subscription](https://developers.redhat.com/articles/ubi-faq#ubi_details).

corgi/collectors/prod_defs.py
jasinner: Good catch. I think that's still OK though as the code will only load products which don't have business unit equal to "Community", so no chance of loading community products into enterprise version. The prod_defs collector will simply return 0 product definitions, which should be a no-op I think.

corgi/tasks/management/commands/loadproductdata.py
jasinner: There is a follow up task for this, CORGI-302. I remove the TODO.

corgi/tasks/management/commands/loadproductdata.py
jasinner: Ok, I'll make that change.

docs/developer_guide.md
jasinner: They are included on lines 42-48 of the guide.

docs/developer_guide.md
jasinner: It's there on line 40.

docs/developer_guide.md
jasinner: They are also there, on lines 69 and 70.

corgi/collectors/prod_defs.py
jasinner: Also before loading local product definitions, we first check if COMMUNITY_MODE is ENABLED. If it's not enabled we just reraise the ConnectionError.

tests/test_brew_collector.py
jasinner: Good catch. Fixed.

jasinner:
> I see a bunch of LFS files that changed. The diff in GitHub only says "Git LFS file not shown" - did you mean to change these, or should they have been completely removed instead?
> 
The old files are not removed, however we shouldn't be using them anymore.

> I think we might be able to run "community mode" without needing a separate settings module or docker-compose.yml file. I only mention because I think you said you would have liked to keep this even simpler (or maybe I'm misremembering?)
> 
> If so, we can add public defaults to internal env vars in docker-compose. The Pypi value will only be used if the env var is unset: ${PIP_INDEX_URL:-"[https://pypi.org/simple"}](https://pypi.org/simple%22%7D)
> 
> And document which env vars should be set / unset for community mode: UMB_BREW_MONITOR_ENABLED turned off COMMUNITY_MODE turned on
> 
> I think that should remove the need for a separate docker-compose and settings file. The Dockerfile might still need some tweaking. I could do some or all of this in a follow-up MR (or push a commit here) if you're interested.

The main reason why we have 2 docker-compose files is so that we can use a separate Dockerfile for each. We need a separate Dockerfile to skip the step where we download and installed the internal CA certificate. Having 2 docker-compose files also allows a separate docker volume for the postgres data, incase you wanted to save the community data but delete the enterprise data.

jasinner:
> > I think we might be able to run "community mode" without needing a separate settings module or docker-compose.yml file. I only mention because I think you said you would have liked to keep this even simpler (or maybe I'm misremembering?)
> > If so, we can add public defaults to internal env vars in docker-compose. The Pypi value will only be used if the env var is unset: ${PIP_INDEX_URL:-"[https://pypi.org/simple"}](https://pypi.org/simple%22%7D)
> > And document which env vars should be set / unset for community mode: UMB_BREW_MONITOR_ENABLED turned off COMMUNITY_MODE turned on
> > I think that should remove the need for a separate docker-compose and settings file. The Dockerfile might still need some tweaking. I could do some or all of this in a follow-up MR (or push a commit here) if you're interested.
> 
> The main reason why we have 2 docker-compose files is so that we can use a separate Dockerfile for each. We need a separate Dockerfile to skip the step where we download and installed the internal CA certificate. Having 2 docker-compose files also allows a separate docker volume for the postgres data, incase you wanted to save the community data but delete the enterprise data.

I resolved this by adding a '| true'  after the download ROOT_CA in the Dockerfile. This allows a community user without that access to skip adding that certificate. However we now have both community and enterprise using the same docker volume.

config/settings/community.py
jasinner: I deleted the community settings file.

PR #268, opened 2022-12-14
Minor: enable gunicorn reload on stage

JimFuller-RedHat:
cause sometimes useful to reload (instead of redeploy) on live stage instance.

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -1,4 +1,4 @@
-from config.utils import running_dev
+from config.utils import running_prod
 
 workers = 4
 worker_class = "gevent"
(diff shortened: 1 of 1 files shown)


juspence (approved):

juspence (commented):
Actually I think this won't work. preload_app will default to False after this change, but the run_service.sh script still won't have the --reload option in it. That option enables gunicorn to reload when some file in the web pod changes, and without it the change here will have no effect.

We can't add the --reload option to that script since that would break prod, where preload_app will still be True. I also don't think we should try to work around this. Let's just keep stage and prod as similar as possible, and if we need to test changes, we can deploy a particular commit ID manually.

JimFuller-RedHat:
ok, I raised this as it seemed this behaviour had changed unexpectedly ... equally happy to close this PR

PR #269, opened 2022-12-14
CORGI-391 - ensure we bias sort el# on latest filter

JimFuller-RedHat:
Further refinement of sorting for 'latest filter' to take into account things like .ael7b vs. .el7 ... el7 should take precedence.

All the gory details described in CORGI-391. 

There maybe future related refinements along these lines though for now it is an explicit choice to not be comprehensive in this specific PR for safety.

Once this is reviewed and merged it probably gives us data parity.

**Note**- also took the opportunity to constrain el# values so we have a clear 'call to action' if/when el10 happens.

**Note** - CI currently fail due to other commits

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -651,7 +651,7 @@ class ProductStream(ProductModel):
                         ),
                         el_match=Func(
                             F("release"),
-                            Value("el([0-9]+)"),
+                            Value(".([a-z]*el[6-9]+)"),
                             function="regexp_match",
                             output=fields.ArrayField(IntegerField()),
                         ),


corgi/core/models.py
```diff
@@ -651,7 +651,7 @@ def get_latest_components(self):
                         ),
                         el_match=Func(
                             F("release"),
-                            Value("el([0-9]+)"),
+                            Value(".([a-z]*el[6,7,8,9]+)"),
```
juspence: If you're trying to match a literal dot . here, I think it needs to be \\. escaped so it isn't treated as part of the regex.

I also think using "[a-z]*" is a little dangerous, if we only want to match a single leading "a"...it feels like the approach here will lead to other side-effects or surprises.

If we want to constrain this to particular .el versions, I think the plus at the end can just be removed. Currently it will match version 66 - 69, 76 - 79, etc. as well as 6 - 9.

So that ends up as:
```suggestion
                            Value("\.(a?el[6-9])"),
```

juspence (approved):
Overall, I'm not opposed to this change if it reaIly gets us to data parity. But if this sorting and filtering logic gets any more complicated, let's make it a Python function with proper comments.

I also have a fix for the tests in https://github.com/RedHatProductSecurity/component-registry/pull/270

corgi/core/models.py
JimFuller-RedHat: thx! safety eg. there was some weird el7#### ... but I guess we can layer on a fix for those if they ever become significant

PR #273, opened 2022-12-16
CORGI-396 update user guide

jasinner:
Fixed some bash escaping issues. Also adjusted purls in the examples after changes in CORGI-224.

I set the description to include CORGI-396 just because an example there now shows how you can filter out container-sources client side.


JimFuller-RedHat (approved):
on scan LGTM ... reminds me ... I have used org-mode tangled mode to autogenerate code results in docs (which are likely to change) ... maybe something to consider for the (far) future.

PR #274, opened 2022-12-16
CORGI-382: Add multidb routing

juspence:
@RedHatProductSecurity/corgi-devs This is ready for discussion and review, but let's wait to merge until after Monday's release.

I see CPU usage in our database spike around 6AM UTC, and then go back to normal around 8AM UTC. (No dashboard link since it's an internal resource). This is the same time that the "fetch_unprocessed_*_relations" tasks run. Those tasks only do two things:

1) Find all the relations which were created after the last successful task run
2) For each relation, schedule a slow_fetch_brew_build task for the associated build ID

This only involves a read from the database and a write to Redis (putting a new task onto the queue). The follow-up slow_fetch_brew_build tasks themselves will obviously write a lot of data, but that's separate from the ProductComponentRelations table reading and Celery task scheduling that happens in the fetch_unprocessed_*_relations tasks.

In order to reduce CPU load on the main / writable DB instance in AWS, we can move all read-only workloads onto the read-only replica DB instance:
1) fetch_unprocessed_relations
2) daily monitoring email (only generates a report, doesn't change data)
3) all GET API endpoints (future PUT API endpoints will use the writable DB instance by default, no special action needed)
4) management commands that e.g. generate reports or schedule Celery tasks only, without changing DB data

This means that if we have many users of our web API (or are doing reconciliation), an overloaded (read-only) database will not cause data loss or interfere with ingestion (make slow_fetch_brew_build tasks fail due to timeouts).

It also means that if we are ingesting large amounts of data, an overloaded (writable) database will not reduce performance of our web APIs or cause an outage that could impact our users.

Finally, by spreading load across both the writable and read-only databases, I hope both "overloaded database" outcomes above will be less likely.

Or if this change doesn't reduce the CPU usage / number of SoftTimeLimitExceeded errors, then I'll need to keep debugging the cause of CORGI-386. But splittling the workloads onto separate DBs will at least help me determine what tasks / queries are overloading our database, with a little better granularity than what we have today.

I've deployed this change and the limit increases from CORGI-375 into stage for testing over the weekend. I'll check the error emails and SignalFX dashboards on Monday morning to see if we had fewer errors or better performance, then add a note here with findings.

### DIFF
--- a/run_celery_cpu.sh
+++ b/run_celery_cpu.sh
@@ -6,4 +6,4 @@
 # ERROR: Pidfile (/tmp/cpu.pid) already exists.
 rm -f /tmp/cpu.pid
 
-exec celery -A config worker -E --loglevel info --pidfile /tmp/cpu.pid -Q cpu -n celery@%h
+exec celery -A config worker -E --loglevel info --pidfile /tmp/cpu.pid -c 1 -Q cpu -n celery@%h
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -269,7 +269,7 @@
         "filename": "tests/test_sca.py",
         "hashed_secret": "b6bed62bf61a2710e03218decfb836a10a7bce0a",
         "is_verified": true,
-        "line_number": 155,
+        "line_number": 159,
         "is_secret": false
       },
       {
@@ -277,10 +277,10 @@
         "filename": "tests/test_sca.py",
         "hashed_secret": "75272b8978226ca0852fd26fb75ed8c2bd4f4680",
         "is_verified": true,
-        "line_number": 181,
+        "line_number": 185,
         "is_secret": false
       }
     ]
   },
-  "generated_at": "2022-12-15T17:39:55Z"
+  "generated_at": "2022-12-15T20:52:11Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -137,14 +137,6 @@
         "is_verified": true,
         "line_number": 16,
         "is_secret": false
-      },
-      {
-        "type": "Secret Keyword",
-        "filename": "docker-compose.yml",
-        "hashed_secret": "e5e9fa1ba31ecd1ae84f75caaa474f3a663f05f4",
-        "is_verified": true,
-        "line_number": 17,
-        "is_secret": false
       }
     ],
     "tests/data/2018747/sources": [
@@ -290,5 +282,5 @@
       }
     ]
   },
-  "generated_at": "2022-12-13T03:51:13Z"
+  "generated_at": "2022-12-15T17:39:55Z"
 }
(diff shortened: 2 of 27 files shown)


juspence:
Overall this looks good after running in stage over the weekend. Some observations:
1) Load is now almost entirely against the read-only replica DB. CPU usage on the writable DB never exceeds 3%.

2) Spikes in load are smaller. Probably there's not as much processing over the weekend, but maybe splitting into read-only and read/write workloads avoids contention / locking issues with the PCR table, reducing overall CPU usage.

3) As mentioned in our weekly meeting, Brew task SoftTimeLimitExceeded / SCA task WorkerLostErrors in monitoring emails are almost / completely gone.

I'm going to leave this open for now since many have focus elsewhere this week. Let's merge ASAP after the new year.

run_celery_cpu.sh
```diff
@@ -6,4 +6,4 @@
 # ERROR: Pidfile (/tmp/cpu.pid) already exists.
 rm -f /tmp/cpu.pid
 
-exec celery -A config worker -E --loglevel info --pidfile /tmp/cpu.pid -Q cpu -n celery@%h
+exec celery -A config worker -E --loglevel info --pidfile /tmp/cpu.pid -c 1 -Q cpu -n celery@%h
```
jasinner, line 9: Why are we adjusting the concurrency of the cpu worker to 1?

jasinner:
I don't have any input on the TODO item you added other to say that I agree with you it requires an investigation. 

Overall this looks good to me, and I think it's great that we are able to make use of both our DB replicas. I just noticed that you dropped the concurrency of the the CPU worker to 1, was that an intentional change? I guess since our load is now quite low there is not much risk of falling behind in processing. I think it would be an issue if we bulk loaded builds though.

jasinner (approved):

run_celery_cpu.sh
juspence: This was the "running out of temp space" bug I mentioned in a few meetings. With this change, I saw no WorkerLostErrors over the weekend.

Today all the workers have to share the memory limit of one pod. If one worker downloads a huge tarball to /tmp, that leaves less temp space (which is an in-memory filesystem) for all the other workers. This can cause WorkerLostErrors when other workers try to save their own data, but fail due to not having enough temp space (AKA memory).

If needed for future ingestion, we can avoid a backlogged CPU queue / too many SCA tasks to process by scaling down the number of slow worker pods from 6 to 3, then scaling up the number of CPU worker pods from 3 to 6. I think "one worker process per pod" makes sense here given how limited our resources are. It will prevent resource contention and seems to reduce the number of errors / failures we see.

mprpic (approved):

juspence:
SonarQube is still complaining even after I removed the TODO comments, because code I added is duplicated. I see no way to mark this failure as ignored and no way to show the exact lines that are a problem. I assume it's the repeated "read_only" strings, but without seeing exact lines / error messages it's hard to know how to fix this.

All other tests are still passing, so going to merge for now. If I can figure this out, I'll open a follow-up PR that fixes it tomorrow.

PR #278, opened 2022-12-21
CORGI-382: Add multidb routing in more places

juspence:
@RedHatProductSecurity/corgi-devs Follow-up MR to shift more load to the read-only DB, which I missed the first time around. Please don't feel rushed to review this - I'm only opening this so I don't forget, but this can certainly wait until after the holidays.

The changes here:
1) Make templates / manifests use the read-only DB, including the cpes ProductModel property, the get_latest_components ProductStream model function, and the get_provides_* Component model functions.

2) Make the builds and coverage ProductModel properties use the read-only DB, which are only used for reporting as far as I can tell.

3) Make the get_sources_* and get_upstreams_* Component model functions use the read-only DB, when they are being used by a serializer / our API.

4) Make the get_sources_* and get_upstreams_* Component model functions use the default writable DB instance, when they're being used by save_component_taxonomy(). This avoids race conditions where we might read stale data from the replica DB and then save it into the primary DB, overwriting more current data.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -413,7 +413,9 @@ class ProductModel(TimeStampedModel):
     def cpes(self) -> tuple[str, ...]:
         """Return CPEs for child streams / variants."""
         # include_self=True so we don't have to override this property for streams
-        descendants = self.pnodes.get_queryset().get_descendants(include_self=True)
+        descendants = (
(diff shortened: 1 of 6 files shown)


jasinner (approved):
LGTM, thanks.

PR #280, opened 2022-12-23
CORGI-302 openstack rdo

jasinner:
This removed Brew build_id as primary key on SoftwareBuild so we can use KOJI and CENTOS build systems in community mode. It also sets up for HACBS in Q1. 

Also this started using build_type in the relations table. Settings as draft till I have more time to test.


JimFuller-RedHat:
@jasinner unsure of status of this (as CI is not passing) ... thinking maybe we hold off until back

JimFuller-RedHat:
ah I just read chat about this - will pick up and fix lint

jasinner:
I fixed flake8 and secrets tox targets now, and also rebased on current main. I'm going to kick off an import in community mode to kick the tyres on this a bit before taking it out of draft.

corgi/core/models.py
```diff
@@ -1227,7 +1256,8 @@ def errata(self) -> list[str]:
         errata_qs = (
```
juspence, line 1667: I can't comment on line 1254 outside this diff window, but I think mypy will pass if you change it to below:
```
if not self.software_build:
```

corgi/api/serializers.py
```diff
@@ -212,7 +217,7 @@ class SoftwareBuildSummarySerializer(serializers.ModelSerializer):
 
     @staticmethod
     def get_link(instance: SoftwareBuild) -> str:
-        return get_model_id_link("builds", instance.build_id)
+        return get_model_id_link("builds", instance.uuid)
```
juspence: We can use ```instance.pk``` here to avoid having to change this in the future.

corgi/api/views.py
```diff
@@ -257,11 +257,10 @@ def get_product_taxonomy(obj: ProductModel) -> tuple[taxonomy_dict_type, ...]:
 class SoftwareBuildViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled until auth is added
     """View for api/v1/builds"""
 
-    queryset = SoftwareBuild.objects.order_by("-build_id").using("read_only")
+    queryset = SoftwareBuild.objects.order_by("-completion_time").using("read_only")
```
juspence: The ordering here needs to be unique (no duplicate values), or else we will have duplicate / inconsistent results across different pages in the API response. OSIDB hit this bug a while back - the solution was to only order by either the primary key, another individual field with a unique=True constraint, or all the fields that appear in a unique_together constraint.

jasinner: I think completion_time is as close as we have to a unique value for this model. Unless we make type and build_id unique together and order_by those 2 fields combined.

corgi/core/models.py
jasinner: Thanks, done.

jasinner:
I tested this some more, but I found any error trying to load builds from CENTOS. After running `loadproductdata` management command, I confirmed there where 869 builds of type CENTOS in ProductComponentRelations table. However when trying to actually load the build data with `loadbrewdata -a` I get an error from the Brew Collector: 

```
2023-01-09T03:43:31.730+00:00 thread=140079497545536, name=celery.app.trace, lineno=265, level=ERROR, app=corgi, environ=dev, msg="Task corgi.tasks.brew.slow_fetch_brew_build[89bbee2b-1479-4c32-9ce4-72f2401c7834] raised unexpected: AssertionError('baseurl argument must not be empty')"
  File "/opt/app-root/src/corgi/tasks/brew.py", line 53, in slow_fetch_brew_build
    component = Brew(b_type).get_component_data(build_id)
  File "/opt/app-root/src/corgi/collectors/brew.py", line 82, in __init__
    self.koji_session = self.get_koji_session(settings.CENTOS_URL)
  File "/opt/app-root/src/corgi/collectors/brew.py", line 90, in get_koji_session
    return koji.ClientSession(url, opts={"serverca": settings.CA_CERT})
  File "/usr/local/lib/python3.9/site-packages/koji/__init__.py", line 2423, in __init__
    assert baseurl, "baseurl argument must not be empty"
AssertionError: baseurl argument must not be empty
```

Therefore I leave this in draft for now. Feel free to pick it up or leave it for when I get back from vacation.

corgi/api/serializers.py
jasinner: Ok, updated.

jasinner:
> I tested this some more, but I found any error trying to load builds from CENTOS. After running `loadproductdata` management command, I confirmed there where 869 builds of type CENTOS in ProductComponentRelations table. However when trying to actually load the build data with `loadbrewdata -a` I get an error from the Brew Collector:
> 
> ```
> 2023-01-09T03:43:31.730+00:00 thread=140079497545536, name=celery.app.trace, lineno=265, level=ERROR, app=corgi, environ=dev, msg="Task corgi.tasks.brew.slow_fetch_brew_build[89bbee2b-1479-4c32-9ce4-72f2401c7834] raised unexpected: AssertionError('baseurl argument must not be empty')"
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 451, in trace_task
>     R = retval = fun(*args, **kwargs)
>   File "/usr/local/lib/python3.9/site-packages/celery/app/trace.py", line 734, in __protected_call__
>     return self.run(*args, **kwargs)
>   File "/usr/local/lib/python3.9/site-packages/celery/app/autoretry.py", line 34, in run
>     return task._orig_run(*args, **kwargs)
>   File "/opt/app-root/src/corgi/tasks/brew.py", line 53, in slow_fetch_brew_build
>     component = Brew(b_type).get_component_data(build_id)
>   File "/opt/app-root/src/corgi/collectors/brew.py", line 82, in __init__
>     self.koji_session = self.get_koji_session(settings.CENTOS_URL)
>   File "/opt/app-root/src/corgi/collectors/brew.py", line 90, in get_koji_session
>     return koji.ClientSession(url, opts={"serverca": settings.CA_CERT})
>   File "/usr/local/lib/python3.9/site-packages/koji/__init__.py", line 2423, in __init__
>     assert baseurl, "baseurl argument must not be empty"
> AssertionError: baseurl argument must not be empty
> ```
> 
> Therefore I leave this in draft for now. Feel free to pick it up or leave it for when I get back from vacation.

Actually this seems to only be a problem with the the `-a` option of loadbrewdata. The `-s openstack-rdo` option it seems to load most builds successfully. Marking as ready.

jasinner:
@juspence I rebased this on the current main branch. Is it OK to merge it now?

corgi/api/views.py
jasinner: I changed the ordered of the view set to be by build_type and build_id and added a unique constraint on those 2 fields.

corgi/api/serializers.py
```diff
@@ -289,12 +289,16 @@ class SoftwareBuildSerializer(IncludeExcludeFieldsSerializer):
...
```
juspence: There were some changes to how we build the URLs and provide these links in one of Jim's recent MRs. Because of that, I'm not sure if this get_model_id_link() function still works correctly. If you test this and it works, I don't have any concerns.

We also said we wanted to use persistent identifiers instead of UUIDs, but if you're fine with this approach I don't see a problem. We won't be reloading data anymore so the UUIDs shouldn't change.

Finally, let's consistently use pk so we don't have to keep changing these (and to match below):
```suggestion
        return get_model_id_link("builds", instance.pk)
```

corgi/api/serializers.py
```diff
@@ -289,12 +289,16 @@ class SoftwareBuildSerializer(IncludeExcludeFieldsSerializer):
...
 
     @staticmethod
     def get_web_url(build: SoftwareBuild) -> str:
         if build.build_type == SoftwareBuild.Type.BREW:
             return f"{settings.BREW_WEB_URL}/brew/buildinfo?buildID={build.build_id}"
+        elif build.build_type == SoftwareBuild.Type.KOJI:
+            return f"{settings.BREW_DOWNLOAD_ROOT_URL}/koji/buildinfo?buildID={build.build_id}"
+        elif build.build_type == SoftwareBuild.Type.CENTOS:
```
juspence, line 308: Should this really use the Brew download URL from our settings? Is there some other Koji or Fedora download URL we should use here instead?

tests/test_sca.py
```diff
@@ -407,7 +407,7 @@ def test_slow_software_composition_analysis(
 
     with open(syft_results, "r") as mock_scan_results:
         mock_syft.return_value = mock_scan_results.read()
-    cpu_software_composition_analysis(build_id)
+    cpu_software_composition_analysis(str(sb.uuid))
```
juspence: Maybe it's better to consistently use sb.pk in all locations:
```suggestion
    cpu_software_composition_analysis(str(sb.pk))
```

corgi/core/migrations/0058_change_software_build_primary_key.py
```diff
@@ -0,0 +1,116 @@
...
+                "ALTER TABLE core_softwarebuild DROP CONSTRAINT core_softwarebuild_pkey CASCADE",
+                "ALTER TABLE core_softwarebuild DROP COLUMN uuid",
+                "ALTER TABLE core_softwarebuild ADD PRIMARY KEY (build_id)",
+                "ALTER TABLE core_softwarebuildtag ADD FOREIGN KEY (tagged_model_id) "
+                "REFERENCES core_softwarebuild (build_id) ON DELETE CASCADE",
+                "ALTER TABLE core_component ADD FOREIGN KEY (software_build_id) "
+                "REFERENCES core_softwarebuild (build_id) ON DELETE CASCADE",
+            ],
+            state_operations=[
+                migrations.RemoveIndex(model_name="softwarebuild", name="core_softwarebuild_pkey"),
```
juspence: I think I see we remove an index here, but don't add it back? And separating the SQL from the Python is a little confusing. I guess it's OK if it makes the migration much faster, but maybe relying on pure-Python / autogenerated migrations is safer.

corgi/core/migrations/0060_softwarebuild_unique_build_id_by_type.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ("core", "0059_add_build_type_to_relations"),
+    ]
+
+    operations = [
+        migrations.AddConstraint(
+            model_name="softwarebuild",
+            constraint=models.UniqueConstraint(
+                fields=("build_id", "build_type"), name="unique_build_id_by_type"
+            ),
```
juspence: Maybe this should be part of migration 0058 - we could add this constraint even before we change which field is the PK.

corgi/core/models.py
```diff
@@ -317,7 +326,9 @@ def save_product_taxonomy(self):
 
 
 class SoftwareBuildTag(Tag):
-    tagged_model = models.ForeignKey(SoftwareBuild, on_delete=models.CASCADE, related_name="tags")
+    tagged_model = models.ForeignKey(
+        SoftwareBuild, on_delete=models.CASCADE, related_name="tags", db_column="tagged_model_uuid"
```
juspence, line 330: If you leave off the db_column= field, shouldn't the SoftareBuildTag's ForeignKey just default to using the related SoftwareBuild's pk?
```suggestion
        SoftwareBuild, on_delete=models.CASCADE, related_name="tags"
```

corgi/core/models.py
```diff
@@ -795,18 +806,37 @@ class Type(models.TextChoices):
     # E.g. product variant for ERRATA, or product stream for COMPOSE
     product_ref = models.CharField(max_length=200, default="")
     build_id = models.CharField(max_length=200, default="")
+    build_type = models.CharField(choices=SoftwareBuild.Type.choices, max_length=20)
```
juspence, line 841: This should work...but eventually we should switch to just using a ForeignKey. I haven't had a chance to get to that ticket yet, sorry :cry:

corgi/core/models.py
```diff
@@ -912,7 +942,11 @@ class Namespace(models.TextChoices):
     # TODO: Or just switch from GenericForeignKey to regular ForeignKey?
     cnodes = GenericRelation(ComponentNode, related_query_name="%(class)s")
     software_build = models.ForeignKey(
-        SoftwareBuild, on_delete=models.CASCADE, null=True, related_name="components"
+        SoftwareBuild,
+        on_delete=models.CASCADE,
+        null=True,
+        related_name="components",
+        db_column="software_build_uuid",
```
juspence, line 996: I think it's better to leave out this change. And I hope that if we change the PK value of another table, the DB will automatically update all the ForeignKeys that link to that table. Did you see issues when testing?

corgi/tasks/brew.py
```diff
@@ -17,18 +17,28 @@
...
+    RETRYABLE_ERRORS,
+    create_relations,
+    get_last_success_for_task,
+)
 from corgi.tasks.errata_tool import slow_load_errata
 from corgi.tasks.sca import cpu_software_composition_analysis
 
 logger = get_task_logger(__name__)
 
+build_type = SoftwareBuild.Type.KOJI if settings.COMMUNITY_MODE_ENABLED else SoftwareBuild.Type.BREW
```
juspence: This is minor, but I think it might be better to use uppercase here so it's clear this is a constant. That also lets you use build_type below instead of b_type, to be consistent with / match build_id:
```suggestion
BUILD_TYPE = SoftwareBuild.Type.KOJI if settings.COMMUNITY_MODE_ENABLED else SoftwareBuild.Type.BREW
```

corgi/tasks/brew.py
```diff
@@ -71,8 +81,8 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
 
     softwarebuild, created = SoftwareBuild.objects.get_or_create(
         build_id=component["build_meta"]["build_info"]["build_id"],
```
juspence: This looks like it's relying on Brew's data format. Should we just use the task's argument here?
```suggestion
        build_id=build_id,
```

corgi/tasks/brew.py
```diff
@@ -118,16 +128,20 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
...
     for b_id in build_ids:
-        logger.info("Requesting fetch of nested build: %s", b_id)
-        slow_fetch_brew_build.delay(b_id, save_product=save_product, force_process=force_process)
+        logger.info("Requesting fetch of nested build: (%s, %s)", b_id, b_type)
+        slow_fetch_brew_build.delay(
+            b_id, b_type, save_product=save_product, force_process=force_process
+        )
 
-    logger.info("Requesting software composition analysis for %s", build_id)
+    logger.info("Requesting software composition analysis for %s", softwarebuild.uuid)
```
juspence: Nit:
```suggestion
    logger.info("Requesting software composition analysis for %s", softwarebuild.pk)
```

tests/test_brew_collector.py
```diff
@@ -782,16 +783,19 @@ def test_fetch_container_build_rpms(mock_fetch_brew_build, mock_load_errata, moc
...
-        tuple(call(build_id, save_product=True, force_process=False) for build_id in RPM_BUILD_IDS),
+        tuple(
+            call(build_id, SoftwareBuild.Type.BREW, save_product=True, force_process=False)
+            for build_id in RPM_BUILD_IDS
+        ),
         any_order=True,
     )
     mock_load_errata.assert_called_with("RHEA-2021:4610", force_process=False)
-    mock_sca.assert_called_with(1781353, force_process=False)
+    mock_sca.assert_called_with(str(softwarebuild.uuid), force_process=False)
```
juspence: Minor nit:
```suggestion
    mock_sca.assert_called_with(str(softwarebuild.pk), force_process=False)
```

corgi/tasks/errata_tool.py
```diff
@@ -17,6 +17,8 @@
 
 logger = get_task_logger(__name__)
 
+build_type = SoftwareBuild.Type.KOJI if settings.COMMUNITY_MODE_ENABLED else SoftwareBuild.Type.BREW
+
```
juspence: Since we need this logic in both corgi.tasks.brew and corgi.tasks.errata_tool, let's move it to the settings module and name it BUILD_TYPE. Making it uppercase makes it clear it's a constant, and also lets us use build_type below instead of b_type (to be consistent with build_id).

corgi/tasks/management/commands/loadbrewdata.py
```diff
@@ -24,6 +25,12 @@ def add_arguments(self, parser: CommandParser) -> None:
             dest="stream",
             help="Fetch latest builds by tag from product stream",
         )
+        parser.add_argument(
+            "-t",
+            "--type",
+            dest="type",
+            help="Fetch latest builds by tag from product stream",
```
juspence: Looks like a typo in the help text here. It's just a copy of the other command above it - not sure if that's right or if it was supposed to say something else.

docs/developer_guide.md
```diff
@@ -43,6 +43,8 @@ export CORGI_PULP_URL=https://rhsm-pulp.example.com/pulp
 # Not used in tests directly, but needed for tests to pass
 export CORGI_BREW_URL=https://koji.fedoraproject.org/kojihub
 export CORGI_BREW_DOWNLOAD_ROOT_URL=https://koji.fedoraproject.org
+export CORGI_CENTOS_URL=https://cbs.centos.org/kojihub
+export CORGI_CENTOS_DOWNLOAD_ROOT_URL=https://cbs.centos.org
```
juspence, line 47: Instead of two different environment variables that are almost the same, what about below?
```suggestion
export CORGI_CENTOS_URL=https://cbs.centos.org
```
Then one of our settings can use the CORGI_CENTOS_URL env var directly, and the other setting can do (in Python):
```
CORGI_CENTOS_OTHER_URL = f"{os.getenv('CORGI_CENTOS_URL')}/kojihub"
```

corgi/tasks/rhel_compose.py
```diff
@@ -2,15 +2,18 @@
...
 from corgi.collectors.rhel_compose import RhelCompose
 from corgi.core.models import ProductComponentRelation, ProductStream, SoftwareBuild
 from corgi.tasks.brew import slow_fetch_modular_build
-from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS, _create_relations
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS, create_relations
 
 logger = get_task_logger(__name__)
 
+build_type = SoftwareBuild.Type.KOJI if settings.COMMUNITY_MODE_ENABLED else SoftwareBuild.Type.BREW
+
```
juspence: As above - let's move it to the settings module and name it BUILD_TYPE.

corgi/tasks/rhel_compose.py
```diff
@@ -53,11 +60,16 @@ def get_builds(
...
+        relations_query.values_list("build_id", "build_type").distinct().iterator()
+    ):
         if not build_id:
             continue
-        if not SoftwareBuild.objects.filter(build_id=int(build_id)).exists():
+        if not SoftwareBuild.objects.filter(build_id=int(build_id), build_type=b_type).exists():
             logger.info("Processing Compose relation build with id: %s", build_id)
+            if build_type == SoftwareBuild.Type.CENTOS:
+                # We would have to update collector modules to avoid using build_id as primary key
+                raise ValueError("We don't support modular builds in composes for CENTOS")
```
juspence, line 74: This shouldn't ever happen, should it? We only set build_type to SoftwareBuild.Type.BREW or SoftwareBuild.Type.KOJI above. Or did you mean b_type here?
```suggestion
            if b_type == SoftwareBuild.Type.CENTOS:
                # We would have to update collector modules to avoid using build_id as primary key
                raise ValueError("We don't support modular builds in composes for CENTOS")
```

corgi/tasks/rhel_compose.py
```diff
@@ -53,11 +60,16 @@ def get_builds(
...
             slow_fetch_modular_build.delay(build_id, force_process=force_process)
```
juspence, line 75: Do we need to pass in the b_type value to slow_fetch_modular_build() here? Either to call Brew() with the right URL (Fedora Koji vs RH Brew), or to set the b_type when saving the newly-fetched build?

corgi/tasks/brew.py
```diff
@@ -494,19 +505,31 @@ def fetch_unprocessed_relations(
...
             logger.info("Processing CDN relation build with id: %s", build_id)
-            slow_fetch_modular_build.delay(build_id, force_process=force_process)
+            if b_type == SoftwareBuild.Type.CENTOS:
+                # This skips use of the Collector models for builds in the CENTOS koji instance
+                # It was done to avoid updating the collector models not to use build_id as
+                # a primary key. It's possible because the only product stream (openstack-rdo)
+                # stored in CENTOS koji doesn't use modules
+                slow_fetch_brew_build.delay(
+                    build_id, SoftwareBuild.Type.CENTOS, force_process=force_process
+                )
```
juspence, line 581: This makes sense, but are we missing some logic at the end? Right now we call slow_fetch_brew_build and slow_fetch_modular_build, when I think we're only supposed to call one:
```suggestion
                slow_fetch_brew_build.delay(
                    build_id, SoftwareBuild.Type.CENTOS, force_process=force_process
                )
                processed_builds += 1
                continue
```

corgi/tasks/brew.py
```diff
@@ -118,16 +128,20 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
...
+            b_id, b_type, save_product=save_product, force_process=force_process
+        )
 
-    logger.info("Requesting software composition analysis for %s", build_id)
+    logger.info("Requesting software composition analysis for %s", softwarebuild.uuid)
     if settings.SCA_ENABLED:
-        cpu_software_composition_analysis.delay(build_id, force_process=force_process)
+        cpu_software_composition_analysis.delay(
+            str(softwarebuild.uuid), force_process=force_process
+        )
```
juspence: Do we also need to pass the b_type value to the SCA task?
```suggestion
        cpu_software_composition_analysis.delay(
            str(softwarebuild.uuid), b_type, force_process=force_process
        )
```

corgi/tasks/brew.py
```diff
@@ -463,22 +477,19 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
...
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def load_brew_tags() -> None:
+    b_type = build_type
+    brew = Brew(build_type)
     for ps in ProductStream.objects.get_queryset():
-        brew = Brew()
+        # This should really be a property in Product Definitions
+        if settings.COMMUNITY_MODE_ENABLED and ps.name == "openstack-rdo":
+            brew = Brew(SoftwareBuild.Type.CENTOS)
+            b_type = SoftwareBuild.Type.CENTOS
```
juspence: This looks like a bug. At first, we set the "brew" variable to use build_type (either KOJI or BREW) outside the for loop.

We loop over many streams, always using the same brew variable with the same build_type (either KOJI or BREW).

Eventually, we get to the "openstack-rdo" stream, and set the "brew" variable to use the CENTOS build_type.

On the next loop iteration, we'll process some other stream that isn't "openstack-rdo", but the "brew" variable will still be set to the old (wrong) build_type, CENTOS (instead of either KOJI or BREW).

corgi/tasks/brew.py
```diff
@@ -463,22 +477,19 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
...
         for brew_tag, inherit in ps.brew_tags.items():
             # Always load all builds in tag when saving relations
-            # TODO: Use _create_relations here and in other places
```
juspence, line 517: Thank you for fixing this :heart:

juspence (commented):
I only reviewed part of this. I will finish the rest tomorrow since it's very late here, but I think there are some bugs which need to be fixed before this can be merged.

Or maybe I'm misunderstanding - either way, this was a massive amount of work. Great job and thank you for working on this. Community-powered Corgi sounds very exciting!

corgi/tasks/yum.py
```diff
@@ -7,17 +7,24 @@
...
-    _create_relations,
+    create_relations,
     get_last_success_for_task,
 )
 from corgi.tasks.pulp import fetch_unprocessed_relations
 
 logger = get_task_logger(__name__)
 
+build_type = SoftwareBuild.Type.KOJI if settings.COMMUNITY_MODE_ENABLED else SoftwareBuild.Type.BREW
+
```
juspence: As above - let's make this a setting we can reuse.

corgi/tasks/management/commands/loadunprocessedrelations.py
```diff
@@ -15,24 +15,38 @@ def handle(self, *args, **options):
...
+                    self.style.SUCCESS(f"Loading build with id: {build_id}, and type: {b_type}")
+                )
+                if b_type == SoftwareBuild.Type.CENTOS:
+                    # This skips use of the Collector models for builds in the CENTOS koji instance
+                    # It was done to avoid updating the collector models not to use build_id as a
+                    # primary key
+                    # It's possible because the only product stream (openstack-rdo) stored in
+                    # CENTOS koji
+                    # doesn't use modules
+                    slow_fetch_brew_build.delay(build_id, SoftwareBuild.Type.CENTOS)
```
juspence: As above - maybe we're supposed to skip the slow_fetch_modular_build() call. I also saw similar code in some other file - maybe we should import and reuse that:
```suggestion
                    slow_fetch_brew_build.delay(build_id, SoftwareBuild.Type.CENTOS)
                    processed_builds += 1
                    continue
```

corgi/tasks/management/commands/loadbrewdata.py
```diff
@@ -78,7 +99,14 @@ def handle(self, *args, **options) -> None:
         )
 
         for build_id in build_ids:
+            build_type = (
+                SoftwareBuild.Type.KOJI
+                if settings.COMMUNITY_MODE_ENABLED
+                else SoftwareBuild.Type.BREW
+            )
```
juspence: Let's make this a reusable setting.

corgi/tasks/management/commands/loadbrewdata.py
```diff
@@ -48,7 +61,15 @@ def handle(self, *args, **options) -> None:
...
-            brew = Brew()
+            if settings.COMMUNITY_MODE_ENABLED:
+                # This should be a property of product-definitions
+                if ps.name == "openstack-rdo":
+                    brew = Brew(SoftwareBuild.Type.CENTOS)
+                else:
+                    brew = Brew(SoftwareBuild.Type.KOJI)
+            else:
+                brew = Brew(SoftwareBuild.Type.BREW)
+
```
juspence: Let's set ```brew = Brew(SoftwareBuild.Type.CENTOS) if ps.name == "openstack-rdo" else Brew(settings.BUILD_TYPE)```.

corgi/api/serializers.py
jasinner: Ok, I will update this.

corgi/api/serializers.py
jasinner: As mentioned in the developer guide, the value for BREW_DOWNLOAD_ROOT_URL should be set to "https://koji.fedoraproject.org" unless you wish to override it, for example when running this code inside the Red Hat firewall.

corgi/core/migrations/0058_change_software_build_primary_key.py
jasinner: Reviewing the documention on RunSQL and StateOperations again I don't think I need the RemoveIndex there. The index will be removed, and recreated by Postgres with the given SQL, but I don't think results in a state change according to Django.

jasinner: The reason for creating the SQL here is so that we can populate the foreign key relationships from softwarebuild_tag and component to the new softwarebuild primary key manually. Maybe I could have done this in Python too, but doing in SQL seems easier to me.

corgi/core/migrations/0060_softwarebuild_unique_build_id_by_type.py
jasinner: Ok, I can do that.

corgi/core/models.py
jasinner: It's the to_field attribute which does that. The db_column attribute defines the name of the foreign key column name. Since I used SQL directly to create the foreign key, I think it's better to explicitly set the column name to useto match the SQL.

corgi/core/models.py
jasinner: All good. I hesitate to include in this PR as this one is already too big. Although I think the approach to changing it could be the same.

corgi/core/models.py
jasinner: @JimFuller-RedHat Can you comment on this? If you change the primary key of a table, does Postgres automatically update all the ForeignKeys that link to that table?

I didn't try just updating the primary key. I did however do some research on it, and found that using the SQL as in migration 0058 was an approach used by others.

corgi/tasks/brew.py
jasinner: Good catch, I'll move these variables inside the for loop to make sure they are reset for each iteration of the product streams.

corgi/tasks/brew.py
jasinner: Good catch again. I'll update it using your suggestion.

corgi/tasks/errata_tool.py
jasinner: So that we don't need to import the SoftwareBuild into the setting module I used tasks.common instead.

corgi/tasks/management/commands/loadbrewdata.py
jasinner: I'll removed the type option as it's not used. Instead pass the centos option to load brew data from centos Koji instance.

corgi/tasks/management/commands/loadbrewdata.py
jasinner: Ok, using the BUILD_TYPE setting in tasks.common

corgi/tasks/management/commands/loadunprocessedrelations.py
jasinner: Good catch, thanks.

corgi/tasks/rhel_compose.py
jasinner: Done.

corgi/tasks/rhel_compose.py
jasinner: Yes, updated to use build_type as the name instead of build_type as the constant in this module is now moved to tasks.common and renamed BUILD_TYPE

corgi/tasks/rhel_compose.py
jasinner: I think it's not required because we don't use composes for Fedora in product-definitions.

corgi/tasks/yum.py
jasinner: Done.

docs/developer_guide.md
jasinner: I reason I did it this way is because the BREW urls don't share a value, and this keeps all 3 instances consistent.

tests/test_brew_collector.py
jasinner: Updated.

tests/test_sca.py
jasinner: OK.

corgi/tasks/rhel_compose.py
juspence: OK, that makes sense. I also see we rely on a default value for build_type when we call slow_fetch_brew_build() from slow_fetch_modular_build(), so we don't need to pass in the value here. I'm fine with whatever you think is best, but maybe we should make the build_type argument required instead of optional.

Then it's impossible to accidentally forget it, so we won't ever call slow_fetch_brew_build with the wrong build type. Today we might default to BREW or KOJI as the build_type, when really we meant to pass in some other value but forgot to add the argument.

That change would require an update to slow_fetch_modular_build, as well as the UMB listener that calls slow_fetch_brew_build. Up to you whether you think that approach or the current code is better, I have no strong opinions.

corgi/core/migrations/0058_change_software_build_primary_key.py
juspence: My concern here is that separating the SQL and Python means Django may not be aware of the changes to our DB. So data could be inconsistent or get corrupted, and it's much harder for me to review and check the SQL above to make sure this won't happen.

Reviewing changes to our model definitions and the Python migrations they generate is much easier. I guess I'm fine with this approach if you've tested locally and are sure that the DB state really does match the Django model state we defined in Python. Let's just take a backup of stage before we merge this, so we can roll back easily if there are issues.

corgi/tasks/brew.py
```diff
@@ -464,21 +478,18 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def load_brew_tags() -> None:
     for ps in ProductStream.objects.get_queryset():
-        brew = Brew()
+        build_type = BUILD_TYPE
+        brew = Brew(BUILD_TYPE)
+        # This should really be a property in Product Definitions
+        if settings.COMMUNITY_MODE_ENABLED and ps.name == "openstack-rdo":
+            brew = Brew(SoftwareBuild.Type.CENTOS)
+            build_type = SoftwareBuild.Type.CENTOS
```
juspence, line 533: My old review comment seems to have disappeared...noting that there's still a bug here so I don't lose track of it.

corgi/tasks/brew.py
```diff
@@ -118,16 +130,18 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
...
+        logger.info("Requesting fetch of nested build: (%s, %s)", b_id, build_type)
+        slow_fetch_brew_build.delay(
+            b_id, build_type, save_product=save_product, force_process=force_process
+        )
 
-    logger.info("Requesting software composition analysis for %s", build_id)
+    logger.info("Requesting software composition analysis for %s", softwarebuild.pk)
     if settings.SCA_ENABLED:
-        cpu_software_composition_analysis.delay(build_id, force_process=force_process)
+        cpu_software_composition_analysis.delay(str(softwarebuild.pk), force_process=force_process)
```
juspence, line 142: Some of my comments disappeared...we probably need to pass build_type to the SCA task, so it can look up the correct build.

corgi/tasks/errata_tool.py
```diff
@@ -112,18 +127,23 @@ def slow_load_errata(erratum_name, force_process: bool = False):
...
     )
     # Convert them to ints for use in queries and tasks
     # We don't store them as int because another build system not use ints
-    return set(int(b_id) for b_id in relation_build_ids if b_id is not None)
+    builds: list[tuple[int, str]] = []
+    for b_id, build_type in relation_builds:
+        if b_id is None:
+            pass
+        builds.append((int(b_id), str(build_type)))
+    return builds
```
juspence: I think you meant "continue" here instead of "pass", and it's much more efficient to use a comprehension instead of building the list one-by-one.

I also think that converting the build_type to a string isn't needed - it should already be a string, right? Or maybe this was just to quiet mypy - that's fine if so.

The "if b_id is not None" was just to avoid a mypy bug that got fixed when we upgraded to the latest version. b_id should never be None since the SoftwareBuild.build_id field doesn't have None=True / isn't nullable:
```suggestion
    builds = tuple((int(b_id), build_type) for (b_id, build_type) in relation_builds)
    return builds
```

corgi/tasks/errata_tool.py
```diff
@@ -112,18 +127,23 @@ def slow_load_errata(erratum_name, force_process: bool = False):
     logger.info("Finished processing %s", erratum_id)
 
 
-def _get_relation_build_ids(erratum_id: int) -> set[int]:
+def _get_relation_builds(erratum_id: int) -> list[tuple[int, str]]:
```
juspence: To make above suggestion not throw a mypy error, and to improve memory usage / speed when returning and iterating over the builds:
```suggestion
def _get_relation_builds(erratum_id: int) -> tuple[tuple[int, str]]:
```

corgi/tasks/errata_tool.py
```diff
@@ -112,18 +127,23 @@ def slow_load_errata(erratum_name, force_process: bool = False):
...
     # Get all the PCR with errata_id
-    relation_build_ids = (
+    relation_builds = (
         ProductComponentRelation.objects.filter(
             type=ProductComponentRelation.Type.ERRATA, external_system_id=erratum_id
         )
-        .values_list("build_id", flat=True)
+        .values_list("build_id", "build_type")
         .distinct()
```
juspence: I think you can get right of the .distinct() here (and save some load on the DB) since you added the unique_together constraint for ("build_id", "build_type").

corgi/tasks/errata_tool.py
```diff
@@ -52,45 +52,60 @@ def slow_load_errata(erratum_name, force_process: bool = False):
...
+    # like HACBS to errata. If that changes, we should review this task for correctness
+    if len(build_types) > 1:
+        raise ValueError("Multiple build types found for errata %s", erratum_name)
 
     # Check is we have software builds for all of them.
     # Most of the time will not have all the builds
     # so we don't load all the software builds at this point
-    no_of_processed_builds = SoftwareBuild.objects.filter(build_id__in=relation_build_ids).count()
+    no_of_processed_builds = SoftwareBuild.objects.filter(
+        build_id__in=build_ids, build_type=BUILD_TYPE
```
juspence: Should we use build_type from the above for-loop? We're guaranteed to only have one value here, but I don't think that value is necessarily the same as BUILD_TYPE.

corgi/tasks/errata_tool.py
```diff
@@ -52,45 +52,60 @@ def slow_load_errata(erratum_name, force_process: bool = False):
...
             for build_obj in build_objects:
                 for build_id, errata_components in build_obj.items():
                     # Add to relations list as we go, so we can fetch them below
-                    relation_build_ids.add(int(build_id))
+                    build_ids.add(int(build_id))
                     ProductComponentRelation.objects.update_or_create(
                         external_system_id=erratum_id,
                         product_ref=variant_id,
                         build_id=build_id,
+                        build_type=BUILD_TYPE,
```
juspence: As above - let's make sure we use the right build_type. Or maybe we only expect to call this code for RH Brew, since the logic here is for Errata Tool? In that case, let's raise an error if the one build_type we have isn't equal to SoftwareBuild.Type.BREW.

corgi/tasks/errata_tool.py
```diff
@@ -99,7 +114,7 @@ def slow_load_errata(erratum_name, force_process: bool = False):
             logger.info("Calling slow_fetch_brew_build for %s", build_id)
             app.send_task(
                 "corgi.tasks.brew.slow_fetch_brew_build",
-                args=(build_id,),
+                args=(build_id, BUILD_TYPE),
```
juspence: As above - let's either use the build_type we found, or raise an error if it isn't the same as BUILD_TYPE.

corgi/tasks/errata_tool.py
```diff
@@ -36,12 +36,12 @@ def save_variant_cdn_repo_mapping() -> None:
 
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def slow_save_errata_product_taxonomy(erratum_id: int):
-    logger.info(f"slow_save_errata_product_taxonomy called for {erratum_id}")
-    relation_build_ids = _get_relation_build_ids(erratum_id)
-    for b in relation_build_ids:
-        logger.info("Saving product taxonomy for build %s", b)
+    logger.info(f"save_errata_product_taxonomy called for {erratum_id}")
```
juspence: Minor nit: looks like a bad merge conflict resolution here.
```suggestion
    logger.info(f"slow_save_errata_product_taxonomy called for {erratum_id}")
```

juspence (approved):
LGTM here as long as we test some more locally, make sure all the known bugs are fixed, and take a backup of stage before merging.

corgi/core/models.py
juspence: I'm fine with this approach if you think it's OK. It just makes me nervous to run migrations manually in SQL instead of letting Django generate them to handle this for us.

corgi/core/models.py
juspence: Yeah, definitely not something for this PR. Maybe after OpenLCS I'll get time to work on this.

corgi/core/models.py
juspence: Ack...up to you, but if it were me I might try below on a test branch:

1. Generating a very small amount of data, like 1 Component, 1 SoftwareBuild, and 1 SoftwareBuildTag, that are all linked together
2. Adding a new SoftwareBuild.build_type field with default=SoftwareBuild.Type.BREW, and a SoftwareBuild.uuid field with unique=True default=uuid.uuid4, to our Python model definitions only (no SQL / migrations yet)
3. Adding the unique_together constraint for ("build_id", "build_type") in Python (and maybe an index too)
4. Letting Django automatically generate the migrations in Python using python3 manage.py makemigrations
5. Running the migrations using python3 manage.py migrate

Above is just prep work that shouldn't cause any issues. Once all the new fields and constraints are in place, you could do below:
6. Change SoftwareBuild.build_id to say pk=False
7. Change SoftwareBuild.uuid to say pk=True
8. Run python3 manage.py makemigrations again, then see what the output is
9. Run python3 manage.py migrate again to actually switch the primary key on the test SoftwareBuild data
10. Confirm the SoftwareBuildTag and Component are still linked to the SoftwareBuild you added in step 1

Above might be a good test for the SQL too (just skip the Python parts). Or maybe you already did this - I'm fine if you've tested migrating data locally already.

corgi/tasks/errata_tool.py
juspence: Ack and good catch re: circular imports.

corgi/core/models.py
jasinner: I tried it out, and got an error during the migration at step 9.

```
django.db.utils.InternalError: cannot drop constraint core_softwarebuild_pkey on table core_softwarebuild because other objects depend on it
DETAIL:  constraint core_component_software_build_id_4f87341b_fk_core_soft on table core_component depends on index core_softwarebuild_pkey
```

We won't be able to remove the foreign key without losing data. That's why with the SQL I first create and populate the uuid and new uuid fk fields before removing the foreign key.

corgi/core/migrations/0058_change_software_build_primary_key.py
jasinner: Definitely a good idea to take a backup of stage before we merge this.  I tested the migration on the production data and it seems to work well.

corgi/tasks/rhel_compose.py
jasinner: I think the existing approach makes sense for the Brew task and collectors.

corgi/tasks/management/commands/loadbrewdata.py
jasinner: Good idea

jasinner:
> > I tested this some more, but I found any error trying to load builds from CENTOS. After running `loadproductdata` management command, I confirmed there where 869 builds of type CENTOS in ProductComponentRelations table. However when trying to actually load the build data with `loadbrewdata -a` I get an error from the Brew Collector:
> 
> Were you able to reproduce this error, or is it fixed now?

Yes, I verified that `loadbrewdata -a` works as expected now.



> > I tested this some more, but I found any error trying to load builds from CENTOS. After running `loadproductdata` management command, I confirmed there where 869 builds of type CENTOS in ProductComponentRelations table. However when trying to actually load the build data with `loadbrewdata -a` I get an error from the Brew Collector:
> 
> Were you able to reproduce this error, or is it fixed now?

I verified 'loadbrewdata -a' works in both enterprise and community mode.

corgi/tasks/brew.py
jasinner: What's the bug? I think I resolved the issue where the brew variable was updated to CENTOS for the iterations after `ps.name == "openshift-rdo"`. I resolved it by setting `brew = Brew(BUILD_TYPE)` inside the for loop instead of outside of it.

corgi/tasks/brew.py
jasinner: We don't need to do that because we use the build uuid not the build_id/type combination.

corgi/tasks/errata_tool.py
jasinner: Ok, done.

corgi/tasks/errata_tool.py
jasinner: yup, I need to set it to this though, I guess because the tuple could be any length?
`_get_relation_builds(erratum_id: int) -> tuple[tuple[int, str],...]`

corgi/tasks/errata_tool.py
jasinner: Good idea.

corgi/tasks/errata_tool.py
jasinner: Actually I need to check what the default value for build_type in the relations table is. I think the value might default to None or an empty string which is not really correct. Maybe we can default to to BREW since we don't have any KOJI data yet.

corgi/tasks/errata_tool.py
jasinner: I think it's better to use the one we found. While today this code should only be used in with BREW type, in future ET might support other types.

corgi/tasks/errata_tool.py
jasinner: I'll switch this and the BUILD_TYPE above to be the one we found.

corgi/tasks/errata_tool.py
jasinner: updated.

jasinner:
> > I tested this some more, but I found any error trying to load builds from CENTOS. After running `loadproductdata` management command, I confirmed there where 869 builds of type CENTOS in ProductComponentRelations table. However when trying to actually load the build data with `loadbrewdata -a` I get an error from the Brew Collector:
> 
> Were you able to reproduce this error, or is it fixed now?

It's fixed now. Verified in both enterprise and community mode.

corgi/tasks/brew.py
jasinner: No need, we lookup the the SoftwareBuild there by unique UUID, not id/type combination.

corgi/tasks/brew.py
jasinner: That makes sense to me. Updated.

corgi/tasks/errata_tool.py
juspence: Yes, oops. I always forget about that special syntax for tuples of any length.

corgi/tasks/brew.py
juspence: The bug is in my eyeballs, which makes it hard to see things :) LGTM

corgi/core/models.py
juspence: Thank you for testing! It's frustrating Django doesn't automatically migrate them for you, and great job with the SQL you wrote to do this. I guess to do it in Django you'd have to:

1. Add the build_type and UUID fields, in the first part above
2. Update the Component and SoftwareBuildTag model ForeignKeys to use to_field=SoftwareBuild.uuid
3. Change the primary key on the SoftwareBuild from build_id to uuid, in the second part above

And above is exactly what your SQL does, so LGTM. Thanks again for testing this out...I feel much more confident about this migration now.

corgi/tasks/errata_tool.py
jasinner: The default value is set correctly in the 0059 migration to BREW.

jasinner:
Moving this back to draft for now to avoid merging until after we do the LCS release (post 1.2.1) next week.

corgi/api/views.py
```diff
@@ -257,11 +257,10 @@ def get_component_taxonomy(
 class SoftwareBuildViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled until auth is added
     """View for api/v1/builds"""
 
-    queryset = SoftwareBuild.objects.order_by("-build_id").using("read_only")
+    queryset = SoftwareBuild.objects.order_by("build_type", "build_id").using("read_only")
```
jasinner, line 260: We should probably keep the build_id in descending order.

corgi/collectors/brew.py
```diff
@@ -77,12 +77,17 @@ class Brew:
...
+        elif source and source not in (SoftwareBuild.Type.BREW, SoftwareBuild.Type.KOJI):
+            raise ValueError(f"Tried to create Brew collector with invalid type: {source}")
+        else:
+            self.koji_session = self.get_koji_session()
 
     @staticmethod
-    def get_koji_session():
-        return koji.ClientSession(settings.BREW_URL, opts={"serverca": settings.CA_CERT})
+    def get_koji_session(url: Optional[str] = settings.BREW_URL):
+        return koji.ClientSession(url, opts={"serverca": settings.CA_CERT})
```
jasinner: We shouldn't use the CA_CERT from settings when creating a ClientSession to Koji (upstream)

corgi/core/models.py
```diff
@@ -395,7 +406,7 @@ def builds(self) -> QuerySet:
         if product_refs:
             return (
                 ProductComponentRelation.objects.filter(product_ref__in=product_refs)
-                .values_list("build_id", flat=True)
+                .values_list("build_id", "build_type")
```
jasinner, line 409: Test using the suggestion from the TODO in this function: 

`return self.components.values_list("software_build", flat=True).distinct()`

corgi/core/models.py
```diff
@@ -260,8 +260,10 @@ class SoftwareBuild(TimeStampedModel):
     class Type(models.TextChoices):
         BREW = "BREW"  # Red Hat Brew build system
         KOJI = "KOJI"  # Fedora's Koji build system, the upstream equivalent of Red Hat's Brew
+        CENTOS = "CENTOS"  # Used by OpenStack RDO
 
-    build_id = models.IntegerField(primary_key=True)
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    build_id = models.IntegerField()
```
jasinner: PNC build IDs are not numeric. This needs to be a string.

corgi/tasks/management/commands/loadunprocessedrelations.py
```diff
@@ -15,23 +14,39 @@ def handle(self, *args, **options):
...
+            .values_list("build_id", "build_type")
             .distinct()
             .using("read_only")
             .iterator()
         ):
             if not build_id:
                 continue
-            if not SoftwareBuild.objects.filter(build_id=int(build_id)).using("read_only").exists():
-                self.stdout.write(self.style.SUCCESS(f"Loading build with id: {build_id}"))
+            if (
```
jasinner: We should be able to extract out the common code from here in tasks/brew.py::fetch_unprocess_relations function to a common function for re-use.

jasinner (commented):
Looking over this again since we're trying to get it merged before we bring in PNC builds such as Quarkus. The main thing that sticks out is that we still have SoftwareBuild.build_id as an integer where PNC uses alphanumeric build ids, eg 'AWYWRYDD3AQAA'.

corgi/api/views.py
jasinner: With build_id becoming string ordering build_id doesn't really make sense anymore.

corgi/core/models.py
jasinner: Updated build_id to be a string in the latest version.

corgi/core/models.py
jasinner: On second though, if we did that it would mean you could not have a SoftwareBuild linked to a Product without also having a component. I prefer it the way it is, without the dependency on a Component.

corgi/tasks/management/commands/loadunprocessedrelations.py
jasinner: Done. In the process I think I found the cause of CORGI-546. Anyway, I'll defer a fix for that to another PR.

corgi/collectors/brew.py
jasinner: I updated the init function here force the selection of a BUILD_TYPE on construction of the Brew class. I also only used CA_CERT for BREW build type (default in enterprise mode).

jasinner:
> Looking over this again since we're trying to get it merged before we bring in PNC builds such as Quarkus. The main thing that sticks out is that we still have SoftwareBuild.build_id as an integer where PNC uses alphanumeric build ids, eg 'AWYWRYDD3AQAA'.

I updated the build_id field of SoftwareBuild model to be type string in the latest version of this PR.

@RedHatProductSecurity/corgi-devs please review the last 3 commits here.

JimFuller-RedHat:
not that I think it should matter but worth stating that changing type of build_id may impact performance and/or implied sorting - alternately we could consider current build_id as an external_system_id and add other 'label id' for other build types - then add a UUID as the true primary key (which maybe we should have done at the beginning) - this would impacts downstream to the REST API, etc.

PR #282, opened 2023-01-03
Upgrade packages to remove setuptools dependency and fix CVEs

juspence:
@RedHatProductSecurity/corgi-devs This will fix the Dependabot alerts for a setuptools CVE. I also upgraded wheel which has a separate CVE / dependabot alert. The PR that dependabot opened won't work, because dependabot doesn't handle the same vulnerable package version in multiple requirements files.

Dependabot also doesn't check for dependency conflicts when upgrading packages. I had to upgrade Celery to the latest 5.2.7 version (we were using 5.2.3, only minor bugfixes since then) to remove the setuptools dependency.


JimFuller-RedHat (approved):
thx for doing this - LGTM

PR #283, opened 2023-01-03
CORGI-420: Fix errors in emails and /taxonomy endpoint

juspence:
@RedHatProductSecurity/corgi-devs This PR adds RedisConnectionError (high-level Redis error) to the list of exceptions we retry if a Celery task fails.

It also removes low-level errors like SSLError and TimeoutError from the list of retryable errors. Those exceptions should eventually raise some higher-level error like InterfaceError or RequestException. Retrying only high-level errors from specific libraries makes our task retry logic more predictable.

The changes in corgi/api/views.py fix recent bugs / errors seen in emails for the taxonomy endpoints. I also added type hints and combined some identical code, to avoid fixing the same things in multiple places and hopefully prevent these kinds of issues in the future.

Finally, I updated yet another model property to use the read-only DB, since .errata is only used by the API serializers.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -211,7 +211,7 @@ def recursive_product_node_to_dict(node):
 def get_component_taxonomy(obj: Component, component_types: list[str]):
     """Look up and return the taxonomy for a particular Component."""
     root_nodes = cache_tree_children(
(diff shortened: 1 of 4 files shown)


JimFuller-RedHat (approved):
LGTM

PR #284, opened 2023-01-04
CORGI-420: Remove broken /taxonomy endpoint from ProductModel subclasses

juspence:
@RedHatProductSecurity/corgi-devs It doesn't work and probably hasn't for a while (ever since we added channels). There are no links to the /taxonomy endpoints from any detail view, so users probably aren't even aware that these pages exist. Or if anyone was using them, we'd be seeing error emails every time they visited that page :)

They also aren't mentioned in our user guide, although the endpoints are documented in the OpenAPI schema.

Fixing the /taxonomy endpoint is possible, but seems pointless since the same information (plus extra info) is already available on the detail view for a particular ProductModel subclass.

The /taxonomy endpoint was also never implemented for Channels. Again, fixing this is possible but duplicates info in the detail view.

The /taxonomy endpoint for Components will stay since it's working.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -18,15 +18,12 @@ from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
 
 from config import utils
 from corgi import __version__
-from corgi.core.constants import NODE_LEVEL_MODEL_MAPPING
 from corgi.core.models import (
     AppStreamLifeCycle,
     Channel,
     Component,
     ComponentNode,
     Product,
-    ProductModel,
-    ProductNode,
     ProductStream,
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat (approved):
Do we have an issue for this work ? - otherwise we might add a 'bucket' of these kinds of items (ex. datascore, etc).

otherwise LGTM

juspence:
CORGI-420 is just for the /taxonomy bugfixes. Feel free to file a general cleanup ticket if you'd like to see other tech debt get removed.

JimFuller-RedHat:
thx I just noticed CORGI-420 which does the needful ;)

PR #285, opened 2023-01-05
CORGI-228 enable persistent url for product streams and components

JimFuller-RedHat:
As per CORGI-228, resources should be addressable with consistent URLs instead of redirecting to their uuid identities.

For product streams this would be with name or ofuri
```
http://localhost:8000/api/v1/product_streams?name=ansible_automation_platform-2.1
http://localhost:8000/api/v1/product_streams?ofuri=o:redhat:ansible_automation_platform:2.1
```

For components this would be with purl
```
http://localhost:8000/api/v1/components?purl=pkg%3Arpm%2Fredhat%2Fpython-inflection%400.5.1-2.el8pc%3Farch%3Dsrc
```

Note- if this looks good then I can expand this to products, product versions, etc

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -69,7 +69,6 @@ class ComponentFilter(FilterSet):
 class ProductDataFilter(FilterSet):
     """Class that filters queries to Product-related list views."""
 
-    name = CharFilter()
     re_name = CharFilter(lookup_expr="regex", field_name="name")
     re_ofuri = CharFilter(lookup_expr="regex", field_name="ofuri")
     tags = TagFilter()
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -5,6 +5,7 @@ from typing import Type, Union
 import django_filters.rest_framework
 from django.db import connections
 from django.db.models import QuerySet
+from django.http import Http404
 from django.utils import timezone
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema
@@ -284,12 +285,23 @@ class ProductViewSet(ProductDataViewSet):
         ofuri = req.query_params.get("ofuri")
         if not ofuri:
             return super().list(request)
-        p = Product.objects.filter(ofuri=ofuri).using("read_only").first()
-        if not p:
-            return Response(status=404)
-        response = Response(status=302)
-        response["Location"] = f"/api/{CORGI_API_VERSION}/products/{p.uuid}"
-        return response
+        return super().retrieve(request)
+
+    def get_object(self):
+        req = self.request
+        p_ofuri = req.query_params.get("ofuri")
+        p_name = req.query_params.get("name")
+        try:
+            if p_name:
+                p = Product.objects.db_manager("read_only").get(name=p_name)
+            elif p_ofuri:
+                p = Product.objects.db_manager("read_only").get(ofuri=p_ofuri)
+            else:
+                pk = req.path.split("/")[-1]  # there must be better ways ...
+                p = Product.objects.db_manager("read_only").get(uuid=pk)
+            return p
+        except Product.DoesNotExist:
+            raise Http404
 
     @action(methods=["get"], detail=True)
     def taxonomy(self, request: Request, uuid: Union[str, None] = None) -> Response:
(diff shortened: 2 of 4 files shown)


corgi/api/filters.py
```diff
@@ -69,7 +69,7 @@ def filter_productmodel(queryset, name, value):
 class ProductDataFilter(FilterSet):
     """Class that filters queries to Product-related list views."""
 
-    name = CharFilter()
+    # name = CharFilter()
```
juspence: Not sure I understand why this isn't needed anymore. But if it needs to go away, we should fully remove it instead of commenting it out.

corgi/api/views.py
```diff
@@ -342,20 +342,24 @@ class ProductStreamViewSetSet(ProductDataViewSet):
...
             return super().list(request)
-        ps = ProductStream.objects.filter(ofuri=ofuri).using("read_only").first()
-        if not ps:
+        try:
+            if ps_name:
+                ps = ProductStream.objects.get(name=ps_name)
+            elif ps_ofuri:
+                ps = ProductStream.objects.get(ofuri=ps_ofuri)
+            serializer = ProductStreamSerializer(ps)
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
juspence: Here we're returning a single stream's data as part of the list view. But I think we should really be redirecting to a detail view / particular ProductStream instance on a separate page, like we did before.

Is there a way to use an ofuri / purl in those detail view URLs instead of a UUID? We may need to add unique=True to the ofuri / purl fields to enable this. Or maybe we can only use primary keys in the URLs for detail views - if that's true, using ofuri / purl as the PK might be worth discussing.

corgi/api/views.py
```diff
@@ -342,20 +342,24 @@ class ProductStreamViewSetSet(ProductDataViewSet):
...
-        if not ofuri:
+        if not ps_ofuri and not ps_name:
             return super().list(request)
-        ps = ProductStream.objects.filter(ofuri=ofuri).using("read_only").first()
-        if not ps:
+        try:
+            if ps_name:
+                ps = ProductStream.objects.get(name=ps_name)
+            elif ps_ofuri:
+                ps = ProductStream.objects.get(ofuri=ps_ofuri)
```
juspence:
```suggestion
                ps = ProductStream.objects.db_manager("read_only").get(name=ps_name)
            elif ps_ofuri:
                ps = ProductStream.objects.db_manager("read_only").get(ofuri=ps_ofuri)
```

corgi/api/views.py
```diff
@@ -468,12 +472,12 @@ def list(self, request: Request, *args: tuple, **kwargs: dict) -> Response:
         # We re-encode the purl here to ensure each segment of the purl is url encoded,
         # as it's stored in the DB.
         purl = f"{PackageURL.from_string(purl)}"
-        component = Component.objects.filter(purl=purl).using("read_only").first()
-        if not component:
+        try:
+            component = Component.objects.get(purl=purl)
```
juspence:
```suggestion
            component = Component.objects.db_manager("read_only").get(purl=purl)
```

corgi/api/views.py
```diff
@@ -468,12 +472,12 @@ def list(self, request: Request, *args: tuple, **kwargs: dict) -> Response:
...
+            serializer = ComponentSerializer(component)
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
juspence: As above...data for a particular component should be shown on the detail view, not as part of the list view (so we should use a redirect instead).

corgi/api/filters.py
JimFuller-RedHat: oh that was unintentional ... or more like I did not understand, will investigate

corgi/api/views.py
JimFuller-RedHat: even better ;) will investigate

corgi/api/views.py
JimFuller-RedHat: ah, thats how you that ... thx

corgi/api/views.py
JimFuller-RedHat: hmm after playing about - this was the best I could come up with (eg. call super.retrieve() and override with get_object) ... otherwise there is no redirect we are performing as we want to preserve the URL ... in this case identity is both a pk (eg. a /{uuid}) and ?name= or ?ofuri= for product stream ... getting all three variants to work with minimal code looks like this for me.

it is likely those who know django/drf better will come up with something cleaner.

juspence: I think we just need to set lookup_field to purl (for Components) or ofuri (for ProductModel subclasses). From DRF docs (https://www.django-rest-framework.org/api-guide/generic-views/#attributes):

lookup_field - The model field that should be used for performing object lookup of individual model instances. Defaults to 'pk'. Note that when using hyperlinked APIs you'll need to ensure that both the API views and the serializer classes set the lookup fields if you need to use a custom value.

Then the code here can just redirect to a detail view page when the user gives some name / ofuri. The URL will still be preserved, since the detail view's URL will show the ofuri the user entered. All that's doing is changing "uuid" in the URL to ofuri / purl, AKA making an unstable URL identifier into a stable one.

JimFuller-RedHat: but we are not proposing a /{ofuri} or /{purl} path in the URL... remember both purl and ofuri are also URIs hence passing via param which is where all the awkwardness starts.

We could entertain /{name} for product streams but thought it best to keep url path just for /{uuid}

as for lookup_field I did try that initially and was not able to get the behaviour that preserves url ... will try again (and thx for the inspiration)

juspence: Maybe I'm not getting it, feel free to ignore if so...I thought the whole point of the ticket was to switch from /{uuid} URLs to /{ofuri} or /{purl} URLs.

I realize those are both URIs, but I thought we'd just percent-encode any special chars in the ofuri / purl to make it work as part of the URL.

JimFuller-RedHat: URIs passed in a URI can only be unambiguously encoded within a url param eg. cannot safely be embedded in the url path itself .... hence the need for ?ofuri={ofuri} and ?purl={purl} ... ok I am going to play around a bit more.

PR #286, opened 2023-01-06
CORGI-417: Fix data quality bugs that impact OpenLCS

juspence:
@RedHatProductSecurity/corgi-devs Opening only for discussion, since I want to make sure this is the right solution first. See individual commits for a detailed description.

Once this is merged, I'll need to run the updatecomponenttaxonomy management command for everything currently in the DB. That will save the component information from the MPTT tree relationships into the ForeignKey fields. Until that completes, the relationships in stage will look like they have disappeared. They haven't, and the MPTT data is still correct. But the API will start reporting data in the ForeignKey fields, which haven't been populated yet.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -274,11 +274,15 @@ class ComponentSerializer(ProductTaxonomySerializer):
 
     @staticmethod
     def get_provides(instance: Component) -> list[dict[str, str]]:
-        return get_component_data_list(instance.get_provides_purls())
+        return get_component_data_list(
+            instance.provides.values_list("purl", flat=True).using("read_only")
+        )
 
     @staticmethod
     def get_sources(instance: Component) -> list[dict[str, str]]:
-        return get_component_data_list(instance.get_sources_purls())
+        return get_component_data_list(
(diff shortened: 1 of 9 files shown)


JimFuller-RedHat (approved):
LGTM, before we do update might be worth taking a backup (or maybe we are ok with prod as backup)

juspence:
> LGTM, before we do update might be worth taking a backup (or maybe we are ok with prod as backup)

No backup needed - this MR involves no migrations. It just switches code to use a field that wasn't being used before, and ensures those field will be automatically populated in the future.

I have to manually populate the fields for the first time once we merge this. If we need to undo this for some reason, I can just revert this MR. I can also empty any data in those fields, but that shouldn't matter since reverting this MR makes the fields unused again (so the left-behind data in them won't matter).

I'm going to kick off the data refresh later this afternoon (have to wait for stage to redeploy). Saving all the data might take a few days, and until it completes the API results will look incorrect. Nothing is missing / changed / removed, we just tweaked how the API serializers fetch the data (to use the new fields, which aren't populated yet).

PR #287, opened 2023-01-13
CORGI-438: constrain arch on component

JimFuller-RedHat:
Added Component.Arch constraint restricting components to the following arch values:

```
src
noarch
i386
ia64
s390
x86_64
s390x
ppc
ppc64
aarch64
ppc64le 
```

**IMPORTANT** -  We need to check that our models are invoking .full_clean() on save (CORGI-439) and probably consider a None value an error.


corgi/core/models.py
```diff
@@ -916,6 +916,20 @@ class Namespace(models.TextChoices):
         UPSTREAM = "UPSTREAM"
         REDHAT = "REDHAT"
 
+    class Arch(models.TextChoices):
+        aarch64 = "aarch64"
+        i386 = "i386"
+        ia64 = "ia64"
+        i686 = "i686"
```
juspence: This is almost alphabetical order (thanks for that), except for these two lines:
```suggestion
        i686 = "i686"
        ia64 = "ia64"
```

corgi/core/migrations/0055_alter_component_arch.py
```diff
@@ -0,0 +1,34 @@
...
+                    ("ia64", "Ia64"),
+                    ("i686", "I686"),
+                    ("noarch", "Noarch"),
+                    ("ppc", "Ppc"),
+                    ("ppc64", "Ppc64"),
+                    ("ppc64le", "Ppc64Le"),
+                    ("s390", "S390"),
+                    ("s390x", "S390X"),
+                    ("src", "Src"),
+                    ("x86_64", "X86 64"),
```
juspence: Not sure why the value on the right has its first letter capitalized.

corgi/core/migrations/0055_alter_component_arch.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0054_auto_20221124_0931"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
juspence, line 13: Here we limit the values for the "arch" field to the new enum. But we don't update any existing components that use a value outside of that list (at least "", maybe others).

juspence (approved):
LGTM once you add logic to handle existing components with "" arch / other invalid values. Other comments are minor.

corgi/core/migrations/0055_alter_component_arch.py
JimFuller-RedHat: these are labels django generates automatically, we could override/define them .. but as we do not define for any other enums we follow the same idiom.

corgi/core/migrations/0055_alter_component_arch.py
JimFuller-RedHat: yes, we need to grok why we have "" values first before we propose a migration to handle them. I will attempt that analysis as part of this issue.

PR #288, opened 2023-01-16
CORGI-443: Fix CVEs by upgrading Python, UBI dependencies

juspence:
@RedHatProductSecurity/corgi-devs This MR upgrades UBI to version 8.7 to fix some CVEs in various RPM dependencies. Version 8.6 which we were using before doesn't have any upgraded version of these packages (and hasn't been updated / rebuilt in three months). I also upgraded Django to 3.2.16 which only includes a CVE fix. This CVE doesn't affect us anyway, but upgrading to fix it will stop some alert spam.

At this point we could also go ahead and switch to UBI version 9. Please let me know if you'd like me to make that upgrade here. Otherwise I think I'll leave that for a future MR, since I'm not sure how much time it will take to identify and debug any issues with our RPM dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8/ubi:8.6
+FROM registry.redhat.io/ubi8/ubi
 
 ARG PIP_INDEX_URL="https://pypi.org/simple"
 ENV PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8/ubi:8.6
+FROM registry.redhat.io/ubi8/ubi
```
JimFuller-RedHat, line 1: do we not want to pin this to 8.7 explicitly ?

JimFuller-RedHat (approved):
LGTM

Dockerfile
juspence: As noted in the commit message, we'll have to explicitly remember to upgrade this to 8.8 in the future. That would be doable if we got some alert about "UBI 8.7 is no longer supported", but we don't.

Today we have to go out and check manually to see if the container is still being built, or if a newer minor version / tag stream is available. Rather than relying on humans to remember and manually do things, let's just automate any upgrades we know we'll eventually need.

Upgrading to a new major version (RHEL9 / UBI9) will still be a manual process anyway. We also explicitly force our Python interpreter (plus all Python package dependencies) to a specific version, so I don't see the UBI container minor version upgrades as being very risky.

JimFuller-RedHat: +1 - happy if you are happy

PR #289, opened 2023-01-17
Minor test enhancements and remove old datascore

JimFuller-RedHat:


pyproject.toml
```diff
@@ -32,7 +32,7 @@ filterwarnings = [
 line-length = 100
 
 [tool.coverage.run]
-omit = ["corgi/core/migrations/*"]
+omit = ["corgi/core/migrations/*","corgi/collectors/migrations/*","corgi/tasks/management/commands/*"]
```
juspence: Not sure if below will work:
```suggestion
omit = ["corgi/**/migrations/*"]
```

I'd like to keep reporting coverage for management commands. The updatecomponenttaxonomy command is currently broken, and when I fix it I'm planning to also add a test, so coverage stats for these are still useful IMO.

tests/test_model.py
```diff
@@ -76,6 +77,9 @@ def test_cpes():
     ps3node = ProductNode.objects.create(parent=pv2node, obj=ps3)
     assert ps3node
 
+    assert ps3node.name == "Brantabernicla"
+    assert ps3node.desc == "The brant or brent goose is a small goose of the genus Branta."
```
juspence: Rather than hardcoding a value, I think it's better to check that the ProductNode object has its name and description copied from the ProductStream object like we expect:
```suggestion
    assert ps3node.name == ps3.name
    assert ps3node.desc == ps3.description
```

tests/test_mixins.py
```diff
@@ -0,0 +1,20 @@
...
+pytestmark = [pytest.mark.unit, pytest.mark.django_db]
+
+
+def test_timestamped_mixin():
+    c1 = ComponentFactory()
+    assert isinstance(c1, TimeStampedModel)
+    c2 = ContainerImageComponentFactory()
+    assert isinstance(c2, TimeStampedModel)
+    ps1 = ProductStreamFactory()
+    assert isinstance(ps1, TimeStampedModel)
```
juspence: This test seems a little odd, since it's asserting a class hierarchy which might change. And the second assert for ContainerImageComponentFactory seems redundant.

We already asserted all Components made from a ComponentFactory() are also TimeStampedModels, so testing a subset of ComponentFactory like ContainerImageComponentFactory doesn't seem necessary.

Is there some specific behavior which should get tested instead? For example, that Components (or the parent TimeStampedModel class) define timestamp fields, and that the field values are set the way we expect / contain the right type of data?

juspence (approved):
LGTM overall.

tests/test_mixins.py
JimFuller-RedHat: this is just a regression test eg. we want to know if product stream or component ever changes in this respect ..

PR #291, opened 2023-01-20
CORGI-432: Allow force-reprocessing child tasks

juspence:
@RedHatProductSecurity/corgi-devs When reingesting builds for CORGI-432, I saw that only the first slow_fetch_brew_build task runs with force_process set to True.

Child tasks like slow_load_errata (which can itself call slow_fetch_brew_build), other slow_fetch_brew_build tasks, and the SCA task either don't support the force_process flag or always set it to False. Here I change the behavior slightly to always pass the flag to these child tasks, whenever it's set in the parent slow_fetch_brew_build task.

When force-processing SCA, we always call both the component and product taxonomy-saving methods just in case the taxonomies changed.

When force-procesing an erratum, we try to recreate the relations in case new ones were created remotely and need to be added to Corgi. We also update each relation's list of components (in the meta_attr) if they changed. Next, we always schedule the child slow_fetch_brew_build tasks (with force_process set to True) even for builds we've already processed. Finally, we save the errata's product taxonomy again just in case it changed.

I also moved save_errata_product_taxonomy to the slow queue. This task will run once per (new or force-reprocessed) erratum, and we don't want many tasks to get spawned in our fast queue and block other, more time-sensitive things like the daily load tasks.

I also fixed some minor issues that I found while examining code or testing - see individual commit messages for details.


mprpic (approved):
LGTM!

JimFuller-RedHat:
LGTM

PR #296, opened 2023-01-26
CORGI-432: Minor bugfixes and ingestion optimizations

juspence:
@RedHatProductSecurity/corgi-devs Some minor bugfixes / tweaks I tested while reingesting data for CORGI-432.

I think that the change to save_product_taxonomy should save 1 query per component, and I deployed this to stage for testing, but I couldn't easily measure the actual DB calls made or if it improved performance. Curious if others have any thoughts about this change - happy to remove if we want to leave alone for now.


JimFuller-RedHat:
LGTM - unsure if the change with save_product_taxonomy is worth doing at this moment in time - we could hold off until @jasinner is back but equally ok if it goes in if you are confident this is well tested.

PR #305, opened 2023-02-03
CORGI-437: API / logic changes needed for OSC integration

juspence:
@RedHatProductSecurity/corgi-devs This is mostly ready for review. The approach shouldn't change and there are detailed comments on the more important / complicated commits, but I still need to:

1. Compare the reconciliation report before and after the latest filter changes, to confirm we have no regressions in coverage (but I'm not expecting any).
2. Confirm there are no performance regressions. Everything was much faster when I tested locally (with large amounts of data), but things were still slow when I deployed to stage for the demo.
3. Add some tests for the new ?include_fields and ?exclude_fields parameters, once I finish above.

I already tested the actual logic, and then demoed to OSC yesterday. They seem happy with the overall approach, but I did emphasize this wasn't finalized yet in case we need to make any changes.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -403,19 +403,6 @@ class ProductModel(TimeStampedModel):
         return ProductComponentRelation.objects.none()
 
     @property
-    def coverage(self) -> float:
-        pnode_children = self.pnodes.get_queryset().get_descendants().using("read_only")
-        # Evaluate queryset once, so it's cached when we reuse it below
-        pnode_children_count = len(pnode_children)
-        if pnode_children_count == 0:
-            return 0
-        build_count = 0
-        for pn in pnode_children:
-            if pn.obj.builds.exists():
-                build_count += 1
-        return round(build_count / pnode_children_count, 2)
-
-    @property
     def cpes(self) -> tuple[str, ...]:
         """Return CPEs for child streams / variants."""
         # include_self=True so we don't have to override this property for streams
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -54,6 +54,8 @@ class ComponentFilter(FilterSet):
     upstreams = CharFilter(lookup_expr="purl__icontains")
     re_upstream = CharFilter(lookup_expr="purl__regex", field_name="upstreams")
 
+    el_match = CharFilter(label="RHEL version for layered products", lookup_expr="icontains")
+
 
 class ProductDataFilter(FilterSet):
     """Class that filters queries to Product-related list views."""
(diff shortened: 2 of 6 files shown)


JimFuller-RedHat:
LGTM ... part of this code is the idiom we applied in osidb so its already gone through a review there as well - we are slower then it could be (but still ok processing time) this is due to yet another slow count which I will eventually fix   ... I would rather get this in sooner rather then later though I wonder if the change in semantics for latest filter match expectations of all prodsec ... lets just push to stage and start the conversation. Very nice work !

corgi/core/models.py
```diff
@@ -876,6 +817,22 @@ def srpms(self) -> models.QuerySet["Component"]:
...
 
+    def latest_components(self) -> models.QuerySet["Component"]:
+        """Return latest versions of components in some queryset."""
+        return self.annotate(
+            latest=models.Subquery(
+                self.filter(
+                    name=models.OuterRef("name"),
+                    el_match=models.OuterRef("el_match"),
+                )
+                .order_by(
```
JimFuller-RedHat, line 828: what I do not understand is why we are no longer using software build completion date in this calculation ...

juspence: Yes, see point number 1 in the MR description above. I'm working on the analysis now and will drop that commit / discuss a new solution with OSC if it causes issues.

The ask here is to be able to see the latest version of all components in some stream, not just the latest version of only root components. Non-root components have no linked SoftwareBuild, so we can't ```.exclude(software_build__isnull=True)``` or ```.order_by(software_build__completion_time)```.

JimFuller-RedHat: for safety ... I propose we keep latest filter we have today asis (at least until Jason gets back and gets a chance to review the rationale of this change) - and provide an OSC filter as you have defined (which makes sense for them) ... it does not matter which you decide to make the default (or how they are named) as long as we can reach either of them in a sensible manner. make sense ?

juspence: I don't think that having two filters is a good idea...users will get confused about which one does what. And if one isn't user-facing / is always turned on for some queries, the second filter won't work.

I'll discuss these latest filter changes when Jason's back, but as of right now I don't see any impact to coverage, and OSC was the team that originally requested we filter by latest. AKA there's no one else depending on the latest filter as far as I know - just coverage reports and OSC reports. Please shout if you know of any other users / usecases.

In the meantime, I added tests and pushed just the commits with the API / serializer changes (include and exclude fields). OSC and OpenLCS would like to get these in ASAP. So I'm going to merge those, then open follow-up MRs as needed once Jason's back. Thanks for reviewing!

JimFuller-RedHat:
I am off the project and its completely your (and Jason's call) to merge eg. it might be that replacing with new filter causes you/Jason to have re perform analysis -  but generally you know that ;)

PR #310, opened 2023-02-09
Fix error when view=summary and component has no linked SoftwareBuild

juspence:
@RedHatProductSecurity/corgi-devs Please review. Originally I thought this was due to recent serializer changes in my MR for CORGI-437, but after looking I don't think that's the case.

I guess maybe we just never used the ?view=summary that often, so we never noticed it wouldn't work when listing out non-root components. I tested manually and this does fix the issue. I really should add a test here but I have a bit too much on my plate at the moment.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 from abc import abstractmethod
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat:
makes sense and LGTM

PR #311, opened 2023-02-15
CORGI-437: Tweak latest filter to be user-controllable and a little faster

juspence:
@jasinner FYI this is the code that I mentioned yesterday. More detailed comments are given on the most important commits.

I did some basic analysis of our coverage report before / after this change and didn't see any major issues. I still need to sit down and do a much more thorough line-by-line review / manually test some more before I feel comfortable merging this.

In the meantime, feel free to look this over and suggest any improvements. Or feel free to ignore if you're busy. It's not urgent at all, and I won't do more work on this unil after my PTO anyway.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -1,9 +1,10 @@
 import logging
 
 from django.core.validators import EMPTY_VALUES
+from django.db.models import QuerySet
 from django_filters.rest_framework import BooleanFilter, CharFilter, Filter, FilterSet
 
-from corgi.core.models import Channel, Component, SoftwareBuild
+from corgi.core.models import Channel, Component, ComponentQuerySet, SoftwareBuild
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


jasinner:
Looks good to me. Does it mean that we can now use the root, and latest filtering on product and product version filters too, not just product stream?

I'm kind of surprised that the coverage report doesn't changing given the removal of ordering by completion time, but if that's the case I'm happy to merge this change.

juspence:
> Looks good to me. Does it mean that we can now use the root, and latest filtering on product and product version filters too, not just product stream?

Yes, these are generic filters that work on any ProductModel subclass, and they now have API documentation too.

> I'm kind of surprised that the coverage report doesn't changing given the removal of ordering by completion time, but if that's the case I'm happy to merge this change.

That surprised me too :smile: I think it's a bug / human error when I was running the report. So I definitely don't want to merge this until I can look more closely. Thanks for reviewing!

juspence:
OSC has decided to get this information using Griffon and will not develop a script of their own, so they don't need "everything to be available in one API call" anymore. So we don't need the root components / latest filters to be separate and user-controllable anymore. Closing.

PR #313, opened 2023-02-15
CORGI-472: Fix related url for SRPMs and upstream containers

juspence:
@jasinner Quick fix for a bug that caused related_urls to be unset for SRPMs and upstream containers. I'm opening this now so you can review and I can merge tomorrow to partially unblock the OSC team.

I'll also open another MR tomorrow that generates the related_url information for remote-source components (with tests).

I haven't figured out how we're going to fix the existing data yet. I recall at one point, OSIDB had a very long-running migration that tried to touch every row in some table, which caused issues for them. Do you think it's safe to try to fix this data as part of a migration?

Is it better to merge this as-is, then manually reingest all the builds which link to SRPM components or upstream container components where the name starts with "github.com"? I can kick that off in my morning, then keep an eye on it throughout the day and revoke (cancel) all the slow_fetch_brew_build() tasks if the DB gets overloaded.

Eventually, I need to do a larger refactoring as part of CORGI-313. That will include cleaning up and / or combining our many different codepaths, plus adding more thorough tests that will catch these kinds of issues in the future.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -56,6 +56,42 @@ from .serializers import (
 
 logger = logging.getLogger(__name__)
 
+INCLUDE_FIELDS_PARAMETER = OpenApiParameter(
+    "include_fields",
+    type={"type": "array", "items": {"type": "string"}},
+    location=OpenApiParameter.QUERY,
+    description=(
+        "Include only specified fields in the response. "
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat (approved):
nice idea with decorator, overall LGTM

PR #315, opened 2023-02-16
CORGI-268: Use exact match on product / version / stream / variant name, not ofuri

juspence:
@jasinner I felt bad Hardik had reported this several months ago and we still hadn't done it yet. It's a very simple change so I just went ahead and switched it. Of course if I misunderstood his requirement, then this gets more complicated :stuck_out_tongue:

Before, Hardik would only see results with a URL like below:
https://corgi-stage.prodsec.redhat.com/api/v1/components?products=o:redhat:rhel

Searching for "rhel", "rh", or "o:redhat:rh" wouldn't return any results.

Now, Hardik will only see results with a URL like below:
https://corgi-stage.prodsec.redhat.com/api/v1/components?products=rhel

Searching for "rh", "o:redhat:rh", or "o:redhat:rhel" won't return any results.

I used an exact lookup instead of an icontains lookup to avoid confusion. With an icontains lookup, searching for "rh" or "rhel" would both show results for "rhel". But users would also see results for "rhel-br" which is a completely different product.

This gets worse if you want to find exactly one stream. Searching for something like "openshift-4" would show results for both base streams and z-streams like "openshift-4.7", "openshift-4.7.z", "openshift-4.8", "openshift-4.8.z", and so on. Users probably meant "openshift-4.0" for the original release.

There are other ways of doing this pattern match if people really want it. I think Hardik's complaint is just that the ofuri format is unfamiliar and a little clunky to use in a search. But he knows the name of the product / stream he's interested in, and expects our filters to just find products with matching names.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.core.validators import EMPTY_VALUES
+from django.db.models import QuerySet
 from django_filters.rest_framework import BooleanFilter, CharFilter, Filter, FilterSet
 
 from corgi.core.models import Channel, Component, SoftwareBuild
@@ -62,10 +63,10 @@ class ComponentFilter(FilterSet):
 
     # User gave a filter like ?ofuri= in URL, assume they wanted a stream
     ofuri = CharFilter(field_name="productstreams", lookup_expr="ofuri")
-    products = CharFilter(lookup_expr="ofuri")
-    product_versions = CharFilter(field_name="productversions", lookup_expr="ofuri")
-    product_streams = CharFilter(field_name="productstreams", lookup_expr="ofuri")
-    product_variants = CharFilter(field_name="productvariants", lookup_expr="ofuri")
+    products = CharFilter(method="filter_ofuri_or_name")
+    product_versions = CharFilter(field_name="productversions", method="filter_ofuri_or_name")
+    product_streams = CharFilter(field_name="productstreams", method="filter_ofuri_or_name")
+    product_variants = CharFilter(field_name="productvariants", method="filter_ofuri_or_name")
     channels = CharFilter(lookup_expr="name")
 
     sources = CharFilter(lookup_expr="purl__icontains")
@@ -84,6 +85,22 @@ class ComponentFilter(FilterSet):
         field_name="license_concluded_raw",
         label="Show only unscanned components (where license concluded is empty)",
     )
+
+    @staticmethod
+    def filter_ofuri_or_name(
+        queryset: QuerySet["Component"], name: str, value: str
+    ) -> QuerySet["Component"]:
+        """Filter some field by a ProductModel subclass's ofuri
+        Or else by a name, depending on the user's input"""
+        if value.startswith("o:redhat:"):
+            # User provided an ofuri
+            # Filter using "products__ofuri", "productversions__ofuri", etc.
+            lookup_expr = f"{name}__ofuri"
+        else:
+            # We don't have a valid ofuri, so assume the user gave a name
+            # Filter using "product__name", "productversions__name", etc.
+            lookup_expr = f"{name}__name"
+        return queryset.filter(**{lookup_expr: value})
 
 
 class ProductDataFilter(FilterSet):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -748,17 +748,20 @@ def test_product_components_versions(client, api_path):
     curl_srpm = SrpmComponentFactory(name="curl", software_build=sb2)
     curl_srpm.productstreams.add(ps1)
 
-    response = client.get(f"{api_path}/components?product_streams=o:redhat:rhel:8")
+    response = client.get(f"{api_path}/components?product_streams={ps2.ofuri}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 2
+    response = client.get(f"{api_path}/components?product_streams={ps2.name}")
     assert response.status_code == 200
     assert response.json()["count"] == 2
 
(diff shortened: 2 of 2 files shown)


corgi/api/filters.py
```diff
@@ -62,10 +62,10 @@ class Meta:
 
     # User gave a filter like ?ofuri= in URL, assume they wanted a stream
     ofuri = CharFilter(field_name="productstreams", lookup_expr="ofuri")
-    products = CharFilter(lookup_expr="ofuri")
-    product_versions = CharFilter(field_name="productversions", lookup_expr="ofuri")
-    product_streams = CharFilter(field_name="productstreams", lookup_expr="ofuri")
-    product_variants = CharFilter(field_name="productvariants", lookup_expr="ofuri")
+    products = CharFilter(lookup_expr="name")
```
juspence: If Hardik really does want a substring / pattern match, where searching for "rhel" shows both "rhel" and "rhel-br" too, switch this code to below:
```suggestion
    products = CharFilter(lookup_expr="name__icontains")
```

jasinner:
I don't think we should change the ofuri filter to use product name for a few reasons:

- ofuri accepts any of product, product version, or product stream. Changing it to name only means we can only search products, not versions or streams.
- ofuri should eventually make more sense once we have services as well

I think what Hardik wanted was to also have an alias for the ofuri search which accepted:

- product stream at least
- perhaps also product version and product
- Have it work the same way as ofuri, (filter latest).

juspence:
> I don't think we should change the ofuri filter to use product name for a few reasons:
> 
> ofuri accepts any of product, product version, or product stream. Changing it to name only means we can only search products, not versions or streams.
> ofuri should eventually make more sense once we have services as well
> I think what Hardik wanted was to also have an alias for the ofuri search which accepted:
> 
> product stream at least
> perhaps also product version and product
> Have it work the same way as ofuri, (filter latest).

Replying via email, so apologies for the formatting. I also am not looking at code right now, but I don't think I changed the ?ofuri= filter at all. Or "oops" if I did.

I only intended to change the ?products=, ?product_versions=, ?product_streams= and ?product_variants= filters / API parameters. These four filters find components which are linked to a matching "thing", based on the ofuri. These should now take a name instead of an ofuri.

For example, ?products= finds components that are a linked to a certain product, while ?product_versions= finds components that are linked to a certain product version. These previously only worked with an ofuri, and now will only work with a name.

I'm talking about the data format here, not the ?ofuri= filter / API parameter which is separate. That filter takes an ofuri for any product, version, stream, or variant. It's called ofuri, so it should always take an ofuri and never a name. It shouldn't change as part of this MR - "oops" if I did change it by mistake.

Even if we did change ?ofuri to use a name (we won't and we shouldn't), I don't think that would stop it from taking any kind of name. We aren't limited to only using product ofuris - we can search by stream ofuris too. We wouldn't be limited to only using product names - we could search by stream names too, right? We just have to guess the right model to filter on based on the user's input. For example, does this look like a product name like "rhel", or a version like "rhel-8", or a stream name like "rhel-8.1.0", and so on.

JimFuller-RedHat:
On Fri, Feb 17, 2023 at 4:43 AM juspence ***@***.***> wrote:

> Replying via email, so apologies for the formatting. I also am not looking
> at code right now, but I don't think I changed the ?ofuri= filter at all.
> Or "oops" if I did.
>
> I only intended to change the ?products=, ?product_versions=,
> ?product_streams= and ?product_variants= filters / API parameters. These
> four filters find components which are linked to a matching "thing", based
> on the ofuri. These should now take a name instead of an ofuri.
>

The discussion with Hardik was that we should additionally allow names (as
well as ofuris) ... we should be able to detect when an ofuri is being used
or not and in the case it is not then attempt to filter by name.

hope that is clear, Jim

juspence:
> The discussion with Hardik was that we should additionally allow names (as well as ofuris) ... we should be able to detect when an ofuri is being used or not and in the case it is not then attempt to filter by name. hope that is clear,

Sorry, I'm a little confused here. I see your first comment in the ticket:
>a few notes:
>    there exists ?products= , ?product_versions= ,  ?product_streams= and ?product_variants= which also only accept exact match of ofuris
>    we could change the default to a regex match which probably satisfies searching using name (need to confirm)

And your last comment in the ticket:
>I think what is being proposed is
>/api/v1/components?product_streams={product stream name}
>/api/v1/components?ofuri={product stream name}

I changed those four filters to accept an exact match on "name" instead of an exact match on "ofuri", based on the above.

I guess you're saying you'd prefer the regex match on the "ofuri" field, but that won't work. We can't do a regex search for a name like "openshift-4" in the ofuri field, because it's not a substring of an ofuri like "o:redhat:openshift:4".

There's not a  simple, clean way to support both ofuri and name exact matching in the same field, depending on the user's input. But I can certainly add new "exact match on name" filters, while leaving the "exact match on ofuri" filters unchanged if we'd prefer that approach.

jasinner:
Justin and I discussed this and we felt it was better to add additional filters for product_*_name

JimFuller-RedHat:
sorry for the confusion ... let me try to clarify:

1) we should be able to filter by ofuri with those fields (this is how it works today)
2) we should be able to filter by product stream name with those fields

on input we should decide - is this an ofuri or not, if it is not then we should translate name to ofuri and use that as input

how this might work 

```/api/v1/components?product_streams=some-product-stream-name-v1```

would redirect to 

```/api/v1/components?product_streams=o:redhat:some-product-stream-name:v1```

similarly:

```/api/v1/components?ofuri=some-product-stream-name-v1```

would redirect to 

```/api/v1/components?ofuri=:redhat:some-product-stream-name:v1```

sorry if this was unclear ... basically we want to let users input product stream names.

PR #316, opened 2023-02-17
CORGI-472: Ensure related_url is always set using purl2url

juspence:
@RedHatProductSecurity/corgi-devs FYI, this isn't ready for a "full" review yet. Pushing anyway so people can tell me if I'm wildly off-base with this approach. Only look at the changes in models.py - everything else was an unfinished first attempt that I need to go clean up.

For RPMs and some containers, we can set the related_url during ingestion based on values from Brew. For all other (remote-source) components, we want to use purl2url to find some upstream repo URL based on the purl.

purl2url supports most of our types, but needed some tweaking to work around bugs in the library, edge cases in our data, and purl / component types that aren't supported yet.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -321,8 +321,10 @@ class Brew:
         remote_sources: dict[str, Tuple] = {}
         # TODO: Should we raise an error if build_info["extra"] is missing?
         if build_info["extra"]:
-            if "index" in build_info["extra"]["image"]:
-                component["meta"]["digests"] = build_info["extra"]["image"]["index"]["digests"]
+            index = build_info["extra"]["image"].get("index", {})
+            if index:
+                component["meta"]["digests"] = index["digests"]
+                component["meta"]["pull"] = index.get("pull", [])
 
             if "parent_build_id" in build_info["extra"]["image"]:
                 parent_image = build_info["extra"]["image"]["parent_build_id"]
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -2,7 +2,7 @@ import logging
 import re
 import uuid as uuid
 from abc import abstractmethod
-from typing import Any, Union
+from typing import Any, Optional, Union
 
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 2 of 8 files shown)


juspence:
This is more ready than it was...I still need to add tests and check some TODO notes, but I'm not going to be able to finish those before PTO. Up to @jasinner whether we merge this now to unblock OpenLCS or hold off.

Note that purl2url natively supports download URLs and repository URLs for Cargo, Gem, and NPM components. There is a bug with Gem purls which I had to work around.

purl2url natively supports only repository URLs for GitHub and PyPI components, so based on the existing code I added support for download URLs as well.

purl2url does not support generic, Golang, or Maven components, so I added support for these based on a combination of official docs, code in other modules, and unmerged PRs.

JimFuller-RedHat:
Once tests are passing, LGTM ... as @juspence mentioned offline we still need to run a migration/.save() to unblock which I believe is the plan.

juspence:
After looking over these changes again, migrations won't be enough to fix this missing data since we never stored the necessary values from Brew into our meta_attr. We need to reingest all builds that have container components, as well as call .save() on any remote-source components.

Given the number of bugs we've fixed recently, and the relatively low number of remote-source components (aside from Golang), I'd recommend just reingesting any non-RPM builds to handle both containers and remote-source components. I've pinged Jason a script that can be adjusted to do this.

Going to hold off on merging this and let @jasinner decide how to proceed from here. I'm also happy to handle this myself once I'm back from PTO (but OSC will probably be upset).

jasinner:
I had a look at the changes here, and I'd prefer if we had some unit tests before merging this.

jasinner:
@mprpic @kgrant-rh Please check the changes to core/core/models.py in the last commit.
I simplied the GOLANG purl to url logic as it seems overly complex to me. I asked some questions in the [upstream commit](https://github.com/package-url/packageurl-python/pull/113/files) Justin used when adding this code.
Also made some changes to MAVEN purl to url logic to exclude 'REDHAT' from purl namespace as we are pointed to Maven central, not Red Hat Maven repository we also need to remove .redhat- from version.

The examples in the tests/test_model.py::test_purl2url method are from real stage data.

kgrant-rh:
Looks reasonable to me. Maybe it's bikeshedding but should we make `central_maven_server` an config variable or a constant or something?

PR #317, opened 2023-02-22
CORGI-494: Upgrade syft to 0.72.0

mprpic:
Includes support for new package types that we can add later on as well as tons of performance improvements.


mprpic:
@jasinner Please review.

jasinner (approved):
LGTM

PR #318, opened 2023-02-23
CORGI-488: Draft collector for middleware CycloneDX SBOMs

kgrant-rh:
A simple parser to extract component-related information from CycloneDX SBOMs.


jasinner:
Nice work!
Can you add some unit tests which reads the SBOMs (or snippets of) you provided here and validates your code yields that data we expect?

jasinner:
Looks good, just need to fix up the lint tox targets.

juspence:
FYI @kgrant-rh Some of the test data here contains internal URLs. These aren't urgent or serious, so we have a separate task to eventually clean these up and scrub our Git history. But if you get a chance, would you mind replacing these with placeholder values or values derived from environment variables? Thanks in advance.

PR #326, opened 2023-03-03
CORGI-510 fix json decode error for manifest with backslash in value

jasinner:
This escapes backslashes from manifest json values which cause json validation errors during a rendering of a manifest.


JimFuller-RedHat:
LGTM

PR #329, opened 2023-03-06
CORGI-499 return stream before variant in get_model_ofuri_type

jasinner:
This returns stream before variant in get_model_ofuri_type serializers function. Which allows latest filter for cryostat-2 to work correctly.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -95,10 +95,12 @@ def get_model_ofuri_type(ofuri: str) -> tuple[Optional[ProductModel], str]:
         return missing_or_invalid
 
     # ProductVersions and ProductStreams both have 4 parts in their ofuri
-    if version := ProductVersion.objects.filter(ofuri=ofuri).using("read_only").first():
-        return version, "ProductVersion"
-    elif stream := ProductStream.objects.filter(ofuri=ofuri).using("read_only").first():
+    # Looking for matching ProductStreams first where version and stream share an ofuri
+    # See CORGI-499
+    if stream := ProductStream.objects.filter(ofuri=ofuri).using("read_only").first():
         return stream, "ProductStream"
+    elif version := ProductVersion.objects.filter(ofuri=ofuri).using("read_only").first():
+        return version, "ProductVersion"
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat (approved):
LGTM

corgi/api/serializers.py
```diff
@@ -95,10 +95,12 @@ def get_model_ofuri_type(ofuri: str) -> tuple[Optional[ProductModel], str]:
...
     # ProductVersions and ProductStreams both have 4 parts in their ofuri
-    if version := ProductVersion.objects.filter(ofuri=ofuri).using("read_only").first():
-        return version, "ProductVersion"
-    elif stream := ProductStream.objects.filter(ofuri=ofuri).using("read_only").first():
+    # Looking for matching ProductStreams first where version and stream share an ofuri
+    # See CORGI-499
+    if stream := ProductStream.objects.filter(ofuri=ofuri).using("read_only").first():
         return stream, "ProductStream"
+    elif version := ProductVersion.objects.filter(ofuri=ofuri).using("read_only").first():
+        return version, "ProductVersion"
```
juspence, line 103: @jasinner The change here doesn't match the description. The old code returns, in order:
1. A matching Product, like RHEL
2. A matching ProductVersion, like RHEL 8
3. A matching ProductStream, like RHEL 8.5
4. A matching ProductVariant, like RHEL 8.5 for SAP

You said we will return "a stream before a variant" but the old code is already doing this. This change returns a stream before a version - is that what you meant?

This code was also kind of a hack that should go away in the draft MR I added. I still need to catch up on other PRs, but I guess this LGTM if above is correct.

jasinner: I said, "a stream before a version", not variant. And that's what it does. In other words, this code now returns a ProductStream before a ProductVersion, which allows the component?ofuri= query to return latest components in the case of 'cryostat-2' which shares an ofuri between it's ProductStream and ProductVersion.

jasinner: Actually I see that I wrote variant in the PR description. That was my mistake. Sorry for that.

PR #331, opened 2023-03-06
Add support for scanning container images in remote registries

mprpic:
This will allow for manifesting of managed services, which are deployed off of specific Quay.io images and built directly from source (mostly GitHub repos).

### DIFF
--- a/corgi/collectors/syft.py
+++ b/corgi/collectors/syft.py
@@ -51,6 +51,31 @@ class Syft:
         return scan_results
 
     @classmethod
+    def scan_repo_image(cls, target_image: str) -> list[dict[str, Any]]:
+        """Scan a remote container image
+
+        `target_image` can point to a specific tag or a digest, e.g.:
+        quay.io/somerepo/example-container:latest
+        quay.io/somerepo/example-container@sha256:03103d6d7e04755319eb303953669182da42246397e32b30afee3f67ebd4d2bb
+
+        Note that this method assumes authentication credentials are set in ~/.docker/config.json
+        (or a location pointed to by DOCKER_CONFIG) for repositories in registries that are private.
+        """
+        scan_result = subprocess.check_output(  # nosec B603
+            [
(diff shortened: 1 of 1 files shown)


mprpic:

jasinner (approved):
LGTM

corgi/collectors/syft.py
```diff
@@ -50,6 +50,31 @@ def scan_files(cls, target_files: list[Path]) -> list[dict[str, Any]]:
...
+            [
+                "/usr/bin/syft",
+                "packages",
+                "-q",
+                "-o=syft-json",
+                "--exclude=**/vendor/**",
+                f"registry:{target_image}",
+            ],
+            text=True,
+        )
```
juspence, line 74: There should be a helper method in corgi/tasks/common.py that handles some of the subprocess-running logic for you, and raises an exception if the tool produced some warning (stderr output). I think we should reuse that if possible, or feel free to ignore if Syft always produces some dumb stderr warning.

juspence (approved):

corgi/collectors/syft.py
mprpic: Syft does return a non-0 exit code when it fails, which would make `run_external` from common fail in a similar way as here (not showing the actual error). It's only when it fails with a 0 but still emits something on stderr that that helper function becomes useful. I can try and improve the CLI call error handling globally in a separate PR if I get some free time later this week.

juspence: Yes, that's the case I'm trying to optimize here - treating warnings (stderr output) as errors that make things fail and get us tracebacks in logs.

Looking again, I think we're also missing check=True here so this might "succeed" even when Syft exits with a non-0 code. EDIT: Except this is using check_output() instead of run(), so this isn't an issue...oops.

Feel free to make those global improvements if you have time, or else I may drop a quick MR on Friday afternoon.

PR #333, opened 2023-03-08
expose errors when downloading sources in sca task

jasinner:
This is related to ops MR component-registry-ops/-/merge_requests/110


juspence (approved):
LGTM!

PR #335, opened 2023-03-10
CORGI-508: Fix related url bugs

juspence:
@RedHatProductSecurity/corgi-devs I think this is ready for review. But of course it's late on Friday and I just found another last-minute bug, so who knows :shrug: I'll look this over again first thing Monday morning, so there's no rush to review if you're busy. This fixes:

1. Links for any component type that were broken due to internal identifiers (release, namespace) not being removed correctly / in all cases
2. Generic links that wouldn't update after they were first added
3. Github links that didn't include the namespace (user name) by mistake
4. Golang links that failed to generate properly due to some incorrect code copied from upstream, or that pointed to the repo's about page instead of a downloadable binary artifact
5. Maven links that were broken due to a trailing slash, missing data in the purl, or pointing to the wrong Maven server

After this is merged, we need to call .save() again for all remote-source component types, then reload all containers. I'll start that reloading in stage next week if we think that's OK. Or I can wait for others to debug the performance issues first, so I don't muddy the waters with a massive load.

Have a good weekend!

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -3,7 +3,7 @@ import logging
 import re
 import uuid as uuid
 from abc import abstractmethod
-from typing import Any, Optional, Union
+from typing import Any, Union
 
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
@@ -897,8 +897,6 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
 
     RPM_PACKAGE_BROWSER = "https://access.redhat.com/downloads/content/package-browser"
     CONTAINER_CATALOG_SEARCH = "https://catalog.redhat.com/software/containers/search"
-    MAVEN_CENTRAL_SERVER = "https://repo1.maven.org/maven2"
-    REDHAT_MAVEN_INDICATOR = ".redhat-"
 
     REMOTE_SOURCE_COMPONENT_TYPES = (
         Type.CARGO,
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat:
thx for this -  I noticed a few data inconsistencies, for example, there is an xstream component with a version={pom.version}

```
https://corgi-stage.prodsec.redhat.com/api/v1/components?purl=pkg%3Amaven%2Fxstream%2Fxstream%40%24%7Bpom.version%7D
```

does this PR address that and is the data 'fix' just reingesting them or will we have to generate a data report ?

jasinner (approved):
This code makes my head spin a bit. But I trust that you've looked at the mountain of purl data we have and made an overall improvement to accuracy of the related_url and download_urls.

jasinner:
I see alot of error messages in stage like this:

```
django.db.utils.IntegrityError: null value in column "related_url" of relation "core_component" violates not-null constraint
```

Might be something we can include in this PR? At least something we should look into before the next mass data update.

juspence:
> thx for this - I noticed a few data inconsistencies, for example, there is an xstream component with a version={pom.version}
> 
> ```
> https://corgi-stage.prodsec.redhat.com/api/v1/components?purl=pkg%3Amaven%2Fxstream%2Fxstream%40%24%7Bpom.version%7D
> ```
> 
> does this PR address that and is the data 'fix' just reingesting them or will we have to generate a data report ?

pom.version is just whatever version we found in the Brew data. It''s not a bug we can address in Corgi if I'm understanding correctly.

juspence:
> I see alot of error messages in stage like this:
> 
> ```
> django.db.utils.IntegrityError: null value in column "related_url" of relation "core_component" violates not-null constraint
> ```
> 
> Might be something we can include in this PR? At least something we should look into before the next mass data update.

I don't see an obvious cause for this - the code I thought was the issue seems to handle a missing related_url correctly. I'll open a separate PR once I find the cause (and won't do the data reload before that).

> This code makes my head spin a bit. But I trust that you've looked at the mountain of purl data we have and made an overall improvement to accuracy of the related_url and download_urls.

Any suggestions on general improvements or changes? I can add better docstrings or try to reduce the duplication a bit if that will help.

A lot of the complexity is because we have many different component types, and we need to get a download URL / about page URL for each. purl2url supports most of these types natively, but a few aren't supported or require special handling to work around bugs in the upstream library.

There are upstream MRs open for some of this code, but the maintainer hasn't merged them even after he approved the fix :neutral_face: Once the upstream MR is merged, we can upgrade our packageurl-python version and simplify this code (just call purl2url after stripping Red Hat-specific identifiers).

kgrant-rh (approved):

juspence:
Going to merge this as-is given approval - the pipeline failure is just a SonarQube warning and I don't see a way to silence this.

PR #336, opened 2023-03-13
CORGI-507 dont load all sources into memory

jasinner:
Trying to load a component with many sources such as 'pkg:rpm/redhat/cracklib-dicts@2.9.6-15.el8?arch=x86_64' results the gunicorn worker crash, which could lead to 504 errors from haproxy in OpenShift since the worker is not able to service the request within the haproxy timeout window of 300s.

This change reduces the memory required to query a component with many sources by not caching the component.sources attribute in memory.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -129,7 +129,7 @@ def get_channel_data_list(manager: Manager["Channel"]) -> list[dict[str, str]]:
     # And channels have no ofuri, so we return a model UUID link instead
     return [
         {"name": name, "link": get_model_id_link("channels", uuid), "uuid": str(uuid)}
-        for (name, uuid) in manager.values_list("name", "uuid").using("read_only")
+        for (name, uuid) in manager.values_list("name", "uuid").using("read_only").iterator()
     ]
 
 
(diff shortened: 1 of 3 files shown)


jasinner:

JimFuller-RedHat:
LGTM ! iterator FTW!

jasinner:
I added another commit which reduces the gunicorn worker timeout to match the HAProxy timeout in OpenShift.

I also adjust the corgi-web pod memory limit to match that in OpenShift. When accessing component in the browser and therefore using debug_toolbar however I found that the web pod requires extra memory compared to without it.

corgi/api/serializers.py
```diff
@@ -413,7 +413,7 @@ def get_provides(instance: Component) -> list[dict[str, str]]:
     @staticmethod
     def get_sources(instance: Component) -> list[dict[str, str]]:
         return get_component_data_list(
-            instance.sources.values_list("purl", flat=True).using("read_only")
+            instance.sources.values_list("purl", flat=True).using("read_only").iterator()
```
rhyw, line 418: Curious if `prefetch_related` on the ManyToMany could help reducing number of queries in this specific case?

juspence: prefetch_related does reduce the number of queries from ```number of components in results list``` to ```only 1```. I researched this a while back, but didn't add it in case the user gives ```?exclude_fields=sources```.

In that case, we don't want to load the sources field at all, but prefetch_related forces us to still make the query and fetch a large amount of data that the user won't even see. Without prefetch_related, we can completely avoid any queries when the field is excluded.

juspence (approved):

corgi/api/serializers.py
juspence, line 418: Can we preemptively make this change for all the related ManyToManyFields? Sources, provides, upstreams, channels, etc.?

jasinner: I added iterator to those fields too, including the product* fields.

rhyw: Thanks

PR #337, opened 2023-03-13
CORGI-313: Fix some errors in related_url saving code

juspence:
@RedHatProductSecurity/corgi-devs This might fix the recent error in monitoring emails about a related_url that's being set to NULL / None.

The field itself is supposed to default to an empty string, but the JSON dictionary we get this data from might have a related_url key that's present with a value of None. In that case, .get() would return None (as expected) and creating the component inside our DB would then fail. The fix is just to force any None values to an empty string.

There are many other (known) bugs lurking in the codepaths here, so there will be follow-up MRs to address all of those and make our component-saving code more consistent. I'm opening this one separately to see if this actually does fix the issue, and to prepare for reloading containers. I need to start this reload ASAP to unblock OpenLCS.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -530,16 +530,16 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
                                 "build_id": None,
                                 "build_type": None,
                                 "build_source_url": None,
-                                "related_url": None,
-                                "upstream_purl": None,
+                                "related_url": ps_component.related_url,
+                                "download_url": ps_component.download_url,
+                                "upstream_purl": ps_component.upstreams.values_list(
+                                    "purl", flat=True
+                                ).first(),
                             }
                             if ps_component.software_build:
                                 component["build_id"] = ps_component.software_build.build_id
                                 component["build_type"] = ps_component.software_build.build_type
                                 component["build_source_url"] = ps_component.software_build.source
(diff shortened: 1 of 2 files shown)


corgi/api/views.py
```diff
@@ -530,16 +530,14 @@ def list(self, request: Request, *args: tuple, **kwargs: dict) -> Response:
                                 "build_id": None,
                                 "build_type": None,
                                 "build_source_url": None,
-                                "related_url": None,
-                                "upstream_purl": None,
+                                "related_url": ps_component.related_url,
+                                # TODO: Do we need download_url here?
```
juspence: I'm not sure about this new code (didn't review yet since it was merged during my PTO).

We need download_url to always be available for every component, but it looks like we pass a custom list of fields to the Response object directly, instead of using the pre-configured list of fields on the serializer.

download_url is also a generated property instead of a field...the serializer handles this but I'm not really sure how to test this custom code. Can someone please provide a sample URL that hits this codepath?

jasinner: Seems like the code path might be /api/v1/components?name=curl&view=latest

juspence: Yep, that's it, and that view is definitely missing the download_url. I'll push a commit here to quickly fix that issue, but I think long-term we should reuse the serializer code for this latest view.

jasinner (approved):
LGTM

PR #338, opened 2023-03-14
CORGI-516 deeply nested descendant queries

jasinner:
This avoid excessive memory usage in Postgres by the following pattern:

`c.cnodes.get_queryset().get_<descendants/ancestores>()` for components that participate in many trees, such as "pkg:rpm/redhat/python3-pip-wheel@21.2.3-6.el9?arch=noarch".

It breaks manifest generation because I changed what's returned by Component.get_provides_nodes used by the provided_relationships_manifest template.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1334,10 +1334,8 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         """Link related components together using foreign keys. Avoids repeated MPTT tree lookups"""
         upstreams = self.get_upstreams_pks(using="default")
         self.upstreams.set(upstreams)
-        provides = self.get_provides_nodes(using="default").values_list("component__pk", flat=True)
-        self.provides.set(provides)
-        sources = self.get_sources_nodes(using="default").values_list("component__pk", flat=True)
-        self.sources.set(sources)
+        self.provides.set(self.get_provides_nodes(using="default"))
+        self.sources.set(self.get_sources_nodes(using="default"))
 
     @property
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat:
picking this up

JimFuller-RedHat:
@jasinner as discussed ... worked on this today ... pushing this to stage to address 100%CPU ... I have left status quo in terms of get_provides_nodes_queryset (used now in manifest generation) - we can carefully work on optimising that in the next PR.

corgi/core/models.py
```diff
@@ -306,10 +306,12 @@ def save_product_taxonomy(self):
...
-            for d in component.cnodes.get_queryset().get_descendants(include_self=True):
-                if not d.obj:
-                    continue
-                components.add(d.obj)
+            for cnode in component.cnodes.get_queryset().prefetch_related("obj"):
+                components.add(cnode.obj)
+                for d in cnode.get_descendants():
+                    if not d.obj:
+                        continue
+                    components.add(d.obj)
```
juspence, line 314: @jasinner I think pretetch_related here will slow down the query (more data transferred), but the obj is only accessed in a different query so we're not using it here (usually).

Try putting it on the inner query, like below:
```suggestion
            for cnode in component.cnodes.get_queryset():
                for d in cnode.get_descendants(include_self=True).prefetch_related("obj"):
                    if not d.obj:
                        continue
                    components.add(d.obj)
```

jasinner: I tried it and it was slower for Brew build 1875516.

PR #339, opened 2023-03-14
CORGI-516 - optimise manifest generation

JimFuller-RedHat:
Follow up to PR #338

This applies the same principle applied there (for setting provides many to many component relationships) to ComponentNode relationship required by manifest generation.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1681,7 +1681,7 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
                 .filter(type__in=type_list)
                 .prefetch_related("obj")
                 .using(using)
-                .values_list("component__pk", flat=True)
+                .values_list("object_id", flat=True)
             )
         return provides_set
 
@@ -1696,13 +1696,16 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         )
         if include_dev:
             type_list = ComponentNode.PROVIDES_NODE_TYPES
-        return (
-            self.cnodes.get_queryset()
-            .get_descendants()
-            .filter(type__in=type_list)
-            .prefetch_related("obj")
-            .using(using)
-        )
+
+        provides_set = set()
+        for cnode in self.cnodes.get_queryset():
+            provides_set.update(
+                cnode.get_descendants()
+                .filter(type__in=type_list)
+                .using(using)
+                .values_list("pk", flat=True)
+            )
+        return ComponentNode.objects.filter(pk__in=provides_set)
 
     def get_sources_nodes(self, include_dev: bool = True, using: str = "read_only") -> set[str]:
         """Return a set of ancestors ids for all PROVIDES ComponentNodes"""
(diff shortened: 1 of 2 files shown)


corgi/core/models.py
```diff
@@ -1696,13 +1696,17 @@ def get_provides_nodes_queryset(
...
-        )
+
+        provides_set = set()
+        for cnode in self.cnodes.get_queryset():
+            provides_set.update(
+                cnode.get_descendants()
+                .filter(type__in=type_list)
+                .prefetch_related("obj")
+                .using(using)
+                .values_list("id", flat=True)
```
juspence: I think this should be "pk".
```suggestion
                .values_list("pk", flat=True)
```

juspence (approved):

corgi/core/models.py
```diff
@@ -1696,13 +1696,17 @@ def get_provides_nodes_queryset(
...
-            .prefetch_related("obj")
-            .using(using)
-        )
+
+        provides_set = set()
+        for cnode in self.cnodes.get_queryset():
+            provides_set.update(
+                cnode.get_descendants()
+                .filter(type__in=type_list)
+                .prefetch_related("obj")
```
juspence: We're not accessing the obj here anymore, it's only used in the queryset below:
```suggestion
                cnode.get_descendants()
                .filter(type__in=type_list)
```

corgi/core/models.py
```diff
@@ -1696,13 +1696,17 @@ def get_provides_nodes_queryset(
...
+                .using(using)
+                .values_list("id", flat=True)
+            )
+        return ComponentNode.objects.filter(id__in=provides_set)
```
juspence: This queryset will access the obj on every node, so will make 1 DB call per node without .prefetch_related().

Up to you whether you add it here, or leave it off to avoid exhausting memory on the Postgres server. .prefetch_related() on such a large queryset may be the cause of these memory issues and not the query refactoring above (too much data being transferred in one query).

Let's also test before merging. Thanks for figuring this out!

corgi/core/models.py
```diff
@@ -1696,13 +1696,16 @@ def get_provides_nodes_queryset(
...
-        )
+
+        provides_set = set()
+        for cnode in self.cnodes.get_queryset():
+            provides_set.update(
+                cnode.get_descendants()
+                .filter(type__in=type_list)
+                .using(using)
+                .values_list("pk", flat=True)
+            )
```
jasinner, line 1707: Let's just change it to:

provides_set = self.get_provides_nodes()

jasinner:
I think we can update the provided_relationships_manifest template to use node.object_id instead of node.object.uuid to avoid another DB lookup on obj.

corgi/core/models.py
JimFuller-RedHat: ok I did not know if you were going to rationalise into a single func ... will do

PR #340, opened 2023-03-14
CORGI-515: Handle edge case where two components can have same purl

juspence:
@RedHatProductSecurity/corgi-devs As noted in ticket, this is usually a PYPI component issue where a dash / underscore in a component name both get converted to a dash in the purl. There's also a similar issue with uppercase / lowercase PyPI and Github component names always being convered to lowercase purls.

I've tested all other Component types and confirmed they don't have these issues. I also checked the data in stage and saw 115 duplicate purls (all PyPI type, we have no Github components). Most of the components I checked looked like they were created by Syft. So I think:

1) We create a new (non-root) component in the Brew task
2) A Syft scan runs for some root component
3) The existing (non-root) component SHOULD be discovered by Syft and get linked to the root component (create a new node)

3 is where things go wrong...the component names that Syft generates don't quite match what we receive from Brew, so we end up creating a new component with a slightly different name and duplicate purl.

I added a migration here that just deletes the duplicate components Syft created. I also changed the code, so that future components with this issue will discover the existing component with the same purl. This will avoid creating a new component with the same purl and only a slightly different name.

There is one Maven component with a duplicate purl, but this looks like a separate issue. I think this Maven component can just be deleted, since it was created due to a Syft scan of an RPM module. We no longer scan modules using Syft, and the name / version is just a (malformed) duplicate of other data anyway. Thoughts or concerns welcome.


kgrant-rh (approved):
Looks sensible to me. Do we have any plans to raise this issue with syft, or are we just accepting it as a limitation of that data source?

juspence:
> Looks sensible to me. Do we have any plans to raise this issue with syft, or are we just accepting it as a limitation of that data source?

My gut says to just accept it. IMO it's not really a bug in Syft but just a side-effect of having two different data sources. If anything it's a bug in the packageurl-python library when constructing PYPI type purls - other component types with different names don't get their purls normalized like this.

But the library behavior seems acceptable as well. "pip install requests-gssapi", "pip install requests_gssapi" and "import requests_gssapi" are all equally valid, so having dash vs. underscore always generate a consistent purl makes sense to me.

JimFuller-RedHat (approved):
LGTM

PR #343, opened 2023-03-16
CORGI-522 fix broken generic purl

jasinner:
Fixes an error when trying to call download_url for (as used in manifest) for affected components.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -5,7 +5,7 @@ import os
 import re
 from datetime import datetime
 from types import SimpleNamespace
-from typing import Any, Dict, Generator, List, Tuple
+from typing import Any, Dict, Generator, List, Optional, Tuple
 from urllib.parse import urlparse
 
 import koji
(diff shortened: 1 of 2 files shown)


jasinner:
There's no data migration here, but there's a [script](https://issues.redhat.com/browse/CORGI-522?focusedId=21922384&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21922384) to clean up the affected components.

jasinner:
I also removed some misleading comments, see https://issues.redhat.com/browse/CORGI-343?focusedId=21922391&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21922391

corgi/collectors/brew.py
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
+                continue
+            bundled_component = cls._check_maven_component(component, version)
+            if bundled_component:
+                bundled_components.append(bundled_component)
                 continue
-            # Strip right parens
-            component = component.replace(")", "")
             # Split into namespace identifier and component name
-            component_split = re.split(r"([(-])", component, maxsplit=1)
+            component_split = re.split(r"([(:\-])", component, maxsplit=1)
```
juspence: I think we're trying to handle those cases with ":" in the component name here, right? But I see if there's any ":" in the component name, _check_maven_component() above returns a truthy value, so we enter the "if bundled_component" block and the for-loop continues. Then this part of the code is never reached, if I'm reading above correctly.

We should also make sure this Maven check doesn't cause issues for purls that have "https://github.com/user/repo" or "git@github.com:user/repo" in their names. There are a few Golang purls with the first form and a few generic purls with the second form. 

Also FYI, I think you don't have to escape a - dash inside a character class, as long as it's the first or last character in the class. The bug here was just me testing this code locally with an incorrect regex.

corgi/collectors/brew.py
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
                 else:
-                    # Account for bundled deps like "bundled(rh-nodejs12-zlib)" where it's not clear
-                    # what is the component type and what is the actual component name.
-                    if separator == "-":
-                        # E.g. unknown / rh-nodejs12-zlib
-                        component = f"{component_type}-{component}"
-                    component_type = Component.Type.GENERIC
-
-            bundled_components.append((component_type, component, version))
+                    component = c_name
```
juspence: Instead of adding an else block here to set the name, is it possible to keep as much of the old code as possible, including the same variable names? i.e. just reusing the "component" variable for all of these transformations instead of adding a new "c_name" variable.

juspence (approved):
I think I follow the changes here and I like that we've split some of the logic into separate methods. LGTM once tests pass, thank you for looking this over!

corgi/collectors/brew.py
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
+    @classmethod
+    def _bundled_or_golang(cls, component: str) -> Optional[str]:
+        # Process bundled deps only; account for typed golang deps of type:
+        # "golang(golang.org/x/crypto/acme)"
+        if component.startswith("bundled("):
+            c = component.removeprefix("bundled(")
+        elif component.startswith("golang("):
+            c = component
+        else:
+            return None
```
juspence: Actually I see the only issue in tests is mypy. If you return "" here instead of None, I think you can avoid importing Optional. The "if not component: continue" logic below should continue to work, and mypy should be happier because the variable will always be of type str.
```suggestion
            return ""
````

File: corgi/collectors/brew.py [hunk 4]
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
+        # Strip right parens
+        return c.replace(")", "")
+
+    @classmethod
+    def _check_maven_component(cls, component: str, version: str) -> Optional[tuple]:
+        if ":" in component:
+            return (Component.Type.MAVEN, component.replace(":", "/"), version)
+        return None
```
juspence, line 138: As above.
```suggestion
        return ()
```

corgi/collectors/brew.py
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
+            c = component.removeprefix("bundled(")
+        elif component.startswith("golang("):
+            c = component
+        else:
+            return None
+        # Strip right parens
+        return c.replace(")", "")
+
+    @classmethod
+    def _check_maven_component(cls, component: str, version: str) -> Optional[tuple]:
```
juspence: You may need an import like "from typing import Type" here to make mypy happy:
```suggestion
    def _check_maven_component(cls, component: str, version: str) -> Optional[tuple[Type, str, str]]:
```
Or maybe you need to change the return signature:
```suggestion
    def _check_maven_component(cls, component: str, version: str) -> tuple:
```

corgi/collectors/brew.py
```diff
@@ -116,54 +116,72 @@ def _parse_remote_source_url(cls, url: str) -> str:
...
+    @classmethod
+    def _get_bundled_component_type(cls, component_type: str, component: str) -> Optional[str]:
+        if component_type.startswith("python"):
+            return Component.Type.PYPI
+        elif component_type.startswith("ruby"):
+            return Component.Type.GEM
+        elif component_type == "golang":
+            # Need to skip arch names, See CORGI-48
+            if component in ("aarch-64", "ppc-64", "s390-64", "x86-64"):
+                return None
```
juspence, line 151: As above.
```suggestion
                return ""
```

corgi/collectors/brew.py
jasinner: You are right, the colon no longer needs to be in the regex, and we don't need to escape '-' when it's at the end of a class. I updated the code to remove the ':', and the '\' (escape character) from the regex.

The purls with 'https:' prefex are coming from containers with legacy osbs names, not bundled provides. I'll file a follow up PR for CORGI-520 which addresses those.

corgi/collectors/brew.py
jasinner: Ok, I change it so that the split operation overrides the existing component, and component_type variables.

corgi/collectors/brew.py
jasinner: I need to adjust the return types here to keep MyPy happy now that I had a few bundled_components.append statements. I'm still making use of Optional but MyPy is no longer complaining.

corgi/collectors/brew.py
jasinner: I need to adjust the return types here to keep MyPy happy now that I had a few bundled_components.append statements. I'm still making use of Optional but MyPy is no longer complaining.

corgi/collectors/brew.py
jasinner: I updated it to:

```
    def _check_maven_component(
        cls, component: str, version: str
    ) -> Optional[tuple[Component.Type, str, str]]
```

corgi/collectors/brew.py
jasinner: I need to adjust the return types here to keep MyPy happy now that I had a few bundled_components.append statements. I'm still making use of Optional but MyPy is no longer complaining.

PR #345, opened 2023-03-16
CORGI-273: Fix manifest performance

juspence:
@RedHatProductSecurity/corgi-devs Please review some changes, which speed up RHEL manifest generation for me locally (12s instead of 25s or so).

I still need to do more testing locally with OpenShift manifests (only looked at RHEL so far), but opening for review now given timezone differences.


jasinner:
Looks good to me. I tried it and it seems to work well.

jasinner (approved):

PR #346, opened 2023-03-17
CORGI-520 invalid purls with colon

jasinner:
This fixes 2 ways that we can get components with colon in the purl. 

- `git@github.com:` in remote_sources.repo
- `http:` protocol in legacy OSBS builds

I've got a script attached to CORGI-520 which can clean up the data with components of type GENERIC or GOLANG with a colon in their name. It needs to be run to fix up the existing data with the new code in this PR, and https://github.com/RedHatProductSecurity/component-registry/pull/343

I also took the opportunity to improve the PURL we build for remote_sources.repo. I'm not setting the type to GOLANG where it's more appropriate then GENERIC, and switched the namespace from REDHAT to UPSTREAM to more accurately reflect the fact that remote_sources repo is outside the Redhat firewall.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -108,13 +108,31 @@ class Brew:
         raise BrewBuildSourceNotFound(no_source_msg)
 
     @classmethod
-    def _parse_remote_source_url(cls, url: str) -> str:
+    def _parse_remote_source_url(cls, url: str) -> Tuple[str, Component.Type]:
         """Used to parse remote_source repo from OSBS into purl name for github namespace
         ref https://github.com/containerbuildsystem/osbs-client/blob/
         f719759af18ef9f3bb45ee4411f80a9580723e31/osbs/schemas/container.json#L310"""
         parsed_url = urlparse(url)
         path = parsed_url.path.removesuffix(".git")
-        return f"{parsed_url.netloc}{path}"
+        # handle url like git@github.com:rh-gitops-midstream/argo-cd
+        if path.startswith("git@"):
+            path = path.removeprefix("git@")
+            path = path.replace(":", "/")
+        # look for github.com and set ComponentType with modified path
+        type = Component.Type.GENERIC
+        if parsed_url.netloc == "github.com":
+            type = Component.Type.GITHUB
+            # urlparse keeps the leading / on the path component when netloc was found
+            # the purl spec dictates that we remove it for Github purls
+            path = path.removeprefix("/")
+        # no netloc
+        elif path.startswith("github.com/"):
+            path = path.removeprefix("github.com/")
+            type = Component.Type.GITHUB
+        # non github url with netloc
+        else:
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat (approved):
purl generation (and related url) have quite a bit of  parse and presentation logic ... at some point we might consider centralising all this behaviour into its own module/function but I can see how specific it is to each ingestion type - makes sense to me and LGTM.

corgi/collectors/brew.py
```diff
@@ -373,7 +391,9 @@ def get_container_build_data(self, build_id: int, build_info: dict) -> dict:
             # AND handle case when "modules" key is present but value is None
             if go and go.get("modules", []):
                 go_modules = tuple(
-                    module["module"] for module in go["modules"] if module.get("module")
+                    module["module"].removeprefix("https://")
+                    for module in go["modules"]
+                    if module.get("module")
```
juspence, line 396: I think this will break some code in download_url / relative_url building. We look for "https://github.com", and give a Github download URL if that's found.

So if we make this change, we should also update all our existing Golang component data to be consistent, then update the code that builds the relative URLs.

But I think it's better to just leave the https:// prefix (or whatever we receive from Brew). If we strip "https://", we should also strip "http://" and "git://" and "git+https://" and any other values to be consistent, which seems like a lot of effort.

The bug in CORGI-520 only affected a few Golang purls, but lots of them have "https://github.com" so I don't think the "https://" is what causes the issue.

juspence (changes requested):

corgi/collectors/brew.py
jasinner: There are only 8 components in stage which have a colon in their name with a type of GOLANG. They are ones ingested by this code path. I think you are mistaken that we look for "http://github.com" when building download_url, I see the [code here](https://github.com/RedHatProductSecurity/component-registry/blob/main/corgi/core/models.py#L1175) which looks for "github.com", which I am preserving in the name with this change.

```
In [4]: for c in m.Component.objects.filter(type="GOLANG",name__contains=":"):
   ...:     print(c)
   ...: 
https://github.com/red-hat-storage/ramen
https://github.com/cincinnati/cincinnati-operator
https://github.com/red-hat-storage/ocs-operator
https://github.com/red-hat-storage/ocs-osd-deployer.git
https://github.com/red-hat-storage/odf-operator
https://github.com/noobaa/noobaa-operator.git
https://github.com/red-hat-storage/odf-multicluster-orchestrator
https://github.com/openshift/ocs-operator
```

juspence: Yep, I probably got confused (again) - sorry for the trouble and feel free to ignore this.

EDIT: There is some code that looks for a generic-type component where the purl starts with (a url-encoded version of) ```git@github.com:```, but I think it's OK to leave this for now. We can probably remove it once we clean up the existing data (remove : from the purls and set type=GITHUB for these components).

juspence (approved):

PR #347, opened 2023-03-17
CORGI-523 pre-generate manifest and host them using whitenose in static dir

jasinner:
This generates the product streams manifests daily and services them using whitenoise in the static directory. We'll have to adjust stage/prod to set OUTPUT_FILES_DIR to a persistent volume shared by both the celery workers and the web pods.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -34,8 +34,10 @@ logger = logging.getLogger(__name__)
 # TODO remove running_community check once domain is established
 if not utils.running_dev():
     CORGI_API_URL = f"https://{settings.CORGI_DOMAIN}/api/{CORGI_API_VERSION}"
+    CORGI_STATIC_URL = f"https://{settings.CORGI_DOMAIN}{settings.STATIC_URL}"
 else:
     CORGI_API_URL = f"http://localhost:8008/api/{CORGI_API_VERSION}"
+    CORGI_STATIC_URL = f"http://localhost:8008{settings.STATIC_URL}"
 
 
 def get_component_data_list(component_list: Iterable[str]) -> list[dict[str, str]]:
@@ -534,7 +536,9 @@ class ProductModelSerializer(ProductTaxonomySerializer):
 
     @staticmethod
     def get_manifest(instance: ProductStream) -> str:
-        return get_model_id_link("product_streams", instance.uuid, manifest=True)
+        if not instance.components.exists():
+            return ""
+        return f"{CORGI_STATIC_URL}{instance.name}-{instance.pk}.json"
 
     @staticmethod
     def get_relations(instance: ProductModel) -> list[dict[str, str]]:
--- a/corgi/tasks/manifest.py
+++ b/corgi/tasks/manifest.py
@@ -1,0 +1,49 @@
+from celery.utils.log import get_task_logger
+from celery_singleton import Singleton
+from django.conf import settings
+from django.core.management import call_command
+from django.db.models import Count
+
+from config.celery import app
+from corgi.core.models import ProductStream
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+
+logger = get_task_logger(__name__)
+
+
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+)
+def update_manifests():
+    for ps in ProductStream.objects.annotate(num_components=Count("components")).filter(
+        num_components__gt=0
+    ):
+        cpu_update_ps_manifest.delay(ps.name)
+
+
+@app.task(
+    base=Singleton,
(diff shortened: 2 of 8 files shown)


corgi/api/serializers.py
```diff
@@ -534,7 +536,7 @@ def get_build_count(instance: ProductModel) -> int:
 
     @staticmethod
     def get_manifest(instance: ProductStream) -> str:
-        return get_model_id_link("product_streams", instance.uuid, manifest=True)
+        return f"{CORGI_STATIC_URL}{instance.name}.json"
```
juspence: Manifest URLs are supposed to use a UUID per the SPDX spec.

run_service.sh
```diff
@@ -5,10 +5,13 @@
...
-    --ignore '*.xml' \
-    --ignore '*.bz2' \
-    --ignore 'tmp*' \
+    --ignore '.gitignore' \
     --noinput
 
 # start gunicorn
-exec gunicorn config.wsgi --config gunicorn_config.py
+if [[ $1 == reload ]]; then
+    exec gunicorn config.wsgi --config gunicorn_config.py --reload
```
juspence, line 14: FYI our gunicorn_config.py doesn't support --reload in stage, only in dev (to save memory in the worker process). So this change is fine if you just want to run collectstatic in dev, but it won't allow --reload to work in stage.

corgi/tasks/manifest.py
```diff
@@ -0,0 +1,47 @@
...
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+
+logger = get_task_logger(__name__)
+
+
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+    soft_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
```
juspence: Do we really expect all three of these tasks to take 40 minutes? Is there a more reasonable timeout we should use?

corgi/tasks/manifest.py
```diff
@@ -0,0 +1,47 @@
...
+        update_ps_manifest.delay(ps.name)
+
+
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+    software_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
+)
+def update_ps_manifest(product_stream: str):
```
juspence: Maybe this should be slow_update_ps_manifest() since this task will be spawned many times. If we leave it as-is it might clog up the fast queue and block our daily monitoring email.

corgi/api/serializers.py
jasinner: I'll include both the name and the UUID so that the name is still useful to humans, not only machines.

run_service.sh
jasinner: I just want to run collectstatic in dev. I don't think we need --reload in stage.

corgi/tasks/manifest.py
jasinner: I removed the soft_time_limit to revert to the default of 900 s

corgi/tasks/manifest.py
jasinner: Good idea, I moved this and collectstatic to the slow queue.

corgi/tasks/manifest.py
```diff
@@ -0,0 +1,49 @@
...
+    retry_kwargs=RETRY_KWARGS,
+)
+def cpu_update_ps_manifest(product_stream: str):
+    logger.info("Updating manifest for %s", product_stream)
+    ps = ProductStream.objects.get(name=product_stream)
+    # TODO figure out a way skip updating files where the content doesnt need updating
+    # for example we could check for the timestamp of the latest build, and only update if we
+    # have a newer build than the last run.
+    # That will allow clients to continue to be served the same content from the browser cache
+    # over the span of multiple days, until the product stream receives a new build
```
juspence, line 38: The existing staticfiles framework is supposed to handle this, I think. We use the ManifestStaticFilesBackend which appends a unique string to each filename based on (I think) its hash. Then if the static file hasn't changed, running "collectstatic" again generates the same file with the same hash in the name, so clients can continue to serve it from cache. If the file has changed, you get a new filename / hash the client hasn't seen before, which handles cache invalidation.

To take advantage of this, in a template you're supposed to use ```{% staticfiles '/path/to/original_filename.json'}```, and the staticfiles backend will generate a URL using the hash (which is unpredictable so can't be hardcoded). There's also a Django / Python function to do something similar outside of a template, which we would need to use in the "get_manifest" serializer method.

We'd also need to change the "collectstatic" command to not ignore the .json files, but once that's done the caching and cache invalidation should be automatically handled for us as part of the backend, if I'm understanding correctly.

PR #351, opened 2023-03-17
Fix possible license erasure when reprocessing a component

juspence:
@RedHatProductSecurity/corgi-devs Sorry for the noise / user error here, I'm juggling a bit too much at the moment. This fixes a bug I realized might exist during the OpenLCS demo today.

If the license_declared field for a component is empty, OpenLCS will submit the value to us. If the license_declared field already has a value, OpenLCS cannot overwrite it.

If we are creating a component for the first time, we can set the license_declared field from the meta["license"] data, either to a value or to an empty string default.

If we are reprocessing a component and meta["license"] is not empty, we SHOULD overwrite the license_declared that OpenLCS provided. In other words, the value from Brew has higher priority than whatever OpenLCS found during their scanning. This is correct.

But If we are reprocessing a component and meta["license"] is an empty / falsey value, we couldn't find any license value from Brew. In this case, OpenLCS may have already provided a license value to us. We should not overwrite it with an empty string, but currently we do. This is a bug.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -224,6 +224,7 @@ def save_component(
         # Handle case when key is present but value is None
         related_url = ""
 
+    license_declared_raw = meta.pop("license", "")
     # We can't build a purl before the component is saved,
     # so we can't handle an IntegrityError (duplicate purl) here like we do in the SCA task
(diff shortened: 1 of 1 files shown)


jasinner (approved):
LGTM, thanks for catching this.

PR #353, opened 2023-03-20
CORGI-273: Avoid OoM in manifest generation / ordering changes in manifest content

juspence:
@RedHatProductSecurity/corgi-devs Follow-up to previous PR that may help to address OoM issues (CORGI-530). It also fixes a bug that caused the relationships in the manifest to constantly change their ordering.

I tried to add some more optimizations based on the previous MR:
1. Avoid converting \u000A to \n which requires parsing all the JSON again and is slow
2. Add .prefetch_related() to reduce duplicate / similar query count
3. Inline the base_manifest.json template to avoid more templating overhead

These didn't seem to improve performance much, but I can push the commits here if people are interested and want to try them out.


corgi/core/models.py
```diff
@@ -303,14 +303,12 @@ def save_product_taxonomy(self):
...
-        for component in self.components.get_queryset():
+        for component in self.components.get_queryset().iterator():
             # This is needed for container image builds which pull in components not
             # built at Red Hat, and therefore not assigned a build_id
-            for cnode in component.cnodes.get_queryset().prefetch_related("obj"):
+            for cnode in component.cnodes.get_queryset().prefetch_related("obj").iterator():
                 components.add(cnode.obj)
-                for d in cnode.get_descendants():
-                    if not d.obj:
-                        continue
```
juspence, line 313: I'm not sure why we added this originally, but:
1. When I checked last week, there were no ComponentNode objects without a linked Component
2. Whenever we create a ComponentNode, we always create it pre-linked to a Component as part of the creation
3. Whenever we delete a Component, the linked ComponentNodes will be deleted automatically due to the reverse relationship / related_query_name on the cnodes field.

So I think we can save some CPU cycles by skipping this check, and maybe even using a set comprehension here, like ```set.update(d.obj for d in cnode.get_descendants())```, similar to below.

juspence:
@jasinner This should be ready for you to try locally. You could also try changing ```component.name``` to  ```component.name|escapejs``` in the component and product manifests. Adding that extra escaping should prevent a name with whitespace from breaking the manifest rendering / throwing errors due to the invalid JSON.

corgi/core/models.py
jasinner: I tried to see if we have any ComponentNodes with a obj == None, but it's not possible to use a reverse query on this relationship, but if you check object_id=None you don't get any result.

jasinner (approved):
Looks good to me!

PR #354, opened 2023-03-20
CORGI-508: Add a download URL for RPMs

juspence:
Quickly add download URLs for RPMs, using a predictable URL I found while researching other tickets.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -5,6 +5,7 @@ import uuid as uuid
 from abc import abstractmethod
 from typing import Any, Union
 
+from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres import fields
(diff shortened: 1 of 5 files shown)


juspence:
Our existing test for "component names with backslash" failed due to a lack of escaping, now that the SRPM download URL includes the component name. I still need to test that adding the escaping to the downloadUrl field in the manifest doesn't break other, valid URLs.

corgi/core/models.py
```diff
@@ -1576,15 +1577,19 @@ def _build_download_url_for_type(self) -> str:
...
+        # /brewroot/packages/servicemesh-prometheus/2.14.0/8.el8.1/src/
+        # servicemesh-prometheus-2.14.0-8.el8.1.src.rpm
         if self.type == Component.Type.RPM:
-            return self.RPM_PACKAGE_BROWSER
+            if not self.release:
+                # TODO: Special case that should never happen
+                #  Are there any others we should handle? Or should this fallback just go away?
+                return self.RPM_PACKAGE_BROWSER
+            return (
+                f"{settings.BREW_DOWNLOAD_ROOT_URL}/packages/"
```
jasinner, line 1589: Download URLs get included in manifests which are destined for users outside of Red Hat. This would be exposing our internal Brew download URL to those users, which is not only irrelevant to them, but also a security concern.

jasinner (commented):
I think perhaps we could add the BREW URL for a SRPM to the SoftwareBuild object instead? But adding it to the Component like this exposes this internal URL in the manifests and therefore outside the Red Hat firewall.
I think a better approach is keep it pointing to the package browser on the portal but research a way to find the signing key used for the SRPM. Perhaps from Pulp?

corgi/core/models.py
juspence: The way to get a download URL we can share externally is described in the old version of the comment here, and in CORGI-47. But we're already exposing an internal-only download URL (registry-proxy) for OpenLCS, so I thought doing this for RPMs would be OK. Let's discuss more later this week.

PR #355, opened 2023-03-21
CORGI-495: Add Syft method for scanning a Git repo with a specific ref

mprpic:
This allows us to manifest sources in additional to artifacts (like images). See Jira issue for more info on how this relates to managed services.

### DIFF
--- a/corgi/collectors/syft.py
+++ b/corgi/collectors/syft.py
@@ -2,8 +2,10 @@ import json
 import logging
 import subprocess  # nosec B404
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Any
 
+from django.conf import settings
 from packageurl import PackageURL
 
 from corgi.core.models import Component
@@ -43,7 +45,7 @@ class Syft:
                     "-q",
                     "-o=syft-json",
                     # see motivation for excluding test/fixtures in CORGI-510
-                    # For example skip test/fixtures/0-dns/package.json file in the nodejs
+                    # For example skip test/fixtures/0-dns/package.json file in nodejs
                     "--exclude=**/vendor/**",
                     "--exclude=**/test/fixtures/**",
                     f"{scheme}:{target_file}",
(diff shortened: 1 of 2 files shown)


mprpic:

corgi/collectors/syft.py
```diff
@@ -78,6 +80,29 @@ def scan_repo_image(cls, target_image: str) -> list[dict[str, Any]]:
         parsed_components = cls.parse_components(scan_result)
         return parsed_components
 
+    @classmethod
+    def scan_git_repo(
+        cls, target_url: str, target_ref: Optional[str] = None
```
juspence: Not important, but FYI you can set ```target_ref: str = ""``` to avoid needing the Optional import.

corgi/collectors/syft.py
```diff
@@ -78,6 +80,29 @@ def scan_repo_image(cls, target_image: str) -> list[dict[str, Any]]:
...
+        """Scan a source Git repository.
+
+        An optional target ref can be specified that represents a valid committish in the Git
+        repo being scanned.
+        """
+        with TemporaryDirectory(dir=settings.SCA_SCRATCH_DIR) as scan_dir:
+            logger.info("Cloning %s to %s", target_url, scan_dir)
+            subprocess.check_call(
+                ["/usr/bin/git", "clone", target_url, scan_dir], stderr=subprocess.DEVNULL
+            )  # nosec B603
```
juspence, line 94: Any reason you send stderr to /dev/null here? There's code from corgi.tasks.common that wraps the subprocess call for convenience. Part of that code treats stderr output (warnings) as errors that make the task fail.

I might like to reuse that, unless we think every git clone will have some stderr output. AKA just debug stuff, not related to actual errors.

juspence (approved):
LGTM!

corgi/collectors/syft.py
mprpic: `git clone` (and `checkout`) send some output to stderr and it shows up as debugging output in logs, so this essentially hides it:
```
$ git checkout main 1> /dev/null
Already on 'main'
$ git checkout main 2> /dev/null
Your branch is up to date with 'origin/main'.
```

corgi/collectors/syft.py
mprpic: Sure, changed.

PR #356, opened 2023-03-21
Fix monitoring email when too many tasks have failed

juspence:
@RedHatProductSecurity/corgi-devs Quick review please. I've tested locally and this seems to work, but I could also deploy to stage (without merging) if people prefer.

### DIFF
--- a/corgi/tasks/monitoring.py
+++ b/corgi/tasks/monitoring.py
@@ -5,6 +5,7 @@ from celery.utils.log import get_task_logger
 from celery_singleton import Singleton, clear_locks
 from django.conf import settings
 from django.core.mail import EmailMessage
+from django.db.models import Count
 from django.utils import timezone
 from django_celery_beat.models import CrontabSchedule, IntervalSchedule, PeriodicTask
 from django_celery_results.models import TaskResult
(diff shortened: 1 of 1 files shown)


corgi/tasks/monitoring.py
```diff
@@ -83,29 +84,41 @@ def email_failed_tasks():
...
+    failed_tasks = TaskResult.objects.filter(
+        status__exact="FAILURE",
+        date_done__gte=failed_tasks_threshold,
+    ).using("read_only")
 
+    failed_tasks_count = failed_tasks.count()
     subject = (
-        f"Failed Corgi Celery tasks after "
-        f"{failed_tasks_threshold.date()}: {failed_tasks.count()}"
+        f"Failed Corgi Celery tasks after " f"{failed_tasks_threshold.date()}: {failed_tasks_count}"
```
mprpic: Black mangled this a bit, jfyi.

corgi/tasks/monitoring.py
```diff
@@ -83,29 +84,41 @@ def email_failed_tasks():
...
     )
 
-    failed_tasks = "\n".join(
-        f"{task_name}: args={task_args}, kwargs={task_kwargs}\n"
-        f"result={result}\n"
-        f"{traceback}\n"
-        for (task_name, task_args, task_kwargs, result, traceback) in failed_tasks.iterator(
-            chunk_size=500
+    if failed_tasks_count < 5000:
```
mprpic, line 97: 5000 still seems like a lot, will that not go over the email message size limit?

mprpic (approved):
Looks good!

corgi/tasks/monitoring.py
juspence: It depends on the message size (length of args + kwargs + traceback). The largest email I've seen came in with 17k errors, but I could lower to 2k or 1k if you think that's safer.

mprpic: Ha, if 17k worked then I guess 5k is fine :)

PR #357, opened 2023-03-21
CORGI-508: Fix container ingest code so that reloading with more data works

juspence:
@RedHatProductSecurity/corgi-devs Fixes an issue where container metadata from Brew is never updated after the component is first created.

Once this is merged, I'll quickly reprocess all the container builds again. It went very fast on Friday so I'm hopeful this won't take too much time. Or if not, I'll be working on some follow-up MRs here to fix a few other lower-priority bugs in this code (missing namespace, arch, etc.) so the reload going slowly won't really be a problem.


juspence:
This is blocking the container data reload / our next production release, and the change is very small, so going to just merge.

PR #360, opened 2023-03-22
Move manifest-generating tasks to CPU workers with more memory

juspence:
@jasinner This change is the same "solution" as the WorkerLostError for the SCA task. We can't guarantee it will never happen, but we give the SCA worker pods as much memory as possible (2GB) so we can avoid it most of the time.

Better fixes for the amount of memory used are still welcome, of course.

### DIFF
--- a/corgi/tasks/manifest.py
+++ b/corgi/tasks/manifest.py
@@ -20,7 +20,7 @@ def update_manifests():
     for ps in ProductStream.objects.annotate(num_components=Count("components")).filter(
         num_components__gt=0
     ):
-        slow_update_ps_manifest.delay(ps.name)
+        cpu_update_ps_manifest.delay(ps.name)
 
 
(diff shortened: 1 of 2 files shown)


jasinner (commented):
The slow worker is the one which has 2G, not the cpu worker which as 768M

juspence:
> The slow worker is the one which has 2G, not the cpu worker which as 768M

Yes, the CPU worker is supposed to have 2G and the slow worker really doesn't need that much memory. We had WorkerLostErrors (OoM problems) in the SCA task whenever it tried to git clone a large repo, and this caused other issues. I opened a PR in the ops repo to switch this back and tagged you for review there as well.

jasinner (approved):

JimFuller-RedHat (approved):
LGTM

PR #361, opened 2023-03-22
CORGI-313: Fix missing arch and namespace

juspence:
@RedHatProductSecurity/corgi-devs Please review. This should fix some bugs where arch and namespace could be an empty string. Fixing this is required before we can constrain arch to a list of enum values, as attempted in a previous PR (don't have link handy).

I also tried to make sure we always pass enough kwargs to get_or_create() / update_or_create() to hit an index and guarantee the function is atomic. Mostly this just means passing release="" even if we know it should be an empty string / unset for some (e.g. upstream) component. I also made the kwarg ordering consistent just for readability (make it easier to see if something is missing).

Finally, I also switched to using update_or_create consistently so that all component types can be reprocessed successfully. The only case I didn't do this is in the SCA task, where the Brew metadata would get overwritten by the Syft metadata (which we don't want), and in cases where there's nothing to update (e.g. a constant value in the defaults dict that never changes).

We need a migration to fix the existing data - I can handle that here or in a follow-up PR. But I wanted to open this now so I could discuss some questions - see notes below.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -238,7 +238,9 @@ def save_component(
         defaults={
             "description": meta.pop("description", ""),
             "filename": find_package_file_name(meta.pop("source_files", [])),
-            "namespace": component.get("namespace", ""),
+            "namespace": Component.Namespace.REDHAT
+            if component_type == Component.Type.RPM
+            else Component.Namespace.UPSTREAM,
             "related_url": related_url,
             "software_build": softwarebuild,
         },
@@ -289,7 +291,7 @@ def save_srpm(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentNode:
             "description": build_data["meta"].get("description", ""),
             "license_declared_raw": build_data["meta"].get("license", ""),
             "meta_attr": build_data["meta"],
-            "namespace": build_data["namespace"],
+            "namespace": Component.Namespace.REDHAT,
             "related_url": related_url,
             "software_build": softwarebuild,
         },
@@ -355,7 +357,7 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
             "description": build_data["meta"].pop("description", ""),
             "filename": find_package_file_name(build_data["meta"].pop("source_files", [])),
             "meta_attr": build_data["meta"],
-            "namespace": build_data.get("namespace", ""),
+            "namespace": Component.Namespace.REDHAT,
             "related_url": related_url,
             "software_build": softwarebuild,
         },
@@ -402,7 +404,7 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
                 arch=image["meta"].pop("arch"),
                 defaults={
                     "meta_attr": image["meta"],
-                    "namespace": image.get("namespace", ""),
+                    "namespace": Component.Namespace.REDHAT,
                     "software_build": softwarebuild,
                 },
             )
@@ -487,7 +489,7 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
             "description": build_data["meta"].get("description", ""),
             "license_declared_raw": build_data["meta"].get("license", ""),
             "meta_attr": meta_attr,
-            "namespace": build_data["namespace"],
+            "namespace": Component.Namespace.REDHAT,
             "software_build": softwarebuild,
         },
     )
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -73,9 +73,9 @@ def slow_fetch_brew_build(build_id: int, save_product: bool = True, force_proces
         build_id=component["build_meta"]["build_info"]["build_id"],
         defaults={
             "build_type": SoftwareBuild.Type.BREW,
+            "meta_attr": build_meta,
             "name": component["meta"]["name"],
             "source": component["build_meta"]["build_info"]["source"],
-            "meta_attr": build_meta,
         },
         completion_time=completion_dt,
     )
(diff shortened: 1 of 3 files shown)


corgi/tasks/brew.py
```diff
@@ -388,15 +398,16 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
...
+                name=image["meta"].pop("name"),
                 version=image["meta"].pop("version"),
                 release=image["meta"].pop("release"),
+                arch=image["meta"].pop("arch"),
                 defaults={
-                    "software_build": softwarebuild,
                     "meta_attr": image["meta"],
-                    "namespace": image.get("namespace", ""),
+                    # TODO: Check this - should it be UPSTREAM or REDHAT here?
+                    "namespace": Component.Namespace.UPSTREAM,
```
juspence: I'm not sure what "image_components" are in reference to a container. Are these upstream components, or RPMs, or a mix of both?

corgi/tasks/brew.py
```diff
@@ -235,11 +237,11 @@ def save_component(
         name=meta.pop("name", ""),
         version=component_version,
         release=meta.pop("release", ""),
-        arch=meta.pop("arch", ""),
+        arch=meta.pop("arch", "noarch"),
         defaults={
             "description": meta.pop("description", ""),
             "filename": find_package_file_name(meta.pop("source_files", [])),
-            "namespace": component.get("namespace", ""),
+            "namespace": Component.Namespace.REDHAT if rpm else Component.Namespace.UPSTREAM,
```
juspence: Note to self: This "rpm" kwarg should go away - we can just do below instead:
```suggestion
            "namespace": Component.Namespace.REDHAT if component_type = Component.Type.RPM.upper() else Component.Namespace.UPSTREAM,
```

corgi/tasks/brew.py
```diff
@@ -469,17 +482,17 @@ def save_module(softwarebuild, build_data) -> ComponentNode:
...
         release=build_data["meta"].get("release", ""),
+        arch=build_data["meta"].get("arch", "noarch"),
         defaults={
-            "license_declared_raw": build_data["meta"].get("license", ""),
             "description": build_data["meta"].get("description", ""),
-            "software_build": softwarebuild,
+            "license_declared_raw": build_data["meta"].get("license", ""),
             "meta_attr": meta_attr,
-            "namespace": build_data["namespace"],
+            "namespace": Component.Namespace.REDHAT,
```
juspence, line 492: We need to fix existing data that has empty "" strings instead of a real value. For arch, we can just migrate any occurrence of "" to "noarch", and this won't have any effect on the purl. "noarch" and "" are both excluded / don't generate any qualifier like "?arch=example".

Some migrations for namespace are also easy - modules and index containers and SRPMs should all have namespace "REDHAT". Non-root components get "REDHAT" if they're RPMs, or "UPSTREAM" for any remote-source component.

Others are tricky - upstream containers obviously should have namespace "UPSTREAM", but I'm not sure how to identify them. Is it safe to just assume any container, which isn't a root component (isn't an index container), is therefore an upstream container?

Or should we avoid trying to guess in a migration, and just reload the container instead? AKA hit this codepath, and let the fixed logic could decide for us?

corgi/tasks/sca.py
```diff
@@ -83,12 +83,13 @@ def save_component(component: dict[str, Any], parent: ComponentNode) -> bool:
...
     try:
         # Use all fields from Component index and uniqueness constraint
+        # Don't update_or_create since Syft's metadata shouldn't override Brew's
         obj, created = Component.objects.get_or_create(
             type=component["type"],
             name=name,
             version=meta.pop("version", ""),
             release="",
-            arch="",
+            arch="noarch",
```
juspence, line 92: TODO: Oops, I forgot to set the namespace here also. Does below seem right?
```suggestion
            arch="noarch",
            defaults={"meta_attr": meta, "namespace": Component.Namespace.REDHAT if component["type"] == Component.Type.RPM else Component.Namespace.UPSTREAM},
```

corgi/tasks/brew.py
jasinner: We should use REDHAT namespace for the image_components. They are parent images used as the layers to build the current image we are processing.

corgi/tasks/brew.py
jasinner: I think it's safe to assume all container_image components are in the REDHAT namespace.

corgi/tasks/sca.py
jasinner: Looks sane to me. Although your question got me thinking if we are setting arch correctly for RPM components detected by Syft. I guess setting arch to "noarch" instead of "" is an improvment so I'm happy to ahead with this PR as is and revisit the arch for RPMs detected by Syft later.

jasinner (approved):
Great, thanks for fixing this up!

corgi/tasks/sca.py
juspence: Good catch about RPM arches...I guess you're right, any RPM with arch=noarch should only be created here, so we can easily migrate in the future if needed. Thanks for the review!

corgi/tasks/brew.py
juspence: I made a mistake above. I thought that we excluded "noarch" from getting added to all purls, but that's only true for containers. So the RPM purls will get a qualifier like ```?arch=noarch```, unless I also exclude "noarch" from there too. Do we have any preference here?

It looks like other ```_build_*_purl()``` functions don't consider the arch at all.

juspence: From discussion in standup, let's not change the arch value for RPMs. Instead, we should rerun SCA on all of them and have Syft set the correct architecture.

I think this might actually be hard, since Syft will do get_or_create with an ```arch=""``` kwarg. This will find the existing component, but not update it with the correct arch value. Maybe we could rerun SCA to create components with the right arch values, then copy the nodes from the old component to the new one and re-save all the taxonomies.

I need to think more about the right way to clean up this old data, but given approval I'm going to merge the MR without the migration. That will at least prevent more bad data from getting created - we can add a proper migration / reingest data as part of CORGI-438, when we constrain the allowed arch values.

jasinner: What I was thinking was to adjust the Syft collector to set the arch correctly for detected RPM, instead of always setting it to arch="". Then it can to get_or_create() with the correct arch value. It looks like [Syft is capable](https://github.com/anchore/syft/tree/0774ad15e26bc9230dbfc393b3e7d6bd36b93cf7/syft/pkg/cataloger/rpm) of detecting RPM arch. I just neglected to pass it through from the collector in Corgi.

Anyway, I still think this PR is an improvement, so let's create a follow task for the Syft collector RPM arch problem.

JimFuller-RedHat (approved):
LGTM though if we are fixing arch is now the time to consider constraining enum eg. Component.Arch. ?

corgi/tasks/brew.py
juspence: Yes, I think either creating a duplicate component (only arch is different) will fail, or we'll end up with the same component in the DB twice. I was worried about losing data if we deleted the old one with arch="", so was trying to figure out a way to migrate the old data over.

Maybe this isn't an issue, if that component was only created by Syft and doesn't have any other metadata. Or maybe we can just create the new one with the correct arch, migrate the nodes from the old one to the new one, then delete the old one.

PR #362, opened 2023-03-22
CORGI-530 oom openshift manifests

jasinner:
This removes duplicates components from the manifest 'packages' list, and only lists provides once as leaf nodes.

The test seem to work OK, but when I compare the old openshift-4.7 manifest with the current one, there seems to be some components missing. Leaving as draft for now.

new:
```
$ cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | wc -l
313
$ cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | sort | uniq | wc -l
313
```

old:
```
$ cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | wc -l
239537
$ cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | sort | uniq | wc -l
35443
```

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -301,7 +301,6 @@ class SoftwareBuild(TimeStampedModel):
         )
 
         product_details = get_product_details(variant_names, stream_names)
-
         components = set()
         for component in self.components.get_queryset().iterator():
             # This is needed for container image builds which pull in components not
@@ -665,6 +664,7 @@ class ProductStream(ProductModel):
                 Component.objects.filter(
                     ROOT_COMPONENTS_CONDITION, productstreams__ofuri=self.ofuri
                 )
+                .exclude(name__endswith="-container-source")
                 .distinct("name")
                 .values_list("name", flat=True)
                 .using(using)
@@ -1466,7 +1466,6 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         self.productstreams.add(*product_pks_dict["productstreams"])
         self.productvariants.add(*product_pks_dict["productvariants"])
         self.channels.add(*product_pks_dict["channels"])
-
         return None
 
     def get_roots(self, using: str = "read_only") -> list[ComponentNode]:
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -689,6 +689,7 @@ class ProductStream(ProductModel):
             .distinct()
             .order_by("provides__pk")
             .using(using)
+            .iterator()
         )
         return Component.objects.filter(pk__in=unique_provides).using(using)
 
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -679,6 +679,19 @@ class ProductStream(ProductModel):
                 return Component.objects.none()
             return Component.objects.filter(Q(ROOT_COMPONENTS_CONDITION & query)).using(using)
 
+    @property
+    def provides_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
+        """Returns unique aggregate "provides" for the latest components in this stream,
+        for use in templates"""
+        unique_provides = (
+            self.get_latest_components()
+            .values_list("provides__pk", flat=True)
+            .distinct()
+            .order_by("provides__pk")
+            .using(using)
+        )
+        return Component.objects.filter(pk__in=unique_provides).using(using)
+
 
 class ProductStreamTag(Tag):
     tagged_model = models.ForeignKey(ProductStream, on_delete=models.CASCADE, related_name="tags")
--- a/corgi/web/templates/product_manifest.json
+++ b/corgi/web/templates/product_manifest.json
@@ -80,10 +80,5 @@
           "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",
           "relationshipType": "PACKAGE_OF",
           "spdxElementId": "SPDXRef-{{component.uuid}}"
-        },{% endfor %}{% for provided in distinct_provides %}
-        {
-          "relatedSpdxElement": "NONE",{# subcomponent is a leaf node #}
-          "relationshipType": "CONTAINS",
-          "spdxElementId": "SPDXRef-{{provided.pk}}"
         },{% endfor %}
 {% endblock relationships %}
--- a/corgi/web/templates/product_manifest.json
+++ b/corgi/web/templates/product_manifest.json
@@ -1,5 +1,5 @@
 {# ProductModel manifest, obj is a ProductModel subclass here #}{% extends "base_manifest.json" %}
-{% block packages %}{% for component in latest_components %}{# Below will eventually become component.manifest #}
+{% block packages %}{% for component in latest_components %}
     {
         "copyrightText": {% if component.copyright_text %}"{{component.copyright_text}}"{% else %}"NOASSERTION"{% endif %},
         "downloadLocation": {% if component.download_url %}"{{component.download_url}}"{% else %}"NOASSERTION"{% endif %},
@@ -21,7 +21,7 @@
         "SPDXID": "SPDXRef-{{component.uuid}}",
         "supplier": "Organization: Red Hat",
         "versionInfo": "{{component.nevra|escapejs}}"
-    },{% for provided in component.provides_queryset %}
+    },{% endfor %}{% for provided in distinct_provides %}
     {
         "copyrightText": {% if provided.copyright_text %}"{{provided.copyright_text}}"{% else %}"NOASSERTION"{% endif %},
         "downloadLocation": {% if provided.download_url %}"{{provided.download_url}}"{% else %}"NOASSERTION"{% endif %},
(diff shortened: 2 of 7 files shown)


jasinner:
I still get signal 9 for slow workers trying to generate openshift-4 manifests. Investigating now to reduce the memory consumption more.

corgi/web/templates/product_manifest.json
```diff
@@ -21,7 +21,7 @@
         "SPDXID": "SPDXRef-{{component.uuid}}",
         "supplier": "Organization: Red Hat",
         "versionInfo": "{{component.nevra|escapejs}}"
-    },{% for provided in component.provides_queryset %}
+    },{% for provided in distinct_provides %}
```
juspence: I think I'm very confused, but isn't this still going to give duplicates / cause other problems? For example, assume we have a stream with two root components A and B. Let's say that A provides A1 and C1, while B provides B1 and C1.

In the old code, we would list out each provided component in each root component's provides_queryset. So we would list A1 and C1 as part of A's provides_queryset, then B1 and C1 as part of B's provides_queryset. And I thought dupliating the "C1" was the bug we're trying to fix here.

In the new code, we're using the provides_queryset property on the stream instead of on the individual components, to avoid this duplication. But this "for provided in distinct_provides" loop here is still inside of the outer "for component in latest_components" loop.

So we'd list the whole stream's distinct_provides A1, B1, and C1 for the latest_component A, then we'd list the whole stream's distinct_provides A1, B1, and C1 again for the latest_component B. Maybe I've misunderstood somewhere...if not, is below right?
```suggestion
    },{% endfor %}{% for provided in distinct_provides %}
```

If above is correct, we also need to remove one of the two ```{% endfor %}``` blocks on line 46. I can't leave a comment there since it's outside of the diff window.

tests/test_manifests.py
```diff
@@ -101,30 +168,37 @@ def test_product_manifest_properties():
         "spdxElementId": f"SPDXRef-{dev_provided.uuid}",
     }
 
+    print(f"stream uuid:{stream.uuid}")
+    print(f"component uuid:{component.uuid}")
+    print(f"provided: {provided.uuid}")
+    print(f"dev_provided: {dev_provided.uuid}")
+
+    print(manifest["relationships"])
+
```
juspence: Looks like some debug code got left behind.

corgi/core/models.py
```diff
@@ -679,6 +679,18 @@ def get_latest_components(
...
+    def provides_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
+        """Returns unique aggregate "provides" for the latest components in this stream,
+        for use in templates"""
+        unique_provides = (
+            self.get_latest_components()
+            .distinct("provides__pk")
+            .values_list("provides__pk")
+            .using(using)
+        )
+        return Component.objects.filter(pk__in=unique_provides).using(using)
```
juspence, line 694: You might try adding .iterator() to the end here, then getting rid of "distinct_provides" in the template and cirectly using "obj.provides_queryset" instead.

That will mean making the same query twice instead of once, but using an iterator prevents Django from caching all the queryset results in memory, so might fix the OoM errors for the OpenShift manifests.

We could also do the same (use .iterator() and query the DB twice) to save memory in the latest_components queryset as well if you think it might help.

corgi/web/templates/product_manifest.json
```diff
@@ -75,16 +75,15 @@
...
           "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",
           "relationshipType": "PACKAGE_OF",
           "spdxElementId": "SPDXRef-{{component.uuid}}"
-        },
-    {% endfor %}
+        },{% endfor %}{% for provided in distinct_provides %}
+        {
+          "relatedSpdxElement": "NONE",{# subcomponent is a leaf node #}
+          "relationshipType": "CONTAINS",
+          "spdxElementId": "SPDXRef-{{provided.pk}}"
```
juspence: I confused myself a little, but after looking at this way too long, I think the new code is good. In the old code, we say that:
```
This document describes a stream

The subcomponent A1 is built from / contained by the component A
The subcomponent A1 is a leaf node
The subcomponent C1 is built from / contained by the component A
The subcomponent C1 is a leaf node
The root component A is a package of the stream

The subcomponent B1 is built from / contained by the component B
The subcomponent B1 is a leaf node
The subcomponent C1 is built from / contained by the component B
The subcomponent C1 is a leaf node
The root component B is a package of the stream
```

In the new code, we say almost the the same thing, without duplicating "C1 is a leaf node". But after looking this over, these leaf node statements seem to be a large percentage of the relationships section, and I'm not sure they're necessary.

They were part of the example SPDX document in the spec that I used as a reference, but I don't think the spec ever explicitly said they were required. And completely removing them might save a lot of space when the number of provides is very large. I guess it depends on whether we think explicitly indicating leaf nodes is helpful or not.

corgi/core/models.py
```diff
@@ -679,6 +679,18 @@ def get_latest_components(
...
             return Component.objects.filter(Q(ROOT_COMPONENTS_CONDITION & query)).using(using)
 
+    @property
+    def provides_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
+        """Returns unique aggregate "provides" for the latest components in this stream,
+        for use in templates"""
+        unique_provides = (
+            self.get_latest_components()
+            .distinct("provides__pk")
+            .values_list("provides__pk")
```
juspence: I think it's best to use flat=True and put .distinct() after the .values_list(), like below:
```suggestion
            .values_list("provides__pk", flat=True)
            .distinct()
```

juspence (approved):
I see one possible issue, but I might be confused and everything else LGTM. I didn't really look at the tests though - they pass, so I'm happy.

tests/test_manifests.py
```diff
@@ -30,21 +31,87 @@
...
 def test_product_manifest_properties():
     """Test that all models inheriting from ProductModel have a .manifest property
     And that it generates valid JSON."""
-    component, stream, provided, dev_provided = setup_products_and_components()
+    component, stream, provided, dev_provided = setup_products_and_components_provides()
 
-    manifest = json.loads(stream.manifest)
+    manifest = stream.manifest
+    print(manifest)
```
juspence: More debug code that got left behind.

corgi/web/templates/product_manifest.json
jasinner: Ha, you are correct. That is grossly inefficient of me, whoops. I fixed it up so each provided component in the stream is only listed once. I also added order_by like you added on Comonent.get_provides_nodes_queryset to ensure the order of elements in the SPDX document don't change unless the content does.

corgi/web/templates/product_manifest.json
jasinner: Great, I'll remove them which will reduce the memory required quite a bit hopefully.

tests/test_manifests.py
jasinner: Removed.

tests/test_manifests.py
jasinner: I removed it now, thanks.

jasinner:
> I see one possible issue, but I might be confused and everything else LGTM. I didn't really look at the tests though - they pass, so I'm happy.

I've now added some test coverage for the packages array in the manifest which I should have added earlier that would have caught this silly mistake. Thanks.

corgi/core/models.py
jasinner: Good idea. I also added order_by and iterator here.

corgi/core/models.py
jasinner: Since we removed the provided contains nothing loop from the relationships array it's a no-brainer to added iterator now, thanks.

jasinner:
Setting this as ready for merge. I managed to generate an openshift-4.7 stream manifest in the slow worker locally with 2G memory limit set. The number of rpm components in the packages list also is pretty close to the number of deduplicated in an CORGI-530 issue comment. This is a different data set, so that could account for the difference perhaps.

In dev:
```
cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | grep ':rpm/redhat' | wc -l
4270
```
From CORGI-530 comment:
```
$ cat corgi/output_files/openshift-4.7-a6b89c93-7c76-4737-984d-a0f69ac7326f.json | jq '.packages[].externalRefs[].referenceLocator' | grep ':rpm/redhat' | sort | uniq | wc -l
4244
```

JimFuller-RedHat (approved):
LGTM

corgi/web/templates/product_manifest.json
juspence: No worries! I'm just embarrassed I made this mistake (duplicate provides in manifests) in the first place. Hopefully this fixes the issue with OoM, but even if not MR LGTM. Thak you for finding and fixing this bug!

PR #365, opened 2023-03-27
CORGI-536 - change sources, provides and upstreams to do exact match

JimFuller-RedHat:
?sources={}, ?provides={} and ?upstreams={} should perform an exact match .. this has the benefit of avoiding a LIKE in generated SQL (which is seriously slow and expensive query).

in addition we added re_ variants to enable regex matching if that is so desired.

**IMPORTANT** - when doing this I added some more tests but confused by the result of explicitly testing node creation ... needs discussion


JimFuller-RedHat:
going to close this for now

PR #366, opened 2023-03-27
added test for component node creation to make structure explicit

JimFuller-RedHat:
### OCI ComponentNode creation test
Given an OCI component node structure:
```
    /
    └──(SOURCE) root_comp: top level root component
        ├──(SOURCE) upstream_comp
        └──(PROVIDES) dep1_comp
            └──(PROVIDES) dep2_comp
```
The following are expectations in terms of sources, providers and upstreams url params on each of these components.
```
    +--------------------+---------+----------+-----------+
    |                    | sources | provides | upstreams |
    +--------------------+---------+----------+-----------+
    | root_component     | 0       | 2        | 1         |
    +--------------------+---------+----------+-----------+
    | upstream_component | 0       | 0        | 0         |
    +--------------------+---------+----------+-----------+
    | dep1_component     | 1       | 1        | 0         |
    +--------------------+---------+----------+-----------+
    | dep2_component     | 2       | 0        | 0         |
    +--------------------+---------+----------+-----------+
```
-----------------------------
### SRPM ComponentNode creation test
Given a SRPM root component node structure:
```
    /
    └──(SOURCE) root_comp: top level root component
        ├──(SOURCE) upstream_comp
        └──(PROVIDES) dep1_comp
```
Then we expect the following in terms of sources, providers and upstreams url params for each of these components.
```
    +--------------------+---------+----------+-----------+
    |                    | sources | provides | upstreams |
    +--------------------+---------+----------+-----------+
    | root_component     | 0       | 1        | 1         |
    +--------------------+---------+----------+-----------+
    | upstream_component | 0       | 0        | 3         |
    +--------------------+---------+----------+-----------+
    | dep1_component     | 1       | 0        | 1         |
    +--------------------+---------+----------+-----------+
```
-----------------------------

What happens with middleware, services or C/C++ components (and/or GOLANG for that matter) ? 

**Note** - We have other tests giving us coverage in this area - the goal with these tests is to characterise _canonical_ component node creation to ensure everyone shares understanding. We also want to have a clear test to help us tease out component relationships eg.  not conflate rpm or container packaging relationships with runtime/buildtime/devtime relationships.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -786,3 +786,319 @@ def test_product_components(client, api_path):
     response = client.get(f"{api_path}/components?products={rhel_br.name}")
     assert response.status_code == 200
     assert response.json()["count"] == 1
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_oci_component_provides_sources_upstreams(client, api_path):
+    """
+    Given an OCI component node structure:
+    /
+    └──(SOURCE) root_comp: top level root component
+        ├──(SOURCE) upstream_comp
+        └──(PROVIDES) dep1_comp
+            └──(PROVIDES) dep2_comp
+
+    The following are expectations in terms of sources, providers and upstreams on each of these components ?
+
+    +--------------------+---------+----------+-----------+
+    |                    | sources | provides | upstreams |
+    +--------------------+---------+----------+-----------+
+    | root_component     | 0       | 2        | 1         |
+    +--------------------+---------+----------+-----------+
+    | upstream_component | 0       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep1_component     | 1       | 1        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep2_component     | 2       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+
+    The data model does not impose this constraint eg. provides, sources and upstream property queries enforce
+    this behaviour.
+
+    """
+
+    # create a top level root source component
+    root_comp = ComponentFactory(
+        name="root_comp",
+        type=Component.Type.CONTAINER_IMAGE,
+        arch="noarch",
+        namespace=Component.Namespace.REDHAT,
+        related_url="https://example.org/related",
+    )
+    root_node, _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        purl=root_comp.purl,
+        defaults={"obj": root_comp},
+    )
+
+    # create upstream child node
+    upstream_comp = ComponentFactory(
+        name="upstream_comp",
+        type=Component.Type.RPM,
+        namespace=Component.Namespace.UPSTREAM,
+        related_url="https://example.org/related",
+    )
+    ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=root_node,
+        purl=upstream_comp.purl,
+        defaults={"obj": upstream_comp},
+    )
+
+    # create dep component
+    dep_comp = ComponentFactory(
+        name="cool_dep_component",
+        arch="src",
+        type=Component.Type.RPM,
+        namespace=Component.Namespace.REDHAT,
+    )
+    # make it a child of root OCI component
+    dep_provide_node, _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.PROVIDES,
+        parent=root_node,
+        purl=dep_comp.purl,
+        defaults={"obj": dep_comp},
+    )
+    # create 2nd level dep
+    dep2_comp = ComponentFactory(
+        name="cool_dep2_component",
+        type=Component.Type.RPM,
+        arch="aarch64",
+        namespace=Component.Namespace.REDHAT,
+    )
+    # making it a child node of dep1_component
+    ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.PROVIDES,
+        parent=dep_provide_node,
+        purl=dep2_comp.purl,
+        defaults={"obj": dep2_comp},
+    )
+
+    # TODO - investigate if invoking any of the following in any order is stable
+    # it is intentional to invoke these twice during the test
+    upstream_comp.save_component_taxonomy()
+    upstream_comp.save()
+    dep2_comp.save_component_taxonomy()
+    dep2_comp.save()
+    dep_comp.save_component_taxonomy()
+    dep_comp.save()
+    root_comp.save_component_taxonomy()
+    root_comp.save()
+    upstream_comp.save_component_taxonomy()
+    upstream_comp.save()
+    dep2_comp.save_component_taxonomy()
+    dep2_comp.save()
+    dep_comp.save_component_taxonomy()
+    dep_comp.save()
+    root_comp.save_component_taxonomy()
+    root_comp.save()
+
+    assert dep_comp.purl in root_comp.provides.values_list("purl", flat=True)
+    assert dep2_comp.purl in root_comp.provides.values_list("purl", flat=True)
+
+    response = client.get(f"{api_path}/components?namespace=REDHAT")
+    assert response.status_code == 200
+    assert response.json()["count"] == 3
+
+    response = client.get(f"{api_path}/components/{root_comp.uuid}")
+    assert response.status_code == 200
+    assert response.json()["name"] == root_comp.name
+    assert response.json()["related_url"] == root_comp.related_url
+    assert len(response.json()["sources"]) == 0
+    assert len(response.json()["provides"]) == 2
+    assert len(response.json()["upstreams"]) == 1
+
+    response = client.get(f"{api_path}/components/{dep_comp.uuid}")
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 1
+    assert len(response.json()["provides"]) == 1
+    assert len(response.json()["upstreams"]) == 0
+
+    response = client.get(f"{api_path}/components/{dep2_comp.uuid}")
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 2
+    assert len(response.json()["provides"]) == 0
+    assert len(response.json()["upstreams"]) == 0
+
+    response = client.get(f"{api_path}/components/{upstream_comp.uuid}")
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 0
+    assert len(response.json()["provides"]) == 0
+    assert len(response.json()["upstreams"]) == 0
+
+    # retrieve all sources of root_comp
+    response = client.get(f"{api_path}/components?provides={root_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["results"] == []
+    assert response.json()["count"] == 0
+    # retrieve all provides of root_comp
+    response = client.get(f"{api_path}/components?sources={root_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 2
+
+    # retrieve all sources of dep_comp component
+    response = client.get(f"{api_path}/components?provides={dep_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+    # retrieve all provides of dep_comp
+    response = client.get(f"{api_path}/components?sources={dep_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+
+    # retrieve all sources of dep2_comp component
+    response = client.get(f"{api_path}/components?provides={dep2_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 2
+    # retrieve all provides of dep2_comp
+    response = client.get(f"{api_path}/components?sources={dep2_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 0
+
+    # retrieve all components with upstream_comp upstream
+    response = client.get(f"{api_path}/components?upstreams={upstream_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_srpm_component_provides_sources_upstreams(client, api_path):
+    """
+    Given a SRPM root component node structure:
+    /
+    └──(SOURCE) root_comp: top level root component
+        ├──(SOURCE) upstream_comp
+        └──(PROVIDES) dep1_comp
+
+    Then we expect the following in terms of sources, providers and upstreams for each of these components.
+
+    +--------------------+---------+----------+-----------+
+    |                    | sources | provides | upstreams |
+    +--------------------+---------+----------+-----------+
+    | root_component     | 0       | 1        | 1         |
+    +--------------------+---------+----------+-----------+
+    | upstream_component | 0       | 0        | 0        |
+    +--------------------+---------+----------+-----------+
+    | dep1_component     | 1       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+
+    This behaviour is predicated on the following assumptions:
+    * SRPM -> arch RPM
+    * binary RPM never have child dependencies
+
+    The data model does not impose this constraint eg. provides, sources and upstream property queries enforce
+    this behaviour.
+
+    """
+
+    # create a top level root source component
+    root_comp = ComponentFactory(
+        name="root_comp",
+        type=Component.Type.RPM,
+        arch="src",
+        namespace=Component.Namespace.REDHAT,
+        related_url="https://example.org/related",
+    )
+    root_node, _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        purl=root_comp.purl,
+        defaults={"obj": root_comp},
+    )
+
+    # create upstream child node
+    upstream_comp = ComponentFactory(
+        name="upstream_comp",
+        type=Component.Type.RPM,
+        namespace=Component.Namespace.UPSTREAM,
+        related_url="https://example.org/related",
+    )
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat:
@mbaluch

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        purl=root_comp.purl,
+        defaults={"obj": root_comp},
+    )
+
+    # create upstream child node
+    upstream_comp = ComponentFactory(
+        name="upstream_comp",
+        type=Component.Type.PYPI,
```
juspence: For root components which are source RPMs, the code we have today creates the upstream component with type=RPM and arch=src.
```suggestion
        type=Component.Type.RPM,
        arch="src",
```

EDIT: Oops, I originally put "arch=noarch" here, but that's only for remote-source components and index container images, not source RPMs which should always have "arch=src".

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    | dep1_component     | 1       | 1        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep2_component     | 2       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+
+    Are the following statements true ?
+
+    * upstream has 0 sources as it is its own SOURCE
+    * should dep1 or dep2 dynamically resolve the top level component's upstreams ?
+    * should dep2 list 2 sources ?
```
juspence: Yes, it should and does. provides should be the inverse of sources, so any components that report "dep2" in its provides (there are two) should also appear in dep2"s "sources" (there are two).

EDIT: Another typo...I said "upstreams" here instead of "provides" - first word of the second sentence above. Clearly I didn't have enough coffee this morning, sorry for any confusion caused.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    * we should not list upstreams in provides right ?
```
juspence: Yes. The upstream component's node has type=source, and is a child of the root component node. The root component's "provides" field lists any child component which has a node with type=provides. So the child upstream component's node with type=source is ignored and not reported.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 2
+    assert len(response.json()["provides"]) == 0
+    assert len(response.json()["upstreams"]) == 0
+
+    # retrieve all sources of dep_comp component
+    response = client.get(f"{api_path}/components?provides={dep_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+
```
juspence: You might add a test for ```?sources={dep_comp.purl}``` and ```?sources={dep2_comp.purl}```.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    # retrieve all sources of dep2_comp component
+    response = client.get(f"{api_path}/components?provides={dep2_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 2
+
+    # retrieve all provided components
+    response = client.get(f"{api_path}/components?sources={root_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 2
+
```
juspence: We could add a test for ```?provides={root_comp.purl}``` to fully assert what's in the table above.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    Taking the following component node structure:
+    /
+    └──(SOURCE) root_comp: top level root component
+        ├──(SOURCE) upstream_comp
+        └──(PROVIDES) dep1_comp
+
+    where upstream is defined as a child node of a root comp and dep1 is also defined as a root component
+    /
+    └──(SOURCE) dep1_comp: top level root component
+            └──(PROVIDES) dep2_comp
```
juspence: As far as I understand, ```dep1 is also defined as a root component``` should never happen. Root components are only source RPMs, RPM modules, and index container images. These should create exactly one ComponentNode with type=SOURCE, parent=None, and purl= equal to the root Component's purl.

Each root component can provide many child components, and those child components can provide child components of their own. These child components should have exactly one ComponentNode with type=PROVIDES, parent=root_node, and purl= equal to the provided / child component's purl.

For example, an index container image (which is generic / not for a particular architecture) can provide binary RPMs, remote-source components like Python packages (which are not built at Red Hat), and architecture-specific containers.

Clients try to first download the index container (which is just a pointer if I understand correctly). Metadata from the index container tells the client "where is the container for my specific architecture?" and clients then download the architecture-specific container image.

That architecture-specific container will provide child components of its own, for example binary RPMs and Python packages. The list of provides for an architecture-specific container should be exactly the list of dependencies / components that are pulled inside the container at build time. These grandchild components should have exactly one ComponentNode with type=PROVIDES, parent=child_node, and purl= equal to the provided / grandchild component's purl.

```suggestion
    where upstream is defined as a child node of a root comp
    /
    └──(PROVIDES) dep1_comp: child component of top-level root component given above
            └──(PROVIDES) dep2_comp
```

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    | upstream_component | 0       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep1_component     | 1       | 1        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep2_component     | 2       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+
+    Are the following statements true ?
+
+    * upstream has 0 sources as it is its own SOURCE
```
juspence: This is true, but I think it's mostly an implementation detail and not an explicit decision we ever discussed. There was a bug in the past relating to sources not being a perfect inverse of provides.

Fixing that bug caused upstreams to report exactly 0 sources. Before that bugfix, you might have seen the root component listed as a source of the upstream.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    * should dep1 or dep2 dynamically resolve the top level component's upstreams ?
```
juspence: The get_upstreams logic is very complicated, and I'd worry this change would make the code harder to maintain or maybe introduce a regression. I think the better approach, for users who want upstreams, is to:
1) Find some child component they're interested in
2) Find the child component's root component
3) Find the root component's upstream component(s)

This doesn't have to be something the users do directly / themselves, e.g. I think translations like this are why we decided to build Griffon.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=root_node,
+        purl=upstream_comp.purl,
+        defaults={"obj": upstream_comp},
+    )
+
+    # create dep component
+    dep_comp = ComponentFactory(
+        name="cool_dep_component",
+        arch="src",
```
juspence, line 852: For a root component that is a source RPM, the child components are all binary RPMs, where the arch value is something like "aarch64" or "x86_64".
```suggestion
        arch="aarch64"
```

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    )
+    # also make it its own top level root component
+    dep_node, _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        purl=dep_comp.purl,
+        defaults={"obj": dep_comp},
+    )
+
+    # create 2nd level dep
```
juspence, line 863: I don't think SRPMs ever have 2nd level dependencies. If you want to test these, you might try creating:
1) A root-level component for an index container image, with type=OCI, arch=noarch, and namespace=REDHAT
2) First-level dependencies like an architecture-specific container (type=OCI, arch=aarch64, namespace=REDHAT)
3) Second-level dependencies like binary RPMs (type=RPM, arch=aarch64, namespace=REDHAT) and remote-source components (type=PYPI, arch=noarch, namespace=UPSTREAM)

Both the first- and second-level dependencies should appear in the root component's "provides" field. Only the second-level dependencies should appear in the first-level dependencies "provides" field.

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+
+    # create dep component
+    dep_comp = ComponentFactory(
+        name="cool_dep_component",
+        arch="src",
+        type=Component.Type.RPM,
+        namespace=Component.Namespace.REDHAT,
+    )
+    # make it a child of root component
+    ComponentNode.objects.get_or_create(
```
juspence:
```suggestion
    dep_node = ComponentNode.objects.get_or_create(
```

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+        defaults={"obj": dep_comp},
+    )
+    # also make it its own top level root component
+    dep_node, _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        purl=dep_comp.purl,
+        defaults={"obj": dep_comp},
+    )
+
```
juspence: Our code doesn't do this today.
```suggestion
```

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    dep2_comp.save_component_taxonomy()
+    dep2_comp.save()
+    dep_comp.save_component_taxonomy()
+    dep_comp.save()
+    root_comp.save_component_taxonomy()
+    root_comp.save()
+
+    assert dep_comp.purl in root_comp.provides.values_list("purl", flat=True)
+    # why does root component not list this dependency ?
+    assert dep2_comp.purl in root_comp.provides.values_list("purl", flat=True)
```
juspence: I think because you created dep2_comp's node as a child of dep_comp's root node, which doesn't match the real code. Instead, create dep2_comp's node as a child of dep_comp's node that is a child of the root component's root node. Don't create any root node for dep_comp.

juspence (changes requested):

tests/test_api.py
```diff
@@ -786,3 +786,160 @@ def test_product_components(client, api_path):
...
+    Are the following statements true ?
+
+    * upstream has 0 sources as it is its own SOURCE
+    * should dep1 or dep2 dynamically resolve the top level component's upstreams ?
+    * should dep2 list 2 sources ?
+    * we should not list upstreams in provides right ?
+
+    """
+
+    # create a top level root source component
```
juspence, line 820: Instead of creating all the components and nodes here manually, I think there's a helper method you can import from one of the other test modules that will do this for you. That should really get made into a fixture, or we could add ComponentNode creation to our existing ComponentFactory class.

tests/test_api.py
```diff
@@ -786,3 +786,163 @@ def test_product_components(client, api_path):
...
+    assert response.json()["related_url"] == root_comp.related_url
+    assert len(response.json()["sources"]) == 0
+    assert len(response.json()["provides"]) == 2
+    assert len(response.json()["upstreams"]) == 1
+
+    response = client.get(f"{api_path}/components/{dep_comp.uuid}")
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 1
+    assert len(response.json()["provides"]) == 1
+    assert len(response.json()["upstreams"]) == 1
```
juspence: I think this was correct in the first version I reviewed, unless I've misunderstood. Today, only the root component reports an upstream component.
```suggestion
    assert len(response.json()["upstreams"]) == 0
```

tests/test_api.py
```diff
@@ -786,3 +786,163 @@ def test_product_components(client, api_path):
...
+
+    response = client.get(f"{api_path}/components/{dep2_comp.uuid}")
+    assert response.status_code == 200
+    assert len(response.json()["sources"]) == 2
+    assert len(response.json()["provides"]) == 0
+    assert len(response.json()["upstreams"]) == 1
```
juspence: I think this was correct in the first version I reviewed, unless I've misunderstood. Today, only the root component reports an upstream component.
```suggestion
    assert len(response.json()["upstreams"]) == 0
```

tests/test_api.py
```diff
@@ -786,3 +786,163 @@ def test_product_components(client, api_path):
...
+    assert response.json()["count"] == 2
+    # retrieve all provides of dep2_comp
+    response = client.get(f"{api_path}/components?sources={dep2_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 0
+
+    # retrieve all components with upstream_comp upstream
+    response = client.get(f"{api_path}/components?upstreams={upstream_comp.purl}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 3
```
juspence, line 1100: I think this was correct in the first version I reviewed, unless I've misunderstood. Today, only the root component reports an upstream component.
```suggestion
    assert response.json()["count"] == 1
```

JimFuller-RedHat: yes just investigating why  ...

tests/test_api.py
JimFuller-RedHat: good idea will add this as a separate test

juspence: The test here is already creating these second-level dependencies. I think it should be updated to only create first-level dependencies that are appropriate for a source RPM, OR we can make the root component an index container image instead of a source RPM.

I don't think we need a second test, especially since a lot of this logic is already being tested elsewhere.

tests/test_api.py
```diff
@@ -786,3 +786,159 @@ def test_product_components(client, api_path):
...
+    | upstream_component | 0       | 0        | 0         |
+    +--------------------+---------+----------+-----------+
+    | dep1_component     | 1       | 1        | 1         |
+    +--------------------+---------+----------+-----------+
+    | dep2_component     | 2       | 0        | 1         |
+    +--------------------+---------+----------+-----------+
+
+    This behaviour is predicated on the following assumptions:
+    * SRPM -> arch RPM
+    * binary RPM never have child dependencies
```
juspence: The first version I reviewed was more accurate. The table above describes a root component with two levels of dependencies, but this behavior is only seen for index container images and not for source RPMs.

These statements could get reverted to their previous version, or the table could be updated to display relationships for a source RPM:
root_component with 0 sources, 2 provides, and 1 upstreams
upstream_component with 0 sources, provides, and upstreams
dep1_component with 1 sources, 0 provides, and 1 upstreams
dep2_component, a sibling of dep1_component, with 1 sources, 0 provides, and 1 upstreams

tests/test_api.py
JimFuller-RedHat: this test is not specifically about adding coverage - it is about having a single standalone test to ensure we all share common understanding ... I will make a 2nd test with OCI example

tests/test_api.py
JimFuller-RedHat: but we want a single standalone test for comprehension ... so will do that separately

tests/test_api.py
jasinner: Agree we should probably just rename the components to match the Container structure.

Although a second test which outlines the RPM structure may also be informative. It would be the same as James provides here, just without the 2nd level of provided dependency.

jasinner (approved):

tests/test_api.py
JimFuller-RedHat: done

tests/test_api.py
JimFuller-RedHat: added 2nd test

JimFuller-RedHat:
We might still have a problem (conceptual or bug) ...if we resolve directly from corgi stage a source RPM

```
https://corgi-stage.prodsec.redhat.com/api/v1/components?purl=pkg%3Arpm%2Fredhat%2F389-admin%401.1.42-1.el7dsrv%3Farch%3Dsrc`
```
this lists pkg:rpm/389-admin@1.1.42 as its upstream.

Peeking into any of the purls listed in its **provides**, for example: 
```
https://corgi-stage.prodsec.redhat.com/api/v1/components?purl=pkg%3Arpm%2Fredhat%2F389-admin%401.1.42-1.el7dsrv%3Farch%3Ds390x
```
also lists pkg:rpm/389-admin@1.1.42 as its upstream.

**Note** - in srpm test version I commented out the various .save_component_taxonomy() and save() calls ... removing comments shows de facto behaviour

This implies that our above tables should not list such things ... and also querying why we do not have parity of behaviour between OCI noarch and RPM src.

tests/test_api.py
```diff
@@ -786,3 +786,315 @@ def test_product_components(client, api_path):
...
+    assert response.json()["count"] == 0
+
+    # retrieve all components with upstream_comp upstream
+    response = client.get(f"{api_path}/components?upstreams={quote(upstream_comp.purl)}")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_srpm_component_provides_sources_upstreams(client, api_path):
```
juspence, line 965: This test looks good, thanks for cleaning up the notes / assumptions!

tests/test_api.py
```diff
@@ -786,3 +786,315 @@ def test_product_components(client, api_path):
     response = client.get(f"{api_path}/components?products={rhel_br.name}")
     assert response.status_code == 200
     assert response.json()["count"] == 1
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_oci_component_provides_sources_upstreams(client, api_path):
```
juspence, line 792: This test looks right now, thank you for fixing!

juspence (approved):
Since this a draft PR I think the pipeline will not run, but the code you've written matches my understanding of how it's supposed to work. If this passes feel free to merge.

JimFuller-RedHat:
thx @juspence for clarification - I agree with you that this is good behaviour with RPM src  ... just wanted to canonize it in a test. ;)

PR #367, opened 2023-03-27
Fix unneeded related object lookups

juspence:
@RedHatProductSecurity/corgi-devs While reviewing some other code, I saw we fetch every upstream ComponentNode's related Component to get only its primary key, which requires one extra query per node.

We can get this Component UUID value directly from the upstream ComponentNode's object_id field without any extra queries. We also don't need to check "if node.obj" or "if node.object_id" since nodes are never created without a linked object, and deleting the linked object also deletes the node.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1778,7 +1778,7 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
 
     def get_upstreams_pks(self, using: str = "read_only") -> tuple[str, ...]:
         """Return only the primary keys from the set of all upstream nodes"""
-        linked_pks = set(node.obj.pk for node in self.get_upstreams_nodes(using=using) if node.obj)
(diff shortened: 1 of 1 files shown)


kgrant-rh (approved):

juspence:
Per discussion in Gchat, I tried to change the type hint for get_upstreams_pks() so that converting uuid.UUID to str wasn't needed. But the name of the "uuid" package that we import conflicts with the name of the "uuid" field that we define on the Component model. I'm also not sure whether passing those raw UUID values directly to save_component_taxonomy() will cause issues.

So for now I'll just keep the type hint unchanged and explicitly cast everything to strings. I think in the old code, this was being done for us automatically when retrieving the Component.uuid value from the UUIDField. Now that we're pulling it from ComponentNode.object_id it seems like this conversion doesn't happen automatically anymore, maybe as a side-effect of being a GenericForeignKey?

PR #369, opened 2023-03-28
CORGI-527 json decode error

jasinner:
Fixes various JSONDecodeErrors in manifest generation:

- Newline in component name
- Trailing backslash in PYPI component version
- Special characters in component copyright_text

Where possible I've used escapejs template tag to prevent future occurances of this issue from breaking manifest generation. However we can't keep newline characters in a component name as that field is used in the documentNamespace field which has to be on a single line.


jasinner:
@RedHatProductSecurity/corgi-devs please review. This also come with a couple of data migration scripts attached to CORGI-527.

jasinner:
I got one JSONDecodeError for a devtools stream in my dev environment after applying the changes from this PR and data migrations. See CORGI-542

tests/test_manifests.py
```diff
@@ -31,6 +31,34 @@
 ]
 
 
+def test_escapejs_on_copyright():
+    # actual copyright_text from snappy rpm
+    c = ComponentFactory(
```
kgrant-rh, line 36: So it was the HTML `&#x27;` type characters that caused the problem? Should we test characters like ©, 🄯, ®, ™?

tests/test_manifests.py
```diff
@@ -31,6 +31,34 @@
...
+        " Inc., Copyright 2011, Google Inc., Copyright 2011 Martin Gieseking &lt;mar"
+        "tin.gieseking@uos.de&gt;, Copyright 2013 Steinar H. Gunderson, Copyright 20"
+        "19 Google Inc., copyright by Cornell University, Copyright Depository of El"
+        "ectronic Materials, Copyright Issues Marybeth Peters, (c) ^P u E2OroCIyV ^T"
+        " C.au, (c) UiL (c)"
+    )
+    try:
+        c.manifest
+    except JSONDecodeError as e:
+        assert False, f"JSONDecoreError thrown by component with special chars in copywrite_txt {e}"
```
kgrant-rh:
```suggestion
        assert False, f"JSONDecodeError thrown by component with special chars in copywrite_txt {e}"
```

kgrant-rh (approved):

tests/test_sca.py
```diff
@@ -500,3 +502,18 @@ def test_save_component_skips_duplicates():
...
+    results = Syft.scan_files([test_archive])
+    for result in results:
+        if "subproject1" in result["meta"]["name"]:
+            assert False
+
+
+def test_remove_whitespace_from_name():
+    test_pom = Path("tests/data/resteasy-1.2.1.GA_CP02_patch02")
+    results = Syft.scan_files([test_pom])
+    assert "spring-hibernate-contacts-persistence" in [result["meta"]["name"] for result in results]
```
juspence: This asserts that the value here is present, but should we also assert that " " spaces or \n newlines are not present?

tests/test_manifests.py
```diff
@@ -41,6 +69,22 @@ def test_latest_components_exclude_source_container():
...
+    sb = SoftwareBuildFactory(
+        completion_time=datetime.strptime("2017-03-29 12:13:29 GMT+0000", "%Y-%m-%d %H:%M:%S %Z%z")
+    )
+    component = SrpmComponentFactory(version="2.8.0 \\", software_build=sb)
+    component.productstreams.add(stream)
+
+    try:
+        stream.manifest
+    except JSONDecodeError:
+        assert False
```
juspence, line 89: Not a problem but just something I don't understand. Why do we catch an exception and "handle" it by asserting False? Can't we just remove the try-block and let the exception get raised normally to make the test fail?

tests/test_manifests.py
```diff
@@ -31,6 +31,34 @@
...
+    except JSONDecodeError as e:
+        assert False, f"JSONDecoreError thrown by component with special chars in copywrite_txt {e}"
+
+
+def test_escapejs_on_version():
+    c = ComponentFactory(type="RPM", name="dyninst", version="12.1.0", release="1.el8", arch="src")
+    manifest = json.loads(c.manifest)
+    assert manifest["packages"][0]["versionInfo"] == "dyninst-12.1.0-1.el8.src"
+
+
```
juspence: I don't understand what this test is doing. There's no special characters here (that I can see without my glasses), and we already test backslash / other escaping below.

corgi/collectors/syft.py
```diff
@@ -119,7 +120,7 @@ def parse_components(cls, syft_json: str) -> list[dict[str, Any]]:
                 typed_component: dict[str, Any] = {
                     "type": pkg_type,
                     "meta": {
-                        "name": artifact["name"],
+                        "name": artifact["name"].strip(),
```
juspence, line 123: Are there any other places we need to sanitize the component names? Or is this just a bug with the Syft data and never with Brew?

corgi/tasks/management/commands/generatemanifests.py
```diff
@@ -0,0 +1,23 @@
+from django.core.management.base import BaseCommand, CommandParser
+
+from corgi.tasks.manifest import cpu_update_ps_manifest, update_manifests
+
+
+class Command(BaseCommand):
+    help = "Fetch component data from Brew for a specific build or tag."
```
juspence: Might want to update the help text here.

corgi/web/templates/component_manifest.json
```diff
@@ -41,7 +41,7 @@
         "packageFileName": {% if obj.filename %}"{{obj.filename}}"{% else %}"NOASSERTION"{% endif %},
         "SPDXID": "SPDXRef-{{obj.uuid}}",
         "supplier": "Organization: Red Hat",
-        "versionInfo": "{{obj.version}}"
+        "versionInfo": "{{obj.nevra|escapejs}}"
```
juspence, line 44: Thanks for fixing this - it was a copy-paste error from v1 to v2 of our templates.

corgi/web/templates/component_manifest.json
```diff
@@ -1,6 +1,6 @@
 {# Component manifest, obj is a Component here #}{% extends "base_manifest.json" %}
 {% block packages %}{% for component in obj.provides_queryset %}{
-        "copyrightText": {% if component.copyright_text %}"{{component.copyright_text}}"{% else %}"NOASSERTION"{% endif %},
+        "copyrightText": {% if component.copyright_text %}"{{component.copyright_text|escapejs}}"{% else %}"NOASSERTION"{% endif %},
```
juspence, line 3: I think we also need to escape obj.copyright_text toward the bottom of this template. I can't comment on that line directly since it's outside the diff window.

Finally, I realize name is part of the documentNamespace which has to be a URI. But I still think we should add escaping to all uses of component.name, obj.name, and provided.name anyway. This will prevent issues with characters that are valid in URIs, but which are invalid in JSON and break the template rendering.

I'm not proposing we **only** escape the names. We do need to clean up the data like you mention in the ticket. But I think we should **also** add this escaping to prevent other, similar bugs in the future, which aren't related to newlines or whitespace.

juspence (approved):

tests/test_manifests.py
jasinner: I'm not sure exactly which character it was. I assumed there were some characters from other non-English language here. I can add tests for those other characters.

tests/test_manifests.py
jasinner: Updated, thanks.

tests/test_manifests.py
jasinner: Not sure with pytest, but in other testing frameworks it's not valid to have a test without an assert statement. In any case I think it reads better this way than as you suggest, as it makes it explicit what we're testing for.

corgi/collectors/syft.py
jasinner: Our data shows us that that we only have newlines in component names coming from Syft. I think we should just address other data bugs as they come up in order to make sure that we're focusing our effort where required, and not creating our own problems.

corgi/tasks/management/commands/generatemanifests.py
jasinner: updated, thanks.

corgi/web/templates/component_manifest.json
jasinner: All good. It had me very confused for a while :)

tests/test_manifests.py
jasinner: I think I added before I realized that we have `obj.version` in the content_manifest template, not `obj.nevra`. I'll remove it now as it's confusing.

corgi/collectors/syft.py
juspence: Ack, not trying to put more work on your plate if it's not needed. I'm just less familiar with this code and couldn't remember if there were other codepaths that might hit this bug. I didn't want something necessary to get missed in this MR if we could avoid it.

tests/test_sca.py
jasinner: Good idea. I'll adjust it.

corgi/web/templates/component_manifest.json
jasinner: Ok, added escapejs to all instances of component.name, obj.name and provided.name in the manifest templates.

jasinner: Also added the other escapejs to obj.copyright_text in the component_manifest.json

PR #370, opened 2023-03-28
CORGI-536 - change sources, provides and upstreams to do exact match

JimFuller-RedHat:
Previously we matched using **icontains** for sources, provides and upstreams url param filters - which resulted in very poor performance as django ORM generated SQL with **LIKE** (with %target%) - these expressions can only be indexed using trigram/full text search. 

We now have **exact** matching for the following endpoint url filter params:
```
api/v1/components?sources={purl}       # return all components provided by this purl
api/v1/components?provides={purl}      # return all sources that use this purl
api/v1/components?upstreams={purl}   # return all components with this upstream purl
```
Much quicker (as we are hitting an index to resolve) and enables reasonable performance (ex. downstream griffon).

In addition, we provide re_ variants for regex search:
```
api/v1/components?re_sources={}
api/v1/components?re_provides={}
api/v1/components?re_upstreams={}
```
which by definition will run slower (presumably by someone who knows that is what they want to do).


tests/test_api.py
```diff
@@ -786,3 +786,158 @@ def test_product_components(client, api_path):
     response = client.get(f"{api_path}/components?products={rhel_br.name}")
     assert response.status_code == 200
     assert response.json()["count"] == 1
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_srpm_component_url_params_provides_sources_upstreams(client, api_path):
```
juspence: Can we merge the other MR first and just rely on the test added there?

juspence (approved):
LGTM, let's just make sure we note the API changes in the release. I still don't think this will solve the performance problem in CORGI-536, though. Repeating some things I've noted in the ticket:

The way our serializers work is that for each Component in the list of results, we have to make one separate query per ManyToMany related model field. This includes provides / sources / upstreams as well as products / versions / streams / variants / channels. Using ?include_fields or ?exclude_fields to leave above fields out of the results avoids these queries.

Finally, the query that has this 502 Bad Gateway error has ?limit set to 120 and ?include_fields contains versions / streams / sources / upstreams. So we'll end up making 480 (120 * 4) queries total. Even if we fix the performance issue with SQL %LIKE% queries, we'll still potentially overload the DB with above.

I think an easier fix is to set ?iimit to a lower value or exclude more fields from the query. I don't think we can guarantee that every API request will succeed without any errors. Once ?limit is set very high (over 100), we risk timeouts just due to the volume of data being fetched from the DB. Avoiding this is why we have the limit parameter (and why it defaults to only 10) in the first place.

JimFuller-RedHat:
some weird git problems with this PR (most likely as I had a previous PR) ... closing this and leaving link in https://github.com/RedHatProductSecurity/component-registry/pull/371

PR #373, opened 2023-03-30
CORGI-341: Fix empty manifest properties

juspence:
@RedHatProductSecurity/corgi-devs Please review a fix for some empty manifest properties, as well as removing internal URLs from our manifests. OpenLCS confirmed they don't need the internal download_url for containers anymore (after some process changes), so we can use a public URL instead now.


jasinner (approved):
LGTM

kgrant-rh (approved):

juspence:
Made a small tweak here to fix an issue with noarch / root containers not having their labels set. Going to merge once the pipeline passes.

PR #376, opened 2023-03-31
CORGI-558, CORGI-559: Add missing filters

juspence:
@jasinner Let's discuss the approach since I'm not sure this is possible. We want to filter for only go-package or gomod type components, using the go_component_type key we set in the meta_attr. But it looks like this key isn't present if:

1) We're creating a golang component as a source of some container
2) We're creating a golang component discovered by Syft
3) Maybe other cases as well

As a side-effect of how I implemented the filter, setting the filter to "YES" means only showing gomod components. go-packages and golang components with an unknown type will be hidden.

Setting the filter to "NO" means only showing go-package components. gomod components and golang components with an unknown type will be hidden.

The golang components with an unknown type will only be shown when the user doesn't click anything in the filters box, or when the user gives a URI like ```?type=GOLANG&gomod_components=``` with an empty gomod_components parameter.

I could adjust this to always treat a golang component with unknown type as a gomod component (or a go-package). Is this worth pursuing further, or should we close this as CANTDO? Or maybe you know of some way to determine whether these unknown golang components are really gomod components or go-packages.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -398,7 +398,10 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
                 version="",
                 release="",
                 arch="noarch",
-                defaults={"namespace": Component.Namespace.UPSTREAM},
+                defaults={
+                    "mets_attr": {"go_component_type": "gomod"},
+                    "namespace": Component.Namespace.UPSTREAM,
+                },
             )
             ComponentNode.objects.get_or_create(
                 type=ComponentNode.ComponentNodeType.SOURCE,
@@ -453,6 +456,9 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
                 # The public repo we want is just github.com/openshift/cluster-api
                 related_url = related_url.replace("openshift-priv", "openshift")
 
+            if source["type"] == Component.Type.GOLANG:
+                # Assume upstream container sources are always go modules, never go-packages
+                source["meta"]["go_component_type"] = "gomod"
             new_upstream, created = Component.objects.update_or_create(
                 type=source["type"],
                 name=component_name,
--- a/corgi/tasks/sca.py
+++ b/corgi/tasks/sca.py
@@ -71,7 +71,9 @@ def find_duplicate_component(meta_name: str, syft_purl: str) -> Component:
         raise ValueError(f"New edge case for duplicate component: {syft_purl}")
 
 
-def save_component(component: dict[str, Any], parent: ComponentNode) -> bool:
+def save_component(
+    component: dict[str, Any], parent: ComponentNode, is_go_package: bool = False
+) -> bool:
     meta = component.get("meta", {})
     if component["type"] not in Component.Type:
         logger.warning("Tried to save component with unknown type: %s", component["type"])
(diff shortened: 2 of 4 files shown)


jasinner:
I think it's safe to assume that a container source is always a go module. Also, Syft [doesn't support detection of go packages](https://github.com/anchore/syft/issues/432), so we could improve the Syft collector to always set the type to that.

However, most of the GOLANG components without go-component-type meta_attr field set are from rpm spectfiles:

```
>>> m.Component.objects.filter(type="GOLANG",meta_attr__contains={"source":["specfile"]}).exclude(meta_attr__has_key="go_component_type").count()
12182
>>> m.Component.objects.filter(type="GOLANG").exclude(meta_attr__has_key="go_component_type").count()
12469
```

I'm not sure if we're ever going to be able to tell for sure if they are modules or packages. Perhaps we should guess and set them to modules if the len of path elements is 2, as in 'k8s.io/metrics/', versus 'k8s.io/metrics/pkg' for a package. However, given the low number of these, and the poor data quality of go packages detected by specfiles I would suggest we just exclude them from the results.

corgi/tasks/brew.py
```diff
@@ -460,6 +463,8 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
                 release="",
                 arch="noarch",
                 defaults={
+                    # TODO: This can be a GOLANG component
```
jasinner: Let's check if the type is golang and set meta_attr.go_component_type to gomod.

corgi/tasks/sca.py
```diff
@@ -95,6 +95,8 @@ def save_component(component: dict[str, Any], parent: ComponentNode) -> bool:
             release="",
             arch="noarch",
             defaults={
+                # TODO: This can be a GOLANG component
```
jasinner: Always set the go_component_type to gomod

jasinner (approved):
Approve once the suggested changes are in place.

juspence:
> However, given the low number of these, and the poor data quality of go packages detected by specfiles I would suggest we just exclude them from the results.

Ack, that's how the filter is currently working. When the user makes a choice (YES or NO in the UI), we only show either gomod or go-package components, respectively. We exclude these Golang components with an unknown type whenever the new Golang component type filter is being used. Users only see these unknown Golang components if they aren't using the new filter at all.

Made above changes so going to merge once the pipeline passes.

corgi/tasks/sca.py
```diff
@@ -199,11 +203,13 @@ def _scan_files(anchor_node, sources) -> int:
     # get go version from container meta_attr
     go_packages = GoList.scan_files(sources)
     _assign_go_stdlib_version(anchor_node.obj, go_packages)
-    detected_components.extend(go_packages)
 
     for component in detected_components:
         if save_component(component, anchor_node):
             new_components += 1
+    for package in go_packages:
```
juspence, line 212: @jasinner The first version you reviewed made the pipeline fail, because not all Golang components in "save_component()" are go modules / detected by Syft. I tweaked it again to fix this problem.

Does "go list" only detect go packages and never go modules? If yes, I think this should work. If "go list" can sometimes find go modules as well, then we might incorrectly report them as go packages.

jasinner: Looking at the go_list collector source and test data, I think it only detects packages, never modules.

PR #378, opened 2023-03-31
CORGI-504: App interface lookup

mprpic:
This interface uses prod-def metadata to extract component composition of
every managed service by listing its source and container image repos.
This metadata can then be used to manifest each component of each service.

### DIFF
--- a/corgi/collectors/app_interface.py
+++ b/corgi/collectors/app_interface.py
@@ -1,0 +1,101 @@
+import logging
+from collections import defaultdict
+
+import requests
+from django.conf import settings
+
+from corgi.core.models import ProductStream
+
+logger = logging.getLogger(__name__)
+
+
+class AppInterface:
+    @classmethod
+    def fetch_service_metadata(cls) -> dict[ProductStream, list[dict]]:
+        repo_query = """
+        {
+            apps_v1 {
+                path
+                name
+                quayRepos {
+                    org {
+                        name
+                        instance {
+                            url
+                        }
+                    }
+                    items {
+                        name
+                        public
+                        description
+                    }
+                }
+                codeComponents {
+                    name
+                    url
+                    resource
+                }
+            }
+        }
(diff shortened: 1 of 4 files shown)


mprpic:

corgi/collectors/app_interface.py
```diff
@@ -0,0 +1,101 @@
...
+            component_name = component["name"]
+            subcomponent_data: defaultdict = defaultdict(dict)
+
+            quay_repos_data = component.get("quayRepos")
+            if quay_repos_data:
+                for org_repos in quay_repos_data:
+                    # Ignore container images in other registries since we assume all images are
+                    # in Quay.io.
+                    if org_repos["org"]["instance"]["url"] != "quay.io":
+                        logger.info(
```
juspence: Should this be a warning?

juspence (approved):
My understanding of this code is very shallow. It might help to have a docstring for the class and the method explaining what they're doing and how this is used. Or maybe a test with sample data would give a better idea of what we're expecting to receive, do, and return. But overall LGTM.

jasinner:
> My understanding of this code is very shallow. It might help to have a docstring for the class and the method explaining what they're doing and how this is used. Or maybe a test with sample data would give a better idea of what we're expecting to receive, do, and return. But overall LGTM.

I agree, it would be nice to have some tests with sample data.

corgi/collectors/app_interface.py
mprpic: Sure, why not. Fixed!

mprpic:
@jasinner @juspence Ack on the test. I'll add one in the next MR (CORGI-505) that will add the Celery task to actually fetch the manifesting data, so that I don't have to bother mocking the ProductStream calls in the test specific to this task. There isn't much to test here though, it's just shuffling data around though I guess there is value in having an example response somewhere in the tests.

PR #379, opened 2023-04-03
CORGI-300 remove schema from runsql migrations

jasinner:
There is a single Postgres DB shared by enterprise and community. They are separated by schema. In these migrations we are including the schema 'public' in the create index commands which only works in the enterprise instance. These migrations fail in community because the schema has a different name. 

Removing the schema has no affect on enterprise, but allows community to deploy correctly.


jasinner:
I verified no changes to migrations are detected with this change in stage (enterprise). Also, the migration does pass in community stage.

PR #382, opened 2023-04-04
Fix community corgi checks

juspence:
@RedHatProductSecurity/corgi-devs Quick review please, this should fix a writable API endpoint on the stage community version. Both community and enterprise use the same DJANGO_SETTINGS_MODULE (whether stage or prod), so our existing running_prod() check isn't enough and we also need to check for COMMUNITY_MODE_ENABLED.

I also tweaked some settings to avoid a too-permissive Host: header check and avoid duplicating some logic. The enterprise version should only allow connections that specify an enterprise Host: header, and vice-versa for the community version.

I'm not sure if this is exploitable or not, but the Django docs say this is needed to prevent CSRF attacks and other manipulations of the Host: header:
https://docs.djangoproject.com/en/4.1/ref/settings/#allowed-hosts

I'm not sure how to actually deploy this - manually into the community namespace? Maybe we should add a step to the CI pipeline to deploy the community resources, to keep everything in sync with the enterprise resources.

We should also make sure that CORGI_DOMAIN is set correctly for the community version. Currently it's set to an internal URL. This makes the web pod generate API links with that internal URL.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -28,7 +28,7 @@ DOCS_URL = os.getenv("CORGI_DOCS_URL")
 OPENSHIFT_BUILD_COMMIT = os.getenv("OPENSHIFT_BUILD_COMMIT")
 PRODSEC_EMAIL = os.getenv("PRODSEC_EMAIL", "dev@example.com")
 
-CORGI_DOMAIN = os.getenv("CORGI_DOMAIN")
+CORGI_DOMAIN = os.getenv("CORGI_DOMAIN", "")
 if CORGI_DOMAIN:
     CSRF_COOKIE_DOMAIN = CORGI_DOMAIN
     LANGUAGE_COOKIE_DOMAIN = CORGI_DOMAIN
@@ -383,8 +383,12 @@ UMB_BROKER_URL = os.getenv("CORGI_UMB_BROKER_URL")
 # https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool
 UMB_BREW_MONITOR_ENABLED = strtobool(os.getenv("CORGI_UMB_BREW_MONITOR_ENABLED", "true"))
 
-# Set to True to turn on loading of community products from product-definitions.
 COMMUNITY_MODE_ENABLED = strtobool(os.getenv("CORGI_COMMUNITY_MODE_ENABLED", "false"))
+
+if CORGI_DOMAIN.endswith(".fedoraproject.org"):
+    CORGI_DOMAIN_BASE = ".fedoraproject.org"
+else:
+    CORGI_DOMAIN_BASE = ".prodsec.redhat.com"
 
 # Brew
 BREW_URL = os.getenv("CORGI_BREW_URL")
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -3,6 +3,7 @@ import logging
 from typing import Type, Union
 
 import django_filters.rest_framework
+from django.conf import settings
 from django.db import connections
 from django.db.models import QuerySet, Value
 from django.http import Http404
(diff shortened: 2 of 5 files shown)


corgi/api/views.py
```diff
@@ -596,7 +597,7 @@ def olcs_test(self, request: Request, uuid: Union[str, None] = None) -> Response
         """Allow OpenLCS to upload copyright text / license scan results for a component"""
         # In the future these could be separate endpoints
         # For testing we'll just keep it under one endpoint
-        if utils.running_prod():
+        if utils.running_prod() or settings.COMMUNITY_MODE_ENABLED:
```
jasinner, line 600: Thanks for catching this!

config/settings/base.py
```diff
@@ -384,7 +384,12 @@
 UMB_BREW_MONITOR_ENABLED = strtobool(os.getenv("CORGI_UMB_BREW_MONITOR_ENABLED", "true"))
 
 # Set to True to turn on loading of community products from product-definitions.
-COMMUNITY_MODE_ENABLED = strtobool(os.getenv("CORGI_COMMUNITY_MODE_ENABLED", "false"))
+if CORGI_DOMAIN.endswith(".fedoraproject.org"):
+    COMMUNITY_MODE_ENABLED = True
+    CORGI_DOMAIN_BASE = ".fedoraproject.org"
+else:
+    COMMUNITY_MODE_ENABLED = False
+    CORGI_DOMAIN_BASE = ".redhat.com"
```
jasinner: Unfortunately we can't set the COMMUNITY_MODE_ENABLED flag in this check because stage still uses an internal URL. I didn't request an external cname and SSL cert for community stage at this time. We could optionally request that and have this change work as written. But for now, I think we should change it to just set the CORGI_DOMAIN_BASE setting.

```suggestion
COMMUNITY_MODE_ENABLED = strtobool(os.getenv("CORGI_COMMUNITY_MODE_ENABLED", "false"))

if CORGI_DOMAIN.endwith(".fedoraproject.org"):
    CORGI_DOMAIN_BASE = ".fedoraproject.org"
else:
    CORGI_DOMAIN_BASE = ".redhat.com"
    
```

jasinner (changes requested):
Let's just remove the check of domain before setting community_mode_enabled

config/settings/base.py
juspence: I was planning to set CORGI_DOMAIN myself to some placeholder URL during the deployment, like "corgi-stage-community-url-placeholder.fedoraproject.org". Without a CORGI_DOMAIN ending in .fedoraproject.org, the logic you suggest here for CORGI_DOMAIN_BASE won't work.

I'll just tweak this to keep COMMUNITY_MODE_ENABLED unchanged, then add ```".fedoraproject.org" if COMMUNITY_MODE_ENABLED else ".redhat.com"``` to our ALLOWED_HOSTS. Sound good?

juspence: Actually having an internal CORGI_DOMAIN will cause other issues. All the links the API generates are going to use this internal CORGI_DOMAIN.

@jasinner Can you please file that ticket to get an external cname and cert for community-stage? We can keep this code as-is, then merge and redeploy with an updated CORGI_DOMAIN env var once the cname is available.

juspence: Or if you'd prefer, you could open a new MR with just the first commit from this branch and merge that right away. Then we can merge the second commit only once we have the stage cname and cert set up (and I think this fix is less urgent anyway). I need to log off because it's very late - have a good day!

jasinner: The API is deployed on the internal CORGI_DOMAIN for community-stage instance, so I'm going to merge this updated as I suggested here. We can always migrate to an external host for community-stage if required later.

jasinner (approved):

PR #383, opened 2023-04-04
CORGI-313: Fix hopefully last bugs in ingest logic and simplify code

juspence:
@RedHatProductSecurity/corgi-devs Last MR for CORGI-313, then this can finally close. This fixes a few bugs:
1. Missing filename on all source RPMs
2. Duplicated values between component fields and meta_attr (might be more instances of this)
3. A typo I made when saving meta_attr for a container's upstream_go_modules

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -268,44 +268,36 @@ def save_component(
 
 
 def save_srpm(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentNode:
+    name = build_data["meta"].pop("name")
+    version = build_data["meta"].pop("version")
     related_url = build_data["meta"].pop("url", "")
     if not related_url:
         # Handle case when key is present but value is None
         related_url = ""
 
+    extra = {
+        "description": build_data["meta"].pop("description", ""),
+        "filename": find_package_file_name(build_data["meta"].pop("source_files", [])),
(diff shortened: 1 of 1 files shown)


corgi/tasks/brew.py
```diff
@@ -442,15 +377,9 @@ def save_container(softwarebuild: SoftwareBuild, build_data: dict) -> ComponentN
...
-                parent=root_node,
-                purl=obj.purl,
-                defaults={
-                    "object_id": obj.pk,
-                    "obj": obj,
-                },
-            )
+            image_arch_node = save_node(ComponentNode.ComponentNodeType.PROVIDES, root_node, obj)
+            # TODO: These are build-time dependencies, right?
+            #  Should we use PROVIDES_DEV here?
```
jasinner: This is partially true based on a conversation I had with the container factory team. All image components are build-time dependencies except for the ones from the base layer (ie UBI). That is because the other layers are part of multi-stage build and are discarded after the build. For example an image containing golang binaries often uses an intermediate layer to install the go compiler to build a go-binary, but that go compiler is no longer required at runtime, nor is it included in the final image.
I'm fine with leaving this as a TODO item for now, but perhaps update the comment with the extra information from this comment?

jasinner (approved):
Looks good, thanks.

PR #385, opened 2023-04-10
Test that components with many sources display quickly (CORGI-507)

juspence:
@RedHatProductSecurity/corgi-devs Please review this basic skeleton for performance testing. These performance tests should run automatically in our CI pipeline on any MR, but must be run manually in a local development environment, using ```tox -e corgi -- -m performance```.

This is because the performance tests run against a live environment (against stage in CI), to catch e.g. slow queries, missing indexes, API errors, and other data-dependent issues. If you want to run the performance tests against a local dev environment, you'll need to load certain data first so that the perf tests have something to measure. Also make sure that CORGI_DOMAIN is unset in your venv and .env file.


mprpic (approved):
LGTM! (after lint error is fixed)

PR #386, opened 2023-04-10
CORGI-524: Tweak test options to meet best practices

juspence:
@RedHatProductSecurity/corgi-devs Not urgent, please review only when you're not busy. There are a few changes here:

1. Remove some tests that require Syft to be installed locally. I think these tests are redundant given other tests (both ours and upstream's), but feedback appreciated.
2. Upgrade some dependencies to prepare for below
3. Report branch coverage and include more files, but skip reporting empty files, fully-covered files, and some boilerplate Web page templates that we don't test
4. When running the CI pipeline, make tests fail if coverage is below 70%. This change doesn't apply locally so that running only a few tests, like ```tox -e corgi -- tests/test_brew_collector.py```, continues to work instead of failing every time.
5. Fail the tests if any warnings occur, and fix a few minor ones
6. Make sure that output files / pre-generated manifests are placed in the correct path (vs. OpenShift PVC config)


tests/test_sca.py
```diff
@@ -504,22 +504,33 @@ def test_save_component_skips_duplicates():
...
-        assert "\n" not in name
-        assert " " not in name
+    try:
+        results = Syft.scan_files([test_pom])
+        result_names = [result["meta"]["name"] for result in results]
+        for name in result_names:
+            assert "\n" not in name
+            assert " " not in name
+    except FileNotFoundError:
+        pytest.xfail("Syft not installed locally")
```
juspence: This test is one of two that requires Syft to be installed locally. xfail() keeps the entire test suite from failing when everything else works fine but Syft is missing. But maybe we should let upstream take care of fixing / testing this bug for us.

tests/test_sca.py
```diff
@@ -504,22 +504,33 @@ def test_save_component_skips_duplicates():
...
-        if "subproject1" in result["meta"]["name"]:
-            assert False
-
-
+    try:
+        results = Syft.scan_files([test_archive])
+        for result in results:
+            assert "subproject1" not in result["meta"]["name"]
+    except FileNotFoundError:
+        assert "--exclude=**/src/test/resources/**" in Syft.SYFT_COMMAND_ARGS
```
juspence: This test is one of two that requires Syft to be installed locally. I think this is testing the same condition as above, unless I've misunderstood something. So we could remove this and let the test suite continue to work even without a local copy of Syft.

juspence:
Coverage check says "70%" on my machine, but "69.66%" in the CI pipeline. AKA this will start passing as soon as I figure out how to round this to the nearest percent.

EDIT: https://coverage.readthedocs.io/en/latest/config.html#report-precision says the default should already be 0 decimal places. It looks like --fail-under always uses 2 decimal places for its calculation even when we force precision to 0 explicitly.

So I'm going to punt and add more tests tomorrow to get our coverage over the 70% mark. I'll also make sure the displayed report percentage always matches the actual fail-under percentage, to avoid this kind of confusion. In the meantime this should be ready for review.

tests/test_sca.py
jasinner: Fine to remove it.

tests/test_sca.py
jasinner: It's fine to remove it.

jasinner (approved):
LGTM

PR #387, opened 2023-04-11
CORGI-542 escapejs on download and related urls

jasinner:
This adds escapejs on Django templates used to generate product stream and component manifests so that special characters in JSON do not break validation of manifests.

### DIFF
--- a/tests/test_manifests.py
+++ b/tests/test_manifests.py
@@ -73,7 +73,7 @@ def test_latest_components_exclude_source_container():
     assert not components.first().name.endswith("-container-source")
 
 
-def test_manifest_backslash():
+def test_stream_manifest_backslash():
     """Test that a tailing backslash in a purl doesn't break rendering"""
 
     stream = ProductStreamFactory()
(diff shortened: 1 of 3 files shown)


tests/test_manifests.py
```diff
@@ -89,6 +89,20 @@ def test_manifest_backslash():
...
+    component = ComponentFactory(
+        version="\\9.0.6.v20130930",
+        type=Component.Type.MAVEN,
+        name="org.eclipse.jetty/jetty-webapp",
+    )
+    assert component.download_url
+    try:
+        component.manifest
+    except JSONDecodeError:
+        assert False
```
juspence, line 103: I think it's better to just let component.manifest raise its own JSONDecodeError. That way the last line in the traceback / the error message says something about the invalid JSON (hopefully with a line number) instead of "assertion failed".

juspence (approved):
Not sure if this is ready for review (no @ corgi-dev) or if you were still working on more tests, but LGTM as-is.

tests/test_manifests.py
jasinner: Then it wouldn't be clear what we where testing.

PR #388, opened 2023-04-11
CORGI-551 always run collectstatic as startup

jasinner:
This updates the start_server script so that collectstatic is always run on startup. This prevents a 500 error when restarting the web application after all the files in the staticfiles directory have been deleted. The 500 error is caused by missing css and javascript files places in staticfiles for Django apps that we include such as drf_spectacular.

We now exclude *.json files as those take a long time to collect and process, which delays the startup of the web application on every restart. Those will be collected by the scheduled task.

### DIFF
--- a/corgi/tasks/manifest.py
+++ b/corgi/tasks/manifest.py
@@ -46,4 +46,4 @@ def cpu_update_ps_manifest(product_stream: str):
     retry_kwargs=RETRY_KWARGS,
 )
 def collect_static():
-    call_command("collectstatic", verbosity=0, interactive=False, ignore_patterns=[".gitignore"])
+    call_command("collectstatic", verbosity=1, interactive=False)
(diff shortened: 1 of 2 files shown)


mprpic (approved):
LGTM

This means the `collect_static()` task in `corgi.tasks.manifest` can probably be removed since the files will always exist when the web pod starts up.

jasinner:
> LGTM
> 
> This means the `collect_static()` task in `corgi.tasks.manifest` can probably be removed since the files will always exist when the web pod starts up.

That one is required to migrate all the *.json manifests into the static folder. Also to keep the static files up to date as new manifests are generated.

mprpic:
> > LGTM
> > This means the `collect_static()` task in `corgi.tasks.manifest` can probably be removed since the files will always exist when the web pod starts up.
> 
> That one is required to migrate all the *.json manifests into the static folder. Also to keep the static files up to date as new manifests are generated.

Ah, right, but if you set `WHITENOISE_AUTOREFRESH = True` then those files will be served dynamically so no need to collect periodically. I think I noted that in some Jira task few weeks back but I guess it never made it to the code.

jasinner:
> WHITENOISE_AUTOREFRESH

The documentation says for performance and security reasons it should not be used in production.

PR #389, opened 2023-04-11
CORGI-505: Manifest services

mprpic:
This adds manifesting of managed service components by analyzing their Quay images and Git repo sources (if either is available). A build is created for each generated manifest of each service (ProductStream). Because these builds are not mapped to an actual build system, they use timestamps as their "version" numbers to  be able to sort them by latest.


corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+            name=service_component["name"],
+            release="",
+            arch="",
+            defaults={
+                "version": faux_version,
+                "namespace": Component.Namespace.REDHAT,
+                "software_build": build,
+            },
+        )
+        root_component.productstreams.add(service)
```
mprpic: @jasinner @juspence I need some help here. What's the magic way to add a ProductStream to a component and then have it magically update the entire product taxonomy with the rest of the product-related models (Product, ProductVersion) based off of the Stream? I tried calling save_product_taxonomy on the Component but that complained it must be done on the build, but that doesn't seem to populate any data for all of the component of that build?

juspence: I think you might need to call ```service.save_product_taxonomy()``` to link the stream to the channels / variants / versions / products first.

I don't recall exactly how ```build.save_product_taxonomy()``` works, but I can try poking at this to see why the components aren't getting linked to that stream correctly.
EDIT: Oh wait, I think this relies on ProductComponentRelations like Brew tags, errata, or Yum repo metadata. And all of these come from the prod-defs data...do we have something similar for services?

Finally, you probably need to call ```component.save_component_taxonomy()``` in the for-loop below, after ```save_component()```, to link the child component to its root component.

I've also toyed with the idea of linking the components together directly (manually setting their related model fields) during creation. This would bypass the MPTT logic and the need to manually call taxonomy-saving methods, but I'm not sure if this is always possible. Let me know if you think this is worth researching.

corgi/tasks/monitoring.py
```diff
@@ -56,6 +56,7 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
         upsert_cron_task("manifest", "update_manifests", hour=4, minute=0)
         upsert_cron_task("manifest", "collect_static", hour=5, minute=0)
         upsert_cron_task("monitoring", "email_failed_tasks", hour=6, minute=0)
+        upsert_cron_task("managed_services", "refresh_service_manifests", hour=7, minute=0)
```
jasinner: I think this needs to be move into the 'else' section below. It's not requied in community_mode, only enterprise.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+        if git_repo:
+            component_data, source_ref = Syft.scan_git_repo(target_url=git_repo)
+            for scanned_component in component_data:
+                scanned_component["meta"]["git_repo"] = f"{git_repo}@{source_ref}"
+            analyzed_components.extend(component_data)
+
+        if not analyzed_components:
+            continue
+
+        faux_version = now.strftime("%Y%m%d%H%M%S")
```
jasinner: If possible, can we use an identifier or timestamp from app_interface such as the timestamp of the repo tag, or timestamp of the git commit instead of minting one on ingestion? 

This version will be used to determine the 'latest' root component in the product stream, and today we are using RPM version semantics to compare the versions. I'm not sure if that will work well with timestamps, but I suspect it will work OK. 

However I'm concerned that our latest version will be 'what we last ingested' instead of 'what was last deployed'. If they are always the same thing, this is probably fine.

jasinner (commented):
As discussed on chat we need to create ProductComponentRelations to link products and components in SoftwareBuild.save_product_taxonomy

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def refresh_service_manifests():
+    services = ProductStream.objects.filter(meta_attr__managed_service_components__isnull=False)
+    service_metadata = AppInterface.fetch_service_metadata(services)
+
+    for service, components in service_metadata.items():
+        manifest_service(service.pk, components)
+        manifest_service.delay(service.pk, components)
```
jasinner: We call the task twice? I think once is enough.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+@transaction.atomic
```
jasinner: I think we don't need to make the task atomic. It will slow down ingestion and I wouldn't expect a partial update to cause much trouble for data readers. We could just pick up the current data on the next run on the task if there was a partial update.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+    for service_component in service_components:
+        analyzed_components = []
+        quay_repo = service_component.get("quay_repo_name")
+        if quay_repo:
+            quay_repo_full = f"quay.io/{quay_repo}"
+            component_data, scan_source = Syft.scan_repo_image(target_image=quay_repo_full)
+            for scanned_component in component_data:
+                scanned_component["meta"][
+                    "quay_image"
+                ] = f"{quay_repo_full}@{scan_source['target']['imageID']}"
```
jasinner: I think we should use this data on the root component instead of component meta.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+            analyzed_components.extend(component_data)
+
+        git_repo = service_component.get("git_repo_url")
+        if git_repo:
+            component_data, source_ref = Syft.scan_git_repo(target_url=git_repo)
+            for scanned_component in component_data:
+                scanned_component["meta"]["git_repo"] = f"{git_repo}@{source_ref}"
```
jasinner: I think we should use this data on the root component instead of component meta.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        touched_build_ids.append(build.pk)
+
+        root_component, _ = Component.objects.update_or_create(
+            type=Component.Type.CONTAINER_IMAGE if quay_repo else Component.Type.GENERIC,
+            name=service_component["name"],
+            release="",
+            arch="",
+            defaults={
+                "version": faux_version,
```
jasinner: Let's use the container image or git repo coords here instead of consolidating everything under a single root component here. We will use the SoftwareBuild object to tie all these together. That suits better with our existing data model where users will expect the 'sources' list to show tangible artifacts not 'faked' ones.

corgi/tasks/monitoring.py
mprpic: Ah, duh. Fixed!

corgi/tasks/managed_services.py
mprpic: Oops, the previous one was commented out and used for testing locally, removed!

corgi/tasks/managed_services.py
mprpic: What we last ingested is as good as it gets since we only check periodically (thus the scheduled task). We don't have a great way of discovering newly deployed versions so daily checks is good enough (this is how the current manifesting through mangen works as well).

corgi/tasks/managed_services.py
mprpic: Ack!

corgi/tasks/managed_services.py
mprpic: I also originally thought about wiping out the previous runs data prior to loading a new set, but figured maybe we do want to keep a history of changes.

corgi/tasks/managed_services.py
mprpic: Is there a specific attribute you have in mind on the root component? Or `quay_image` in `meta` is fine?

corgi/tasks/managed_services.py
mprpic: Is there a specific attribute you have in mind on the root component? Or `git_repo` in `meta` is fine?

jasinner: If we are saving the same data in the root_component we don't need this meta attribute do we?

corgi/tasks/managed_services.py
jasinner: If we are saving the same data in the root_component we don't need this meta attribute do we?

corgi/tasks/managed_services.py
jasinner: Ok, let's make sure that when we call 'get_latest_components' for these product streams we get the latest root_components.

mprpic: I actually did add the deletion code for past builds because otherwise we'd have an ever-growing mess of components linked to many builds, but with get_or_create the past builds would simply be getting orphaned unless we made their versions unique every time somehow (even though in reality they may not have changed).

I guess this will solve itself once we transition to a properly tracked history against specific builds instead of this periodic refresh.

mprpic:
@juspence @jasinner Ok, I think I addressed all of the issues noted above. I put in the ProductComponentRelation object creation and called all the magic save_X_taxonomy() methods but the relations to products is still not generating for some reason (provides/sources does work, so that's good). Any ideas?

juspence:
> @juspence @jasinner Ok, I think I addressed all of the issues noted above. I put in the ProductComponentRelation object creation and called all the magic save_X_taxonomy() methods but the relations to products is still not generating for some reason (provides/sources does work, so that's good). Any ideas?

Quick thoughts without looking at code:
1. Provides / sources are set by component.save_component_taxonomy(), so that working is good
2. products / streams / etc. are set by build.save_product_taxonomy() to link a SoftwareBuild's components to some product-like things

So to fix 2, I'd check that you created the build / called .save() on it first and that you set build_type to BREW.

Next, check that you created the component / called .save() on it and that it's linked to the build. The component's software_build property should be set with the PK or object instance of the parent SoftwareBuild.

Finally, make sure you created the relation / called .save() on it, that the build ID matches the real build, and that the product ID is right.

Your new relation type needs to be either a stream type or variant type, as given in the ProductComponentRelation model (and make sure you add it to this list):
```
    # Below not defined in constants to avoid circular imports
    # ProductComponentRelation types which refer to ProductStreams
    STREAM_TYPES = (Type.BREW_TAG, Type.COMPOSE, Type.YUM_REPO)

    # ProductComponentRelation types which refer to ProductVariants
    VARIANT_TYPES = (Type.CDN_REPO, Type.ERRATA)
```
I think the product ID to use may depend on which type (stream or variant) you're using. Maybe you need to call fetch_unprocessed_relations? And definitely be sure you're calling build.save_product_taxonomy() last.

I'll look at actual code first thing tomorrow morning.

corgi/tasks/monitoring.py
```diff
@@ -76,6 +76,7 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
         upsert_cron_task("manifest", "update_manifests", hour=10, minute=0)
         upsert_cron_task("manifest", "collect_static", hour=11, minute=0)
         upsert_cron_task("monitoring", "email_failed_tasks", hour=11, minute=45)
+        upsert_cron_task("managed_services", "refresh_service_manifests", hour=12, minute=0)
```
juspence: Can we run the monitoring email at 12:45 now (after refresh_service_manifests), so that we don't have to wait until the next day to see failures in this task?

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+    ProductStream,
+    SoftwareBuild,
+)
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+from corgi.tasks.sca import save_component
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def refresh_service_manifests():
+    services = ProductStream.objects.filter(meta_attr__managed_service_components__isnull=False)
```
juspence, line 22: I wonder if this should be a boolean field on the ProductStream model, like "is_service" or "stream_type" (for future extensibility?) or something. I don't like long-term use of meta_attr stuff, but I guess it's not a huge deal.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+
+            ProductComponentRelation.objects.create(
+                product_ref=service.name,
+                build_id=build_id,
+                build_type=SoftwareBuild.Type.APP_INTERFACE,
+                type=ProductComponentRelation.Type.APP_INTERFACE,
+            )
+
+            build.save_product_taxonomy()
+            for component_obj in build.components.get_queryset():
```
juspence: You might want to use .iterator() here. to avoid a possible OoM if the number of components in a managed-service build is very large.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+            # the last time we looked. Historical data is not needed as of right now.
+            past_builds = SoftwareBuild.objects.filter(
+                name=service_component["name"],
+                build_type=SoftwareBuild.Type.APP_INTERFACE,
+                meta_attr__service=service.name,
+            )
+            for build in past_builds:
+                ProductComponentRelation.objects.filter(build_id=build.build_id).delete()
+                build.components.get_queryset().delete()
+                build.delete()
```
juspence, line 73: You don't need to call build.components.get_queryset().delete() here I think. The Component model's software_build ForeignKeyField has on_delete=CASCADE, so deleting the build should clean up all the components and related cnodes automatically.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+def refresh_service_manifests():
+    services = ProductStream.objects.filter(meta_attr__managed_service_components__isnull=False)
+    service_metadata = AppInterface.fetch_service_metadata(services)
+
+    for service, components in service_metadata.items():
+        manifest_service.delay(str(service.pk), components)
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def manifest_service(product_stream_id: str, service_components: list):
```
juspence: Let's run this task in the CPU queue since it needs to download and scan a potentially large Quay image / Git repo, and since there are many copies of it running at once:
```suggestion
def cpu_manifest_service(product_stream_id: str, service_components: list) -> None:
```

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+    ProductComponentRelation,
+    ProductStream,
+    SoftwareBuild,
+)
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+from corgi.tasks.sca import save_component
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def refresh_service_manifests():
```
juspence: Let's add the return type hint here so that mypy will check this method instead of ignoring it:
```suggestion
def refresh_service_manifests() -> None:
```

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+    services = ProductStream.objects.filter(meta_attr__managed_service_components__isnull=False)
+    service_metadata = AppInterface.fetch_service_metadata(services)
+
+    for service, components in service_metadata.items():
+        manifest_service.delay(str(service.pk), components)
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
```
juspence: Depending on how long this task takes / the number of components for each service, you may want to increase the timeout to settings.CELERY_LONGEST_SOFT_TIME_LIMIT or whatever it's called.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+            build = SoftwareBuild.objects.create(
+                name=service_component["name"],
+                build_type=SoftwareBuild.Type.APP_INTERFACE,
+                build_id=build_id,
+                completion_time=now,
+                meta_attr={"service": service.name},
+            )
+
+            root_component = Component.objects.create(
+                type=Component.Type.CONTAINER_IMAGE if quay_repo else Component.Type.GENERIC,
```
juspence: Are these managed service components ever Github repos, and can we tell if they are? If yes to both, we might sometimes want to use Component.Type.GITHUB instead of GENERIC.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+                name=service_component["name"],
+                version=component_version,
+                release="",
+                arch="",
```
juspence: Let's use "noarch" here please.

corgi/tasks/managed_services.py
```diff
@@ -0,0 +1,109 @@
...
+                namespace=Component.Namespace.REDHAT,
+                software_build=build,
+            )
+            root_component.productstreams.add(service)
```
juspence: Is this still needed, AKA is build.save_product_taxonomy() still not working? I'd rather fix the bug there instead of working around it. I can try to take a look if you're running out of time on this ticket.

I don't see where the ProductStream for a service is created, but it needs to have a ProductNode created and linked as well. Otherwise build.save_product_taxonomy() won't work, because get_product_details() won't find the related products / versions / variants.

juspence (approved):
Haven't reviewed tests, but LGTM so far aside from the issue with ProductStreams not being linked properly.

corgi/tasks/managed_services.py
mprpic: Ideally we'd have a separate product model for services but I think we need to re-do prod-defs for that, which is a longer term project.

corgi/tasks/managed_services.py
mprpic: Fixed!

corgi/tasks/managed_services.py
mprpic: Good call, switched it to GITHUB.

corgi/tasks/managed_services.py
mprpic: Ack, increased.

corgi/tasks/monitoring.py
mprpic: Sure!

corgi/tasks/managed_services.py
mprpic: You're right, not needed, forgot to remove it. Good catch!

jasinner (approved):
LGTM as long as component/product links are populated and CI passes.

corgi/tasks/monitoring.py
```diff
@@ -75,7 +75,8 @@ def upsert_interval_task(module, task, hours=None, minutes=None):
         upsert_cron_task("yum", "fetch_unprocessed_yum_relations", hour=9, minute=0)
         upsert_cron_task("manifest", "update_manifests", hour=10, minute=0)
         upsert_cron_task("manifest", "collect_static", hour=11, minute=0)
-        upsert_cron_task("monitoring", "email_failed_tasks", hour=11, minute=45)
+        upsert_cron_task("managed_services", "refresh_service_manifests", hour=11, minute=45)
+        upsert_cron_task("monitoring", "email_failed_tasks", hour=12, minute=45)
```
juspence: Sorry I didn't catch this earlier - if this managed services are just streams in prod-defs, we should probably refresh_service_manifests before we update_manifests. Otherwise the manifest we generate and serve might have stale data:
```suggestion
        upsert_cron_task("managed_services", "refresh_service_manifests", hour=10, minute=0)
        upsert_cron_task("manifest", "update_manifests", hour=11, minute=0)
        upsert_cron_task("manifest", "collect_static", hour=12, minute=0)
        upsert_cron_task("monitoring", "email_failed_tasks", hour=12, minute=45)
```

corgi/collectors/app_interface.py
```diff
@@ -11,7 +11,9 @@
 
 class AppInterface:
     @classmethod
-    def fetch_service_metadata(cls) -> dict[ProductStream, list[dict]]:
+    def fetch_service_metadata(
+        cls, services: list[ProductStream]
```
juspence, line 15: To make mypy pass, I think services should have type QuerySet[ProductStream], or Iterable[ProductStream] if you need to pass both lists and querysets:
```suggestion
        cls, services: QuerySet[ProductStream]
```

juspence (approved):

corgi/tasks/monitoring.py
mprpic: Ack, shuffled around!

corgi/collectors/app_interface.py
mprpic: I made it pass a list.

PR #390, opened 2023-04-13
CORGI-464: Clean up duplicate tests

juspence:
@RedHatProductSecurity/corgi-devs Per agreement in the recent testing ADR, we should remove tests that don't run automatically, in favor of OpenShift probes on our web pod. So this MR deletes the integration tests which only check API availability and never run anyway.

The existing code in the ops repo already configures periodic checks on our web pod, and will:
1. Roll back to a previous deployment, if the current deployment fails to bring up the web pods within 10 minutes (default behavior for OpenShift deployments).
2. Refuse to send web traffic / API requests to a container that isn't ready for it (readiness probes are configured)
3. Restart a container, whenever it becomes unresponsive or is unable to handle more requests (liveness probes are configured)

See https://developers.redhat.com/blog/2020/11/10/you-probably-need-liveness-and-readiness-probes for more info about available probe types in OpenShift.

I also added a test for manifest performance, but there might still be some slowness here. I'll handle that in a follow-up MR if needed.


juspence:
A performance test that was passing is now failing. Probably the size of the manifest increased after some recent updates.

juspence:
Going to merge since it's just changes to test code and we've already seen this new test catch real issues.

PR #391, opened 2023-04-13
Add OIDC authentication

kgrant-rh:
Adds basic infrastructure for OIDC authentication. See corgi.api.views.ControlledAccessTestView for an example of restricting a specific view to a specific role.


kgrant-rh:
I guess "sonarqube" has never heard of https://www.django-rest-framework.org/api-guide/views/#api_view !

juspence:
> I guess "sonarqube" has never heard of https://www.django-rest-framework.org/api-guide/views/#api_view !

Ah yes, this bug. You can mark those warnings as ignored, SonarQube only understands the Flask and native Django ways of doing this, not the Django-REST-Framework way.

config/settings/base.py
```diff
@@ -249,6 +251,7 @@
         "django": {"handlers": ["console"], "level": "WARNING"},
         # Mail errors only, but set level=WARNING here to pass warnings up to parent loggers
         "django.request": {"handlers": ["mail_admins"], "level": "WARNING"},
+        "mozilla_django_oidc": {"handlers": ["console"], "level": "WARNING", "propagate": False},
```
juspence: I think this logger's configuration matches the root logger, right? Print any warnings or higher to the console? Mailing an error should only happen if you add mail_admins to the list of handlers.

corgi/core/models.py
```diff
@@ -1916,3 +1916,17 @@ class Meta:
...
+
+class RedHatUser(models.Model):
+    """Additional information provided by Red Hat SSO, used for access controls"""
+
+    rhat_uuid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    rhat_roles = models.TextField(default="")
+    # Storing CN instead of trying to split it into Django's given/first/family/last
+    # bc https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/
+    cn = models.CharField(max_length=256, default="")
```
juspence: cn is a little confusing. Can we call it cert_common_name if it's a name from a certificate, or combined_name if it's from a human?

And if it's a common name from a certificate, should we be enforcing uniqueness here? I hope our CA is sane and we won't ever see two different users with the same CN, We might see two different certs with the same CN, though, if a user with an older expired cert tries to log in with a newly-issued cert. Both certs would still have the same common name I think. Unless the user recently changed their name (marriage, divorce, etc.)....ugh.

If you think this is correct, feel free to ignore me and leave as-is.

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+        RedHatUser.objects.filter(user=user).update(
+            rhat_uuid=claims["rhatUUID"],
+            rhat_roles=claims.get("groups", ""),
+            cn=claims.get("cn", ""),
+        )
+
+        return user
+
+
+class RedHatRolePermission(BasePermission):  # type: ignore
```
juspence: I think this is the metaclass error you mentioned in Gchat, right? It's fine to ignore, but please add a more specific ignore for a particular mypy error code, like ```# type: ignore[attr-defined]```, and leave a comment explaining why it's here.

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+        rhat_uuid = claims.get("rhatUUID")
+
+        if not rhat_uuid:
+            return self.UserModel.objects.none()
+
+        try:
+            rhat_user = RedHatUser.objects.get(rhat_uuid=rhat_uuid)
+            return [rhat_user.user]
+
+        except ObjectDoesNotExist:
```
juspence: Should this be the more specific RedHatUser.DoesNotExist, or is there some other class (outside of the inheritance hierarchy) that can throw DoesNotExist errors we need to handle?
```suggestion
        except RedHatUser.DoesNotExist:
```

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+        return user
+
+
+class RedHatRolePermission(BasePermission):  # type: ignore
+    """A permission class that only grants access to users with a given role.
+    Nb: Users are only required to have ONE of the specified roles, if more than one
+    are specified."""
+
+    def has_permission(self, request, view):
+        assert hasattr(view, "roles_permitted")
```
juspence: No asserts outside of test code please, they can get optimized out in some cases:
```suggestion
        if not hasattr(view, "roles_permitted"):
            raise ValueError(f"{view} doesn't define any permitted roles")
```

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            rhat_user = RedHatUser.objects.get(user=request.user)
+        except ObjectDoesNotExist:
```
juspence: As above, could be RedHatUser.DoesNotExist (unless this is needed for some other reason).

corgi/api/views.py
```diff
@@ -207,6 +216,36 @@ def list(self, request: Request) -> Response:
...
+    """
+    View to determine whether you are authenticated with an account that has a specific
+    role.
+    """
+
+    authentication_classes = [OIDCAuthentication]
+    permission_classes = [RedHatRolePermission]
+    roles_permitted = ["prodsec-dev"]
+
+    def get(self, request, format=None):
```
juspence, line 244: Let's add some type hints here if possible.

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+            return self.UserModel.objects.none()
+
+        return self.UserModel.objects.none()
+
+    def create_user(self, claims):
+        """Rather than changing the existing Django user model, this stores Red Hat SSO
+        claims in a separate model keyed to the created user."""
+        assert "rhatUUID" in claims
+        user = super(CorgiOIDCBackend, self).create_user(claims)
```
juspence, line 46: Not a problem necessarily but what's the difference between ```super()``` and ```super(CorgiOIDCBackend, self)```?

corgi/api/views.py
```diff
@@ -207,6 +216,36 @@ def list(self, request: Request) -> Response:
...
+    return Response(content)
+
+
+class ControlledAccessTestView(APIView):
+    """
+    View to determine whether you are authenticated with an account that has a specific
+    role.
+    """
+
+    authentication_classes = [OIDCAuthentication]
```
juspence, line 240: Should one or both of the views here be using CorgiOIDCAuthentication instead? Or the authentication backend in settings? I don't see anywhere that actually uses our custom class.

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+
+        except ObjectDoesNotExist:
+            return self.UserModel.objects.none()
+
+        return self.UserModel.objects.none()
+
+    def create_user(self, claims):
+        """Rather than changing the existing Django user model, this stores Red Hat SSO
+        claims in a separate model keyed to the created user."""
+        assert "rhatUUID" in claims
```
juspence: Let's raise an exception if this isn't true instead of asserting.

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+        return verified and "rhatUUID" in claims
+
+    def filter_user_by_claims(self, claims):
+        """The default behavior is to use e-mail, which may not be unique.
+        Instead, we use Red Hat UUID, which should be unique and persistent
+        between changes to other user claims."""
+        rhat_uuid = claims.get("rhatUUID")
+
+        if not rhat_uuid:
+            return self.UserModel.objects.none()
```
juspence, line 30: I think mypy might have complained at you before because this returns an empty UserModel queryset here and in the below except block, but a RedHat user queryset (with a value) on success. See if below fixes those mypy errors and lets you add type hints to all your methods:
```suggestion
            return RedHatUser.objects.none()
```
Above also avoids needing the self arg, so this can become a static method. Or maybe we should always use self.UserModel, and the code on line 29 below that accesses RedHatUser directly should be changed?

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Create a Red Hat User for this user
+        _ = RedHatUser.objects.create(
+            rhat_uuid=claims["rhatUUID"],
+            rhat_roles=claims.get("groups", ""),
+            cn=claims.get("cn", ""),
+            user=user,
+        )
+
+        return user
```
juspence, line 56: I don't know much about how the OIDCAuthenticationBackend parent class is supposed to work, but should we return the RedHatUser we created here instead of the user created by the parent class?

corgi/core/authentication.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def update_user(self, user, claims):
+        RedHatUser.objects.filter(user=user).update(
+            rhat_uuid=claims["rhatUUID"],
+            rhat_roles=claims.get("groups", ""),
+            cn=claims.get("cn", ""),
+        )
+
+        return user
```
juspence, line 65: As above - I don't know how this works in the base class, but typically in Python you'd return None if you're only mutating an object.

juspence (approved):
LGTM, thanks for adding this feature! Most comments are minor suggestios or questions, feel free to ignore if they don't make sense.

juspence:
Not sure if it's valuable or not, but maybe we could add a simple unit test on the auth views. Type hints on a few more methods would also be nice if possible, but don't stress too much about these.

corgi/core/models.py
kgrant-rh: It's actually a holdover from LDAP's "common name", which could refer to a person or any other object, but here it refers to a person. How about something like `full_name`?

corgi/api/views.py
kgrant-rh: The terminology is convoluted, but this setting just determines what kind of authentication to use (OIDC), and the configuration for OIDC defines the backend to use (CorgiOIDCBackend). The backend is specified in `settings.base.OIDC_DRF_AUTH_BACKEND`.

corgi/core/models.py
juspence: Ether full_name or common_name is fine. Maybe just add a comment with the explanation above.

corgi/core/authentication.py
kgrant-rh: I may have chosen a poor name for `RedHatUser`, which is not actually a User object at all, just a separate object that stores additional information from Red Hat SSO. This method wants a queryset, but now that I look more closely at it, I realize I had a typo that could have been an issue. Anyway, I'll rename `RedHatUser` to `RedHatProfile` which should make the distinction clearer.

juspence: The issue here is not so much the naming, but the return type. Because we return different queryset types on success vs. failure, mypy will complain.

I think you can fix this by always returning a RedHatUser.objects.none() on failure instead of self.UserModel.objects.none(). Even if the first is really a profile and not a user, I think it's OK since we're just returning an empty query set anyway.

kgrant-rh: I added some type annotations here to hopefully clear things up a bit. The return type of this method is Iterable[User]. On line 34, the return is actually the User object referenced by the Profile, not the Profile itself.

corgi/core/authentication.py
kgrant-rh: The parent class just wants the user. The purpose in deriving from that class is to create a profile when the user is created, but the profile is just a side effect.

corgi/core/authentication.py
kgrant-rh: Yeah, this is another consequence of deriving from `OIDCAuthenticationBackend`.

corgi/core/authentication.py
kgrant-rh: They're probably equivalent here. I'll be honest, I used the explicit form only because [that's what the docs do](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#changing-how-django-users-are-created).

juspence: Following the docs is a good reason. TIL, thanks for letting me know 🙂

config/settings/base.py
```diff
@@ -369,6 +373,18 @@
 }
 
 
+# OIDC Authentication
+# https://mozilla-django-oidc.readthedocs.io/
+OIDC_AUTH_ENABLED = strtobool(os.getenv("CORGI_OIDC_AUTH_ENABLED", "false"))
+OIDC_OP_JWKS_ENDPOINT = os.getenv("CORGI_OIDC_OP_JWKS_ENDPOINT", "")
```
jasinner, line 379: Could you perhaps open a component-registry-ops PR with this values populated so that we could know how to test this code in DEV? It will be required for this feature to work in stage anyway.

I found only one value for client secret in the Vault.

jasinner (approved):

config/settings/base.py
kgrant-rh: Because SSO restricts the URIs that can receive tokens, you can't use the stage values for local testing. I've added the configuration for stage and prod to component-registry-ops.

PR #392, opened 2023-04-14
CORGI-546 low coverage

jasinner:
Fixes bad linking between product streams and variants. The variant 8Base-Quay-3 for example was reused in both the quay-3.8 and quay-3.6 product streams. We can determine what's in those stream by using brew tags only. We don't need the variant, so I removed the problem linking for that Product Version, and also 4 others as explained in CORGI-546.

This change also comes with a script which cleans up the bad data produced by this bug. I verified it worked for the 8Base-Quay-3 variant. See CORGI-546 for the script.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -514,11 +514,21 @@ def fetch_modular_builds(relations_query: QuerySet, force_process: bool = False)
 
 
 def fetch_unprocessed_relations(
-    relation_type: ProductComponentRelation.Type,
     created_since: Optional[datetime] = None,
+    product_ref: Optional[str] = "",
+    relation_type: Optional[ProductComponentRelation.Type] = None,
     force_process: bool = False,
 ) -> int:
-    query = Q(type=relation_type)
+    """Load Brew builds for relations which don't have an associated SoftwareBuild.
+    Accepts optional arguments product_ref and relation_type which add query filters"""
+    query = Q()
+    if relation_type:
+        query &= Q(type=relation_type)
+        logger.info(f"Processing relations of type {relation_type}")
+    if product_ref:
+        query &= Q(product_ref=product_ref)
+        logger.info(f"Processing relations with product reference {product_ref}")
+
     if created_since:
         query &= Q(created_at__gte=created_since)
     relations_query = (
@@ -527,7 +537,7 @@ def fetch_unprocessed_relations(
         .distinct()
         .using("read_only")
     )
-    logger.info(f"Processing relations of type {relation_type}")
+
     processed_builds = 0
     for build_id, build_type in relations_query.iterator():
         if not build_id:
@@ -570,7 +580,7 @@ def fetch_unprocessed_brew_tag_relations(
             "corgi.tasks.brew.fetch_unprocessed_brew_tag_relations"
         )
     return fetch_unprocessed_relations(
-        ProductComponentRelation.Type.BREW_TAG,
+        relation_type=ProductComponentRelation.Type.BREW_TAG,
         force_process=force_process,
         created_since=created_dt,
     )
--- a/corgi/tasks/management/commands/loadunprocessedrelations.py
+++ b/corgi/tasks/management/commands/loadunprocessedrelations.py
@@ -1,4 +1,4 @@
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandParser
 
 from corgi.core.models import ProductComponentRelation
 from corgi.tasks.brew import fetch_unprocessed_relations
@@ -7,17 +7,41 @@ from corgi.tasks.brew import fetch_unprocessed_relations
 class Command(BaseCommand):
     help = "Fetch unprocessed builds from the relations table"
 
-    def handle(self, *args, **options):
-        self.stdout.write(
-            self.style.SUCCESS(
-                "Loading unprocessed relations",
-            )
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument(
+            "-p",
+            "--product_ref",
+            dest="product_ref",
+            help="Fetch unprocessed relations for product reference",
         )
+
+    def handle(self, *args, **options):
         processed_builds = 0
-        for relation_type in ProductComponentRelation.objects.values_list(
-            "type", flat=True
-        ).distinct():
-            processed_builds += fetch_unprocessed_relations(relation_type=relation_type)
+        if options["product_ref"]:
+            if (
+                not ProductComponentRelation.objects.db_manager("read_only")
+                .filter(product_ref=options["product_ref"])
+                .exists()
+            ):
+                self.out.write(
+                    self.style.ERROR(
+                        f"Could not find relations with product ref: {options['product_ref']}"
+                    )
+                )
+                exit(1)
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Loading unprocessed relations with product ref: {options['product_ref']}",
+                )
+            )
+            processed_builds = fetch_unprocessed_relations(product_ref=options["product_ref"])
+        else:
+            self.stdout.write(
+                self.style.SUCCESS(
+                    "Loading unprocessed relations",
+                )
+            )
+            processed_builds = fetch_unprocessed_relations()
         self.stdout.write(
             self.style.SUCCESS(
                 f"Loaded {processed_builds} unprocessed relations",
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -130,8 +130,10 @@ def update_products() -> None:
                             "obj": product_stream,
                         },
                     )
-
-                    if len(brew_tags) > 0:
+                    # quay-3 Errata Tool product version Quay-3-RHEL-8 list too many brew tags
+                    # Linking quay streams to the 8Base-Quay-3 variant here via brew tags leads
+                    # to builds from later streams being included in earlier ones, see PROJQUAY-5312.
+                    if len(brew_tags) > 0 and product_version.name != "quay-3":
                         logger.debug(
                             "Found brew tags (%s) in product stream: %s",
                             brew_tags,
--- a/corgi/tasks/pulp.py
+++ b/corgi/tasks/pulp.py
@@ -34,7 +34,7 @@ def fetch_unprocessed_cdn_relations(
     else:
         created_dt = get_last_success_for_task("corgi.tasks.pulp.fetch_unprocessed_cdn_relations")
     return fetch_unprocessed_relations(
-        ProductComponentRelation.Type.CDN_REPO,
+        relation_type=ProductComponentRelation.Type.CDN_REPO,
         force_process=force_process,
         created_since=created_dt,
     )
--- a/corgi/tasks/yum.py
+++ b/corgi/tasks/yum.py
@@ -35,7 +35,7 @@ def fetch_unprocessed_yum_relations(
     else:
         created_dt = get_last_success_for_task("corgi.tasks.yum.fetch_unprocessed_yum_relations")
     return fetch_unprocessed_relations(
-        ProductComponentRelation.Type.YUM_REPO,
+        relation_type=ProductComponentRelation.Type.YUM_REPO,
         force_process=force_process,
         created_since=created_dt,
     )
--- a/tests/test_brew_collector.py
+++ b/tests/test_brew_collector.py
@@ -9,8 +9,14 @@ from django.db import IntegrityError
 from yaml import safe_load
 
 from corgi.collectors.brew import Brew, BrewBuildTypeNotSupported
-from corgi.core.models import Component, ComponentNode, SoftwareBuild
+from corgi.core.models import (
+    Component,
+    ComponentNode,
+    ProductComponentRelation,
+    SoftwareBuild,
+)
 from corgi.tasks.brew import (
+    fetch_unprocessed_relations,
     find_package_file_name,
     save_component,
     slow_fetch_brew_build,
@@ -25,6 +31,7 @@ from tests.data.image_archive_data import (
 )
 from tests.factories import (
     ContainerImageComponentFactory,
+    ProductComponentRelationFactory,
     SoftwareBuildFactory,
     SrpmComponentFactory,
 )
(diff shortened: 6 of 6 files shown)


jasinner:
In Dev the 'secrets' task is failing with:

```
The baseline file was updated.
Probably to keep line numbers of secrets up-to-date.
Please `git add .secrets.baseline`, thank you.


ERROR: InvocationError for command /usr/bin/bash -c 'detect-secrets-hook --baseline .secrets.baseline  $(git diff --name-only "origin/main..HEAD" || git ls-files)' (exited with code 3)
_____________________________________________________________________________________________________________________ summary _____________________________________________________________________________________________________________________
ERROR:   secrets: commands failed
```

I'm not sure how to fix it.

juspence:
> With an iterator this doesn't cause OOM problems which was why it was designed to only check since the last run initially.

I think I added the "get_last_success_for_task" logic to avoid slowness / too many DB queries, if we processed all the relations every day. I never saw any out-of-memory issues here as far as I recall. I do see that in stage, we have 2.5 million relations, so I'm not even sure we'll finish processing those in one day before the next day's processing begins.

If we want to make our PCR processing more robust, I think the right way is to finish CORGI-352 (which I have no time to commit to - feel free to pick it up). Then we can efficiently filter the relations, and process only the ones which have null / empty ForeignKey links to a SoftwareBuild or ProductStream / ProductVariant (one SQL query to find all relations that need processing).

Right now we would have to iterate through all 2.5 million relations, fetch the build ID and product ID from each, then make 2 separate SQL queries per relation to see if those builds / products exist. So we'd make 5 million extra queries every day just to find all relations that need processing. That all happens before we can actually process the relations themselves - this is just to find which ones need processing.

To make the secrets check pass, you just need to commit the (changed) .secrets.baseline file. This is exactly the same as needing to commit the changed openapi.yml schema file whenever we add new API parameters or endpoints.

corgi/tasks/brew.py
```diff
@@ -514,43 +513,57 @@ def fetch_modular_builds(relations_query: QuerySet, force_process: bool = False)
...
             logger.info("Processing %s relation build with id: %s", relation_type, build_id)
             if build_type == SoftwareBuild.Type.CENTOS:
-                # This skips use of the Collector models for builds in the CENTOS koji instance
-                # It was done to avoid updating the collector models not to use build_id as
-                # a primary key. It's possible because the only product stream (openstack-rdo)
-                # stored in CENTOS koji doesn't use modules
-                slow_fetch_brew_build.delay(
-                    build_id, SoftwareBuild.Type.CENTOS, force_process=force_process
-                )
+                slow_fetch_brew_build.delay(build_id, SoftwareBuild.Type.CENTOS)
```
juspence: Looks like we lost the force_prcess arg here.

corgi/tasks/management/commands/loadunprocessedrelations.py
```diff
@@ -7,19 +7,111 @@
...
         )
+
+    def check_relation_type(self, options):
+        if (
+            not ProductComponentRelation.objects.db_manager("read_only")
+            .filter(type=options["relation_type"])
+            .exists()
+        ):
+            self.out.write(
+                self.style.ERROR(f"Could not find relations of type: {options['relation_type']}")
```
juspence: Instead of doing a DB query here, shouldn't we just see if the type given is in ProductComponentRelation.STREAM_TYPES or ProductComponentRelation.VARIANT_TYPES?

corgi/tasks/management/commands/updatecomponenttaxonomy.py
```diff
@@ -22,20 +29,49 @@ def handle(self, *args, **options):
...
                 self.style.SUCCESS(
                     "updating all builds component taxonomies",
                 )
             )
-            builds = SoftwareBuild.objects.get_queryset()
-        for sb in builds:
-            self.stdout.write(
-                self.style.SUCCESS(
-                    f"updating {sb.build_id}: {sb.name}",
+            for build in SoftwareBuild.objects.db_manager("read_only").all().iterator():
```
juspence: If we want to let slow_fetch_brew_build handle the taxonomy-saving for us, and we don't need to call sb.save_product_taxonomy() directly, we can use .values_list("build_id", "build_type") to avoid fetching extra data from the DB:
```suggestion
            for build in SoftwareBuild.objects.db_manager("read_only").values_list("build_id", "build_type").iterator():
```
Or if .db_manager() doesn't support this method, you can do below instead:
```suggestion
            for build in SoftwareBuild.objects.values_list("build_id", "build_type").using("read_only").iterator():
```

But I think the old way was better, since that just saved the taxonomy directly instead of adding many tasks onto the slow queue. We could still add an .iterator() to the two queries there to avoid OoM issues when processing many builds or components.

corgi/tasks/prod_defs.py
```diff
@@ -130,57 +129,6 @@ def update_products() -> None:
...
-                                        },
-                                    )
-                                    ProductNode.objects.get_or_create(
-                                        object_id=product_variant.pk,
-                                        defaults={
-                                            "parent": product_stream_node,
-                                            "obj": product_variant,
-                                        },
-                                    )
-                                    product_variant.save_product_taxonomy()
```
juspence: I don't really understand this change or your comment in the MR description. By "bad linking", do you mean that two different streams both pointed to the same variant? If that's the case, I think we should add a field like "variant_source" to distinguish whether a variant came from a Brew tag or directly from variant definitions in Errata Tool.

I don't want to stop creating the variant, since this seems likely to introduce regressions in other places.

Maybe we could use some existing field in the Brew / Errata Tool data to distinguish these two variants, instead of adding a custom field ourselves (e.g. name alone isn't enough to make them unique). Or maybe this is a bug in the Brew tag / Errata Tool variant definitions, and we should report it to the Quay engineering team?

juspence:
I haven't looked over the cleanup script, but had a question. Let's assume the quay-3.6 variant and quay-3.8 variant (which are the same in our DB) both had some components linked to them, and that the quay-3.8 variant is the one that came from a Brew tag.

The code change above stops creating variants that came from Brew tags. And I think your comment in the MR description is saying that we can use the PCR table to find all the builds / components related to that Brew tag / the quay-3.8 variant. Then we would delete only those relations, builds, and components, and we'd be left with only the quay-3.6 variant and all its relations / builds / components.

Is that right? But if so, how do we load the correct quay-3.8 builds and components (which have this Brew tag) in the future? There's no longer a variant or any relations to link them to.

juspence:
> We don't need the variant, so I removed the problem linking.

In the future we need to be able to manifest variants, per Rogue's comments in another meeting. This used to be possible, but implementation details in the latest filter broke this and it needs to get fixed. If we stop creating variants that are based on Brew tags, manifesting these variants won't be possible. Unless they also have a matching ET variant, but that seems rare.

Or maybe matching ET variant names / Brew tag names isn't rare? I think we should check all the Brew tags we loaded vs. the ET variants to see if this naming collision happens frequently / if there are any other variants which are referred to by multiple streams. The DB enforces that one variant always has exactly one stream, and we use update_or_create with only a variant name kwarg. So I think the ET variants will always "win".

In other words, the Brew tag variant is created first, then creating the ET variant fails (matching name already exists). So the preexisting variant (for the Brew tag) will be updated to point to the ET variant's stream, instead of the Brew tag's stream.

So we should be able to check that all streams have exactly one variant for each of their Brew tags. Any stream that's missing a variant for one of its Brew tags would be affected by this bug.

juspence (changes requested):
There's too much here I don't understand. If I'm missing something obvious, would you mind a quick chat to disscuss? I don't think I'm understanding the issue just looking at code / text.

corgi/tasks/management/commands/loadunprocessedrelations.py
```diff
@@ -7,19 +7,111 @@
...
+            self.check_relation_type(options)
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Loading relations with product ref: {options['product_ref']},"
+                    f" and type {options['relation_type']}",
+                )
+            )
+            processed_builds = fetch_unprocessed_relations(
+                product_ref=options["product_ref"],
+                relations_type=options["relation_type"],
```
juspence: Looks like SonarQube is complaining about a typo here:
```suggestion
                relation_type=options["relation_type"],
```

jasinner:
> If we want to make our PCR processing more robust, I think the right way is to finish CORGI-352

I found through testing these updated tasks in dev that the DB queries don't take that long. But since the biggest cause of low coverage seems to be bad variant matching I'm going to take the commits out of this PR which updated the unprocess_relations task, and just keep the update to the management command which allows us to process a single product_ref, so we can update those when required. 

We should implement CORGI-352 before we update all the history, as you say.

corgi/tasks/management/commands/loadunprocessedrelations.py
jasinner: I removed this commit from the PR

corgi/tasks/management/commands/loadunprocessedrelations.py
jasinner: I removed this commit from the PR

corgi/tasks/brew.py
jasinner: I removed this commit from the PR

jasinner:
> I haven't looked over the cleanup script, but had a question. Let's assume the quay-3.6 variant and quay-3.8 variant (which are the same in our DB) both had some components linked to them, and that the quay-3.8 variant is the one that came from a Brew tag.
> 
> The code change above stops creating variants that came from Brew tags. And I think your comment in the MR description is saying that we can use the PCR table to find all the builds / components related to that Brew tag / the quay-3.8 variant. Then we would delete only those relations, builds, and components, and we'd be left with only the quay-3.6 variant and all its relations / builds / components.
> 
> Is that right? But if so, how do we load the correct quay-3.8 builds and components (which have this Brew tag) in the future? There's no longer a variant or any relations to link them to.

The relations for the quay-3.8 brew tag will still exist.

jasinner:
> In Dev the 'secrets' task is failing with:
> 
> ```
> The baseline file was updated.
> Probably to keep line numbers of secrets up-to-date.
> Please `git add .secrets.baseline`, thank you.
> 
> 
> ERROR: InvocationError for command /usr/bin/bash -c 'detect-secrets-hook --baseline .secrets.baseline  $(git diff --name-only "origin/main..HEAD" || git ls-files)' (exited with code 3)
> _____________________________________________________________________________________________________________________ summary _____________________________________________________________________________________________________________________
> ERROR:   secrets: commands failed
> ```
> 
> I'm not sure how to fix it.

Resolved in the latest version of this PR

corgi/tasks/management/commands/updatecomponenttaxonomy.py
jasinner: I removed this commit from the PR

jasinner:
> I haven't looked over the cleanup script, but had a question. Let's assume the quay-3.6 variant and quay-3.8 variant (which are the same in our DB) both had some components linked to them, and that the quay-3.8 variant is the one that came from a Brew tag.
> 
I had a closer look at it and it seems only the quay-3.6 stream has the 8Base-Quay-3 variant linked to it, not both. I think this is because we use update_or_create to create the link so whatever stream you process last gets the variant link, even though the prod_defs task should link both streams to be variant via brew tags.

> The code change above stops creating variants that came from Brew tags. And I think your comment in the MR description is saying that we can use the PCR table to find all the builds / components related to that Brew tag / the quay-3.8 variant. Then we would delete only those relations, builds, and components, and we'd be left with only the quay-3.6 variant and all its relations / builds / components.
> 
Brew tags are used to link builds to builds to streams in the PCR table, variants are not required. In Deptopia Variants are only used for linking when a stream has an errata_info stanza.

> Is that right? But if so, how do we load the correct quay-3.8 builds and components (which have this Brew tag) in the future? There's no longer a variant or any relations to link them to.

Since we created relations for the quay-3.x streams using brew tags, we don't need to link builds via variants.

juspence:
> Brew tags are used to link builds to builds to streams in the PCR table, variants are not required. In Deptopia Variants are only used for linking when a stream has an errata_info stanza.
> 
> Since we created relations for the quay-3.x streams using brew tags, we don't need to link builds via variants.

Thank you for explaining! This is the part I was missing, and I think given above this change shouldn't cause any issues.

corgi/tasks/prod_defs.py
```diff
@@ -130,7 +130,11 @@ def update_products() -> None:
                             "obj": product_stream,
                         },
                     )
-
+                    # Some Product Versions in Errata Tool list too many brew tags
+                    # Linking these Product Versions here leads to builds from later streams
+                    # being included in earlier ones, see CORGI-546.
+                    if product_version.name in ["quay-3", "rhscl-3", "dts-11", "dts-12", "rhn_satellite_6"]:
+                        continue
                     if len(brew_tags) > 0:
```
juspence: After your explanation, I agree it's OK to skip creating the variant from the Brew tag, since we only need to link the Brew tag to a stream and not a variant.

But the check here will also skip creating any variants from the errata_info stanza in prod-defs for these streams (line 189 below). Should we move the check a little lower? Or should we try to get the data in ET updated, so that these PVs don't have incorrect Brew tags?

I might be confused and we don't want any variants here, either from Brew tags or from errata_info. In that case feel free to ignore.
```suggestion
                    # Some Product Versions in Errata Tool list too many brew tags
                    # Linking these Product Versions here leads to builds from later streams
                    # being included in earlier ones, see CORGI-546.
                    if len(brew_tags) > 0 and product_version.name not in ("quay-3", "rhscl-3", "dts-11", "dts-12", "rhn_satellite_6"):
```

juspence:
> I had a closer look at it and it seems only the quay-3.6 stream has the 8Base-Quay-3 variant linked to it, not both. I think this is because we use update_or_create to create the link so whatever stream you process last gets the variant link, even though the prod_defs task should link both streams to be variant via brew tags.

Yes, in a comment above I note that the last stream processed will always win. We can't link two different streams to the same variant, because each variant must have exactly one parent stream. So the 8Base-Quay-3 variant must be linked to either the quay-3.6 or quay-3.8 stream, and cannot be linked to both.

This was discussed a while ago when we switched to using ForeignKeys. I asked if I should use a ForeignKey field (which only supports linking one stream to many variants) or a ManyToManyField (which would support linking many streams to many variants).

We could switch Variants to support many parent streams instead of only one, but this change is probably going to require a bit of work. We'd need to update at least the models and serializers, and maybe other places too.

juspence (approved):
I think this will work based on your explanation. There might be follow-up work needed if I'm understanding your comments correctly, but I might be confused so feel free to just merge this.

jasinner:
Abandoning this in favor of: https://github.com/RedHatProductSecurity/component-registry/pull/394

PR #393, opened 2023-04-17
CORGI-576: Don't report unreliable / human-provided upstream URLs for RPMs

juspence:
@RedHatProductSecurity/corgi-devs Please review this fix for a few internal URLs in manifests. Manifest generation is currently broken in stage, so I can't verify where these problematic / internal URLs (mentioned in ticket) are coming from or if this PR fixes all the issues. But this should fix at least one known issue, and matches what we discussed in a previous meeting.

EDIT: The example URL mentioned in the ticket is not actaully an internal URL. I can resolve it off VPN, just was having DNS issues when I first checked it. I think we still want to exclude these RPM upstream URLs though since they can be literally anything and there's no quality control or standards here.

### DIFF
--- a/corgi/web/templates/base_manifest.json
+++ b/corgi/web/templates/base_manifest.json
@@ -11,7 +11,7 @@
     "documentDescribes": [
         "SPDXRef-{{obj.uuid}}"
     ],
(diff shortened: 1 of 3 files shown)


mprpic (approved):
LGTM!

PR #394, opened 2023-04-17
CORGI-546 quay stream overzealous linking

jasinner:
Replaces https://github.com/RedHatProductSecurity/component-registry/pull/392

Fixes bad linking between product streams and variants. The variant 8Base-Quay-3 was reused in both the quay-3.8 and quay-3.6 product streams. We can determine what's in those stream by using brew tags only. We don't need the variant, so I removed the problem linking for that Product Version, and filed a task (PROJQUAY-5312) to fix the Variant usage from 3.9.

This change also comes with a script which cleans up the bad data produced by this bug. I verified it worked for the 8Base-Quay-3 variant. See CORGI-546 for the script.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -518,6 +518,7 @@ def fetch_unprocessed_relations(
     product_ref: Optional[str] = "",
     relation_type: Optional[ProductComponentRelation.Type] = None,
     force_process: bool = False,
+    save_only: bool = False,
 ) -> int:
     """Load Brew builds for relations which don't have an associated SoftwareBuild.
     Accepts optional arguments product_ref and relation_type which add query filters"""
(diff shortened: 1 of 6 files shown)


corgi/tasks/brew.py
```diff
@@ -527,14 +538,15 @@ def fetch_unprocessed_relations(
...
         if not build_id:
             # build_id defaults to "" and int() will fail in this case
             continue
         if (
-            not SoftwareBuild.objects.filter(build_id=build_id, build_type=build_type)
+            save_only
+            or not SoftwareBuild.objects.filter(build_id=build_id, build_type=build_type)
             .using("read_only")
             .exists()
         ):
```
juspence, line 552: We used to call int() on build_id on line 549, but it seems like we lost this after some refactoring. Do we need to add that back, or can we get rid of the first if statement on line 544?

juspence (commented):
LGTM, sorry for not getting to this earlier.

PR #396, opened 2023-04-18
CORGI-578: Add upstreams info (for containers only) to manifests

juspence:
@RedHatProductSecurity/corgi-devs Please review this addition to our manifests. We've been given this requirement from a few different teams, and ideally we'd include this upstream info before we start publishing manifests for customers.

We exclude RPM upstream data because it isn't reliable enough to report, and RPM modules have no upstreams. Remote-source components don't have upstreams, since they are already upstream components. So the only upstream data we can report is for containers.

Containers are also a little special, since they can be built from multiple upstream components, for example upstream go modules. So we can't report just a single URL for the upstream project on e.g. Github. We have to iterate over the linked upstream components and report all of their data.

### DIFF
--- a/corgi/web/templates/component_manifest.json
+++ b/corgi/web/templates/component_manifest.json
@@ -20,7 +20,27 @@
         "SPDXID": "SPDXRef-{{component.uuid}}",
         "supplier": "Organization: Red Hat",
         "versionInfo": "{{component.nevra|escapejs}}"
-    },{% endfor %}
+    },{% endfor %}{% if obj.type != obj.Type.RPM %}{% for upstream in obj.upstreams.get_queryset.iterator %}{# RPM upstream data is human-generated and unreliable #}{
+        "copyrightText": {% if upstream.copyright_text %}"{{upstream.copyright_text|escapejs}}"{% else %}"NOASSERTION"{% endif %},
+        "downloadLocation": {% if upstream.download_url %}"{{upstream.download_url|escapejs}}"{% else %}"NOASSERTION"{% endif %},
+        "externalRefs": [
+            {
+                "referenceCategory": "PACKAGE_MANAGER",
+                "referenceLocator": "{{upstream.purl|safe}}",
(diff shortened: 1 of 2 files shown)


corgi/web/templates/component_manifest.json
```diff
@@ -20,6 +20,27 @@
...
+        "filesAnalyzed": false,
+        "homepage": {% if upstream.related_url and upstream.type != upstream.Type.RPM %}"{{upstream.related_url|escapejs}}"{% else %}"NOASSERTION"{% endif %},
+        "licenseComments": "Licensing information is automatically generated and may be incomplete or incorrect.",
+        "licenseConcluded": {% if upstream.license_concluded %}"{{upstream.license_concluded}}"{% else %}"NOASSERTION"{% endif %},
+        "licenseDeclared": {% if upstream.license_declared %}"{{upstream.license_declared}}"{% else %}"NOASSERTION"{% endif %},
+        "name": "{{upstream.name|escapejs}}",
+        "originator": "NOASSERTION",
+        "packageFileName": {% if upstream.filename %}"{{upstream.filename}}"{% else %}"NOASSERTION"{% endif %},
+        "SPDXID": "SPDXRef-{{upstream.uuid}}",
+        "supplier": "Organization: Red Hat",
```
juspence: Technically we're still "supplying" this upstream component to the customer (with source code included in / as part of some Red Hat product), but I could leave off this "supplier" key for the upstream component if others prefer.

mprpic: Let's remove it, I think that's a more correct approach though it's debatable.

mprpic (approved):
Apart from the supplier, LGTM!

corgi/web/templates/component_manifest.json
juspence: Ack, I've set supplier to "NOASSERTION" and removed originator for all the upstream components.

PR #397, opened 2023-04-18
CORGI-336: Stop reporting internal ofuris that don't map to other IDs

juspence:
@mprpic Quick thumbs-up please, or let me know if you think we should continue to report the ofuri for internal customers. The CPE is already reported in the manifest when present, we just have some other bug that causes it to never be loaded.

Since you'll add the CPEs manually, I'll investigate + fix that later. Next up is the filter to report only released components instead of all components.


mprpic (approved):
LGTM! Thanks for the quick fix!

PR #401, opened 2023-04-19
CORGI-592 refresh build/component product taxonomy after creating a new relation

jasinner:
In CORGI-592 it was discovered that if a SoftwareBuild was created before we created a ProductComponentRelation for that build we did not update the build and it's related components product taxonomy.

This updates the create_relation function to accept a callback function to be called when a new relation is created. The callback function then is responsible for calling SoftwareBuild.save_product_taxonomy function for that build. We call that from slow_fetch_brew_build even if a SoftwareBuild object for the build already exists.

The reason for not calling SoftwareBuild.save_product_taxonomy directly in create_relation is too allow flexibility in how we reprocess builds.  In also ensures that any modules are represented correctly as components and the component taxonomy is still up to date.

### DIFF
--- a/corgi/tasks/common.py
+++ b/corgi/tasks/common.py
@@ -68,7 +68,9 @@ def get_last_success_for_task(task_name: str) -> datetime:
     )
 
 
-def create_relations(build_ids, build_type, external_system_id, product_ref, relation_type) -> int:
+def create_relations(
+    build_ids, build_type, external_system_id, product_ref, relation_type, refresh_task
+) -> int:
     no_of_relations = 0
     for build_id in build_ids:
         _, created = ProductComponentRelation.objects.get_or_create(
@@ -79,6 +81,15 @@ def create_relations(build_ids, build_type, external_system_id, product_ref, rel
             defaults={"type": relation_type},
         )
         if created:
+            # Similar to fetch_unprocessed_relations
+            # This skips use of the Collector models for builds in the CENTOS koji instance
+            # It was done to avoid updating the collector models not to use build_id as
+            # a primary key. It's possible because the only product stream (openstack-rdo)
+            # stored in CENTOS koji doesn't use modules
+            if build_type == SoftwareBuild.Type.CENTOS:
+                refresh_task.delay(build_id=build_id, build_type=SoftwareBuild.Type.CENTOS)
+            else:
+                refresh_task.delay(build_id=build_id)
             no_of_relations += 1
     return no_of_relations
 
--- a/tests/test_rhel_compose.py
+++ b/tests/test_rhel_compose.py
@@ -142,8 +142,9 @@ def mock_fetch_rpm_data(compose_url, variants):
 
 
 @pytest.mark.django_db
+@patch("corgi.tasks.brew.slow_fetch_modular_build.delay")
 @patch("corgi.collectors.rhel_compose.RhelCompose._fetch_rpm_data", new=mock_fetch_rpm_data)
-def test_save_compose(requests_mock):
+def test_save_compose(mock_fetch_modular_build, requests_mock):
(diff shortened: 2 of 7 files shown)


jasinner:

tests/test_rhel_compose.py
```diff
@@ -142,8 +142,9 @@ def mock_fetch_rpm_data(compose_url, variants):
 
 
 @pytest.mark.django_db
+@patch("corgi.tasks.brew.slow_fetch_modular_build.delay")
 @patch("corgi.collectors.rhel_compose.RhelCompose._fetch_rpm_data", new=mock_fetch_rpm_data)
-def test_save_compose(requests_mock):
+def test_save_compose(mock_fetch_modular_build, requests_mock):
```
juspence, line 147: It looks like you added a new mock, but didn't assert_called_once_with(args)?

corgi/tasks/common.py
```diff
@@ -79,6 +81,15 @@ def create_relations(build_ids, build_type, external_system_id, product_ref, rel
             defaults={"type": relation_type},
         )
         if created:
+            # Similar to fetch_unprocessed_relations
+            # This skips use of the Collector models for builds in the CENTOS koji instance
+            # It was done to avoid updating the collector models not to use build_id as
+            # a primary key. It's possible because the only product stream (openstack-rdo)
+            # stored in CENTOS koji doesn't use modules
```
juspence, line 88: I don't think I understand the logic here. slow_fetch_brew_build() can create many relations, then each relation will immediately call slow_fetch_brew_build() when it's created. Is it possible this will cause our Celery queue to fill up with too many tasks all at once?

Also, I don't know that we forgot to call slow_fetch_brew_build() or save the taxonomies here. I thought we split processing of relations into two separate tasks on purpose, to avoid overloading Celery like I mentioned above.

I thought that create_relations was only supposed to populate the PCR table. Then fetch_unprocessed_relations runs once a day, and processes all the relations (fetches Brew builds and saves taxonomies) that were created since the last successful task run.

jasinner: The problem with the current design is that that the fetch_unprocessed_relation task doesn't update the product taxonomy for a build if that build already exists. With the modifications in this PR we now update the product taxonomy for a build even if it already exists, whenever a relation is created.

I think we could probably remove the scheduled fetch_unprocessed_relations tasks, but I hesitated to do that straight away, as I wanted to monitor the task output to make sure they were always return 0 (indicated they didn't find any relations without build objects) because removing them.

We could change the design so that instead of passing a callback function to create_relations we always call slow_fetch_*_build even if a SoftwareBuild exists in order to make sure that builds product taxonomy is up to date. However I think the current design increases cohesion by keeping creation of relations and their representation in builds together.

If we call a task which is already in the backlog the Celery Singleton mechanism kicks in and ensures we only have one instance of that task in the backlog, so it shouldn't cause a redis backlog overload.

juspence: My concern isn't about duplicate tasks, it's about trying to reprocess 1 build and ending up with 5,000 tasks in our queue. AKA we'll process everything all at once, instead of creating relations and processing them later like we did before.

If you think this design is better I'm not really opposed to it, I guess I'm just confused about why we separated create_relations and fetch_unprocessed_relations in the first place. Maybe that's a sign you're right and we should just combine them :slightly_smiling_face:

juspence (approved):

tests/test_rhel_compose.py
jasinner: I just added the mock so that the task didn't fail because celery/redis is not setup for accepting delayed tasks. The focus of this test is on the save_compose part, not the create_relation it calls. I added a new test for that.

PR #402, opened 2023-04-19
CORGI-591: Add filter for released components

juspence:
@mprpic Quick review please. I still need to add tests, but if this looks sane I'll manually run the same logic I added to the Brew collector. That will process the list of tags on each build and figure out which are released errata IDs, then update the build's meta_attr with this list.

Once this is merged, I'll find the latest builds in some stream and update only those, then use the API filters here to confirm their components show up as "released". Once I have a few builds updated and working, I'll submit another PR for the Celery task bits and to make manifests use this filter.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -19,7 +19,7 @@ from corgi.core.models import Component, SoftwareBuild
 
 logger = logging.getLogger(__name__)
 
-ADVISORY_REGEX = re.compile(r"RH[BES]A-[12]\d{3}:\d{4,6}")
+ADVISORY_REGEX = re.compile(r"RH[BES]A-[12]\d{3}:\d{4,}")
 
 
 class BrewBuildTypeNotFound(Exception):
@@ -737,7 +737,18 @@ class Brew:
             match = ADVISORY_REGEX.match(tag)
             if match:
                 advisory_ids.add(match.group())
-        return list(advisory_ids)
+        return sorted(advisory_ids)
+
+    @staticmethod
+    def _parse_advisory_ids(errata_tags: list[str]) -> list[str]:
+        """From a Brew build's list of Errata tags, return tags with released (4-digit) IDs"""
+        # released errata always have 4-digit IDs, e.g. RHBA-2023:1234
+        # unreleased errata have 5-digit IDs or greater
+        # e.g. RHEA-2023:12345 or RHSA-2023:123456
+        # tags in Brew also have a -released, -dropped, or -pending suffix
+        # but our ADVISORY_REGEX strips this to get just the friendly advisory name
+
+        return [errata_tag for errata_tag in errata_tags if len(errata_tag.split(":")[-1]) == 4]
 
     @staticmethod
     def get_module_build_data(build_info: dict) -> dict:
@@ -843,6 +854,7 @@ class Brew:
         tags = self.koji_session.listTags(build_id)
         build["tags"] = [tag["name"] for tag in tags]
         build["errata_tags"] = self._extract_advisory_ids(build["tags"])
+        build["released_errata_tags"] = self._parse_advisory_ids(build["errata_tags"])
 
         # TODO: handle wrapper RPM builds:
         # brew buildID=1839210
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -19,7 +19,7 @@ from corgi.core.models import Component, SoftwareBuild
 
 logger = logging.getLogger(__name__)
 
-ADVISORY_REGEX = re.compile(r"RH\wA-[12]\d{3}:\d{4,6}")
+ADVISORY_REGEX = re.compile(r"RH[BES]A-[12]\d{3}:\d{4,6}")
 
 
 class BrewBuildTypeNotFound(Exception):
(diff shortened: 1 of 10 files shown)


juspence:
Existing tests are all passing, just need to add new tests to get coverage back above 70%, which is expected / I'll do after cleaning up a small amount of data.

corgi/collectors/brew.py
```diff
@@ -730,14 +730,27 @@ def get_maven_build_data(build_info: dict, build_type_info: dict) -> dict:
...
             if match:
                 advisory_ids.add(match.group())
         return list(advisory_ids)
 
+    @staticmethod
+    def _parse_advisory_ids(errata_tags: list[str]) -> list[str]:
+        """From a Brew build's list of Errata tags, return tags with released (4-digit) IDs"""
+        # released errata always have 4-digit IDs, e.g. RHBA-2023:1234
+        # dropped errata always have 5-digit IDs, e.g. RHEA-2023:12345
+        # pending errata always have 6-digit IDs, e.g. RHSA-2023:123456
```
mprpic: This isn't quite right. Any 5+ digit ID is just the non-live one, 6-digit IDs is just the latest errata since we've passed 99,999 errata now.

corgi/core/models.py
```diff
@@ -920,11 +920,36 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+
+    def root_components(self, include: bool = True) -> models.QuerySet["Component"]:
+        """Show only root components by default, or only non-root components if include=False"""
+        if include:
+            # Truthy values return the filtered queryset (only root components)
+            return self.filter(ROOT_COMPONENTS_CONDITION)
+        # Falsey values return the excluded queryset (only non-root components)
+        return self.exclude(ROOT_COMPONENTS_CONDITION)
+
+    def srpms(self, include: bool = True) -> models.QuerySet["Component"]:
```
mprpic, line 945: What is this filter used for?

openapi.yml
```diff
@@ -378,6 +378,16 @@ paths:
         name: release
         schema:
           type: string
+      - in: query
+        name: released_components
+        schema:
+          type: boolean
+        description: Show only released components
```
mprpic, line 403: It feels a bit unintuitive that the opposite of showing only released components is showing only unreleased components. I would expect it to simply show all components and a different param for showing only unreleased components. I guess when I think of boolean query params for filtering, I expect the truthy value to do something extra and the falsy be the same as if I didn't specify the param at all.

The alternative design for something like this is to not use a boolean value but an enum: "?component_status=released" vs "?component_status=unrealeased"; not specifying shows both.

corgi/core/models.py
```diff
@@ -920,11 +920,36 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
-        return self.filter(ROOT_COMPONENTS_CONDITION)
+    """Helper methods to filter QuerySets of Components"""
+
+    def released_components(self, include: bool = True) -> models.QuerySet["Component"]:
+        """Show only released components by default, or unreleased components if include=False"""
+        # TODO: I could make below into separate ArrayFields on the SoftwareBuild model
+        #  like brew_tags and released_errata, but this is all Brew-specific
+        #  it doesn't make sense for other build systems
+        missing_released_errata = Q(software_build__meta_attr__released_errata_tags__isnull=True)
+        empty_released_errata = Q(software_build__meta_attr__released_errata_tags=())
```
mprpic, line 930: Doesn't this also catch the previous condition?

corgi/core/models.py
juspence: You're commenting on a queryset here, but I think you mean the root_components filter in corgi/api/filters.py? That's laying groundwork for a future refactoring of filter_latest_components(). And that refactoring is needed to be able to manifest variants again.

openapi.yml
juspence: This is consistent with the other filters we've added. There's hints in the UI for the user, where they see a label like "include / exclude some components (YES / NO)" above each of the boolean controls.

An enum is possible, but requires our users to know the enum values (active / all, released / unreleased, etc.), so when there's only two choices I've typically just used a boolean filter.

corgi/core/models.py
juspence: I think it does, this was left over from a previous version of the code. Originally I was filtering on errata_tags, and some code implied that this was either a list or a string. But after I looked more at our data, the key is always present as a list, never a string, None, or missing. So only the second case should be needed here. Thanks for catching!

corgi/collectors/brew.py
juspence: I can't post the internal link here, but the 5-digit errata tags I looked at always ended with "-dropped", while the 6-digit errata always ended with "-pending". Is below correct?
```suggestion
        # pending errata and dropped errata may sometimes have 5-digit IDs, e.g. RHSA-2023:12345
        # pending errata and dropped errata may sometimes have 6-digit IDs, e.g. RHSA-2023:123456
```

corgi/core/models.py
mprpic: Ah, nvm, I didn't notice the `def srpms(self)` already existed above but the diff made it look like a brand new method.

corgi/collectors/brew.py
mprpic: You can just group it into one: "unreleased errata have 5-or-more-digit IDs". Dropped is just a different state of an unreleased erratum, same as QE for example. The only difference is that the QE state may result in an erratum being released whereas dropped means it will always remain unreleased.

openapi.yml
mprpic: Well, those enums are present in the OpenAPI schema and I think the web UI even turns those fields into dropdowns where you can select the value, but I guess if we're committed to boolean yes/no values then leave it as is.

mprpic (approved):

openapi.yml
juspence: Ah, you're right. I forgot about the ComponentType enum and was thinking of the ugly active/all enum filter for product streams, which isn't in the UI at all.

PR #403, opened 2023-04-19
Deconflict migrations

kgrant-rh:
Fix a migration conflict.


juspence (approved):
LGTM! If the pipeline fails on the performance tests, you can add  ```allow_failure: true``` to the test-performance stage that starts on line 131 of .gitlab-ci.yml.

What might have happened earlier is if you force-pushed a commit to add / change something you forgot, the "old" pipeline result still looks like everything is working. If you merge the PR right away and don't wait for the pipeline to run again, you may not realize that black / flake8 / migrations check is failing until the pipeline runs on the main branch.

I know this beause it happens to me all the time :facepalm:

PR #405, opened 2023-04-20
Don't try to look up anonymous users

kgrant-rh:
Fixes an issue where views with `RedHatRolePermission` would fail when unauthenticated users tried to connect.

### DIFF
--- a/corgi/core/authentication.py
+++ b/corgi/core/authentication.py
@@ -1,12 +1,14 @@
+import logging
 from collections.abc import Iterable
 from typing import Any
 
 from django.contrib.auth.models import User
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
-from rest_framework.exceptions import AuthenticationFailed
 from rest_framework.permissions import BasePermission
 
 from corgi.core.models import RedHatProfile
+
+logger = logging.getLogger(__name__)
 
 
 class CorgiOIDCBackend(OIDCAuthenticationBackend):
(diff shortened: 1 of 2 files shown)


kgrant-rh:
@RedHatProductSecurity/corgi-devs FYI

corgi/core/authentication.py
```diff
@@ -75,6 +75,9 @@ def has_permission(self, request: Any, view: Any) -> bool:
         if not hasattr(view, "roles_permitted"):
             raise ValueError(f"View {view} doesn't define any permitted roles")
 
+        if not request.user.is_authenticated:
+            return False
+
```
juspence: What about when a request doesn't have the user attribute defined? Or maybe that's not possible. It would also be great to add unit tests for this case if at all possible.

juspence (approved):

corgi/core/authentication.py
kgrant-rh: According to [The DRF docs](https://www.django-rest-framework.org/api-guide/authentication/) :

> If no class authenticates, `request.user` will be set to an instance of `django.contrib.auth.models.AnonymousUser`, and `request.auth` will be set to None.

That being the case, I think request.user will always be an actual user or `AnonymousUser`, for whom `is_authenticated` returns false. I'll make sure there are tests for unauthenticated users though.

kgrant-rh:

corgi/core/authentication.py
kgrant-rh: The tests I've added explicitly use `AnonymousUser`, which is what the DRF authentication will supply to the authentication backend in the case that there's no authentication.

tests/test_authentication.py
```diff
@@ -0,0 +1,88 @@
...
+            "id_token": "id_token",
+            "access_token": "access_granted",
+        }
+        requests_mock.post.return_value = post_json_mock
+
+        response = authentication_status(req)
+        self.assertEqual(response.status_code, 200)
+
+        response = ControlledAccessTestView.as_view()(req)
+        # This should maybe be a 401? Needs more research.
```
juspence, line 75: I think returning a 403 is correct. You've provided mocked credentials so the user isn't completely unauthenticated (401), but their profile doesn't have any rhat_roles defined so using this endpoint is unauthorized / forbidden (403).

And of course you have the test for unauthenticated users above that returns 401, then you have the test with an authenticated + authorized user below that returns 200. Thank you for adding comprehensive tests!

juspence (approved):
Still LGTM. Feel free to just merge approved MRs once you resolve all conversations, I don't think a second round of review is required. Although I'm happy to look if you just want a second pair of eyes.

PR #406, opened 2023-04-20
CORGI-591: Make manifests use filter for released components

juspence:
@mprpic No rush to look at this. I checked some OpenShift data in the API and saw the filter appeared to be working properly, so I've added the same filtering to our ProductStream manifest templates.

### DIFF
--- a/corgi/core/files.py
+++ b/corgi/core/files.py
@@ -67,11 +67,12 @@ class ProductManifestFile(ManifestFile):
     def render_content(self) -> str:
 
         latest_components = self.obj.get_latest_components()  # type: ignore[attr-defined]
+        released_components = latest_components.released_components()
         distinct_provides = self.obj.provides_queryset  # type: ignore[attr-defined]
 
         kwargs_for_template = {
             "obj": self.obj,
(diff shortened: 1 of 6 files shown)


corgi/core/files.py
```diff
@@ -67,11 +67,12 @@ class ProductManifestFile(ManifestFile):
     def render_content(self) -> str:
 
         latest_components = self.obj.get_latest_components()  # type: ignore[attr-defined]
```
mprpic, line 69: Is this unused now then?

mprpic (approved):
LGTM!

corgi/core/files.py
juspence: No, see the line below it. We need to get all the latest components in a stream, then find which of those are released. We can't do the other way around - find all the components that are released, then take their latest builds / versions - as a side-effect of how this filter was implemented. Refactoring is planned.

mprpic: Ah, fail, I didn't realize released_ was being run on top of latest_. Ignore me! :)

PR #409, opened 2023-04-22
fix component taxonomy view

JimFuller-RedHat:
As part of CORGI-586 - component taxonomy view should create a nested structure representing dependency tree.

**Note** -  Pure tree generation (without any calls to node.obj) is 1-2 magnitudes faster if we do not access obj - though we can use prefetch_related("obj") which makes them good enough - so I left in and expanded on Component data we return - if we need/want 'lightning speed' we can remove resolving any data from obj.


jasinner (approved):
LGTM

PR #411, opened 2023-04-24
CORGI-591: Fix filter for released components

juspence:
@mprpic Quick review please. There are some user-facing changes here, but I tried to keep these minimal and sane. I'd prefer to take some time to look this over and add tests, but please let me know if I really need to merge this ASAP to generate updated manifests.

@RedHatProductSecurity/corgi-devs FYI, you may want to look over the changes to our latest filter here. This is very badly needed ASAP so we can publish manifests for customers without missing data. But please let me know if you think there's a better way.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -379,7 +379,7 @@ class ProductModel(TimeStampedModel):
 
     @property
     @abstractmethod
-    def components(self) -> models.Manager["Component"]:
+    def components(self) -> "ComponentQuerySet":
         pass
 
     @property
@@ -525,7 +525,7 @@ class Product(ProductModel):
 
     # implicit "components" field on Product model
     # is created by products field on Component model
-    components: models.Manager["Component"]
+    components: "ComponentQuerySet"
 
     def get_ofuri(self) -> str:
         """Return product URI
@@ -562,7 +562,7 @@ class ProductVersion(ProductModel):
 
     # implicit "components" field on ProductVersion model
     # is created by productversions field on Component model
-    components: models.Manager["Component"]
+    components: "ComponentQuerySet"
 
     def get_ofuri(self) -> str:
         """Return product version URI.
@@ -609,7 +609,7 @@ class ProductStream(ProductModel):
 
     # implicit "components" field on ProductStream model
     # is created by productstreams field on Component model
-    components: models.Manager["Component"]
+    components: "ComponentQuerySet"
 
     def get_ofuri(self) -> str:
         """Return product stream URI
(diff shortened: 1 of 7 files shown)


corgi/core/models.py
```diff
@@ -923,19 +861,51 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
 class ComponentQuerySet(models.QuerySet):
     """Helper methods to filter QuerySets of Components"""
 
-    def released_components(self, include: bool = True) -> models.QuerySet["Component"]:
+    def filter_latest_nevra_by_name(self, component_name: str) -> str:
+        nevras = self.filter(name=component_name).values_list("nevra", flat=True)
+        if nevras:
+            # Get the latest NVR using python. This only works for valid RPM NEVRAs
+            return sorted(nevras, key=functools.cmp_to_key(compare_packages))[-1]
```
mprpic, line 868: I guess this may not work so well for components with versions that are just commit hashes but let's file that as a separate issue.

corgi/core/models.py
```diff
@@ -923,19 +861,51 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+            return ""
+
+    def latest_components(
+        self,
+        include: bool = True,
+    ) -> "ComponentQuerySet":
+        """Return components from latest builds."""
+        names = self.values_list("name", flat=True).order_by().distinct().iterator()
+        query = Q()
+        for name in names:
```
mprpic, line 879: Is this ever called on *all* components? I see in the manifest file we call it on the already filtered set of components of the stream or component we are generating the manifest for, but if this is exposed over the API to allow filtering on all components it will be very slow since it has to iterate over all component names, no?

juspence: Yes, it could in theory time out. .iterator() should prevent any DoS / OoM issues with the web pod itself, though, so I'm not sure it's worth fixing right now.

There are some other issues with tests, so going to slow-walk this one to avoid breaking things. Sorry for the delay - I'll ping as soon as I have an updated manifest for you. This will hopefully be tomorrow before lunch.

mprpic (approved):
LGTM!

corgi/core/models.py
juspence: Filed CORGI-604 for this.

juspence:
@mprpic Tests passed and there are no regressions that I can see, issues earlier were just due to human error. The deploy to stage is running now, so when that finishes you can run below in an OCP shell:
```
from corgi.tasks.tasks import *
stream = ProductStream.objects.get(name="openshift-4.12")
cpu_update_ps_manifest.delay(stream.name)
```
That should get you an updated manifest with correctly-filtered data. I think Jason also fixed the bug that was causing 4.12 to run out of memory, but if not just change above "openshift-4.12" to "openshift-4.12.z".

I'll do above first thing tomorrow morning if you've already stepped out for the day. There's one more change needed here to keep the tags updated automatically, so I'll throw in more tests for the filtering changes here as part of that PR.

PR #412, opened 2023-04-24
Exclude component names that use relative paths

mprpic:
...specifically in generated manifests. Solving the root cause of exluding them at ingestion is tracked in CORGI-605 and requires research as to how this is happening.

### DIFF
--- a/corgi/core/files.py
+++ b/corgi/core/files.py
@@ -4,6 +4,7 @@ from abc import ABC, abstractmethod
 
 import jsonschema
 from django.conf import settings
+from django.db.models import Q
 from django.template.loader import render_to_string
 
 from corgi.core.mixins import TimeStampedModel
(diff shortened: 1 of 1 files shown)


corgi/core/files.py
```diff
@@ -65,9 +66,16 @@ class ProductManifestFile(ManifestFile):
...
-
         components = self.obj.components  # type: ignore[attr-defined]
-        components = components.exclude(name__endswith="-container-source").using("read_only")
+        excluded_components = (
+            # Exclude source containers because they would only duplicate the content of regular
+            # container images.
+            Q(name__endswith="-container-source")
+            # Exclude components with relative paths in their name; see also CORGI-605
+            | Q(name__contains="./")
+            | Q(name__contains="../")
```
juspence: Accidentally commented on the Gitlab commit instead of the PR here :facepalm: 

This mostly LGTM, although I think the ./ case on line 75 also covers the ../ case below it. Can you just confirm we don't have some weird edge-case Golang component with a truncated namespace followed by other stuffl? Like ```namespace.k8s.io./component-name```?

Actualy, since this is just a Golang bug, let's make the above condition (type=GOLANG, name__contains="./") just for safety.

juspence (approved):
LGTM!

corgi/core/files.py
mprpic: I didn't see any other malformed components that would have a trailing dot.

Condition updated.

PR #413, opened 2023-04-25
Enhance get_provides, get_sources and get_upstreams to use include/exclude fields

JimFuller-RedHat:
Components return provides, sources and upstream components. Currently we only return component purls but as we use internal component names (eg. ps_component) for raising affects it is cumbersome to get this rote list of purls and then have to separately retrieve associated internal component names.

This PR (unblocks griffon) extends include_filter, exclude_filter paradigm to get_provides, get_sources and get_upstreams we are able to select them - while retaining status quo if no include/exclude filters are provided.

```
http://localhost:9000/api/v1/components?name=grafana&include_fields=purl,link,sources.purl,sources.name
```

**Note** - need to figure out the right way to annotate this construct for openapi schema and mypy and add tests.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -405,20 +405,51 @@ class ComponentSerializer(ProductTaxonomySerializer):
     def get_link(instance: Component) -> str:
         return get_component_purl_link(instance.purl)
 
-    @staticmethod
-    def get_provides(instance: Component) -> list[dict[str, str]]:
+    # @staticmethod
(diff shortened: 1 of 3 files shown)


jasinner:
Seems good in theory, but we really need some tests before merging.

jasinner (approved):
LGTM

JimFuller-RedHat:
to be clear - this preserves existing codepath so that status quo usage (without include_fields) makes no change in terms of performance ... merging

PR #414, opened 2023-04-25
CORGI-591: Add automation to keep Brew / Corgi build tags in sync

juspence:
@RedHatProductSecurity/corgi-devs Please review. This change is needed to keep the tags up-to-date on Corgi builds whenever they change in Brew. Having up-to-date tags ensures that manifests always represent the latest released content for some stream and are never stale.

I didn't add tests for the Celery task yet since I'm not sure about the logic. Is it enough to call slow_load_errata() when errata tags get added to some build? Or is there something else I need to do, so that relations are created and the build is correctly linked to the erratum's variant?

I'm also not sure how to correctly clean up the relations, whenever some build has an errata tag removed. If someone has an easy idea for a solution, I can handle that as part of this MR. I could also file a backlog ticket, or we could avoid doing this at all if we never want to delete the relations (but that doesn't seem right).


corgi/monitor/management/commands/runmonitor.py
```diff
@@ -11,6 +11,6 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         if settings.UMB_BREW_MONITOR_ENABLED:
             try:
-                BrewUMBListener.consume()
+                BrewBuildUMBListener.consume()
```
mprpic, line 14: You added two new listeners but they aren't actually being started to consume the messages from their configured topics, no? Or am I missing something?

corgi/monitor/consumer.py
```diff
@@ -77,19 +84,69 @@ def on_message(self, event: Event):
...
+class BrewTagUMBReceiverHandler(UMBReceiverHandler):
+    """Handle messages about Brew builds that have tags added or removed"""
+
+    def on_message(self, event: Event) -> None:
+        logger.info("Received UMB event on %s: %s", self.virtual_topic_address, event.message.id)
+        message = json.loads(event.message.body)
+        build_id = message["build"]["build_id"]
+
+        tag_added_or_removed = message["tag"]["name"]
+        if self.virtual_topic_address.endswith(".tag"):
```
mprpic, line 96: Can this instead check the class name (or type) instead of an arbitrary part of the virtual topic?

corgi/tasks/brew.py
```diff
@@ -590,3 +591,37 @@ def fetch_unprocessed_brew_tag_relations(
...
+        if not build:
+            logger.warning(f"Brew build with matching ID not ingested (yet?): {build_id}")
+            return
+
+        if tag_added:
+            build.meta_attr["tags"].append(tag_added)
+            errata_tag = ADVISORY_REGEX.match(tag_added)
+            if errata_tag:
+                # Below should automatically create new relations for this build / erratum
+                slow_load_errata.delay(errata_tag.group())
```
mprpic, line 617: Should this be scheduled only after the build is saved?

corgi/monitor/management/commands/runmonitor.py
juspence: You are correct...thank you for catching this. I need to go reread all this code apparently, no rush to review the rest here when I missed something this obvious :facepalm:

corgi/tasks/brew.py
juspence: Probably, yes. I don't think this will cause an issue with the task or creating the relations, but better safe than sorry.

corgi/monitor/consumer.py
juspence: I was going to make this change, but looking again I don't think it's possible. The class here is BrewTagUMBReceiverHandler, which is registered to either BrewTagUMBListener or BrewUntagUMBListener.

From the listener, you can check the handler class, but I don't think going the other way around is possible. The handler class also has no useful variables to indicate which receiver is using it.

The message itself has this data in the headers (topic name and type like "Tag" or "Untag"), but at that point we may as well inspect the virtual_topic_address variable instead of a message format that might change.

I could split the handler class into two - one for tag and one for untag events. But the message format and info provided are exactly the same in both, so that feels like pointless duplication.

mprpic: Gotcha, keep it as is then.

corgi/monitor/management/commands/runmonitor.py
juspence: After rereading, I actually think this whole approach seems questionable :frowning:. The consume() method we call here is just a wrapper around Container(SomeMessagingHandler).run(). Docs aren't very detailed:
https://qpid.apache.org/releases/qpid-proton-0.38.0/proton/python/docs/proton.reactor.html#proton.reactor.Container.run

But based on that run() method's code and the try/except example code we copied, the listener will enter an infinite loop and process each message as it comes in (until it's interrupted like with Ctrl-C). So I don't think we can run multiple listeners in the same Python process / module's "handle" function / management command.

Our run_umb_monitor.sh script also execs on this command ("python3 manage.py runmonitor"). I could change the bash script to fork a different process for each listener we want to run. But then we run into issues with multiple processes inside a container, lack of PID reaping, failure to catch errors when one process dies, etc.

Or I could change the Python code to run mutliple listener.consume() calls in the same process, but different threads. Each listener is fully independent, and the code should be I/O bound rather than CPU-bound, so this shouldn't cause conflicts or performance issues. But it doesn't reduce the overall complexity, and I have no idea if we'll handle errors in one thread in a sane way.

Running 3 copies (for high availability) of a UMB pod for each listener (builds completed, tags added, tags removed) seems like a ridiculous amount of overhead.

We already have a Celery task that pre-generates a manifest for some stream. Maybe in that task we should just manually refresh tags on the latest builds in that stream, before we build the manifest?

juspence:
Based on silly mistakes I made and Jason's feedback in standup, the approach here will change to support:
1. Processing UMB messages from the ET topic instead of the Brew topic, to support other build systems like PNC in the future
2. Listening to multiple topics / creating multiple receivers for one handler. This change isn't complicated (actually already done) but needs testing and not rushing.

So going to close this for now and come back to it (hopefully) later this week.

PR #415, opened 2023-04-25
CORGI-595 switch from version_utils python package to python3-rpm package

jasinner:
The version_utils python package does not support comparison of modular rpms. This change switches to the python3-rpm system package preinstalled in UBI-8 and Fedora. We install the 'rpm' python package to ensure we have access to the package in a virtual environment.


jasinner:

JimFuller-RedHat:
LGTM

corgi/core/models.py
```diff
@@ -861,11 +861,36 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+        return (
+            epoch,
+            evr[2],
+            evr[3],
+        )
+
     def filter_latest_nevra_by_name(self, component_name: str) -> str:
-        nevras = self.filter(name=component_name).values_list("nevra", flat=True)
+        nevras = self.filter(name=component_name).values_list(
+            "nevra", "meta_attr__epoch", "version", "release"
```
juspence, line 888: Can we migrate these epoch values (when present) to a real model field, and default the field to 0 if unset? I think that's simpler than looking up the meta_attr data (which we're not supposed to rely on directly) then trying to sanitize it on every call to the latest filter.

juspence, line 888: Is the first value in this list supposed to be "name" instead of nevra?

juspence (approved):
Don't have time for an in-depth review right now, but we have tests here and this will fix an existing bug, so I guess LGTM.

I'm a little worried this new library may not handle edge-cases the old one did (given we have so much quirky data), but that's not a problem we can (or should) solve right now.

corgi/core/models.py
jasinner: I guess we could select the name and build an NVR out of the values. I just stuck with the original field 'nevra' as the value to return.

jasinner: The problem I see with that is that most Components won't have this value set. It's only very rarely used, even for RPMs. Therefore I think meta_attr is a suitable place for it.

juspence: Maybe I'm confused but looking this over again, do you need the nevra / name value at all? You pass a tuple of (nevra, epoch, version, release) to the _compare_packages function.

The same (nevra, epoch, version, release) tuple gets passed to the ensure_epoch function, which always returns a tuple of only (epoch, version, release) to be compared by labelCompare.

PR #418, opened 2023-04-26
Add UUID based permission class

kgrant-rh:
Adds a new permission class to allow permitting specific users by their UUID. 

See https://issues.redhat.com/browse/CORGI-610

### DIFF
--- a/corgi/core/authentication.py
+++ b/corgi/core/authentication.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from collections.abc import Iterable
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review when you have a chance.

juspence:
Thanks for adding the tests, but I'm not sure I understand the context here. Can you add the Jira ticket number to the PR title, or add a detailed commit message that explains why this change is needed?

corgi/core/authentication.py
```diff
@@ -83,3 +84,45 @@ def has_permission(self, request: Any, view: Any) -> bool:
...
         return set(user_roles).intersection(set(view.roles_permitted)) != set()
+
+
+# drf's BasePermission seems to use metaclasses in a way mypy doesn't like
+class RedHatUUIDPermission(BasePermission):  # type: ignore[misc]
+    """A permission class that grants access to users specified by rhatUUID.
+    More than one user can be specified in a non-nested list, set, or tuple
+    of strings or UUIDs."""
+
+    ALLOWED_UUID_TYPES = (list, tuple, set, str, uuid.UUID)
```
juspence: Python has "Iterable" and "Collection" generic types. You might be able to just specify ```ALLOWED_UUID_TYPES = (collections.abc.Collection, str, uuid.UUID)``` instead.

corgi/core/authentication.py
```diff
@@ -83,3 +84,45 @@ def has_permission(self, request: Any, view: Any) -> bool:
...
+
+    def has_permission(self, request: Any, view: Any) -> bool:
+        if not hasattr(view, "uuids_permitted"):
+            raise ValueError(f"View {view} doesn't define any permitted UUIDs")
+
+        if type(view.uuids_permitted) not in RedHatUUIDPermission.ALLOWED_UUID_TYPES:
```
juspence:
```suggestion
        if not isinstance(view.uuids_permitted, RedHatUUIDPermission.ALLOWED_UUID_TYPES):
```

corgi/core/authentication.py
```diff
@@ -83,3 +84,45 @@ def has_permission(self, request: Any, view: Any) -> bool:
...
+            raise TypeError(
+                f"View {view} specifies UUIDs with unsupported type {type(view.uuids_permitted)}"
+            )
+
+        if type(view.uuids_permitted) in (list, tuple, set):
+            if not all([type(x) in (str, uuid.UUID) for x in view.uuids_permitted]):
```
juspence: Breaking this into multiple lines and adding a comment makes this a little clearer, assuming I understood this correctly:
```suggestion
        # if uuids_permitted is a collection, iterate over it to check all members are scalar (str or UUID)
        is_collection = isinstance(view.uuids_permitted, collections.abc.Collection)
        if is_collection and not all(isinstance(member, (str, uuid.UUID)) for member in view.uuids_permitted):
```

Checking each view dynamically seems a little odd. Maybe it's better to assert (in unit tests) that any view requiring authentication defines the uuids_permitted attribute, and populates it with the appropriate types of values.

But the current approach is also fine if you prefer this way, or if I missed some library behavior and this is required.

corgi/core/authentication.py
```diff
@@ -83,3 +84,45 @@ def has_permission(self, request: Any, view: Any) -> bool:
...
+                raise TypeError(
+                    f"Something other than str or UUID provided for permission in View {view}"
+                )
+
+        # Converting str to UUID can raise if the string is not a valid hexadecimal UUID
+        if isinstance(view.uuids_permitted, uuid.UUID):
+            permitted_uuids = [view.uuids_permitted]
+        elif isinstance(view.uuids_permitted, str):
+            permitted_uuids = [uuid.UUID(view.uuids_permitted)]
```
juspence: Rather than converting the values here, can we enforce that uuids_permitted is always a list, even if it contains only a single item? Like ```[uuid.UUID("str_val")]``` instead of ```uuid.UUID("str_val")```.

corgi/core/authentication.py
```diff
@@ -83,3 +84,45 @@ def has_permission(self, request: Any, view: Any) -> bool:
...
+        else:
+            str_uuids = [uuid.UUID(s) for s in view.uuids_permitted if isinstance(s, str)]
+            nonstr_uuids = [u for u in view.uuids_permitted if not isinstance(u, str)]
+            permitted_uuids = str_uuids + nonstr_uuids
```
juspence: Assuming we control these uuids_permitted values on each view, let's just set them to UUIDs to begin with if we know we need that format.

kgrant-rh: You're right, since we're the only users of this class, it makes everything easier to just require views that use it to declare permitted UUIDs as UUIDs.

kgrant-rh:
> Thanks for adding the tests, but I'm not sure I understand the context here. Can you add the Jira ticket number to the PR title, or add a detailed commit message that explains why this change is needed?

Sorry, I realized I added the issue ID in the branch name but forgot to add it to the description. It's https://issues.redhat.com/browse/CORGI-610 . The essential concern here is that it might not always be feasible to get certain roles added to certain users by certain deadlines, so this class lets us add permissions based on the rhatUUID claim of a user or application.

jasinner:
I guess it's good to have the capability to add specific user permissions to views. However I'm a little worried about how we are going to keep this up to date.

juspence (approved):
LGTM, thanks for explaining the context here and making those changes! Hopefully we'll never need this, but I guess it's good to have in case we ever need it.

PR #419, opened 2023-04-27
Fix community data

jasinner:
While testing the temporary community database I noticed we don't have daily jobs to create the yum and brew_tag relations, so I added those.


JimFuller-RedHat (approved):
LGTM ... a very simple change

PR #420, opened 2023-04-27
Fix softwarebuilds view

jasinner:
This allows lookup of softwarebuild by build_id which we lost when switching primary key to UUID. It also fixes an issue I got with the tests/test_api.py::test_latest_components_filter where if the type gets randomly selected as MAVEN we fail with uniqueness constraint on component Purl.


jasinner:

corgi/core/models.py
```diff
@@ -277,7 +277,7 @@ class Type(models.TextChoices):
     components: models.Manager["Component"]
 
     class Meta:
-        indexes = (models.Index(fields=("completion_time",)),)
```
JimFuller-RedHat, line 280: is this index not needed ?

juspence (approved):

corgi/core/models.py
jasinner: We don't sort by completion_time anyway, and that's why it was added, so I'd say we remove it for now.

PR #422, opened 2023-04-27
CORGI-591: Add automation to keep Brew / Corgi build tags in sync

juspence:
@RedHatProductSecurity/corgi-devs Please review. This change is needed to keep the tags up-to-date on Corgi builds whenever they change in Brew. Having up-to-date tags ensures that manifests always represent the latest released content for some stream and are never stale.

I added tests for the Celery task, but I'm not sure about the logic. Is it enough to call slow_load_errata() when errata tags get added to some build? Or is there something else I need to do, so that relations are created and the build is correctly linked to the erratum's variant?

I'm also not sure how to correctly clean up the relations, whenever some build has an errata tag removed. If someone has an easy idea for a solution, I can handle that as part of this MR. I could also file a backlog ticket, or we could avoid doing this at all if we never want to delete the relations (but that doesn't seem right).

### DIFF
--- a/corgi/monitor/consumer.py
+++ b/corgi/monitor/consumer.py
@@ -15,11 +15,13 @@ logger = logging.getLogger(__name__)
 class UMBReceiverHandler(MessagingHandler):
     """Handler to deal with received messages from UMB."""
 
-    def __init__(self, virtual_topic_address: str, selector: Optional[str] = None):
+    def __init__(self, virtual_topic_addresses: dict[str, str], selector: Optional[str] = None):
+        """Set up a handler that listens to many topics and processes messages from each"""
         super(UMBReceiverHandler, self).__init__()
 
-        # A virtual topic address determined by a specific listener.
-        self.virtual_topic_address = virtual_topic_address
+        # A mapping of virtual topic addresses to functions that handle topic messages
+        # as determined by a specific listener.
+        self.virtual_topic_addresses = virtual_topic_addresses
 
         # Set of URLs where UMB brokers are running. Use AMQP protocol port numbers only. See
         # specific URLs in settings; use env vars to select the appropriate broker.
@@ -48,16 +50,44 @@ class UMBReceiverHandler(MessagingHandler):
         # message that is accepted automatically (this is the default value).
         self.auto_settle = True
 
-    def on_start(self, event: Event):
+    def on_start(self, event: Event) -> None:
+        """Connect to UMB broker(s) and set up a receiver for each virtual topic address"""
         recv_opts = [self.selector] if self.selector is not None else []
         logger.info("Connecting to broker(s): %s", self.urls)
         conn = event.container.connect(urls=self.urls, ssl_domain=self.ssl_domain, heartbeat=500)
-        event.container.create_receiver(
-            conn, self.virtual_topic_address, name=None, options=recv_opts
-        )
+        for virtual_topic_address in self.virtual_topic_addresses:
+            event.container.create_receiver(
+                conn, virtual_topic_address, name=None, options=recv_opts
+            )
 
-    def on_message(self, event: Event):
-        logger.info("Received UMB event on %s: %s", self.virtual_topic_address, event.message.id)
+    def on_message(self, event: Event) -> None:
+        """Route message to a handler function, based on the virtual topic it was received on"""
+        logger.info("Received UMB event on %s: %s", event.message.address, event.message.id)
+        # Convert general topic address into specific address we listen on
+        address = event.message.address or ""
+        address = address.replace("topic://", f"Consumer.{settings.UMB_CONSUMER}.")
+
+        # Turn function name (str) into callable so we can pass event to it below
+        # We don't pass the callable itself because it needs a self arg
+        callback_name = self.virtual_topic_addresses.get(address, "")
+        callback_function = getattr(self, callback_name, None)
+
+        if not address:
+            raise ValueError(f"UMB event {event.message.id} had no address!")
+        elif callback_function:
+            callback_function(event)
+        else:
+            raise ValueError(
+                f"UMB event {event.message.id} had unrecognized address: {event.message.address}"
+            )
+
+
+class BrewUMBReceiverHandler(UMBReceiverHandler):
+    """Handle messages about completed Brew builds, tagged builds, and untagged builds"""
+
+    def handle_builds(self, event: Event) -> None:
+        """Handle messages about completed Brew builds"""
+        logger.info("Handling UMB event for completed builds: %s", event.message.id)
         message = json.loads(event.message.body)
         build_id = message["info"]["build_id"]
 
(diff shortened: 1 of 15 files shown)


corgi/monitor/consumer.py
```diff
@@ -48,16 +50,44 @@ def __init__(self, virtual_topic_address: str, selector: Optional[str] = None):
...
         recv_opts = [self.selector] if self.selector is not None else []
         logger.info("Connecting to broker(s): %s", self.urls)
         conn = event.container.connect(urls=self.urls, ssl_domain=self.ssl_domain, heartbeat=500)
-        event.container.create_receiver(
-            conn, self.virtual_topic_address, name=None, options=recv_opts
-        )
+        for virtual_topic_address in self.virtual_topic_addresses:
+            event.container.create_receiver(
+                conn, virtual_topic_address, name=None, options=recv_opts
+            )
```
mprpic, line 61: Did you test that this works? The docs around Receivers aren't super clear (and the code is not super readable either) so it's not apparent to me whether a Container supports multiple Receivers over a single Link. My worry here would be that if it only supports one, you're essentially overwriting receivers in a single container until you reach the last one.

juspence: Not yet, but if the overall MR looks good, I'm definitely planning on deploying to stage and testing before I merge.

mprpic: You can probably test this locally using your own UMB certs (or using the app stage ones) to connect to UMB stage :wink:

juspence: I tried a while ago to set up my own certs, but hit some issues with RHEL CSB I think. And I think if I use the stage ones, it will end up interfering with the queues / messages that stage should be processing.

tests/test_tasks.py
```diff
@@ -0,0 +1,127 @@
...
+    assert build.meta_attr["released_errata_tags"] == clean_released_tags
+
+
+def test_slow_update_brew_tags_removed():
+    """Test that builds have their tags updated, but not relations"""
+    tag_removed = RELEASED_TAGS[-1]
+    tag_removed_without_suffix = tag_removed.rsplit("-", maxsplit=1)[0]
+    build = SoftwareBuildFactory(build_type=SoftwareBuild.Type.BREW, meta_attr={"tags": CLEAN_TAGS})
+
+    with patch("corgi.tasks.brew.slow_load_errata.delay") as mock_load_errata:
```
jasinner, line 87: nitpik: In other tests we're using the @patch decorator, perhaps you should use that to keep the test consistent? Or did you have a specific reason to use this context method to patch the task?

corgi/tasks/brew.py
```diff
@@ -595,3 +596,40 @@ def fetch_unprocessed_brew_tag_relations(
...
+        else:
+            build.meta_attr["tags"].remove(tag_removed)
+            # TODO: Clean up old relations for some build / erratum when a tag is removed
+
+        build.meta_attr["errata_tags"] = Brew.extract_advisory_ids(build.meta_attr["tags"])
+        build.meta_attr["released_errata_tags"] = Brew.parse_advisory_ids(
+            build.meta_attr["errata_tags"]
+        )
+        build.save()
+        return f"Added tag {tag_added} or removed tag {tag_removed} for build {build_id}"
```
jasinner, line 635:
```suggestion
        if tag_added:
            return "empty tag, not adding"

        def add_if_not_existing(value: str, target: list[str]) -> tuple[list[str], bool]:
            if value not in target:
                target.append(value)
                return sorted(target), True
            return target, False


        build.meta_attr["tags"], tags_updated = add_if_not_existing(tag_added, build.meta_attr["tags"])

        # Check if it's an errata_tag
        errata_tag = ADVISORY_REGEX.match(tag_added)
        if not errata_tag:
            if tags_updated:
                build.save()
            return f"{tag_added} not an errata tag"

        errata_tag = errata_tag.group()
        build.meta_attr["errata_tags"], _ = add_if_not_existing(errata_tag, build.meta_attr["errata_tags"])

        # We might consider refactoring this code from brew collector to a common base package
        if len(errata_tag.split(":", maxsplit=1)[-1]) != 4:
            build.save()
            return f"{tag_aded} not a released errata tag"

        build.meta_attr["released_errata_tags"], released_updated = add_if_not_existing(errata_tag, build.meta_attr["released_errata_tags"])
        if release_updated:
            slow_load_erata.delay(released_errata_tag)

        # If we got this far either one of errata_tags or released_errata_tag meta_attr values needs to be saved
        build.save()
        return f"Added tag {tag_added} or removed tag {tag_removed} for build {build_id}"
```

jasinner (changes requested):
Calling slow_load_errata from the slow_update_brew_tags function is the correct approach to ensure that the builds have the correct product taxonomy. 

I think we should modify the slow_update_brew_tags as suggested. This might mean we also need to update when we call slow_load_errata in the slow_fetch_brew_build job so that it's only called for release_errata tags, not all errata tags.

corgi/tasks/brew.py
juspence: I'm not sure I understand this suggestion, but I think you're proposing:

1. Only add tags, never remove them
2. Only update errata_tags if the tag really is an erratum name
3. Only update released_errata_tags if the tag really is released
4. Only call .save() if something was actually added in 1-3 above
5. Only call slow_load_errata() / create relations for released errata, not all errata

Number 4 makes sense, but the existing code should handle this as far as I understand. We shouldn't get messages about added tags which are already present/ removed tags which are already gone. So we'll always need to call .save(), because at least meta_attr["tags"] will get updated every time.

You added a comment, ```# We might consider refactoring this code from brew collector to a common base package```. But the custom code you added is already part of a separate method that we're reusing. Brew.extract_advisory_ids() and Brew.parse_advisory_ids() are the same methods we use during ingestion to set the errata_tags and released_errata_tags, respectively.

If the tag we added (or removed) isn't an erratum name or isn't released, the list of errata_tags and released_errata_tags won't change, respectively. So calling these methods might be a no-op, and they handle checking the tag format for us. I don't think having separate code to check "is this tag an erratum name?" is really needed in this task. AKA numbers 2 and 3 are handled - the methods we reuse are fairly lightweight no-ops in this case.

Number 5 makes sense, but doesn't match what we do during ingestion. We create relations for all errata, not just the released errata. I'm not sure where else we rely on the ERRATA-type relations, or what the impact would be if only released errata had relations created instead of all errata. Could this cause problems for Incident Response? I assume that when they filter by a particular variant, they want to see components / builds that aren't released yet but which are planned to be released in that variant.

I'm happy to make the change in both this task and the ingestion logic, but maybe we need a discussion about when we want to relate components to variants (at release time or at build time). For now only the manifests filter on "released components" and exclude unreleased components.

I don't really understand number 1 (don't remove tags) - can you clarify a bit? Or we can discuss on a call if you'd prefer.

tests/test_tasks.py
juspence: I like the context manager because it makes it clear which mock is patching which call. This makes longer tests more readable, and avoids issues where you might want to patch some calls but not others. For example, some of the UMB monitor tests have more-complex patching needs.

corgi/monitor/consumer.py
juspence: I see this is working after deploying to stage temporarily for testing. Splunk logs indicate that we're receiving, handling, and logging messages from multiple topics. I can't put the internal Splunk search link on this public discussion page, but you can search for "Handling UMB" to see the different brew.build.complete and brew.build.tag messages we're processing.

corgi/tasks/brew.py
juspence: @jasinner After thiking about it, number 1 above (don't remove tags) doesn't seem like the correct behavior, and number 5 (create relations for only released errata, instead of all errata) seems outside the scope of this ticket. 2 - 4 are already handled by the current code.

I've tested and confirmed multiple receivers / topics per handler works, so I've filed CORGI-618 as a followup ticket to handle number 5, and I'm going to merge the rest as-is. Let's please discuss if you have any concerns.

jasinner (approved):

PR #423, opened 2023-04-28
CORGI-613: Add Django/DRF Token Authentication

kgrant-rh:
https://issues.redhat.com/browse/CORGI-613

This adds an example view and tests to demonstrate using DRF Token-based authentication only on write operations.


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review. My intent is that we can add this type of authentication transparently to endpoints that need write capabilities.

juspence (approved):
LGTM! Only two minor notes, feel free to ignore and just merge:

1. It might be good to add type hints on the new view methods.

2. In the tests we import the view directly, build a request using RequestFactory, and call some_view.as_view()(constructed_request). This is fine for testing the view logic, but it occurs to me that (I think) we aren't testing the API's routers / URL-handling logic. Or maybe I'm mistaken, I don't know much about Django's built-in RequestFactory.

In other tests we just do requests.get(test_host + "/path/to/endpoint") directly. That ensures we actually register the view and can direct traffic / requests to it. No need to make changes here, but might be something to keep in mind for the future.

PR #424, opened 2023-05-01
Enable write for authenticated users

kgrant-rh:
https://issues.redhat.com/browse/CORGI-234

These changes enable authenticated users to write component data.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -559,10 +559,6 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
     filterset_class = ComponentFilter
     lookup_url_kwarg = "uuid"
 
-    # Require authentication for write operations
-    authentication_classes = [TokenAuthentication]
-    permission_classes = [IsAuthenticatedOrReadOnly]
-
     def get_queryset(self) -> QuerySet[Component]:
         # 'latest' and 'root components' filter automagically turn on
         # when the ofuri parameter is given
@@ -649,8 +645,13 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
         manifest = json.loads(obj.manifest)
         return Response(manifest)
 
-    @action(methods=["put"], detail=True)
-    def olcs_test(self, request: Request, uuid: Union[str, None] = None) -> Response:
+    @action(
+        methods=["put"],
+        detail=True,
+        authentication_classes=[TokenAuthentication],
+        permission_classes=[IsAuthenticatedOrReadOnly],
+    )
+    def update_license(self, request: Request, uuid: Union[str, None] = None) -> Response:
         """Allow OpenLCS to upload copyright text / license scan results for a component"""
         # In the future these could be separate endpoints
         # For testing we'll just keep it under one endpoint
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -655,10 +655,6 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
         """Allow OpenLCS to upload copyright text / license scan results for a component"""
         # In the future these could be separate endpoints
         # For testing we'll just keep it under one endpoint
-        if settings.COMMUNITY_MODE_ENABLED:
-            # This is only temporary for OpenLCS testing
-            # Do not enable in production until we add OIDC authentication
-            return Response(status=status.HTTP_403_FORBIDDEN)
         component = self.queryset.filter(uuid=uuid).using("default").first()
         if not component:
             return Response(status=status.HTTP_404_NOT_FOUND)
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -27,7 +27,6 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 from rest_framework.viewsets import GenericViewSet, ReadOnlyModelViewSet
 
-from config import utils
 from corgi import __version__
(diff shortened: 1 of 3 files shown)


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review. NB TokenAuthentication will be swapped out for OIDC as soon as is practicable.

corgi/api/views.py
```diff
@@ -560,6 +559,10 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
     filterset_class = ComponentFilter
     lookup_url_kwarg = "uuid"
 
+    # Require authentication for write operations
+    authentication_classes = [TokenAuthentication]
+    permission_classes = [IsAuthenticatedOrReadOnly]
+
```
juspence: Just to confirm my own understanding...I think this will allow any logged-in user to write to any endpoint in this viewset (including ones we add in the future) that supports PUT / POST / PATCH actions. Is that right? And right now, only the olcs_test endpoint should support a PUT action / define a method to handle these PUT requests.

I think our intention was to only allow OpenLCS to PUT updated license data on an existing component, and not to POST a completely new component. So I think this is OK for now, but maybe in the future we'll need to make sure that:
1. Other users (not OpenLCS) can't PUT license data
2. OpenLCS can't POST a new component

Do we also need to define uuids_permitted on this viewset, so that OpenLCS can upload data?
EDIt: Ah nvm, I guess that's not relevant for TokenAuthentication, so we'd only add that in the future when we switch back to OIDCAuthentication.

corgi/api/views.py
```diff
@@ -651,7 +654,7 @@ def olcs_test(self, request: Request, uuid: Union[str, None] = None) -> Response
         """Allow OpenLCS to upload copyright text / license scan results for a component"""
```
juspence, line 655: It's outside the diff window so I can't comment on it directly, but the name of this method should probably be changed if we're enabling this in production. We could call it "update_license" or something similar. And we probably need to coordinate with OpenLCS for testing, but I assume you're handling that already.

tests/test_api.py
```diff
@@ -388,11 +403,12 @@ def test_component_detail_dev(client, api_path):
 @pytest.mark.django_db
 def test_component_write_not_allowed(client, api_path):
```
juspence: Let's test what happens If the user authenticates. AKA does the PUT / POST succeed, or does it continue to fail with the same 405 error as below?
```suggestion
@pytest.mark.django_db
def test_component_write_not_allowed_if_authenticated(client, api_path):
    c = ComponentFactory(name="curl")

    # User for OpenLCS authentication
    olcs_user = User.objects.create_user(username="olcs", email="olcs@example.com")
    olcs_token = Token.objects.create(user=olcs_user, key="mysteries_quirewise_volitant_woolshed")
    
    # Authenticate
    client.credentials(HTTP_AUTHORIZATION=f"Token {olcs_token.key}")

    response = client.put(f"{api_path}/components/{c.uuid}", params={"name": "wget"})
    assert response.status_code == 405
    response = client.post(f"{api_path}/components", data={"name": "curl"})
    assert response.status_code == 405


@pytest.mark.django_db
def test_component_write_not_allowed_if_unauthenticated(client, api_path):
```

juspence (approved):
LGTM with two minor change to tests and the endpoint name. You probably need to run "tox -e schema" and commit the updated openapi.yml file before the pipeline will pass.

corgi/api/views.py
kgrant-rh: I've restricted the authentication to just the license API endpoint, so write methods on other endpoints should fail now even for authenticated users. Also, yeah, we'll need to specify either the appropriate SSO groups or UUIDs, but only when we make the change to use that form of authentication.

corgi/api/views.py
```diff
@@ -646,12 +645,17 @@ def manifest(self, request: Request, uuid: str = "") -> Response:
...
+    )
+    def update_license(self, request: Request, uuid: Union[str, None] = None) -> Response:
         """Allow OpenLCS to upload copyright text / license scan results for a component"""
         # In the future these could be separate endpoints
         # For testing we'll just keep it under one endpoint
-        if utils.running_prod() or settings.COMMUNITY_MODE_ENABLED:
+        if settings.COMMUNITY_MODE_ENABLED:
             # This is only temporary for OpenLCS testing
             # Do not enable in production until we add OIDC authentication
             return Response(status=status.HTTP_403_FORBIDDEN)
```
jasinner: We can remove this check and failure for COMMUNITY_MODE_ENABLED too as the endpoint I think. Since we don't have CORGI_OIDC_AUTH_ENABLED set in community environment the OIDC_AUTH_ENABLED django setting is correctly being set to False. That should mean we always get a 403 for this endpoint right?

```
$ oc project
Using project "corgi--community" on server "example.com:6443".
$ oc rsh dc/corgi-web
$ python3 manage.py shell
>>> from django.conf import settings
>>> settings.OIDC_AUTH_ENABLED
0
```

juspence: I don't think OIDC_AUTH_ENABLED controls anything, since we're using the TokenAuthentication class / backend and not OIDCAuthentication.

But I think you're right that we can remove the community check, since the endpoint should require a Django token even in community mode. @kgrant-rh Did I miss something?

kgrant-rh: Nope, @juspence is correct in that token authentication would still nominally let users of the community instance write if they authenticated since OIDCAuthentication is not currently enabled, however, since we don't have any users in our community instances, it shouldn't be an issue.

PR #425, opened 2023-05-02
make latest filter only return root components

jasinner:
This modifies the latest filter so that always filters root_component as well. Comparing components which are not root_components doesn't really makes sense for a few reasons:

- We are using RPM schematics to compare things, so comparing anything but RPM, or OCI types in the REDHAT namespace doesn't make sense.  
- All arches should be considered equivalent in terms of latest.

Leaving this in draft as I got a failing test locally:
```
>       assert manifest["relationships"][0] == provided_contained_by_component
E       AssertionError: assert {'relatedSpdx...32b765159d68'} == {'relatedSpdx...32b765159d68'}
E         Omitting 2 identical items, use -vv to show
E         Differing items:
E         {'relatedSpdxElement': 'SPDXRef-e24f6913-d568-4e9e-a359-570e53706ad3'} != {'relatedSpdxElement': 'SPDXRef-95582f58-12c8-4fdd-bf3e-86a8a72a2724'}
E         Use -v to get more diff

tests/test_manifests.py:200: AssertionError
```
Hopefully that is just a change in the order.

### DIFF
--- a/corgi/core/files.py
+++ b/corgi/core/files.py
@@ -68,7 +68,7 @@ class ProductManifestFile(ManifestFile):
 
         components = self.obj.components  # type: ignore[attr-defined]
         components = components.exclude(name__endswith="-container-source").using("read_only")
-        released_components = components.root_components().released_components().latest_components()
+        released_components = components.released_components().latest_components()
         distinct_provides = self.obj.provides_queryset  # type: ignore[attr-defined]
 
         kwargs_for_template = {
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -890,8 +890,10 @@ class ComponentQuerySet(models.QuerySet):
         )
 
     def filter_latest_nevra_by_name(self, component_name: str) -> str:
-        nevras = self.filter(name=component_name).values_list(
-            "nevra", "meta_attr__epoch", "version", "release"
+        nevras = (
+            self.root_components()
+            .filter(name=component_name)
+            .values_list("nevra", "meta_attr__epoch", "version", "release")
         )
         if nevras:
             # Get the latest NEVRA using python. This only works for valid RPM epoch/version/release
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -184,8 +184,8 @@ def test_component_detail(client, api_path):
 
 @pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
 def test_latest_components_filter(client, api_path):
-    older_component = ComponentFactory(type=Component.Type.RPM, release="9")
-    newer_component = ComponentFactory(
+    older_component = SrpmComponentFactory(type=Component.Type.RPM, release="9")
+    newer_component = SrpmComponentFactory(
         type=older_component.type,
         name=older_component.name,
         version=older_component.version,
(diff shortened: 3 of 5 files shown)


JimFuller-RedHat:
picking this up

JimFuller-RedHat:
this looks fine to me - though conceptually (for downstream like griffon) will need to do more ... for example

```
/api/v1/components?name=grafana&latest_components=True
```

Only returns a single src ... presumably this is the _latest component of the latest stream_. 

What we need is _latest component(s) by stream_ as we have not constrained using **&product_streams=** or **&ofuri=** we want a list of components existing across different product streams - which as discussed with @jasinner will do as a different PR.

juspence:
> This modifies the latest filter so that always filters root_component as well. Comparing components which are not root_components doesn't really makes sense for a few reasons:
> 
>     * We are using RPM schematics to compare things, so comparing anything but RPM, or OCI types in the REDHAT namespace doesn't make sense.

Instead of always filtering root_components when looking for latest_components, can we filter by ```type__in("RPM", "OCI")```? That limits the latest filter to types where it actually works, but doesn't bake in the "root components" logic into a separate filter. There have been requests in the past where teams wanted to see the latest binary RPM / arch-specific container, and not only the root components.

jasinner:
> > This modifies the latest filter so that always filters root_component as well. Comparing components which are not root_components doesn't really makes sense for a few reasons:
> > ```
> > * We are using RPM schematics to compare things, so comparing anything but RPM, or OCI types in the REDHAT namespace doesn't make sense.
> > ```
> 
> Instead of always filtering root_components when looking for latest_components, can we filter by `type__in("RPM", "OCI")`? That limits the latest filter to types where it actually works, but doesn't bake in the "root components" logic into a separate filter. There have been requests in the past where teams wanted to see the latest binary RPM / arch-specific container, and not only the root components.

The problem with that is we might return any arch as the latest, as they are all considered equal. For example we could randomly select the x86_64 arch as the latest, so we'd have to add some logic filter the results to only include RPM/src and noarch containers. That is definitely possible, but just not something we have time for at the moment.

juspence:
> The problem with that is we might return any arch as the latest, as they are all considered equal. For example we could randomly select the x86_64 arch as the latest, so we'd have to add some logic filter the results to only include RPM/src and noarch containers. That is definitely possible, but just not something we have time for at the moment.

Yes, this is what I keep offering to help with. I know exactly where this is needed and left a comment describing that in the Jira ticket. I can run my "refresh tags" script in the background while working on fixing this bug I introduced - please let me know if I should pick it up tomorrow.

jasinner:
> > The problem with that is we might return any arch as the latest, as they are all considered equal. For example we could randomly select the x86_64 arch as the latest, so we'd have to add some logic filter the results to only include RPM/src and noarch containers. That is definitely possible, but just not something we have time for at the moment.
> 
> Yes, this is what I keep offering to help with. I know exactly where this is needed and left a comment describing that in the Jira ticket. I can run my "refresh tags" script in the background while working on fixing this bug I introduced - please let me know if I should pick it up tomorrow.

We could abandon this change in light of the Summit focus, but if you think it won't take you long and it's worth while, please take a stab at it. Otherwise, let's just file a follow up task to make sure we don't use latest filter on non-RPM versioned component types. Jim confirmed he doesn't need this change to resolve CORGI-609.

juspence:
I think it's needed for the Summit / manifesting focus, since right now the latest_components() filter gets used by manifests and is sometimes returning the wrong result. So going to pick this up.

juspence:
Fixed latest filter regression in #434 and made tests much more thorough to catch these kinds of bugs.

PR #426, opened 2023-05-02
add include/exclude filters to component products

JimFuller-RedHat:
Recently we added include/exclude filters to **sources**, **provides** and **upstreams** ... this PR extends the same idiom to other product related sub entities eg. **products**, **product_versions**, **product_streams**, **product_variants** and **channels** enabling selection of these entity attributes.

```
/api/v1/components?name=grafana&include_fields=purl,name,product_streams.name
```

which is very handy for downstream clients (like griffon).

**Note** - this PR maintains the status quo if no product specific include/excludes are defined.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -268,6 +268,19 @@ class IncludeExcludeFieldsSerializer(serializers.ModelSerializer):
         else:
             return True
 
+    def get_include_exclude_serializer(self, fieldname, serializer, instance, many=True):
+        """return include exclude Serializer when specified"""
+        if self._next_level_include_fields.get(
+            fieldname, []
+        ) or self._next_level_exclude_fields.get(fieldname, []):
+
+            context = {
+                "include_fields": self._next_level_include_fields.get(fieldname, []),
+                "exclude_fields": self._next_level_exclude_fields.get(fieldname, []),
+            }
+            return serializer(instance=instance, many=many, read_only=True, context=context)
+        return None
+
 
 class TagSerializer(serializers.Serializer):
     name = serializers.SlugField(allow_blank=False)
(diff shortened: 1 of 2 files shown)


tests/test_api.py
```diff
@@ -805,6 +805,17 @@ def test_product_components_ofuri(client, api_path):
...
+    response = client.get(
+        f"{api_path}/components?ofuri=o:redhat:rhel:8.6.0&include_fields=purl,product_streams.name"
+    )
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+    assert "purl" in response.json()["results"][0]
+    assert len(response.json()["results"][0]["product_streams"]) == 1
+    assert "name" in response.json()["results"][0]["product_streams"][0]
+    assert "purl" not in response.json()["results"][0]["product_streams"][0]
+    assert response.json()["results"][0]["product_streams"][0]["name"] == "rhel-8.6.0"
```
juspence, line 853: Minor nit: Let's just convert ```response = response.json()``` one time to make this a little more readable:
```suggestion
    response = response.json()

    assert response["count"] == 1
    result = response["results"][0]

    assert "purl" in result
    assert len(result["product_streams"]) == 1
    assert "name" in result["product_streams"][0]

    assert "purl" not in result["product_streams"][0]
    assert reult["product_streams"][0]["name"] == "rhel-8.6.0"
```

corgi/api/serializers.py
```diff
@@ -338,30 +338,74 @@ class Meta:
...
+            }
+            serializer = ProductVariantSerializer(
+                instance=instance.productvariants, many=True, read_only=True, context=context
+            )
+            return serializer.data
         return get_product_data_list("product_variants", instance.productvariants)
 
-    @staticmethod
-    def get_channels(instance: Union[Component, ProductModel]) -> list[dict[str, str]]:
+    def get_channels(self, instance: Union[Component, ProductModel]):
```
juspence: It looks like you've removed the return type hint on all the updated methods here. Was that by mistake and this still returs a ```list[dict, str, str]```? Or can we no longer guarantee what's being returned?

corgi/api/serializers.py
```diff
@@ -338,30 +338,74 @@ class Meta:
...
+        if self._next_level_include_fields.get(
+            "channels", []
+        ) or self._next_level_exclude_fields.get("channels", []):
+            context = {
+                "include_fields": self._next_level_include_fields.get("channels", []),
+                "exclude_fields": self._next_level_exclude_fields.get("channels", []),
+            }
+            serializer = ChannelSerializer(
+                instance=instance.channels, many=True, read_only=True, context=context
```
juspence: We call ChannelSerializer with many=True, but instance.channels (or instance.productstreams, etc.) can return only one object. This happens when we're accessing .channels from a Channel, .productstreams from a stream, etc. It just returns a single self object in that case.

I think this case is handled in your test below, since you're using ?include_fields=product_streams.name on a ProductStream ofuri. But FYI in case you weren't aware - I'm not sure if there's other code here that would need to handle this case.

juspence (commented):
You've added a test for the product_streams case, but we have the same code being added for all ProductModel subclasses (as well as Channels). So test coverage is going to drop, and the non-stream methods will break (without making tests fail) if we have any typo there, or add one in the future.

Instead of adding lots of redundant tests that all check the same thing for each method, can we factor out the common code into a single method? It could just take a string "productmodel_type" as argument, like "products", "product_versions", etc.

I'm thinking of something similar to ```get_product_data_list("product_versions", instance.productversions)```, but for this new code, e.g. something like "get_productmodel_serializer_data". Then the one test we have will cover this method without needing to write new ones, and future changes only happen in one place so we won't have to remember to update others.

corgi/api/serializers.py
JimFuller-RedHat: adding back ;)

jasinner (approved):
LGTM

corgi/api/serializers.py
```diff
@@ -338,30 +351,52 @@ class Meta:
...
+        )
+        if include_exclude_serializer:
+            return include_exclude_serializer.data
         return get_product_data_list("product_variants", instance.productvariants)
 
-    @staticmethod
-    def get_channels(instance: Union[Component, ProductModel]) -> list[dict[str, str]]:
+    def get_channels(self, instance: Union[Component, ProductModel]) -> list[dict[str, str]]:
+        include_exclude_serializer = self.get_include_exclude_serializer(
+            "channels", ProductVariantSerializer, instance.channels, many=False
```
juspence, line 396: @JimFuller-RedHat Should this be ChannelSerializer?

JimFuller-RedHat: yes, nice spot, will raise PR

PR #427, opened 2023-05-02
CORGI-609: add latest_components_by_stream) view

JimFuller-RedHat:
Adds **?latest_components_by_stream=True** view which is used by downstream clients (like griffon).

For example, to return latest components with 'grafana' in the component internal name, across product streams
```
/api/v1/components?re_name=grafana&latest_components_by_streams=False&include_fields=purl,product_streams,namespace,arch,nvr&limit=200&namespace=REDHAT
```

**Note** - this isolates code path from latest_components which is probably safer.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -927,6 +927,55 @@ class ComponentQuerySet(models.QuerySet):
                 return self
             return self.exclude(query)
 
+    def latest_components_by_streams(
+        self,
+        include: bool = True,
+    ) -> "ComponentQuerySet":
+        """Return components from latest builds ROOT COMPONENTS by product streams."""
+        product_stream_uuids = set(
+            self.using("read_only")
+            .root_components()
+            .exclude(productstreams__isnull=True)
+            .values_list("productstreams__uuid", flat=True)
+            .distinct()
+        )
+        query = Q()
+        for ps_uuid in product_stream_uuids:
+            if ps_uuid:
+                names = set(
+                    self.using("read_only")
+                    .root_components()
+                    .prefetch_related("productstreams")
+                    .filter(productstreams=ps_uuid)
+                    .values_list("name", flat=True)
+                    .distinct()
+                )
+                for name in names:
+                    latest_nevra = (
+                        self.using("read_only")
+                        .root_components()
+                        .prefetch_related("productstreams")
+                        .filter(name=name, productstreams=ps_uuid)
+                        .filter_latest_nevra_by_name(component_name=name)
+                    )
+                    if latest_nevra:
+                        query |= Q(nevra=latest_nevra)
+        if include:
+            # Show only the latest components
+            if not query:
+                # No latest components to show??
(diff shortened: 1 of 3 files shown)


corgi/core/models.py
```diff
@@ -927,6 +927,38 @@ def latest_components(
...
             return self.exclude(query)
 
+    def latest_components_by_streams(
+        self,
+        include: bool = True,
+    ) -> "ComponentQuerySet":
+        """Return components from latest builds ROOT COMPONENTS by product streams."""
+        product_stream_uuids = set(
+            self.using("read_only")
+            .filter(ROOT_COMPONENTS_CONDITION)
```
juspence: It's not required, but it might be better to reuse ```.root_components()``` here to keep the logic of "what is a root component?" isolated to that one method / helper queryset.

corgi/core/models.py
```diff
@@ -927,6 +927,38 @@ def latest_components(
...
+            .distinct()
+        )
+        query = Q()
+        for ps_uuid in product_stream_uuids:
+            for name in names:
+                latest_nevra = (
+                    self.using("read_only")
+                    .filter(ROOT_COMPONENTS_CONDITION)
+                    .exclude(productstreams__isnull=True)
+                    .filter(productstreams=ps_uuid)
```
juspence: You shouldn't need both of these - when you filter for a particular productstream UUID, that should automatically show only components with a matching (non-null) UUID.

corgi/core/models.py
```diff
@@ -927,6 +927,38 @@ def latest_components(
...
+                    .filter_latest_nevra_by_name(component_name=name)
+                )
+                if latest_nevra:
+                    query |= Q(nevra=latest_nevra)
+        return self.using("read_only").filter(query).distinct()
```
juspence: If we have no latest components / NEVRAs in this queryset, we'll end up returning all the components, right? That's fine if it's what you're expecting.

But I think the .distinct() here isn't needed, since you're filtering by NEVRA and we enforce a uniqueness constraint on those. And leaving .distinct() might cause issues if we try to call it on all the components, when we had no latest_nevras / the Q() object is empty.

juspence (commented):
Needs to have tests added per recent decision in ADR. That will prevent us from breaking this (again) during some refactoring.

corgi/core/models.py
JimFuller-RedHat: thx, will adjust to use root_components() as we do want minimum set of ps streams to crunch on.

corgi/core/models.py
JimFuller-RedHat: we should not need it - though local tests showed product_streams=[] somehow passing through ... just occurred to me maybe we are getting an empty value in product_streams_uuids and looping through that ... will investigate

corgi/core/models.py
JimFuller-RedHat: ah this was left in as I was trying to determine leakage (mentioned in previous PR comment) ... will remove

jasinner (approved):
LGTM

PR #428, opened 2023-05-02
CORGI-622: Add Celery task to manually refresh tags on container builds

juspence:
@jasinner Please feel free to review, merge, and start running this task when you feel ready. No tests here since:
1. This code is temporary and should be removed after this one-time refresh
2. This is almost identical to code I've been running manually in a shell (and which I tested locally beforehand)
3. We're reusing existing functions to parse the list of tags (and those functions already have unit tests)

The task will time out - there are over 190,000 builds that need to be refreshed. You can just keep rerunning it. It will safely pick up where it left off, refreshing just the tags on container builds which don't define any released_errata_tags.

I don't want you to spend too much time on this, so please feel free to ignore if you have other more important tasks. I can continue refreshing the OCI builds when I get back in tomorrow morning.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -633,3 +633,43 @@ def slow_update_brew_tags(build_id: str, tag_added: str = "", tag_removed: str =
         )
         build.save()
         return f"Added tag {tag_added} or removed tag {tag_removed} for build {build_id}"
+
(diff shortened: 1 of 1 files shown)


jasinner (approved):
Ok, I will merge and run it today.

PR #430, opened 2023-05-03
CORGI-624 ensure that the OUTPUT_FILES_DIR exists on first run 

jasinner:
Since its now a subdirectory on the PVC, we need to ensure the OUTPUT_FILES_DIR exists before we starting writting working manifests to it.

### DIFF
--- a/corgi/tasks/manifest.py
+++ b/corgi/tasks/manifest.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


corgi/tasks/manifest.py
```diff
@@ -17,6 +19,8 @@
     retry_kwargs=RETRY_KWARGS,
 )
 def update_manifests():
+    # Ensure the OUTPUT_FILES_DIR exists on first run
+    Path(settings.OUTPUT_FILES_DIR).mkdir(exist_ok=True)
```
JimFuller-RedHat, line 23: alternately things like directory creation might be an activity to consider for component-registry-ops ansible ... otherwise LGTM

JimFuller-RedHat (approved):
LGTM

corgi/tasks/manifest.py
jasinner: Noted.

PR #434, opened 2023-05-03
CORGI-609: Fix latest filter regression

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have time. The latest filter refactoring we needed for manifests uncovered a bug, which only happens when we have multiple namespace / arch values for the same component name.

Previously, we always filtered for root_components() first. So the namespace was always "REDHAT", and the arch was always "src" for source RPMs or "noarch" for index containers, and the bug never occurred.

During the refactoring, we decoupled the root_components() and latest_components() filters based on user requests. This lets users see the latest versions of binary RPMs and arch-specific containers directly, without having to first find the latest root component then look at its "provides" field.

That decoupling caused us to sometimes report a binary RPM as the only "latest" component, when we had both a source RPM and a binary RPM with the same NVR. If the user wanted to filter the results further, they might end up with no results (e.g. if they applied the root_components filter, but there are only non-root results here).

The fix is to just filter on all three of (namespace, name, arch), instead of only filtering on name, to decide which component is latest. Now we'll report three results (one for an upstream component, one for a source component, one for a binary component) that the user can filter further based on their needs (show only upstream or root or non-root components).

In the future, we may also want to filter on the major RHEL version for layered products. For example, el_match[0] == "8" or "9", so that users can see the latest version when the product is running on top of RHEL8 vs. the latest version when the product is running on top of RHEL9. Since the two base products have different support lifecycles, we've been asked previously to support showing only the components that are in a particular version of the base product.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -20,6 +20,7 @@ from rpm import labelCompare
 from corgi.core.constants import (
     CONTAINER_DIGEST_FORMATS,
     EL_MATCH_RE,
+    EPOCH_TYPE,
     MODEL_NODE_LEVEL_MAPPING,
     RELEASE_VERSION_DELIM_RE,
     ROOT_COMPONENTS_CONDITION,
@@ -868,14 +869,15 @@ class ComponentQuerySet(models.QuerySet):
     """Helper methods to filter QuerySets of Components"""
 
     def _compare_packages(
-        self, nevr_1: tuple[str, str, str, str], nevr_2: tuple[str, str, str, str]
+        self, nevr_1: tuple[str, EPOCH_TYPE, str, str], nevr_2: tuple[str, EPOCH_TYPE, str, str]
     ) -> int:
         """Use the rpm library to compare 2 epoch/version/release tuples"""
         evr_1 = self._ensure_epoch(nevr_1)
         evr_2 = self._ensure_epoch(nevr_2)
         return labelCompare(evr_1, evr_2)
 
-    def _ensure_epoch(self, evr: tuple[str, str, str, str]) -> tuple[str, str, str]:
+    @staticmethod
+    def _ensure_epoch(evr: tuple[str, EPOCH_TYPE, str, str]) -> tuple[str, str, str]:
         """Ensure the epoch value is set to '0' if unset and also drop the nevra at index 0"""
         # Test for None or 0 int value
         if not evr[1]:
@@ -908,7 +910,7 @@ class ComponentQuerySet(models.QuerySet):
         self,
         include: bool = True,
     ) -> "ComponentQuerySet":
-        """Return components from latest builds."""
+        """Return components from latest builds across all product streams."""
         distinct_components = (
             self.values_list("namespace", "name", "arch").order_by().distinct().iterator()
         )
@@ -937,7 +939,7 @@ class ComponentQuerySet(models.QuerySet):
         self,
         include: bool = True,
     ) -> "ComponentQuerySet":
-        """Return components from latest builds ROOT COMPONENTS by product streams."""
+        """Return only root components from latest builds for each product stream."""
         product_stream_uuids = set(
             self.using("read_only")
             .root_components()
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -940,23 +940,30 @@ class ComponentQuerySet(models.QuerySet):
         include: bool = True,
     ) -> "ComponentQuerySet":
         """Return only root components from latest builds for each product stream."""
-        product_stream_uuids = set(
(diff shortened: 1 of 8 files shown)


corgi/core/models.py
```diff
@@ -940,23 +940,31 @@ def latest_components_by_streams(
...
+        product_stream_uuids = (
             self.using("read_only")
             .root_components()
             .exclude(productstreams__isnull=True)
             .values_list("productstreams__uuid", flat=True)
+            # Clear ordering inherited from parent Queryset, if any
+            # So .distinct() works properly and doesn't have duplicates
+            .order_by()
             .distinct()
+            .iterator()
```
JimFuller-RedHat, line 951: why would we use an iterator for a small flat list ? set is faster and I doubt the memory savings is anything but very small (we only have 300+ potential uuids)

juspence: You're right, I was quickly scanning this method and just replaced any set() I saw with .iterator() to match other code. This one case is probably fine, so I'll put it back.

EDIT: Although FYI I see 1200+ streams locally, but still probably not an issue.

jasinner:
I think we should return an error (bad request) when a user tries to use the latest_components API filter without specifying a name.

corgi/core/models.py
```diff
@@ -904,75 +910,84 @@ def latest_components(
         self,
         include: bool = True,
     ) -> "ComponentQuerySet":
-        """Return components from latest builds."""
-        names = self.values_list("name", flat=True).order_by().distinct().iterator()
+        """Return components from latest builds across all product streams."""
+        distinct_components = (
+            self.values_list("namespace", "name", "arch").order_by().distinct().iterator()
+        )
```
jasinner, line 916: There's no restriction on type, so if I include a name in the filter such as 'cloud.google.com/go/monitor', I get results such as 'pkg:golang/cloud.google.com/go/monitoring@v1.1.0' which uses semantic versioning, not RPM versioning, so the results could be incorrect.

```suggestion
        # This should probably go in constants.py
        RPM_SCHEMANTIC_CONDITION = Q(type="RPM") | Q(type="OCI", namespace="REDHAT")
        distinct_components = (
            self.filter(RPM_SCHEMANTIC_CONDITION).values_list("namespace", "name", "arch").order_by().distinct().iterator()
        )
```

juspence: If we're comparing e.g. two Golang components, the epoch should always be "0" and the release should always be an empty "" string, right? In that case, we'd compare the two components using only their version strings.

From what I remember of the two standards, I think the version comparison logic for a semver is simpler than the version comparison logic for an RPM version. I'm happy to add a test for this case, but I'm 99% sure that the filter / the library we use will give the correct results for semantic versions.

The user request was to see the latest version of all components in a stream, not just the latest version of RPM / container components, so I don't want to limit the results we return unless necessary. I'd also rather keep the scope of this ticket focused on fixing this particular bug (where we don't split results by arch or namespace). I don't want to change the logic of what the filter returns until we're sure there's a bug.

Happy to discuss in Gchat or on a call if you'd like - please just let me know.

juspence:
> I think we should return an error (bad request) when a user tries to use the latest_components API filter without specifying a name.

I don't understand this. Even before my changes, we've previously supported filtering all the components in a stream (with the ?ofuri parameter) and showing the latest version of each. We also need this latest_components() filter for manifests, again to show only the latest versions of all component in a stream (or in a variant, in the future - see CORGI-590).

Are you talking about the latest_components_for_stream filter that Griffon uses? I know in the old view code, and in the examples Jim gave in the ticket, we always called that filter for a particular name, e.g. webkitgtk. But I don't want to touch any filtering logic here or change what would be returned. I'm just trying to fix a very specific bug, where if we have a source RPM and a binary RPM with the same NVR, we only report one as "latest" when really both are "latest".

EDIT: Discussed on a call. For now we'll just remove this filter from the API, and only use this code for manifests / the ofuri parameter by calling it directly / as a queryset.

corgi/core/models.py
juspence: Discussed on a call. We probably need more complex logic to support this usecase. We could have different version-comparison logic for different component types (which use different version types, e.g. semver).

For now we'll only use the latest filter on RPMs and REDHAT containers, which we know use RPM-style versions. Containers used by managed services (and stored in Quay) may be of type UPSTREAM if a service just uses an upstream container directly without repackaging / rebuilding it in Brew / the REDHAT namespace. So we also need to filter on ```Q(namespace=REDHAT, type=OCI)```.

juspence:
Merging based on conversation with Jason. Pipeline failure is just due to an outage in the internal PyPI mirror, tests still pass locally.

PR #435, opened 2023-05-03
CORGI-336: Set ProductVariant CPE on creation, from CollectorErrataProductVariant

juspence:
@RedHatProductSecurity/corgi-devs This turned out to be nice and simple, so I've probably messed up :slightly_smiling_face: Is the Collector model the right value to use here? It looks like a full CPE, and not just a pattern like in prod-defs.

Looking at the stage data, I see 42 cases of a None / null CPE, and 22 of an empty "" string CPE, versus 1740 Collector variants that do have a CPE. I also don't think we ever clean up the old Collector models, so I'm assuming those are for old / stale variants that can be ignored.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -5,7 +5,10 @@ from celery_singleton import Singleton
 from django.db import transaction
 
 from config.celery import app
-from corgi.collectors.models import CollectorErrataProductVersion
+from corgi.collectors.models import (
+    CollectorErrataProductVariant,
+    CollectorErrataProductVersion,
+)
 from corgi.collectors.prod_defs import ProdDefs
 from corgi.core.models import (
     Product,
@@ -177,6 +180,7 @@ def update_products() -> None:
                                         name=et_variant.name,
                                         defaults={
                                             "version": "",
+                                            "cpe": et_variant.cpe,
                                             "description": "",
                                             "products": product,
                                             "productversions": product_version,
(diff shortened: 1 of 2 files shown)


mprpic:
@juspence Have you tried looking at the streams that we used for the beta SBOMs and seeing if after loading CPEs for all variants we get the same ones as those in CORGI-623 ?

I'd also be very surprised if all ps_update_stream have a correct list of Variants or Brew tags defined on them. Taking the preliminary list of streams in CORGI-627, can we see how many are missing these?

corgi/tasks/prod_defs.py
```diff
@@ -177,6 +177,7 @@ def update_products() -> None:
                                         name=et_variant.name,
                                         defaults={
                                             "version": "",
+                                            "cpe": et_variant.cpe,
```
juspence, line 183: Note to self: I should add a migration to actually set these values. They should also get refreshed during the daily load task but it shouldn't hurt to force setting them.

juspence: Not going to bother, I'll just manually run the task to update the prod-defs data since I have to manually regenerate all the manifests anyway.

juspence:
Data quality review is finished. Detailed notes are in CORGI-623 and CORGI-627, but we have new tests that are passing, and this looks like a significant improvement with no regressions / unexpected values that I can see, so going to merge as-is. We can file follow-up tickets if needed for any variants that are still missing CPEs.

PR #436, opened 2023-05-03
Add UMB listener for sbomer SBOM availability

kgrant-rh:


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review. Note that the parsing of the SBOMs is stubbed out, to be completed in CORGI-488.

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        # TODO: Be paranoid and restrict the URL to a specific host or domain?
+        r = requests.get(sbom_data["link"])
+        r.raise_for_status()
+    except HTTPError:
+        logger.warning("SBOM fetch failed to fetch SBOM: %s", sbom_data["link"])
+        return
+
+    try:
+        for component in CycloneDxSbom.parse(r.text):
```
mprpic: You can just pass `r.json()` here and then not worry about a manual `json.loads` in the parser.

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@app.task(base=Singleton)
+def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
+    logger.info("Fetching PNC SBOM %s for purl %s", sbom_data["id"], purl)
+    """Fetch a PNC SBOM from sbomer"""
+
+    # Validate the supplied product information
+    try:
+        product_variant = ProductVariant.objects.get(name=product_data["productVariant"])
```
mprpic: Based on the discussion with Justin last week in chat about variants, I think our ProductVariants don't necessarily match to ET Variants. Is there any model that maps to ET Variants directly @jasinner?

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,62 @@
...
+    except ProductVariant.DoesNotExist:
+        logger.warning(
+            "SBOM fetch request for nonexistant product variant: %s", product_data["productVariant"]
+        )
+        return
+
+    if product_data["productVersion"] != product_variant.productversions.name:
```
mprpic: This and the below won't ever match since our product model is based off of prod-defs while the values SBOMer is using is ET products. Unless we load the entire hierarchy of products, their versions and variants from ET, this comparison is probably not needed. We really only need the variant name so it's OK to ignore the rest I think (I noted that in the gdoc where the UMB message format was being discussed as well).

corgi/monitor/consumer.py
```diff
@@ -163,3 +164,32 @@ class BrewUMBListener(UMBListener):
...
+
+
+class SbomerUMBHandler(UMBReceiverHandler):
+    """Handle messages about new SBOMs available from PNC"""
+
+    def sbom_complete(self, event: Event) -> None:
+        logger.info(f"Handling UMB message for PNC SBOM {event.message.id}")
+        message = json.loads(event.message.body)
+        try:
+            slow_fetch_pnc_sbom(
```
mprpic: This should be scheduled as an async task (`delay` or `apply_async`) otherwise it will block on releasing the message until the SBOM is processed. If the fetch/process fails in the async task, we'll see it in the error log and it can be retried later.

corgi/monitor/consumer.py
```diff
@@ -163,3 +164,32 @@ class BrewUMBListener(UMBListener):
...
+                message["sbom"],
+            )
+        except Exception as e:
+            logger.error(f"Failed to schedule fetch PNC SBOM {event.message.id}: {str(e)}")
+            self.release(event.delivery, delivered=True)
+        else:
+            self.accept(event.delivery)
+
+
+class SbomerUMBListener(UMBListener):
```
mprpic: Note that if this is a standalone listener, it'll need its own process and pod to run in (check out the runmonitor.py management command). We could also make the `BrewUMBListener` into a general listener and include the SBOM topic into its mapping.

kgrant-rh: I think I'd rather do something like the latter, but I'd rather just make like a (non-abstract) `UMBDispatcher` or something that routes any UMB messages Corgi is interested in to the appropriate handler.

corgi/tasks/pnc.py
jasinner: The names of ET Variants should match what we use in the name field of the ProductVariant core models. There are 2 ways we create those models. But matching brew_tags from product_definitions with those from ET, in which case we use the ET Variant name to create the ProductVariant core model. The other way is by directly reading and creating errata_info from product_definitions. In this case if product_definitions has bad data we could get something which doesn't not exist in ET, we don't do validation. We have the data in the CollectorErrataProductVariant collector model, so we could do such validation.

As mentioned in my other comment below, we won't match any existing ProductVariant here until we first create that ProductVariant in Corgi. Those only get created uses the 2 methods I describe below, so I think we should update product_definitions for Quarkus as least to add an errata_info stanza.

corgi/tasks/pnc.py
jasinner: I think that SBOMer is using the correct approach by using ET product taxonomy. What we are probably missing in product_definitions and therefore Corgi is errata_info for each product_stream we expect to receive UMB message from SBOMer. That will allow us to link the ProductVariant to the rest of the product models in Corgi.

I agree we don't need any additional check on ET products here, unless SBOMer sometimes doesn't populate the ET variant fields. We don't current link any errata_info apart from ProductVariant but we do record those values in the variant's meta_attr.

kgrant-rh:
@RedHatProductSecurity/corgi-devs This change now also refactors UMB message handling into something that can be extended with individual handlers. It does not address CORGI-650, which I would like to handle separately after this change.

jasinner (approved):
LGTM

kgrant-rh:
Closing this in favor of making two smaller changes, the first in https://github.com/RedHatProductSecurity/component-registry/pull/491 will introduce the UMBDispatcher, and a subsequent PR will add a UMBHandler for sbomer specifically.

PR #439, opened 2023-05-05
CORGI-624: Change daily task timing to avoid manifest truncation

juspence:
@RedHatProductSecurity/corgi-devs Please review. I'm not sure this will fix the issue, but I don't have any better ideas and I don't think it will hurt to try.


juspence:
This is wrong, it's not a truncation issue with the file size I think.

PR #441, opened 2023-05-05
CORGI-590: Fix manifesting for variants

juspence:
Just a draft / PoC for now. I started on this while waiting for the Nexus outage to resolve, but the same outage is blocking me from testing locally (my dependencies are out of date) and I know there are some TODOs left here.

We need to be able to manifest particular releases of products, like OpenShift 4.12.10, which is neither the latest 4.12.12 release in the z-stream nor the very first 4.12.0 release as tracked in the base stream. I haven't thought about this much, but we relate errata only to variants and never to streams. Each build also has a list of released errata tags.

So we should be able to identify "which errata we want" in the API, then pass that list of errata names to the manifest-generating logic when creating a manifest for some variant. We can then filter "which components we want" based on whether or not that component's linked build has a matching released_errata_tag.

This also fixes some other minor issues related to manifesting, mainly CORGI-599. Note that Rogue says "GENERATED_FROM" in that ticket, but the correct relationship is "GENERATES". You must read the SPDX relationships section from bottom to top, e.g. the spdxElement "for an UPSTREAM component" -> relationshipType "GENERATES" -> relatedSpdxElement "a REDHAT component".

Finally, I tried to figure out why we have duplicates in the SPDX relationships section, but the queryset we're using already has .distinct() and I don't see any obvious reason why this happens. Given the time it will take to dig into our data and find the bug's root cause, I'll have to pick this up after PTO.

### DIFF
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -496,7 +496,7 @@ class ProductStreamViewSetSet(ProductDataViewSet):
         obj = self.queryset.filter(uuid=uuid).first()
         if not obj:
             return Response(status=status.HTTP_404_NOT_FOUND)
-        manifest = json.loads(obj.manifest)
+        manifest = json.loads(obj.manifest())
         return Response(manifest)
 
 
@@ -535,7 +535,8 @@ class ProductVariantViewSetSet(ProductDataViewSet):
         obj = self.queryset.filter(uuid=uuid).first()
         if not obj:
             return Response(status=status.HTTP_404_NOT_FOUND)
-        manifest = json.loads(obj.manifest)
+        errata_ids = request.query_params.get("errata_ids", "")
+        manifest = json.loads(obj.manifest(errata_ids=errata_ids))
         return Response(manifest)
 
 
--- a/corgi/api/views.py
+++ b/corgi/api/views.py
@@ -530,6 +530,14 @@ class ProductVariantViewSetSet(ProductDataViewSet):
         except ProductVariant.DoesNotExist:
             raise Http404
 
+    @action(methods=["get"], detail=True)
+    def manifest(self, request: Request, uuid: str = "") -> Response:
+        obj = self.queryset.filter(uuid=uuid).first()
+        if not obj:
+            return Response(status=status.HTTP_404_NOT_FOUND)
+        manifest = json.loads(obj.manifest)
+        return Response(manifest)
+
 
 @INCLUDE_EXCLUDE_FIELDS_SCHEMA
 class ChannelViewSet(ReadOnlyModelViewSet):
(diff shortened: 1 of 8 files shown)


juspence:
@mprpic Could you give this a quick review when you have time please? I'm refreshing the tag data in stage now, but after that's done I might try to wrap this one up if nothing else interferes.

mprpic:
> @mprpic Could you give this a quick review when you have time please? I'm refreshing the tag data in stage now, but after that's done I might try to wrap this one up if nothing else interferes.

Could you perhaps split this out into separate PRs each covering its own topic? Reviewing this, I'm not sure what change relates to what "problem solved".

juspence:
@mprpic Ack, will split in a bit but (I hope) if you look at the commit-by-commit view it should be a little easier to see what's getting fixed when.

EDIT: Actually most of the changes here are only for 590. I split the very small tweak to relationship types (CORGI-599) into another PR.

The logic here is that we refactored some manifest properties / querysets onto the ProductStream model, when everything was generic before (supported any ProductModel subclass). So I undid that refactoring, and added back a "manifest" property / link only for ProductVariants in the API.

The more important piece here is the last commit, which adds an "errata_ids" parameter when generating a manifest for ProductVariants. See description for full justification, but this should enable manifesting only some errata for a Variant, instead of all of them (which is the default).

If that part looks OK I'll rebase onto main to get the recent manifesting changes, clean this up, and add tests.

corgi/api/views.py
```diff
@@ -530,6 +530,15 @@ def get_object(self):
         except ProductVariant.DoesNotExist:
             raise Http404
 
+    @action(methods=["get"], detail=True)
+    def manifest(self, request: Request, uuid: str = "") -> Response:
+        obj = self.queryset.filter(uuid=uuid).first()
+        if not obj:
+            return Response(status=status.HTTP_404_NOT_FOUND)
+        errata_ids = request.query_params.get("errata_ids", "")
```
mprpic, line 538: This should ideally parse out valid erratum IDs against `ADVISORY_REGEX`, and pass them on as a list instead of having to split the raw text in `files.py`.

mprpic (commented):
Looked this over and conceptually there is nothing wrong with the code here, but I'm not sure it's the right approach to solving our version-specific SBOM issue. Let's discuss more in standup and/or Jira issue.

juspence:
Discussed with Martin - this might be close to the right logic (find errata that released some version of some stream, generate manifest only for builds / components linked to those errata). But we need to do this at the stream level since differences between variant content sets aren't really important.

Since we don't link errata to streams, this needs a larger conversation about how to find the right errata and map them to the versions they release. We also need a solution for "how do we know which streams use errata for releases at all" (e.g. OpenShift) vs. the ones that don't (e.g. RHEL).

PR #443, opened 2023-05-08
CORGI-634 disable django manifest static caching

jasinner:
I was not able to reproduce this issue in Dev, but I suspect it's being caused by the use of `CompressedManifestStaticFilesStorage`, so I'm going to try switching to `CompressedStaticFilesStorage`. This also upgrades to the latest version of Whitenoise and fixes a bug in dev where files written by the celery worker containers cannot be read by the web container.


jasinner:

JimFuller-RedHat:
LGTM

PR #446, opened 2023-05-10
CORGI-634 generate links to manifests with hash

jasinner:
I think CORGI-634 file truncation is because we overwrite the file, but django doesn't re-read the file metadata from disk on every request unless you have [WHITENOISE_AUTOREFRESH](https://whitenoise.readthedocs.io/en/latest/django.html#WHITENOISE_AUTOREFRESH) set. An alternative approach might be to add a [WHITENOISE_IMMUTABLE_FILE_TEST](https://whitenoise.readthedocs.io/en/latest/django.html#WHITENOISE_IMMUTABLE_FILE_TEST)  for the json files.

This switches back to `whitenoise.storage.CompressedManifestStaticFilesStorage` and generates manifest links like this:
such as `openshift-4.13-efe620b4-e61d-4862-a632-81e320bdf6a4.a1f4f6010cda.json`, notice the `a1f4f6010cda` hash in the filename. The next time we run `collectstatic `we'll get a new file and won't overwrite any files. Which should mean no mismatch between file metadata (content-length) and file contents. Those files can be cached forever by clients in the browsers, or anywhere in between the server and browser as the content won't change.

We'll have to keep an eye on the size of our staticfiles PVC and clean it up when it's getting close to capacity. I'll increase the size in an ops repo MR soon. All also see if I can find a metric to monitor the PVC disk usage.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -125,7 +125,7 @@
         "filename": "config/settings/dev.py",
         "hashed_secret": "6adfb183a4a2c94a2f92dab5ade762a47889a5a1",
         "is_verified": true,
-        "line_number": 30,
+        "line_number": 5,
         "is_secret": false
       }
     ],
@@ -282,5 +282,5 @@
       }
     ]
   },
-  "generated_at": "2023-03-14T20:10:12Z"
+  "generated_at": "2023-05-10T20:29:38Z"
 }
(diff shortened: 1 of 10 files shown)


jasinner:
While I was changing things related to manifest I also adjusted the code which generates and builds links to the manifest so that only streams with released products get generated.

jasinner:
@RedHatProductSecurity/corgi-devs please review. I've tested this in stage and it seems to be working well.

JimFuller-RedHat:
LGTM

PR #447, opened 2023-05-11
CORGI-649: add temporary cpe fixup

JimFuller-RedHat:
added temporary cpe fixup for when we generate manifests.


corgi/core/fixups.py
```diff
@@ -0,0 +1,335 @@
...
+    "stf-1.5": ["cpe:/a:redhat:service_telemetry_framework:1.5::el8"],
+    "wto-1.5": ["cpe:/a:redhat:webterminal:1.5::el8"],
+    "rhcertification-6": ["cpe:/a:redhat:certifications:1::el6"],
+    "rhcertification-7": ["cpe:/a:redhat:certifications:1::el7"],
+    "rhcertification-8": ["cpe:/a:redhat:certifications:1::el8"],
+    "rhcertification-9": ["cpe:/a:redhat:certifications:1::el9"],
+}
+
+
+def cpe_lookup(product_stream_name: str) -> List[str]:
```
mprpic, line 331: I think you can just replace calls to this with `cpes.get(stream, [])`.

corgi/core/files.py
```diff
@@ -64,7 +65,7 @@ class ProductManifestFile(ManifestFile):
     # We can subclass this to handle different Product subclasses with different properties
     # Or to handle different ways of generating manifest properties from Product properties
 
-    def render_content(self) -> str:
+    def render_content(self, fixup=False) -> str:
```
mprpic: Can we call this something like `use_cpe_mapping`?

corgi/core/files.py
```diff
@@ -76,6 +77,11 @@ def render_content(self) -> str:
             "released_components": released_components,
             "distinct_provides": distinct_provides,
         }
+
+        if fixup:
+            # apply temporary cpe lookup fixup
+            kwargs_for_template["cpes"] = cpe_lookup(self.obj.name)  # type: ignore
```
mprpic: Let's pull out the CPEs into a "cpes" kwarg depending on the flag value instead. We want one set of CPEs or the other, not both. So:

```
        kwargs_for_template = {
            "obj": self.obj,
            "released_components": released_components,
            "distinct_provides": distinct_provides,
            "cpes": cpes.get(self.obj.name) if use_cpe_mapping else self.obj.cpes,
        }
```

This way you also don't have to modify the template and only change `obj.cpes` to `cpes`.

corgi/tasks/management/commands/generatemanifests.py
```diff
@@ -13,11 +13,17 @@ def add_arguments(self, parser: CommandParser) -> None:
             dest="stream",
             help="Update the manifest for the named product stream",
         )
+        parser.add_argument(
+            "-f",
+            "--fixup",
+            action="store_true",
+            help="Apply manifest fixups",
```
mprpic: Can we add a default of `True` for this?

jasinner (approved):
LGTM, thanks

corgi/web/templates/product_manifest.json
```diff
@@ -66,14 +66,14 @@
...
-        "externalRefs": [{% for cpe in obj.cpes %}
+        "downloadLocation": "NOASSERTION", {% if cpes %}
+        "externalRefs": [{% for cpe in cpes %}
             {
                 "referenceCategory": "SECURITY",
                 "referenceLocator": "{{cpe}}",
                 "referenceType": "cpe22Type"
             }{% if not forloop.last %},{% endif %}{% endfor %}
-        ],{% endif %}
+        ], {% endif %}
```
mprpic, line 76: Can you please trim this? :-)

juspence: Handled in #460.

PR #448, opened 2023-05-12
CORGI-648 collector product variant cpe null

jasinner:
When ET returns null for variant CPE set it to an empty string in the Collector models.

Also fixes an issue running schema and secrets tox targets with the latest tox 4.5.1.


JimFuller-RedHat:
LGTM

jasinner:
I'm going to ignore the Sonarqube failure for now, because I'm pressed for time. We should really address that by splitting sections of the load_et_products function into separate functions in future.

PR #450, opened 2023-05-15
CORGI-652 reverse proxy via nginx

jasinner:
Yet another iteration of the file truncation issue solution. I finally decided to give up on serving static files via gunicorn in favour of NGinx. In this iteration using NGinx as a reverse proxy.

Not that to use NGinx as a reverse proxy in dev you'll need to update the url you access Corgi on from port 8008 to 8080

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -298,11 +298,10 @@ USE_L10N = False
 USE_TZ = True
 
 # Static files (CSS, JavaScript, Images)
-STATIC_URL = f"https://{CORGI_DOMAIN}/static/"
+STATIC_URL = f"https://{CORGI_DOMAIN}/"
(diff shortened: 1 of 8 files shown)


JimFuller-RedHat:
LGTM

jasinner:
I updated this a bit to use the origin ALLOWED_HOSTS config for stage and prod.

I also simplified the nginx config in dev.

JimFuller-RedHat:
LGTM

mprpic:
If we have NGINX serving this content, we don't need whitenoise at all. The manifest generation can just place the files in some director that NGINX is serving it out of and that's it. The path to the files can also then be something more standard like `/files/sbom/<file>`.

PR #452, opened 2023-05-15
CORGI-650: Handle brew tag renames

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have time. If this looks good I'll deploy to stage and test the selector / message filtering logic. In the meantime I'll be writing tests, but I wanted to get review on this before I find out it's the wrong approach. Once tests are written and this is merged, I'll manually refresh tags on all builds for (hopefully) the last time.

I left the Brew tag / untag UMB logic in place, since that works correctly for 2 out of 3 cases (only fails when an existing tag is renamed), and I didn't want to do a major refactoring of this code before summit. We could switch to listening on the errata.v2.builds.changed topic instead, and then all of our tag / untag events would use data from Errata Tool.

Note that this errata.v2.builds.changed topic won't handle / report any messages for non-errata tags that get added to a build. Since we rely on Brew tags to link streams to builds, I think it's important we keep all tags in sync between Brew and Corgi, not just errata tags. But please let me know if I've misunderstood something or there's a better way.

The ET docs for all these UMB topics specifically call out "Brew build IDs" and the "Brew API". So I'm not sure this solution is future-proof / will work for PNC builds attached to errata. My preferred approach here is just to focus on fixing bugs before summit, then refactoring as part of a larger effort in the future if needed, only when we're ready to add broader support for middleware / PNC. Right now we listen for UMB messages about "completed Brew builds" but not "completed PNC builds", so there are other gaps we'll need to close anyway.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -77,7 +77,9 @@ class Brew:
     # A list of component names, for which build analysis will be skipped.
     COMPONENT_EXCLUDES = json.loads(os.getenv("CORGI_COMPONENT_EXCLUDES", "[]"))
 
-    def __init__(self, source: Optional[str] = ""):
+    koji_session: koji.ClientSession = None
+
+    def __init__(self, source: str = ""):
         if source == SoftwareBuild.Type.CENTOS:
             self.koji_session = koji.ClientSession(settings.CENTOS_URL)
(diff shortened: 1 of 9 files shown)


mprpic (approved):
LGTM!

jasinner (approved):
LGTM

PR #453, opened 2023-05-15
CORGI-655: Limit stream manifests to supported

mprpic:
This is a quick and simple way to limit our manifests to those that can be published. SDEngine queries for streams by looking at those that have a manifest link and will automatically publish all manifests that have a link. This ensures only those that are supported (are included in our manually curated list) are published.

Long-term, CPEs should be defined on product streams in prod-defs (or w/e other equivalent exists) and then this method can use the metadata of the ProductStream object directly.


mprpic:
@jasinner Please review. This is a pretty simple solution that works against the static mapping we have available. We'll need to refactor both once we move the data out of Corgi. Wdyt?

JimFuller-RedHat (approved):
LGTM

PR #454, opened 2023-05-15
CORGI-654: Switch to native Django StaticFilesStorage instead of Whitenoise

juspence:
@RedHatProductSecurity/corgi-devs Please review. I'm not sure I understand the new setup, since the nginx pod's "./staticfiles" dir (at least locally / in dev's docker-compose.yml) seems to point to the main corgi folder's "/opt/app-root/src" dir instead of the "/opt/app-root/staticfiles" dir.

We also still need to collect static files on web pod startup in order to get the "base.css" file our HTML templates rely on to be displayed. Not collecting this static file after the PVC is recreated triggers server errors whenever someone tries to view the main Corgi webpage. This eventually fixes itself when the collectstatic task runs, but let's just collect the files we need at startup for safety / to avoid outages.

EDIT: Actually it looks like the current nginx setup is completely broken because the manifest links we generate in the API still rely on Whitenoise's URL-building logic, and we return an empty "" string if the URL isn't found. Have we tested with Whitenoise's ManifestStaticFilesStorage or Django's StaticFilesStorage?

These don't use Gzip compression as far as I understand, so shouldn't be affected by this Gzip bug that causes manifest truncation. That seems like a simpler change than standing up a whole new webserver just to serve the manifests.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -155,7 +155,6 @@ INSTALLED_APPS = [
     "django.contrib.contenttypes",
     "django.contrib.sessions",
     "django.contrib.messages",
-    "whitenoise.runserver_nostatic",
     "django.contrib.staticfiles",
     "django.contrib.humanize",
     "django.contrib.postgres",
@@ -178,7 +177,6 @@ INSTALLED_APPS = [
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
-    "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
@@ -298,10 +296,9 @@ USE_L10N = False
 USE_TZ = True
 
 # Static files (CSS, JavaScript, Images)
-STATIC_URL = f"https://{CORGI_DOMAIN}/"
+STATIC_URL = "/"
 STATIC_ROOT = os.getenv("CORGI_STATIC_FILES_DIR", str(BASE_DIR / "staticfiles"))
 STATICFILES_DIRS = [OUTPUT_FILES_DIR]
-STATICFILES_STORAGE = "whitenoise.storage.CompressedStaticFilesStorage"
 
 # Celery config
 CELERY_BROKER_URL = os.getenv("CORGI_REDIS_URL", "redis://redis:6379")
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -8,9 +8,9 @@ from uuid import UUID
 
 from django.conf import settings
 from django.db.models.manager import Manager
+from django.templatetags.static import static
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
-from whitenoise.storage import CompressedStaticFilesStorage
 
 from corgi.api.constants import CORGI_API_URL
 from corgi.core.constants import MODEL_FILTER_NAME_MAPPING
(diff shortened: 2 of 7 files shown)


JimFuller-RedHat (approved):
+1 to removing whitenoise 

running collectstatic when we run the service itself will presumably delete files ... is that the intent ?

otherwise propose testing this in stage first before we commit to main -  in general LGTM

config/settings/base.py
```diff
@@ -298,10 +296,9 @@
 USE_TZ = True
 
 # Static files (CSS, JavaScript, Images)
-STATIC_URL = f"https://{CORGI_DOMAIN}/"
+STATIC_URL = "/"
```
JimFuller-RedHat, line 299: is the intent to not force https scheme ... perhaps http->https is forced elsewhere in infra.

run_service.sh
```diff
@@ -3,10 +3,16 @@
 # Custom run script for starting corgi django service in corgi-stage and corgi-prod environments.
 # Note - DJANGO_SETTINGS_MODULE env var is required
 
+# collect static files from any app folder ending in /static/
+# (currently only corgi/web/static/base.css)
+python3 manage.py collectstatic \
```
JimFuller-RedHat, line 8: is the intent is to wipe out these files and regenerate ?

run_service.sh
```diff
@@ -3,10 +3,16 @@
...
+    --ignore '*.json' \
+    -v 2 \
+    --noinput
```
mprpic, line 11: We can probably remove the collectstatic Celery task now too since w/e files are collected (just the CSS/JS stuff) at startup will never change.

corgi/api/serializers.py
```diff
@@ -591,7 +591,7 @@ def get_manifest(instance: ProductStream) -> str:
             return ""
         filename = f"{instance.name}-{instance.pk}.json"
         try:
-            manifest_link = CompressedStaticFilesStorage().url(filename)
+            manifest_link = static(filename)
```
mprpic, line 594: I'm not sure what benefit there is going through Django's static middleware now when we dump the files into a dir from where NGINX is serving it. We could just link to `CORGI_DOMAIN/file` and call it done.

juspence: Yes, this was the source of the EDIT: note in the PR description. I threw this in there just to make the pipeline stop failing, but I need to look over the new Nginx logic and understand how we actually generate the manifests + build the links.

run_service.sh
juspence: Yes, see below - I originally had removed the collectstatic task, then added it back because I wasn't sure how we're actually building the links to the static files and if we still needed to move them from where they're generated to where they're stored.

config/settings/base.py
juspence: This setting doesn't force HTTPS connections if I understand correctly, it just means that we'll give a 404 if you use a plaintext / HTTP URL. It also only applies to staticfiles and not to regular requests outside the staticfiles path.

Actually this should probably be changed to be "/static/" like it was before, but I need to look over the PRs that were merged during my PTO. HTTPS is forced by the Django HSTS settings and some OpenShift route annotations.

run_service.sh
juspence: No, it's just to avoid an error when the PVC is recreated and doesn't have the base.css file / other required files to display the Corgi homepage.

config/settings/base.py
juspence: It looks like HSTS settings might not get passed through correctly with the new Nginx config in place. That Nginx server really should only be used for staticfiles and not all requests. I need to look over the ops repo changes, but CORGI-650 is a higher priority so going to close this MR for now.

PR #455, opened 2023-05-16
fix cpe fixup

jasinner:


JimFuller-RedHat:
to get things working LGTM though we may want to revisit in the future (where presumably we are time rich)

mprpic:
I'm not sure why this change is needed. The task runs with `fixup=True` every time anyway.

jasinner:
Yes, but the task doesn't and can't pass the fixup argument through to the manifest property of the product stream. Properties don't accept arguments as far as I can tell.

I discovered this during CORGI-577 because the cpes were not showing up the manifest. With this change the cpes are present.

PR #456, opened 2023-05-16
CORGI-658 Remove provided components with internal urls from manifest

jasinner:
This consolidates use of manifest query to a single queryset call `manifest_components' used when checking which manifest to generate in the cpu_update_ps_manifest task, and them again when actually generating them using the template.

It also remove any internal (redhat.com in name) provided components from the manifests.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -638,13 +638,8 @@ class ProductStream(ProductModel):
         """Returns unique aggregate "provides" for the latest components in this stream,
         for use in templates"""
         unique_provides = (
-            self.components.exclude(
-                type=Component.Type.CONTAINER_IMAGE, name__endswith="-container-source"
-            )
+            self.components.manifest_components()
             .using(using)
-            .root_components()
-            .released_components()
-            .latest_components()
             .values_list("provides__pk", flat=True)
             .distinct()
             .order_by("provides__pk")
@@ -653,7 +648,9 @@ class ProductStream(ProductModel):
         return (
             Component.objects.filter(pk__in=unique_provides)
             # Remove .exclude() below when CORGI-428 is resolved
-            .exclude(type=Component.Type.GOLANG, name__contains="./").using(using)
+            .exclude(type=Component.Type.GOLANG, name__contains="./")
+            .external_components()
+            .using(using)
         )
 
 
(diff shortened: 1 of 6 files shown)


corgi/core/models.py
```diff
@@ -1014,6 +1011,26 @@ def root_components(self, include: bool = True) -> "ComponentQuerySet":
...
         return self.exclude(ROOT_COMPONENTS_CONDITION)
 
+    # See CORGI-658 for the motivation
+    def external_components(self, include: bool = True) -> "ComponentQuerySet":
+        """Show only external components by default, or internal components if include=False"""
+        redhat_com_query = Q(name__contains="redhat.com")
+        if include:
+            # Truthy values return the excluded queryset (only released components)
+            return self.exclude(redhat_com_query)
+        # Falsey values return the filtered queryset (only unreleased components)
```
juspence: Nit: Looks like a copy-paste error in the comments.

tests/test_manifests.py
```diff
@@ -241,6 +234,26 @@ def test_product_manifest_excludes_unreleased_components():
...
 
+def test_product_manifest_excludes_internal_components():
+    """Test that manifests for products don't include unreleased components"""
+    component, stream, provided, dev_provided = setup_products_and_components_provides(
+        internal_component=True
+    )
+
+    manifest = json.loads(stream.manifest)
+
+    # No released components linked to this product
```
juspence: If there are no released components, doesn't that mean we're not testing the internal / external component filtering logic? Maybe we should add 2 of each component type (root / provided / upstream components), one internal and one external, then check that internal components don't end up in any of the three querysets.

corgi/core/models.py
```diff
@@ -653,7 +648,9 @@ def provides_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
         return (
             Component.objects.filter(pk__in=unique_provides)
             # Remove .exclude() below when CORGI-428 is resolved
-            .exclude(type=Component.Type.GOLANG, name__contains="./").using(using)
+            .exclude(type=Component.Type.GOLANG, name__contains="./")
+            .external_components()
```
juspence, line 652: Should this be part of manifest_components below? AKA do we want to always exclude internal components from all three querysets (provided_components, root_components, and upstream_components) that are used in manifests?

juspence (approved):
LGTM if tests are correct.

corgi/core/models.py
jasinner: I originally tried it that way, but it didn't work. I think that's because that query only select root_components. Whereas the package we are trying to filter is included in the manifest as a result of being included in this provides_queryset.

tests/test_manifests.py
jasinner: Another copy and paste error. Updating that comment.

corgi/core/models.py
jasinner: Thanks, updated the comments.

PR #459, opened 2023-05-17
CORGI-596, CORGI-599: Fix relationships in manifests

juspence:
@RedHatProductSecurity/corgi-devs Please review these two small fixes for the SPDX relationships key in our manifests.


mprpic (approved):
LGTM!

PR #460, opened 2023-05-17
CORGI-657: Deduplicate upstream components before adding to manifests

juspence:
@jasinner This should allow us to report different information for upstream vs. Red Hat components, while still fixing the duplication you discovered. I copied the tests from your other MR, and they still pass - if this looks good I can merge as-is and do the release right after. Or please let me know if you'd like some changes here / to keep this out of the release for now. I'm happy to do another release tomorrow or Friday if we want to validate this change in stage first.


jasinner (approved):
LGTM!

PR #462, opened 2023-05-18
Remove license data from manifests

juspence:
This is not the right solution, but it's the only option I can think of that will actually work before summit.


juspence:
Cantfix, wontfix, dontneedtofix, thankgoodnesswerenot"fixing".

PR #464, opened 2023-05-22
CORGI-654: Remove whitenoise and fix Nginx bugs / ESS control violations

juspence:
@RedHatProductSecurity/corgi-devs Please review. This will:

1. Ensure X-Forwarded-For is always set (by the OpenShift router) to the client's external IP, never overwritten (by Nginx) with the router's internal IP
2. Send Nginx logs to Splunk
3. Pass client headers in the request through to Django
4. Fix some differences between local and stage / prod config
5. Automatically collect static files on each web pod startup / avoid UI breakage due to missing dependencies
6. Write manifests directly into the staticfiles PVC / avoid file truncation due to network errors when copying files between PVCs

We also need some corresponding changes in the ops repo - I'll open a PR there in just a bit.

### DIFF
--- a/etc/nginx/proxy.conf
+++ b/etc/nginx/proxy.conf
@@ -5,8 +5,17 @@ access_log /dev/stdout;
 error_log stderr warn;
 
 # Settings to serve static files
-location /static/  {
+location /staticfiles/  {
+    # The staticfiles dir must be mounted inside of below folder
+    # URLs like CORGI_DOMAIN/staticfiles/spdx-22-schema.json will become
+    # filenames like /opt/app-root/src/staticfiles/spdx-22-schema.json
     root               /opt/app-root/src/;
+}
+
+# To hide 404 errors about missing favicon
+location /favicon.ico {
+    default_type        text/plain;
+    return              200;
 }
 
 # For OpenShift health checks
@@ -22,6 +31,7 @@ location / {
     # These are set by the Openshift router and must be logged as received
     # Setting them here will overwrite them with internal router IPs instead
     proxy_pass_request_headers   on;
+    proxy_read_timeout 120s;
     # Not used in production, but required in Dev to get correct links in the /api/v1 menu
     proxy_set_header   Host localhost:8080;
 }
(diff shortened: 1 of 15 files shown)


jasinner (approved):
This is a very well done. Thanks! This also allows us to add `autoindex on;` to the /staticfiles/ location in the NGinx conf. If we also added a link to that path in the web index.html template we'd have a way to see all the maifests in a single listing.

etc/nginx/proxy.conf
```diff
@@ -1,8 +1,23 @@
+# Global settings
+# Send access and error logs to container stdout
+# So they automatically show up in "oc logs" output / Splunk
+access_log /dev/stdout combined;
+error_log /dev/stdout;
```
jasinner: I especially like this, thank you! Bonus points if we could add `app=corgi` and `environ={stage/prod}` to the message by default for filtering in Splunk.

juspence: I tried, but the log_format directive isn't allowed in this proxy configuration file. It has to be present in the outermost http{} block in the default configuration file, which we aren't overriding. I don't want to maintain this / keep backporting changes from the default config into our custom one.

We aren't logging this app / environ info in gunicorn either, although maybe we should be. Gunicorn's default access_log_format matches the Nginx "combined" log format I've used here, so following requests between Nginx and Gunicorn should be nice and simple.

Before we added the app=corgi parameter, I think just searching for "corgi" was enough to find our logs in Splunk. That value should be in the name of the Python module which gets logged, like corgi.tasks.brew, or in the hostname of the request logged by Gunicorn.

juspence:
Oops, performance test failure is expected. I had to change the test so it will pass after this MR, and it passed yesterday with this code deployed to stage for testing. But it fails with the current code on main deployed to stage, which is expected.

I didn't add autoindex since it's considered an insecure option by some people. Adding that will reveal the contents of library directories like staticfiles/rest_framework/ or staticfiles/mptt/, not just the manifests stored in the staticfiles/ root. Going to merge as-is given approval.

PR #466, opened 2023-05-23
CORGI-667: Fix ValueError when tag to be removed not found

juspence:
@RedHatProductSecurity/corgi-devs Please review this fix for an error seen in recent daily monitoring emails. We fail with an error if some tag to be removed from a build is not actually present in that build's list of tags. This is probably due to a tag rename from X to Y, and then trying to remove Y. We receive no UMB events for the rename, so the build's list of tags still has only X and is missing Y.

We handle this case for errata that were just recently released / entered the SHIPPED_LIVE state (by refreshing all tags), so that tags like "stream-candidate" or "stream-pending" are properly stored in Corgi with their final "stream-released" values.

For errata which aren't released / enter the DROPPED_NO_SHIP state, "stream-candidate" is probably renamed to "stream-dropped" (or similar). The solution is the same as above - if removing the tag fails, just refresh all tags on the build so we have the latest values and the Corgi data matches the Brew data.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -624,8 +624,17 @@ def slow_update_brew_tags(build_id: str, tag_added: str = "", tag_removed: str =
                 # Below should automatically create new relations for this build / erratum
                 slow_load_errata.delay(errata_tag.group())
         else:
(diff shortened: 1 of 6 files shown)


jasinner (approved):
LGTM

PR #473, opened 2023-05-25
CORGI-616 brew tag relation timeliness

jasinner:
This saves brew_tag relations everytime we ingest a new build. It's done before save_product_taxonomy so that the new build's components are instantly related to any product streams via it's brew_tags.

I wanted to make use of the existing save_product test function from test_manifest.py so I made it a test fixture. Since that function is called by multiple other functions in test_manifests, I also made all those fixtures as well. One of the fixtures (setup_products_and_components_provides) originally took arguments so I updated it to use pytest.markers instead.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -2,6 +2,7 @@ import re
 from datetime import datetime, timedelta
 from typing import Optional
 
+from celery.local import PromiseProxy
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
@@ -113,8 +114,15 @@ def slow_fetch_brew_build(
     for child_component in component.get("components", []):
         save_component(child_component, root_node, softwarebuild)
 
-    # We don't call save_product_taxonomy by default to allow async call of slow_load_errata task
-    # See CORGI-21
+    # for builds with any tag, check if the tag is used for product stream relations, and create the
+    # relations if so.
+    if not build_meta["tags"]:
+        logger.info("no brew tags")
+    else:
+        new_relations = load_brew_tags(build_id, build_meta["tags"])
+        logger.info(f"Created {new_relations} for brew tags in {build_type}:{build_id}")
+
+    # Allow async call of slow_load_errata task, see CORGI-21
     if save_product:
         softwarebuild.save_product_taxonomy()
         for related_component in softwarebuild.components.get_queryset():
@@ -494,28 +502,59 @@ def save_node(
     retry_kwargs=RETRY_KWARGS,
     soft_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
 )
-def load_brew_tags() -> None:
+def load_stream_brew_tags() -> None:
     for ps in ProductStream.objects.get_queryset():
-        build_type = BUILD_TYPE
-        brew = Brew(BUILD_TYPE)
-        refresh_task = slow_fetch_modular_build
-        # This should really be a property in Product Definitions
-        if settings.COMMUNITY_MODE_ENABLED and ps.name == "openstack-rdo":
-            brew = Brew(SoftwareBuild.Type.CENTOS)
-            build_type = SoftwareBuild.Type.CENTOS
-            refresh_task = slow_fetch_brew_build
+        brew, build_type, refresh_task = _relation_context_for_stream(ps.name)
         for brew_tag, inherit in ps.brew_tags.items():
-            # Always load all builds in tag when saving relations
             builds = brew.get_builds_with_tag(brew_tag, inherit=inherit, latest=False)
-            no_of_created = create_relations(
-                builds,
-                build_type,
-                brew_tag,
-                ps.name,
-                ProductComponentRelation.Type.BREW_TAG,
-                refresh_task,
-            )
-            logger.info("Saving %s new builds for %s", no_of_created, brew_tag)
+            _create_relations_for_tag(builds, brew_tag, build_type, ps.name, refresh_task)
+
+
+def load_brew_tags(build_id: str, brew_tags: list[str]) -> int:
+    all_stream_tags = ProductStream.objects.exclude(brew_tags__exact={}).values_list(
+        "name", "brew_tags"
+    )
+    no_created = 0
+    distinct_brew_tags = set(brew_tags)
+    for stream_name, stream_tags in all_stream_tags:
+        distinct_stream_tags = set(stream_tags)
+        distinct_stream_tags = distinct_brew_tags.intersection(distinct_stream_tags)
+        for tag in distinct_stream_tags:
+            brew, build_type, _ = _relation_context_for_stream(stream_name)
+            logger.info(f"Creating relations for {stream_name} and {tag}")
+            no_created += _create_relations_for_tag((build_id,), tag, build_type, stream_name, None)
+    return no_created
+
+
+def _create_relations_for_tag(
+    build_ids: tuple,
+    brew_tag: str,
+    build_type: SoftwareBuild.Type,
+    stream_name: str,
+    refresh_task: Optional[PromiseProxy],
+):
+    no_of_created = create_relations(
+        build_ids,
+        build_type,
+        brew_tag,
+        stream_name,
+        ProductComponentRelation.Type.BREW_TAG,
+        refresh_task,
+    )
+    logger.info("Saving %s new builds for %s", no_of_created, brew_tag)
+    return no_of_created
+
+
+def _relation_context_for_stream(stream_name: str):
+    build_type = BUILD_TYPE
+    brew = Brew(BUILD_TYPE)
+    refresh_task = slow_fetch_modular_build
+    # This should really be a property in Product Definitions
+    if settings.COMMUNITY_MODE_ENABLED and stream_name == "openstack-rdo":
+        brew = Brew(SoftwareBuild.Type.CENTOS)
+        build_type = SoftwareBuild.Type.CENTOS
+        refresh_task = slow_fetch_brew_build
+    return brew, build_type, refresh_task
 
 
 def fetch_modular_builds(relations_query: QuerySet, force_process: bool = False) -> None:
--- a/corgi/tasks/common.py
+++ b/corgi/tasks/common.py
@@ -1,7 +1,9 @@
 import logging
 import subprocess
 from datetime import datetime, timedelta
+from typing import Optional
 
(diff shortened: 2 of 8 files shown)


jasinner:

corgi/tasks/common.py
```diff
@@ -69,7 +69,12 @@ def get_last_success_for_task(task_name: str) -> datetime:
 
 
 def create_relations(
-    build_ids, build_type, external_system_id, product_ref, relation_type, refresh_task
+    build_ids: tuple,
+    build_type: SoftwareBuild.Type,
+    external_system_id: str,
+    product_ref: str,
+    relation_type: ProductComponentRelation.Type,
+    refresh_task,
```
juspence: refresh_task can be an Optional[Callable] I think.

corgi/tasks/brew.py
```diff
@@ -488,34 +494,55 @@ def save_node(
...
 
 
-@app.task(
-    base=Singleton,
-    autoretry_for=RETRYABLE_ERRORS,
-    retry_kwargs=RETRY_KWARGS,
-    soft_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
-)
-def load_brew_tags() -> None:
+def load_stream_brew_tags() -> None:
```
juspence, line 505: This is still being run daily as a Celery task, but you've removed the decorator.

tests/test_manifests.py
```diff
@@ -234,11 +223,10 @@ def test_product_manifest_excludes_unreleased_components():
     }
 
 
-def test_product_manifest_excludes_internal_components():
+@pytest.mark.internal_component(True)
```
juspence: Instead of having a pytest fixture that takes arguments based on a test marker, can we just have the fixture create all the components (normal, internal, unreleased) every time?

That would also make the manifest tests more robust (by including more data), so we might catch more bugs (like the duplicate provides / upstreams bugs that we had to fix).

This test should work the same either way - we can ignore the normal and unreleased components, use the internal_provided and internal_dev_provided components, and continue to ```assert "redhat.com/" not in package["name"]```.

tests/test_manifests.py
```diff
@@ -198,11 +188,10 @@ def test_slim_rpm_in_containers_manifest():
     assert manifest["relationships"][-1] == document_describes_product
 
 
-def test_product_manifest_excludes_unreleased_components():
+@pytest.mark.unreleased(True)
```
juspence: As below - let's create all the components, ignore the normal and internal ones, and use only the unreleased components in this test. Then we can ditch the marker and the parameter in the fixture.

pyproject.toml
```diff
@@ -21,6 +21,8 @@ markers = [
     "integration",
     "performance",
     "unit",
+    "unreleased",
+    "internal_component"
```
juspence: Using pytest marks as parameters to a test fixture seems very odd. It also means the list of marks will grow a lot if we keep adding parameters.

I think the simplest solution is just to remove the @pytest.fixture() decorator, import the setup function in each test module that needs to use them, and call the function with whatever arguments it needs, like we were doing before.

Or you could split each parameter (internal, unreleased, or regular components) into its own fixture. Each fixture can just call the main setup function with the right set of parameters for that fixture - for example, the setup_internal fixture would call the setup_base() function with an internal=True argument. This is the "fixture factories" approach described here: http://stackoverflow.com/questions/63425998/#63431123

Or we could combine the manifest tests, then run them with pytest.mark.parametrize and pass different test data to them each time. Then the helper function can remain a function, and we can pass different test data by just calling the function with different parameters.

juspence (commented):
I haven't finished reviewing the Brew tags logic or the tests, but going to submit the rest before I forget.

corgi/tasks/brew.py
```diff
@@ -488,34 +494,55 @@ def save_node(
...
-                ps.name,
-                ProductComponentRelation.Type.BREW_TAG,
-                refresh_task,
-            )
-            logger.info("Saving %s new builds for %s", no_of_created, brew_tag)
+            _create_relations_for_tag(builds, brew_tag, build_type, ps.name, refresh_task)
+
+
+def load_brew_tags(build_id: str, brew_tags: list[str]) -> int:
+    all_stream_tags = ProductStream.objects.exclude(brew_tags__exact={}).values_list(
```
juspence, line 514: Should this be ```brew_tags=[]``` with square brackets to match against an empty list of tags? ```{}``` is Python syntax for an empty set, and I'm not sure that will translate properly in the JSONField.

Or maybe this JSON blob for the Brew tags is always wrapped in an outermost / top-level curly {} braces instead of square [] brackets, even when empty?

corgi/tasks/brew.py
```diff
@@ -488,34 +494,55 @@ def save_node(
...
+    )
+    no_created = 0
+    for stream_name, stream_tags in all_stream_tags:
+        for tag in brew_tags:
+            if tag in stream_tags:
+                brew, build_type, _ = _relation_context_for_stream(stream_name)
+                logger.info(f"Creating relations for {stream_name} and {tag}")
+                no_created += _create_relations_for_tag(
+                    (build_id,), tag, build_type, stream_name, None
+                )
```
juspence: Instead of many nested loops that check all the tags, I think it's more efficient to convert the tags to a set and check the intersection:
```suggestion
    brew_tags = set(brew_tags)
    for stream_name, stream_tags in all_stream_tags:
        build_stream_tags = set(stream_tags)
        build_stream_tags = brew_tags.intersection(build_stream_tags)
        for tag in build_stream_tags:
                brew, build_type, _ = _relation_context_for_stream(stream_name)
                logger.info(f"Creating relations for {stream_name} and {tag}")
                no_created += _create_relations_for_tag(
                    (build_id,), tag, build_type, stream_name, None
                )
```

corgi/tasks/brew.py
```diff
@@ -113,8 +113,14 @@ def slow_fetch_brew_build(
...
         save_component(child_component, root_node, softwarebuild)
 
-    # We don't call save_product_taxonomy by default to allow async call of slow_load_errata task
-    # See CORGI-21
+    # for builds with any tag, check if the tag is used for product stream relations, and create the
+    # relations if so.
+    if not build_meta["tags"]:
+        logger.info("no brew tags")
+    else:
+        load_brew_tags(build_id, build_meta["tags"])
```
juspence: We return no_created from this function, but it's ignored here - did you want to log it somewhere?

tests/test_brew_collector.py
```diff
@@ -844,22 +848,45 @@ def test_fetch_rpm_build(mock_sca, mock_brew):
...
     image_index = Component.objects.get(
         name="subctl-container", type=Component.Type.CONTAINER_IMAGE, arch="noarch"
     )
-    softwarebuild = SoftwareBuild.objects.get(build_id=1781353, build_type=SoftwareBuild.Type.BREW)
 
+    # Check that new components are related to the build via the brew_tag
+    assert ProductComponentRelation.objects.get(
+        build_id="1781353", build_type=SoftwareBuild.Type.BREW
+    )
+    assert image_index.productstreams.get(pk=stream.uuid)
```
juspence: Minor nit: It's more efficient to just ```assert Model.filter(field=condition).exists()``` if we're not going to use the returned value.

Or you can just say ```Model.get(field=condition)``` since that will raise an exception anyway if the object doesn't exist - no need for the assert here.

tests/test_brew_collector.py
```diff
@@ -893,6 +920,34 @@ def test_fetch_container_build_rpms(mock_fetch_brew_build, mock_load_errata, moc
...
+        type=ProductComponentRelation.Type.BREW_TAG,
+    )
+    assert new_brew_tag_relation.product_ref == stream.name
+    mock_fetch_modular_build.assert_called_once()
+
+
+@pytest.mark.django_db
+def test_load_brew_tags():
+    stream = ProductStreamFactory(brew_tags={"rhacm-2.4-rhel-8-container-released": True})
+    load_brew_tags("1", ["rhacm-2.4-rhel-8-container-released"])
```
juspence, line 944: It might be worth patching _create_relations_for _tag so we can assert it's called with a None argument for refresh_task.

Or we could mock both slow_fetch_brew_build and slow_fetch_modular_build, then assert_not_called() on both mocks.

juspence (commented):
I think we should change how we reuse the helper methods in the tests, and we're missing the Celery task decorator now. But the rest of the Brew logic is good, I think.

corgi/tasks/common.py
jasinner: I tried that, but unfortunately Callable doesn't have the '.delay()' function used below.

corgi/tasks/brew.py
jasinner: Oh yes, I'll add it back, thanks.

corgi/tasks/brew.py
jasinner: Good idea. I'll change it.

corgi/tasks/brew.py
jasinner: brew_tags attribute of Product Stream is a dictionary, not a list.

corgi/tasks/brew.py
jasinner: It's a good idea. We already log each time we create a relation, but I guess it's also a good idea to provide this count here so we know where the new relations came from.

tests/test_brew_collector.py
jasinner: Ok, I'll change it to .filter().exists() because it's more explicit what's be testing than throwing an exception in my opinion.

tests/test_brew_collector.py
jasinner: Ok, I like the 2nd approach, as we need the _create_relations_for_tag function to actually make the data changes in order to assert those changes are as we expect.

corgi/tasks/common.py
jasinner: Actually I can use celery.local.PromiseProxy as the type instead of Callable.

tests/test_manifests.py
jasinner: Ok, I just moved the save_product function used by test_brew_collector.py::test_fetch_container_build_rpms into conftest.py, and leave test_manifests.py as it was (without fixtures).

tests/test_manifests.py
jasinner: I removed the markers.

pyproject.toml
jasinner: I removed the markers.

PR #477, opened 2023-05-26
CORGI-671: Fix go component type filter

juspence:
@RedHatProductSecurity/corgi-devs Please review this fix for some bad code and data, which caused Golang components to be created without a "go_component_type" key in their meta_attr. This caused them to get excluded when a user filtered for "?gomod_components=True" (only go modules) or "?gomod_components=False" (only go-packages).

I have a commit that also fixes the filter (include unknown types when gomod_components=False), but I don't want to merge it since it makes the code rather ugly. I think the right solution is to fix the data - we should never have Golang components with an unknown / missing go_component_type.

In that case, the current filter behavior is a useful sanity-check, since we can verify that the "?type=GOLANG" count equals the "?gomod_components=True" count plus the "?gomod_components=False" count. That's actually how this bug was discovered to begin with.


corgi/collectors/brew.py
```diff
@@ -286,6 +273,33 @@ def get_rpm_build_data(self, build_id: int) -> dict:
...
+                "source": ["specfile"],
+            }
+
+            bundled_component: dict[str, Union[str, dict[str, Union[str, list[str]]]]] = {
+                "type": component_type,
+                "namespace": Component.Namespace.UPSTREAM,
+                "meta": bundled_component_meta,
+            }
+            if component_type == Component.Type.GOLANG:
+                # TODO: Do we ever bundle go-packages, or only modules?
```
juspence: Would like a quick sanity-check on this before merge. I think we can only bundle modules, and packages can't be used / downloaded independently of a module.

jasinner: Bundles entries in an RPM spec file are created by a human, so it's hard to predict what will be put in them. If we had to decide here if they where a package or a bundle I think the best predictor is the number of '/' characters. 
Two '/' characters as in 'github.com/gorilla/context' means it's a go module. More than two '/' characters indicates a package, for example 'github.com/stretchr/testify/require'. The same is true for anywhere go dependencies are used, not just in RPM spec files.
It's also true that some go components could have less than 2 '/' characters in the name, for example standard library packages such as 'archive/zip' only have a single '/'.

corgi/core/migrations/0067_fix_missing_go_component_type.py
```diff
@@ -0,0 +1,36 @@
...
+    """Set missing go_component_type for Golang modules
+    bundled in source RPMs, and upstream of container images"""
+    Component = apps.get_model("core", "Component")
+
+    # These are bundled into an SRPM - we assume they are always go modules
+    for component in Component.objects.filter(
+        type="GOLANG", meta_attr__go_component_type__isnull=True, meta_attr__source=["specfile"]
+    ).iterator():
+        with transaction.atomic():
+            component.meta_attr["go_component_type"] = "gomod"
```
jasinner: I'm not sure this is always the case. Perhaps an additional check on the number of '/' characters is required.

corgi/core/migrations/0067_fix_missing_go_component_type.py
```diff
@@ -0,0 +1,36 @@
...
+            component.save()
+
+    # These are always upstreams of a container image
+    # The first cnode has type SOURCE and a non-NULL parent
+    for component in Component.objects.filter(
+        type="GOLANG", meta_attr__go_component_type__isnull=True, meta_attr__source__isnull=True
+    ).iterator():
+        with transaction.atomic():
+            component.meta_attr["go_component_type"] = "gomod"
```
jasinner: Again, I think a check on the number of '/' characters might be a good idea.

jasinner (commented):
FWIW here's the [Fedora Go Packing guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/). Although I'm not sure how useful it is.

corgi/collectors/brew.py
juspence: Discussed in Gchat / meetings - checking the number of '/' characters will give incorrect results in some cases, so we're just going to leave go_component_type unset in this case.

OpenLCS will be unable to scan Go modules bundled into RPMs directly, but that's OK since they already scan the parent SRPM's source code. That scan result should include license / copyright data for all the bundled components inside the SRPM, so nothing should be missing.

juspence:
Thanks for the review and the info about go packaging guidelines. Since we're not able to tell if a bundled Golang component is a module or package, I just removed the related changes (based on our previous discussion).

I kept the fix for upstream_go_modules missing a go_component_type / source value in their meta_attr. That only affects about 2 or 3 hundred components, but it's still good to fix. I also added some other small bugfixes I found while looking over all the related code.

I'm going to merge this PR tomorrow morning since tests pass locally - feel free to review again if you'd like, or you can ignore if you don't have time.

PR #479, opened 2023-05-30
CORGI-680 slow latest queries

jasinner:
Replace the current latest_components filter which results in one query for each root component in a stream, with a single aggregate query which groups components by name (namespace, and arch) for all root components in a stream. This dramatically reduces the number of queries required for getting the latest components in a stream.

I also moved epoch from meta_attr to a Component field because querying by JSONField attributes is cumbersome as evidenced by the _ensure_epoch function. With epoch as a field I was able to remove the ensure_epoch function, and also remove the EPOCH_TYPES constant. However we could remove that commit if we didn't want the epoch field polluting non RPM components without affecting the rest of this PR too much.


jasinner:

corgi/core/models.py
```diff
@@ -930,12 +932,27 @@ def latest_components(
...
+            for nevra in nevra_data:
+                nevras.append(
+                    (
+                        nevra["nevra"],
+                        nevra["epoch"],
+                        nevra["version"],
+                        nevra["release"],
+                    )
+                )
+            latest_nevra = sorted(nevras, key=functools.cmp_to_key(self._compare_packages))[-1]
```
juspence: Let's just make the list we want directly, instead of appending to it:
```suggestion
            nevras = tuple(
                (nevra["nevra"], nevra["epoch"], nevra["version"], nevra["release"])
                for nevra in nevra_data
            )
            latest_nevra = sorted(nevras, key=functools.cmp_to_key(self._compare_packages))[-1]
```

We can also avoid building the whole list in memory and calling sorted(), which will cause OoM if the number of items to sort is very large:
```suggestion
            latest_nevra = nevra_data[0]
            for current_nevra in nevra_data:
                # if the current nevra is newer than the "latest" nevra we already have
                if self._compare_packages(current_nevra, latest_nevra):
                    latest_nevra = current_nevra
```

corgi/core/models.py
```diff
@@ -930,12 +932,27 @@ def latest_components(
...
         """Return components from latest builds across all product streams."""
-        distinct_components = (
-            self.values_list("namespace", "name", "arch").order_by().distinct().iterator()
+        nevras_by_components = self.values_list("namespace", "name", "arch").annotate(
+            nevras=JSONBAgg(
+                RawSQL(
+                    "json_build_object('nevra', nevra, 'epoch', epoch, 'version', core_component.version, 'release', release)",
+                    (),
+                )
+            )
```
juspence: Can you add a comment or detailed commit message that explains what this is doing? I don't really get why we need JSONBAgg and RawSQL.

Looking over all the related changes, I think we're just getting a list of all nevras / versions for each component (namespace / name / arch) triple in the queryset, then sorting by epoch / version / release (using rpm.labelCompare / in Python) and taking only the last / latest nevra.

I think there's a much simpler way to do this - the old code we had used a Django QuerySet with a subquery. It found a list of all nevras / versions for each component (namespace / name / arch) triple in the queryset, then sorted (using SQL / in the DB) by values that didn't give the correct result (e.g. SoftwareBuild completion_time).

If we take that old code and fix the sorting, we could avoid doing the sort in Python / running out of memory when the queryset is very large. I feel like that code was easier to understand, and should work the same as the rpm labelCompare method as long as we sort by epoch first, then version, then release.

I also see we have ```core_component.version```, but only ```release``` instead of ```core_component.release```. Is that a bug?

corgi/core/models.py
```diff
@@ -888,71 +889,67 @@ class ComponentQuerySet(models.QuerySet):
...
+                        "'nevra', nevra, "
+                        "'epoch', epoch, "
+                        "'version', core_component.version, "
+                        "'release', release)",
+                        (),
+                    )
+                )
+            )
+            .order_by()
+            .distinct()
```
juspence, line 948: Does .iterator() not work with .annotate()? I'd prefer to keep using it so that we don't risk an OoM with large querysets.

juspence (commented):
It's hard for me to understand what the new latest filter code is doing, but the epoch changes LGTM. Based on the PR description I get what the latest filter change is supposed to do and why we need it, I'm just struggling to understand the implementation.

Whether we use this new Python-based latest filter or the old SQL-based latest filter, let's add a performance test so that we can see what the improvement is and make sure we don't have regressions in the future.

corgi/core/models.py
jasinner: I will add a comment which explains what this is doing. This would be simpler and not require JSONBAgg or Raw SQL if we just aggregated the nevras. I didn't do that because we don't take the nevras from Brew, but build it ourselves from the name/epoch/version/release/arch values in the get_nevra function of Component. I'm worried we will build it incorrectly which will lead to bad results here. Also, we'd have to strip the name and arch values before using labelCompare because it expects only a tuple of epoch/version/release. Fetching each individual field allows us to ensure we have the correct value for each directly from Brew, and also not have to rebuild the nevra once we find the latest.

The reason we stopped using the database to do the sorting was that we need to use RPM schemantics to compare the versions, see this comment from [RHEL security architect](https://issues.redhat.com/browse/CORGI-391?focusedId=21407021&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21407021). The rules around this are complicated and not something that's easy to replicate in the DB directly.

When I was originally testing this, I added core_component table name to version because the database seemed to be confused about which 'version' field we where after. However I tested removing the table name now and it didn't complain so I'm going to remove that.

corgi/core/models.py
jasinner: It does work so I added it back.

corgi/core/models.py
juspence: I'm not proposing you take only the nevras and ignore the epoch / version / release. Using those epoch / version / release values directly is good - we shouldn't try to parse the NEVRA when we already have the parsed data we need.

I do agree with Tomas's comment in that ticket. The rules for RPM version comparisons / semantics are just:
1. Compare epoch first
2. Compare version next
3. Compare release last

His comment has two components with the same epoch and version - only the release portion is different. If you do below in Python, you get the expected result that matches his comment:
```
>>> "2.el7" > '2.ael7b'
True
```
Doing this epoch, version, release ordering / comparison in the DB should work, right? The problem with our old DB sorting logic was that we were using SoftwareBuild completion timestamps, which we should just ignore (build timestamps have nothing to do with versions).

corgi/core/models.py
jasinner: I don't think the first suggestion here will make any difference to the memory consumption, and you seem to be missing the type cast of the epoch value to string. I thought this loop was complicated enough to warrant using a traditional for loop.

I like the 2nd option, I'll update the loop to compare during the iterator to avoid building the list.

corgi/core/models.py
jasinner: Actually I got the error again about version. So I'm going to qualify the table.

`psycopg2.errors.AmbiguousColumn: column reference "version" is ambiguous
LINE 1: ...object('nevra', nevra, 'epoch', epoch, 'version', version,`

juspence: I looked over my reference again:
https://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/#how-rpm-compares-version-parts

The first part of the RPM version logic is pretty simple - compare epoch, then version, then release. But the actual logic of how to do those comparisons is more complicated than I remember, e.g. the magic ~ tilde character.

So I guess doing this in Python is our only choice, and in that case this LGTM as long as we're careful to avoid OoM issues with sorted().

corgi/core/models.py
juspence: Yes, I was commenting on an older commit (looking at each commit individually to understand better) so the diff here doesn't have the epoch-to-str cast. The first suggestion is just a strong style preference. The second one will avoid out of memory issues. Thanks for fixing!

juspence (approved):
LGTM as long as we:

1. Can use .iterator() and avoid sorted() so that OoM shouldn't happen
2. Add a comment explaining what the filter is doing
3. Minor: add a performance test to prevent future regressions, if you have time

Thanks for making those changes!

PR #480, opened 2023-06-02
CORGI-597: Fix relationship type for arch-specific children of noarch containers

juspence:
@mprpic Please review (you're given as the stakeholder in the ticket). This change should give us SPDX relationships in our manifests like:

1. VARIANT_OF for containers which are child / provided components (purl starts with "pkg:oci/")
2. DEV_DEPENDENCY_OF for any non-container dev_provided component (purl does not start with "pkg:oci/", node has PROVIDES_DEV type)
3. CONTAINED_BY for all other provided components (which are not containers or dev dependencies)

The first commit is the logic for above, the second was just whitespace changes to help me debug / diff the manifests locally while testing this.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1871,7 +1871,7 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         return (
             ComponentNode.objects.filter(pk__in=provides_set)
             .using(using)
-            .values_list("type", "object_id")
+            .values_list("purl", "type", "object_id")
             # Ensure generated manifests only change when content does
             .order_by("object_id")
             .distinct()
--- a/corgi/web/templates/base_manifest.json
+++ b/corgi/web/templates/base_manifest.json
@@ -1,29 +1,26 @@
-{# Used as the base of all manifest.json templates, obj is either a Component or a ProductModel subclass #}
-{
-    "creationInfo": {
-        "created": "{% now 'Y-m-d' %}T{% now 'H:i:00' %}Z",
-        "creators": [
-            "Organization: Red Hat Product Security (secalert@redhat.com)"
-        ],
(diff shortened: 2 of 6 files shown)


mprpic (approved):
Just one clarifying question: the only container image relationships we have in place is that a noarch container provides (i.e. ComponentNodeType.PROVIDES) an arch-specific container, right? We don't represent relationships between images and their base images, or images and their source containers. Because VARIANT_OF might not be the best SPDX relationship type for those relationships.

Code looks good!

juspence:
> Just one clarifying question: the only container image relationships we have in place is that a noarch container provides (i.e. ComponentNodeType.PROVIDES) an arch-specific container, right?

I had the same question while implementing. I think that only arch-specific containers will appear in the "provides" queryset, while the noarch container will always be part of the parent queryset (either as "root components in some stream" or "this component that we're manifesting").

> We don't represent relationships between images and their base images, or images and their source containers. Because VARIANT_OF might not be the best SPDX relationship type for those relationships.

We definitely aren't linking regular containers with source containers. We looked at doing this a while ago, but we don't have any hard requirements and implementation might be tricky.

We have the base image info in the meta_attr if memory serves, but we're not using it today. If we start linking base images with their children in the future and don't update our manifests, I think that e.g. the root / noarch openssl-container will be reported as a VARIANT_OF the base image / parent ubi9-container. So if we're OK with using VARIANT_OF for all provided container children, this shouldn't need any changes.

Or we could extend this logic to check the purl names, not just the purl types. If the name in the purl is different from the component being manifested (e.g. we have "openssl-container" in some node's purl, but we're manifesting "ubi9-container"), we could report GENERATED_FROM or DESCENDANT_OF instead.

PR #481, opened 2023-06-02

juspence:
@jasinner Would you mind looking over this change to the Cachito manifesting logic? It's not complete, since I still need to fix all the existing tests.

I also need to add new test data for Cachito manifests that use the "yarn" package manager, write a test for the missing transitive dependencies case, and figure out why our existing tests didn't catch this / how we're getting the magic numbers we assert on in the tests.

Before I spend all the effort to do above, I'd like to make sure this isn't the completely wrong approach. In other words, is there some reason that we look at both the .packages and .dependencies keys in Cachito manifests for Golang, but only the .packages key for other types?

I assume that's a bug and it's why dependencies are missing, but the tests make it hard to tell. Maybe it's not a bug, and the "new components" I've discovered are really just duplicates of existing components that got added twice?


juspence:
@RedHatProductSecurity/corgi-devs Please review. This fixes a few bugs:

1. Go packages have "vendor/" stripped from their name if they're children of Go modules (which are children of another component), but don't have "vendor/" stripped if they're direct children of that other component. There are 628 packages that will be updated by this migration.
2. We would silently fail to process unknown Cachito package types / skip saving these components, instead of raising an error.
3. We didn't correctly map transitive dependency types in Cachito to Corgi types. We only mapped root / top-level dependency types in Cachito to Corgi types. This may or may not be the root cause of the bug Rogue reported.

I split this PR since I want to merge this change and test the last two bugfixes / reingest a few builds.

There are probably more bugfixes needed in our Cachito manifest-processing logic, but I'm going to handle those in a follow-up PR since they're more complex and need lots of test changes I don't fully understand.

kgrant-rh (approved):

PR #487, opened 2023-06-06
CORGI-601: Fix RPM filenames in manifests

juspence:
@RedHatProductSecurity/corgi-devs Please review. I'm still not sure this is the right approach, e.g. if there's ever some RPM which doesn't use its NEVRA as the filename customers download in Pulp / other locations.

Pulling this data from Brew would require changes to logic in a few different places - we get data for each RPM, but return only the build ID and not any other data (including the filename). So for now I've taken the simple approach and just set the filename on save using the NEVRA we already built.

I also cleaned up some unused fields like version_arr and release_arr while I was looking at the NEVRA field and .save() code. The el_match filter needs to stay, and needs to work for any component, not just root components.

The el_match field / filter parses any release value with ".el" in it, so that users can find "which RHEL version was this component released for", when looking at component data for a layered product like OpenShift. That supports showing "the latest components shipped to RHEL 8" vs. "the latest components shipped to RHEL 9" when the user turns on this filter.


mprpic (approved):
This looks correct; RPMs should always use `<nevra>.rpm` as the file name.

LGTM!

PR #489, opened 2023-06-08
CORGI-662 create variants for et relations

jasinner:
Associate streams using BREW_TAG relations with ProductVariant where there are erratum that link the two. For example if there is a ProductStream eg. `openshift-4.13.z`, which uses brew_tags to associate builds with that ProductStream, when one of the builds gets released with the `8Base-RHOSE-4.13` ProductVariant the stream will then be associated with that ProductVariant.

This should allow us to associate builds and their components with ProductVariants where that wasn't possible before with stream using BREW_TAG relations. It also allows us to build a list of CPEs for streams which don't have errata_info, composes, or yum_repository relations.

### DIFF
--- a/corgi/tasks/errata_tool.py
+++ b/corgi/tasks/errata_tool.py
@@ -1,16 +1,21 @@
+from collections import defaultdict
+
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
 from django.db import transaction
-from django.db.models import QuerySet
 
 from config.celery import app
 from corgi.collectors.errata_tool import ErrataTool
-from corgi.collectors.models import CollectorRPMRepository
+from corgi.collectors.models import (
+    CollectorErrataProductVariant,
+    CollectorRPMRepository,
+)
 from corgi.core.models import (
     Channel,
     ProductComponentRelation,
     ProductNode,
+    ProductStream,
     ProductVariant,
     SoftwareBuild,
 )
@@ -35,15 +40,83 @@ def save_variant_cdn_repo_mapping() -> None:
     ErrataTool().save_variant_cdn_repo_mapping()
 
 
+def associate_variant_with_build_stream(
+    build_id: str, build_type: str, variant_names: list[str]
+) -> bool:
+    try:
+        software_build = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
+    except SoftwareBuild.DoesNotExist:
+        logger.warning(f"SoftwareBuild with id {build_id} of type {build_type} does not exist.")
+        return False
+
+    # This function is design to discover new variant to stream mappings
+    # If a variant already exists, we assume it is already correct mapped to a stream and give up
+    # This avoids the slow build_streams query below when it's not necessary
+    if ProductVariant.objects.filter(name__in=variant_names).exists():
+        logger.debug(f"Not associating Product Variants {variant_names}")
+        return False
+
+    build_streams = ProductStream.objects.filter(components__software_build=software_build)
+    any_created = False
+
+    build_streams_count = len(build_streams)
+    # Only associate this variant to a single stream to keep one-to-many cardinality between
+    # ProductStream and ProductVariant
+    if build_streams_count != 1:
+        return any_created
+
+    stream = build_streams[0]
+    for variant_name in variant_names:
+        # The CollectorErrataProductVariant should always exist, let the DoesNotExist error
+        # propagate if it doesn't
+        try:
+            errata_variant = CollectorErrataProductVariant.objects.get(name=variant_name)
+            cpe = errata_variant.cpe
+        except CollectorErrataProductVariant.DoesNotExist:
+            logger.warning(f"CollectorErrataProductVariant with name {variant_name} does not exist")
+            cpe = ""
+        # We don't use update_or_create here because we checked earlier in this function
+        # ProductVariant objects with this name do not already exists.
+        ProductVariant.objects.create(
+            name=variant_name,
+            cpe=cpe,
+            products=stream.products,
+            productversions=stream.productversions,
+            productstreams=stream,
+            meta_attr={
+                "build_id": build_id,
+                "build_type": build_type,
+            },
+        )
+        any_created = True
+        # CORGI-703 - Also add channels for each new variant here
+    return any_created
+
+
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
-def slow_save_errata_product_taxonomy(erratum_id: int) -> None:
-    logger.info(f"slow_save_errata_product_taxonomy called for {erratum_id}")
-    relation_builds = _get_relation_builds(erratum_id)
-    for build_id, build_type in relation_builds:
+def slow_save_errata_product_taxonomy(
+    build_variants: dict[str, list[str]], build_type: str
+) -> bool:
+    logger.info(f"slow_save_errata_product_taxonomy called with build_variants {build_variants}")
+
+    variants_created = False
+    for build_id, variant_ids in build_variants.items():
+        # Only associate stream with variants where an existing relation not of type ERRATA exists
+        if (
+            ProductComponentRelation.objects.filter(
+                build_id=build_id,
+                build_type=build_type,
+            )
+            .exclude(type=ProductComponentRelation.Type.ERRATA)
+            .exists()
+        ):
+            if associate_variant_with_build_stream(build_id, build_type, variant_ids):
+                variants_created = True
         logger.info("Saving product taxonomy for build (%s, %s)", build_id, build_type)
         # once all build's components are ingested we must save product taxonomy
         sb = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
         sb.save_product_taxonomy()
+    return variants_created
 
 
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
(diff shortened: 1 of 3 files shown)


jasinner:
@RedHatProductSecurity/corgi-devs please review the design of this. It's not functionally complete, but remaining changes should be small.

corgi/tasks/errata_tool.py
```diff
@@ -1,16 +1,22 @@
+from collections import defaultdict
+from typing import Tuple
```
juspence: You don't need to import this - you can always use the built-in tuple[type1, type2, ...] in place of typing.Tuple[type1, type2, ...]

corgi/tasks/errata_tool.py
```diff
@@ -125,16 +152,34 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
-        slow_save_errata_product_taxonomy.delay(erratum_id)
+        for build_id in build_variants.keys():
+            sb = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
+            sb.save_product_taxonomy()
 
     logger.info("Finished processing %s", erratum_id)
 
 
-def _get_relation_builds(erratum_id: int) -> QuerySet:
+def _get_relation_builds(erratum_id: int) -> Tuple[dict[str, list[str]], str]:
```
juspence:
```suggestion
def _get_relation_builds(erratum_id: int) -> tuple[dict[str, list[str]], str]:
```

corgi/tasks/errata_tool.py
```diff
@@ -125,16 +152,34 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
+    for build_id, build_type, variant_id in relation_builds:
+        build_variants[build_id].append(variant_id)
+        build_types.add(build_type)
+
+    # Errata are not used in community products, and we are not yet attaching other builds types
+    # like HACBS to errata. If that changes, we should review slow_load_errata for correctness
+    if len(build_types) > 1:
+        raise ValueError("Multiple build types found for errata %s", erratum_id)
+    elif len(build_types) == 0:
+        build_type = BUILD_TYPE
```
juspence, line 208: It took me a minute to figure out the 0 build_types case exists, so let's add a comment here since it's not immediately obvious. 
```suggestion
        # if the queryset above had no results
        build_type = BUILD_TYPE
```

corgi/tasks/errata_tool.py
```diff
@@ -35,15 +41,41 @@ def save_variant_cdn_repo_mapping() -> None:
...
 
-@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
-def slow_save_errata_product_taxonomy(erratum_id: int) -> None:
-    logger.info(f"slow_save_errata_product_taxonomy called for {erratum_id}")
-    relation_builds = _get_relation_builds(erratum_id)
-    for build_id, build_type in relation_builds:
-        logger.info("Saving product taxonomy for build (%s, %s)", build_id, build_type)
-        # once all build's components are ingested we must save product taxonomy
-        sb = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
-        sb.save_product_taxonomy()
```
juspence: I think there might be other modules that use this code, and we have two different codepaths below that both depend on the logic here. But after your changes, only one of the codepaths below has the new logic to link streams to variants.

So I'd prefer we keep this task separate from the slow_load_errata task, and just update the parameters. We can pass build_id, build_type, and the variant_ids to associate if a matching Brew tag relation exists. The task can link the stream to the variant if needed, and save the product taxonomy as well. That avoids duplicating code in slow_load_errata.

corgi/tasks/errata_tool.py
```diff
@@ -35,15 +41,41 @@ def save_variant_cdn_repo_mapping() -> None:
...
+    )
+    any_created = False
+    for variant_name in variant_names:
+        # The CollectorErrataProductVariant should always exist, let the DoesNotExist error
+        # propagate if it doesn't
+        errata_variant = CollectorErrataProductVariant.objects.get(name=variant_name)
+        stream = build_streams.first()
+        # Only associate this variant to a single stream to keep one-to-many cardinality between
+        # ProductStream and ProductVariant
+        if stream and len(build_streams) == 1:
```
juspence: We're going to repeatedly make DB calls for the stream inside the for-loop, but they're the same for each variant name. Let's move them out of the loop instead. We can also avoid looking up the variant when the stream condition is false / we don't do any linking:
```suggestion
    # This evaluates the build_streams queryset once and avoids further DB calls
    build_streams_count = len(build_streams)
    stream = None
    if build_streams_count:
        # Reuse the evaiuated queryset instead of calling .first() / doing a new DB lookup
        stream = build_streams[0]
    for variant_name in variant_names:
        # Only associate this variant to a single stream to keep one-to-many cardinality between
        # ProductStream and ProductVariant
        if stream and build_streams_count == 1:
            # The CollectorErrataProductVariant should always exist, let the DoesNotExist error
            # propagate if it doesn't
            errata_variant = CollectorErrataProductVariant.objects.get(name=variant_name)
```

corgi/tasks/errata_tool.py
```diff
@@ -35,15 +41,41 @@ def save_variant_cdn_repo_mapping() -> None:
...
+            _, created = ProductVariant.objects.get_or_create(
```
juspence: Should we do update_or_create here like we do in other places? I guess that would wipe out the existing meta_attr, but if some errata_variant's CPE changes, I think we want to get the updated value in Corgi.

corgi/tasks/errata_tool.py
```diff
@@ -35,15 +41,41 @@ def save_variant_cdn_repo_mapping() -> None:
...
+                name=errata_variant.name,
+                defaults={
+                    "cpe": errata_variant.cpe,
+                    "products": stream.products,
+                    "productversions": stream.productversions,
+                    "productstreams": stream,
+                    "meta_attr": {
+                        "associated_by_build": build_id,
+                        "associated_by_build_type": build_type,
```
juspence: I get why we want this, but I'm not sure we can rely on it. We also create Variants in the prod-defs collector based on the Brew tag names and ET variant data, but we don't set this key to None / an empty "' string in those cases.

Maybe it's better to have a "source" key in the Variant meta-attr just like the Component meta_attr, and to always set it to one of f"brew_tag: {tag_name}", f"build_id: {build_id}", or f"variant_name: {variant.name}". That handles the Brew tag case in the prod-defs collector, the variant-to-stream linking based on build ID here, and the ET Variant case in the prod-defs collector.

Do we need any other updates to that prod-defs collector code? We use update_or_create there (and shouldn't change this), so that code might wipe out this meta_attr we set here. We can leave meta_attr out of defaults, do update_or_create in all places, then check the keys in the old vs. new meta_attr and merge the two if needed.

corgi/tasks/errata_tool.py
```diff
@@ -96,19 +113,29 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
+    elif len(build_variants) == no_of_processed_builds:
+        for build_id, variant_ids in build_variants.items():
+            # Only associate stream with variants where an existing relation of type BREW_TAG exists
+            # Other types should already have variants associated with them
+            if ProductComponentRelation.objects.filter(
+                build_id=build_id,
+                build_type=build_type,
+                type=ProductComponentRelation.Type.BREW_TAG,
+            ).exists():
+                associate_variant_with_build_streams(build_id, build_type, variant_ids)
```
juspence: We're relying on BREW_TAG relations to do this linking, and I think we create these relations when processing the brew_tags stanza for some stream defined in prod-defs. What happens if a Brew tag defined there gets removed? Do we need to clean up the old BREW_TAG relations?

In other words, do we need to stop linking tagged builds and their related ProductVariants to this stream? Or maybe Brew tags are never removed from a stream in prod-defs - feel free to ignore me if so.

corgi/tasks/errata_tool.py
```diff
@@ -125,16 +152,34 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
             )
 
     if force_process:
-        slow_save_errata_product_taxonomy.delay(erratum_id)
+        for build_id in build_variants.keys():
+            sb = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
+            sb.save_product_taxonomy()
```
juspence: You added the stream-to-variant linking in the case above, but not this case here. Before this change, both cases were identical and called slow_save_errata_product_taxonomy.

I don't fully understand the differences, but I think we should keep them identical, keep slow_save_errata_product_taxonomy as a separate task, and just add the stream-to-variant linking into that task instead of duplicating the logic here. Please ignore me if I've misunderstood something.

tests/test_errata_data.py
```diff
@@ -201,14 +210,165 @@ def setup_models_for_variant_repos(repos, ps_node, variant, et_id):
...
+    component = ComponentFactory(software_build=sb)
+    component.productstreams.set([stream])
+    assert associate_variant_with_build_streams(sb.build_id, sb.build_type, ["8Base-RHOSE-4.13"])
+    product_variant = ProductVariant.objects.get(name="8Base-RHOSE-4.13")
+    assert product_variant
+    assert product_variant.cpe == "cpe:/a:redhat:openshift:4.13::el8"
+    assert product_variant.productstreams == stream
+
+
+# Assert that we only associate a variant with a stream where the stream is active
```
juspence: What happens when an active stream becomes inactive? If the task reruns, we won't associate the stream with any new Variants (based on tagged builds which are shipped in some errata for those Variants). Do we need to unlink any Variants we already associated?

juspence (approved):
I agree in general that this idea is a good way to avoid the manual CPE mapping. But I'm not sure I understood all the details of how it's implemented, and I didn't look closely at the tests.

jasinner:
I ran this on all the existing relations and it only uncovered new variants for 4 streams. So I don't think it's worth the added complexity and maintenance burden to do this type of linking.

PR #490, opened 2023-06-08
Email failed tasks that end in RETRY and REVOKED

mprpic:
This was specifically causing us to miss the daily failures of the managed services manifesting job because of a network connection issue that would always cause the Celery task to end in a RETRY state.

@RedHatProductSecurity/corgi-devs quick review please.

### DIFF
--- a/corgi/tasks/monitoring.py
+++ b/corgi/tasks/monitoring.py
@@ -1,5 +1,6 @@
 from datetime import timedelta
 
+from celery import states as celery_states
 from celery.signals import beat_init
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton, clear_locks
@@ -94,7 +95,7 @@ def email_failed_tasks():
 
     failed_tasks_threshold = get_last_success_for_task("corgi.tasks.monitoring.email_failed_tasks")
     failed_tasks = TaskResult.objects.filter(
-        status__exact="FAILURE",
+        status__in=(celery_states.FAILURE, celery_states.RETRY),
         date_done__gte=failed_tasks_threshold,
     ).using("read_only")
 


corgi/tasks/monitoring.py
```diff
@@ -94,7 +95,7 @@ def email_failed_tasks():
 
     failed_tasks_threshold = get_last_success_for_task("corgi.tasks.monitoring.email_failed_tasks")
     failed_tasks = TaskResult.objects.filter(
-        status__exact="FAILURE",
+        status__in=CELERY_TASK_EXCEPTION_STATES,
```
juspence: I'm about to go revoke about 30,000 tasks (for container data reload) since I've discovered more bugs. There's no point reloading the current data when not all the bugs are fixed yet - we'll need to reload again anyway after the next round of bugfixing.

If we email revoked tasks and not just retried tasks, cases like this are going to flood the number of "issues" that we report. Since revoking is only done manually for a specific reason, I think we should avoid reporting these REVOKED tasks to avoid spamming the monitoring email even more.

juspence: I'm also not sure that we should report retried tasks, but I'm probably not understanding the Celery task states. I thought that we'd create a RETRY TaskResult for any retried attempts, and a SUCCESS / FAILURE TaskResult for the final attempt once the maximum number of retries has been reached. In that case, wouldn't we get a lot of email spam for tasks which had some temporary error but retried successfully?

Or if that's wrong and there's only one TaskResult, does its status change to SUCCESS / FAILURE at the end of the retry loop? In that case, our existing monitoring emails should report any failures, right? I'm confused how a TaskResult ended up in the RETRY state. The number of retries should be limited - it's not an infinite retry loop I hope (or if it is, let's fix that bug).

corgi/web/static/cnv-4.8-9ea63480-edda-45e4-bc44-a4980bf5f3de.json
juspence, line 1: Looks like a manifest file that got committed by mistake?

juspence (approved):
I'm not really sure if this is needed, but assuming I'm confused and this really is a problem, the solution LGTM.

corgi/web/static/cnv-4.8-9ea63480-edda-45e4-bc44-a4980bf5f3de.json
mprpic: :man_facepalming:

Will remove.

corgi/tasks/monitoring.py
juspence: Discussed in Gchat / standup - we're not sure where the RETRY tasks come from, but something weird is happening with the TaskResults so reporting these until we fix that other bug LGTM.

juspence: Discussed in Gchat / standup - we'll just report tasks in the FAILURE or RETRY states for now, and ignore any tasks in the REVOKED state.

PR #491, opened 2023-06-08
Add a UMB dispatcher which routes messages to many handlers

kgrant-rh:
Refactor UMB handling by creating a dispatcher class which maintains a registry of handlers which each specify the topics they're listening for and methods to handle messages on those topics.

### DIFF
--- a/corgi/monitor/consumer.py
+++ b/corgi/monitor/consumer.py
@@ -1,6 +1,6 @@
 import json
 import logging
-from typing import Optional
+from collections.abc import Callable
 
 from django.conf import settings
 from proton import Event, SSLDomain
@@ -12,17 +12,33 @@ from corgi.tasks.errata_tool import slow_handle_shipped_errata
 
 logger = logging.getLogger(__name__)
 
+# A method which receives an event, and returns true if the message was accepted,
+# or false if it should be released back into the queue.
+HandleMethod = Callable[[Event], bool]
 
-class UMBReceiverHandler(MessagingHandler):
-    """Handler to deal with received messages from UMB."""
+VIRTUAL_TOPIC_PREFIX = f"Consumer.{settings.UMB_CONSUMER}.VirtualTopic.eng"
 
-    def __init__(self, virtual_topic_addresses: dict[str, str], selectors: dict[str, str]):
-        """Set up a handler that listens to many topics and processes messages from each"""
-        super(UMBReceiverHandler, self).__init__()
 
+class UMBHandler:
+    """Handler to deal with received messages from UMB. Defines topics to listen to and
+    methods to invoke for handle messages in those topics."""
+
+    def __init__(self, addresses: dict[str, HandleMethod], selectors: dict[str, str]):
         # A mapping of virtual topic addresses to functions that handle topic messages
         # as determined by a specific listener.
-        self.virtual_topic_addresses = virtual_topic_addresses
+        self.virtual_topic_addresses = addresses
+
+        # A set of filters used to narrow down the received messages from UMB for this handler, if
+        # none, handle all messages
+        self.selectors = selectors
+
+
+class UMBDispatcher(MessagingHandler):
+    """Maintains a collection of UMBHandlers and dispatches messages to them"""
+
+    def __init__(self):
+        """Set up a handler that listens to many topics and processes messages from each"""
+        super(UMBDispatcher, self).__init__()
 
         # Set of URLs where UMB brokers are running. Use AMQP protocol port numbers only. See
         # specific URLs in settings; use env vars to select the appropriate broker.
@@ -38,10 +54,8 @@ class UMBReceiverHandler(MessagingHandler):
         self.ssl_domain.set_trusted_ca_db(settings.CA_CERT)
         self.ssl_domain.set_peer_authentication(SSLDomain.VERIFY_PEER)
 
-        # A set of filters used to narrow down the received messages from UMB; see individual
-        # listeners to see if they define any selectors or consume all messages without any
-        # filtering.
-        self.selectors = selectors
+        # A list of UMBHandlers to which messages will be dispatched
+        self.handlers: list[UMBHandler]
 
         # Ack messages manually so that we can ensure we successfully acted upon a message when
         # it was received. See accept condition logic in the on_message() method.
@@ -50,6 +64,33 @@ class UMBReceiverHandler(MessagingHandler):
         # Each message that is accepted also needs to be settled locally. Auto-settle each
         # message that is accepted automatically (this is the default value).
         self.auto_settle = True
+
+        # Handlers must be registered before the listener is started
+        self.started = False
+
+    def register_handler(self, handler: UMBHandler):
+        """Register an additional handler. Handlers should be registered before the dispatcher
+        is started."""
+        if self.started:
+            raise RuntimeError("Handlers must be added before UMBDispatcher is started")
+        elif not handler.virtual_topic_addresses:
+            raise ValueError("Handler has no addresses")
+        else:
+            self.handlers.append(handler)
+
+    @property
+    def virtual_topic_addresses(self) -> dict[str, HandleMethod]:
+        return {
+            addr: handler.virtual_topic_addresses[addr]
+            for handler in self.handlers
+            for addr in handler.virtual_topic_addresses
+        }
+
+    @property
+    def selectors(self) -> dict[str, str]:
+        return {
+            addr: handler.selectors[addr] for handler in self.handlers for addr in handler.selectors
+        }
 
     def on_start(self, event: Event) -> None:
         """Connect to UMB broker(s) and set up a receiver for each virtual topic address"""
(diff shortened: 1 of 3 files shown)


corgi/monitor/consumer.py
```diff
@@ -12,17 +12,33 @@
 
 logger = logging.getLogger(__name__)
 
+# A method which receives an event, and returns true if the message was accepted,
+# or false if it should be released back into the queue.
+HandleMethod = Callable[[Event], bool]
```
juspence, line 17: Nice job with the type hints here - I didn't realize you could qualify Callable like this.

corgi/monitor/consumer.py
```diff
@@ -129,14 +197,11 @@ def handle_shipped_errata(self, event: Event) -> None:
...
-            # message will be re-delivered to any available client again.
-            self.release(event.delivery, delivered=True)
+            return False
         else:
-            # Accept the delivered message to remove it from the queue.
-            self.accept(event.delivery)
+            return True
 
-    def handle_tags(self, event: Event) -> None:
+    def handle_tags(self, event: Event) -> bool:
```
juspence: Just returning a bool is a much better idea than having the same accept / release logic duplicated into every method - thank you for cleaning it up.

That lets us pass callables directly instead of string function names (which I think you're doing - I haven't finished looking), so this can be a static method now. Maybe that lets us simplify things further.

corgi/monitor/consumer.py
```diff
@@ -51,6 +65,33 @@ def __init__(self, virtual_topic_addresses: dict[str, str], selectors: dict[str,
...
+
+    def register_handler(self, handler: UMBHandler):
+        """Register an additional handler. Handlers should be registered before the dispatcher
+        is started."""
+        if self.started:
+            raise RuntimeError("Handlers must be added before UMBDispatcher is started")
+        elif not handler.virtual_topic_addresses:
+            raise ValueError("Handler has no addresses")
+        else:
+            self.handlers.append(handler)
```
juspence: This makes sense, but are we setting self.started to True anywhere? Should it be at the end of this method, or somewhere else?

corgi/monitor/management/commands/runmonitor.py
```diff
@@ -12,6 +12,8 @@ class Command(BaseCommand):
     def handle(self, *args: str, **options: dict[str, str]) -> None:
         if settings.UMB_BREW_MONITOR_ENABLED:
             try:
-                BrewUMBListener.consume()
+                dispatcher = UMBDispatcher()
+                dispatcher.register_handler(BrewUMBHandler())
+                dispatcher.consume()
```
juspence: The one thing I don't like about this pattern is that we have to know / choose which handlers we want to register. IMO it's fine for the base class to have that flexibility, but with the old way calling a subclass like BrewUMBListener automatically registed the right handlers and ser up the right listening logic.

My fear is that if we have to manually register the handlers, we'll add new topics / handling methods but forget to actually register those new handlers here.

I guess we need the flexibility to deal with SBOMer's / PNC UMB topics, right? And in the future we'd just add ```dispatcher.register_handler(PNCMBHandler())``` to this block?

So the BrewUMBListener class was poorly named, and we should just make a generic listener class for all UMB topics we want to listen to, whether that's SBOMer / PNC or Brew. We'll never run more than one, since ```.consume()``` enters an event loop and never returns.

And I think I finally get it - that's what your new UMBDIspatcher class is doing, right? Since we always call one UMBDispatcher, it has all the logic to listen to any topic using any number of handlers. The name change confused me a little, but I guess it's serving the same purpose as the UMBListener class we had before.

Can we just move the register_handler() logic into the class itself, and keep this file to a one-liner ```UMBDispatcher.consume()``` like we had before? Then when we want to add new topics / handlers / other logic, all the updates we need to make are limited to one file, and we won't need to add things here (so no risk of forgetting).

corgi/monitor/consumer.py
```diff
@@ -51,6 +65,33 @@ def __init__(self, virtual_topic_addresses: dict[str, str], selectors: dict[str,
...
+            addr: handler.virtual_topic_addresses[addr]
+            for handler in self.handlers
+            for addr in handler.virtual_topic_addresses
+        }
+
+    @property
+    def selectors(self) -> dict[str, str]:
+        return {
+            addr: handler.selectors[addr] for handler in self.handlers for addr in handler.selectors
+        }
```
juspence, line 176: I guess these properties are needed to support multiple handlers per dispatcher. But it seems like the old way was simpler - not necessarily a problem, if we need the extra complexity for the future.

I guess it's better to add a separate PNCHandler class with its own handler functions, topics, and mapping, instead of cluttering up the Brew handler with all this non-Brew logic.

corgi/monitor/consumer.py
```diff
@@ -69,25 +110,55 @@ def on_message(self, event: Event) -> None:
...
-
-class BrewUMBReceiverHandler(UMBReceiverHandler):
-    """Handle messages about completed Brew builds, tagged builds, and untagged builds"""
-
-    def handle_builds(self, event: Event) -> None:
+    def consume(self):
+        """Run a single message handler, which can listen to multiple virtual topic addresses"""
+        if not self.handlers:
+            raise ValueError("Dispatcher must have handler(s) registered before consuming")
+
```
juspence: Should we also raise ValueError if not self.started?

juspence (approved):
LGTM. The changes were a little confusing at first, but I like the abstraction you're going for. It does mean we have to rewrite a lot of tests, so if you'd prefer, you could punt and just add a smaller set of changes here instead of a full refactoring. Completely up to you - whatever design you think is best is fine by me, but below may save you some effort on the tests.

I think that you could combine the old BrewUMBListener into the generic UMBListener class. Then you could return bools from all the handler methods, and call self.accept or self.release in on_message(). That means that each handler method can be a staticmethod, and you can pass HandlerClass.function_name callables directly instead of string function names.

That means in the future, we can add PNC / SBOMer topics and functions to the generic UMBListener class's mapping of virtual_topic_addresses. We could keep the actual handler methods themselves in a separate PNCUMBHandler class (since they're now staticmethods and don't need a self arg), which avoids polluting the BrewUMBHandler class with non-Brew logic.

Then just call the generic UMBHandler with the big mapping of all topics to all functions (regardless of what class the function is in), which should spawn one generic Container with many receivers (one per topic) and call .run().

All together, that should support multiple topics / receivers with some abstraction / separation of concerns between Brew, PNC, and other topics, just like the current code does. But it wouldn't require as much refactoring of tests if I'm imagining this correctly.

corgi/monitor/consumer.py
kgrant-rh: Obviated by having the dispatcher register the handlers itself.

corgi/monitor/consumer.py
kgrant-rh: That's the impetus for this change, adding a separate handler for SBOMer will just require defining the addresses and methods to call for messages to those addresses.

corgi/monitor/consumer.py
kgrant-rh: Obviated by having the dispatcher register handlers itself, per our discussion.

kgrant-rh:
@juspence If you have time, I'd appreciate if you'd double-check and make sure I retained the spirit of the tests.

corgi/monitor/consumer.py
juspence: @kgrant-rh Can you add a staticmethod decorator on all the handler functions please?

tests/test_monitor.py
```diff
@@ -3,94 +3,73 @@
...
+
+def test_brew_umb_handler_setup():
+    """Test that the BrewUMBHandler class is set up correctly by the UMBDispatcher"""
+    # Stub out the SSLDomain config class to avoid needing real UMB certs in tests
+    with patch("corgi.monitor.consumer.SSLDomain") as mock_ssl_domain_constructor:
+        dispatcher = UMBDispatcher()
+
+    # The addresses and selectors of the dispatcher should match those of a handler
+    handler = BrewUMBHandler()
+    assert dispatcher.virtual_topic_addresses.keys() == handler.virtual_topic_addresses.keys()
```
juspence, line 73: Any reason why we assert .keys() instead of the whole dict? We should still copy the values / callback functions from the handler to the dispatcher, right?

tests/test_monitor.py
```diff
@@ -3,94 +3,73 @@
...
+    assert dispatcher.auto_settle is True
+
+
+def test_dispatcher_requires_addresses():
+    """Test UMBHandler raises ValueError for missing addresses"""
+
+    class NoAddressHandler(UMBHandler):
+        def __init__(self):
+            with pytest.raises(ValueError):
+                super(UMBHandler, self).__init__(None, None)
```
juspence: This test makes sense, but I think the implementation won't work - the NoAddressHandler class is only defined and never used by the test. So we'll never call its \_\_init\_\_ method and check the with pytest.raises statement.

Can you just call UMBHandler directly with those None args instead?
```suggestion
def test_dispatcher_requires_addresses():
    """Test UMBHandler raises ValueError for missing addresses"""

    with pytest.raises(ValueError):
        UMBHandler(None, None)
```

corgi/monitor/consumer.py
```diff
@@ -12,82 +12,50 @@
...
-        """Set up a handler that listens to many topics and processes messages from each"""
-        super(UMBReceiverHandler, self).__init__()
 
+class UMBHandler:
+    """Handler to deal with received messages from UMB. Defines topics to listen to and
+    methods to invoke for handle messages in those topics."""
+
+    def __init__(self, addresses: dict[str, HandleMethod], selectors: dict[str, str]):
+        if not addresses:
+            raise ValueError("UMBHandler has no addresses")
```
juspence, line 28: We could check that we define a callback function for each key in the dict, if you want to be really thorough. Just an idea that occurred to me, not really necessary.

juspence (approved):
LGTM if test coverage is reasonably close to what it was before. Thank you for the really nice PR!

tests/test_monitor.py
kgrant-rh: The comparison is between the BrewUMBHandler that UMBDispatcher creates itself, and another one created outside of it. The addresses and the function types are the same, but they differ in address.

tests/test_monitor.py
kgrant-rh: You're right, I forgot to instantiate the class. :expressionless:  I could call the UMBHandler constructor directly, but I'd prefer to test it with a subclass since that's the intended use.

tests/test_monitor.py
juspence: Can we use the dispatcher's instance of handler, using e.g. ```dispatcher.handlers[0].virtual_topic_addresses```? That would let us confirm the actual callback functions / dict.values() are getting set correctly in the dispatcher's virtual_topic_addresses property, which I think is important.

EDIT: Oops, didn't see this was already merged. It's important but not super important, so I may just tweak the assert if I ever touch this code again. No need for you to change anything here.

PR #493, opened 2023-06-09
UI tweaks

mprpic:
Just some minor (hopefully) improvements to the main page UI:

image

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -23,7 +23,8 @@ ADMINS = tuple(
     for name_and_email in os.getenv("CORGI_ADMINS", ";root@localhost").split(",")
 )
 
-DOCS_URL = os.getenv("CORGI_DOCS_URL")
+# Default to public Jira project if we can't point at any ops docs.
+DOCS_URL = os.getenv("CORGI_DOCS_URL", "https://issues.redhat.com/projects/SECDATA")
 OPENSHIFT_BUILD_COMMIT = os.getenv("OPENSHIFT_BUILD_COMMIT")
 PRODSEC_EMAIL = os.getenv("PRODSEC_EMAIL", "dev@example.com")
 
--- a/corgi/web/templates/rest_framework/api.html
+++ b/corgi/web/templates/rest_framework/api.html
@@ -1,0 +1,5 @@
+{% extends "rest_framework/base.html" %}
+
+{% block branding %}
+  <a class="navbar-brand" href="/">Component Registry</a>
+{% endblock %}
--- a/tests/data/image_archive_data.py
+++ b/tests/data/image_archive_data.py
@@ -1,5 +1,3 @@
-import os
-
 TEST_IMAGE_ARCHIVE = {
     "btype": "image",
     "btype_id": 4,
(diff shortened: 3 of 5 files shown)


config/settings/base.py
```diff
@@ -23,7 +23,8 @@
     for name_and_email in os.getenv("CORGI_ADMINS", ";root@localhost").split(",")
 )
 
-DOCS_URL = os.getenv("CORGI_DOCS_URL")
+# Default to public Jira project if we can't point at any ops docs.
+DOCS_URL = os.getenv("CORGI_DOCS_URL") or "https://issues.redhat.com/projects/SECDATA"
```
juspence: .getenv() defaults to None and supports a second arg for a custom default, just like .get(). So this can just be:
```suggestion
DOCS_URL = os.getenv("CORGI_DOCS_URL", "https://issues.redhat.com/projects/SECDATA")
```

tests/data/image_archive_data.py
```diff
@@ -1,5 +1,3 @@
-import os
-
```
juspence, line 2: Do you know why flake8 didn't catch this unused import? I usually run "flake8 ." at the project root, so it's not like it's only looking at "flake8 corgi/" and ignoring the tests/ or something.

corgi/web/templates/rest_framework/api.html
```diff
@@ -0,0 +1,5 @@
+{% extends "rest_framework/base.html" %}
+
+{% block branding %}
+  <a class="navbar-brand" href="/">Component Registry</a>
+{% endblock %}
```
juspence, line 5: Super minor, but browsers make lots of spammy requests for a favicon even when one doesn't exist / isn't defined on the page. If we're going to override the defaults, we may as well fix this issue too:
```suggestion
{% extends "rest_framework/base.html" %}

{% block style %}
{{ block.super }}
  <link rel="icon" href="data:," />
{% endblock %}

{% block branding %}
  <a class="navbar-brand" href="/">Component Registry</a>
{% endblock %}
```
Reference:
https://stackoverflow.com/questions/46842938/how-do-i-edit-the-favicon-in-the-browsable-api-in-django-rest-framework
https://stackoverflow.com/questions/1321878/how-to-prevent-favicon-ico-requests

juspence (approved):
LGTM!

corgi/web/templates/rest_framework/api.html
mprpic: We don't have a favicon on the main page either so I guess I'd prefer to leave this as is and try and add one (that works in both places) in a separate PR.

tests/data/image_archive_data.py
mprpic: Because we exclude this directory in:

```
[flake8]
exclude = .git/,venv/,.tox/,tests/data/
```

corgi/web/templates/rest_framework/api.html
juspence: Ack, FYI you could also add the "empty favicon" snippet above to the index.html of the main page - anywhere in the \<head\> element will work. But leaving for the future and trying to use a real favicon is also fine.

PR #495, opened 2023-06-13

juspence:
@RedHatProductSecurity/corgi-devs Please review another bugfix for these issues. This should fix two issues:

1. In the old code, we would process both .packages (all parent components) and .dependencies (all child components) keys in the Cachito manifests for Golang components only. But the .dependencies results would always overwrite the .packages results, so we're missing all the parent Golang components.
```
                    (
                        source_component["components"],
                        remote_source.packages,
                    ) = self._extract_golang(remote_source.packages, go_stdlib_version)
                    # This assignment overwrites above
                    # So we lose the .packages and keep only the .dependencies 
                    (
                        source_component["components"],
                        remote_source.dependencies,
                    ) = self._extract_golang(remote_source.dependencies, go_stdlib_version)
```
2. We would sometimes overwrite the non-Golang results (pypi, npm, yarn, etc.) with the Golang results.
```
                    # This assignment overwrites anything already in source_component["components"]
                    # If we process the Golang components first, that's OK
                    # But if we already processed pypi / npm / yarn components, they are overwritten
                    (
                        source_component["components"],
                        remote_source.dependencies,
                    ) = self._extract_golang(remote_source.dependencies, go_stdlib_version)```
```
Most of the changes are to the tests, with minimal changes to code. Based on previous discussions, the numbers there are what we expected the correct values to be. Now we process the whole manifest and assert that these values are actually correct / nothing is missing.

Specifically, we assert that all .dependencies children of all .packages are equal to all the top-level .dependencies. So processing only .packages and their children doesn't cause us to miss anything.

There are some TODO notes here about another bug for Golang, but I think that's an existing bug and not affected by my changes here. I'm working on a follow-up PR to fix this (along with adding more tests), and submitting this fix now to keep the changes smaller and more comprehensible.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -247,6 +247,8 @@ class Brew:
                     "arch": rpm_info["arch"],
                     "source": ["koji.listRPMs", "koji.getRPMHeaders"],
                     "rpm_id": rpm_id,
+                    # Our custom "source" key conflicts with the "SOURCE" RPM header
+                    # So we call it "source_files" here instead
                     "source_files": headers["source"],
                 },
             }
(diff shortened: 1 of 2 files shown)


juspence:
Disregard, I missed a test I commented out that still needs work. Sorry for the spam.

PR #497, opened 2023-06-14

juspence:
@RedHatProductSecurity/corgi-devs Quick review please. There will be at least 2 follow-up PRs for this issue, so I'm splitting this change out to make it easier to review. The first commit is general cleanup and the second is the actual bugfix, where we would sometimes overwrite the non-Golang results (pypi, npm, yarn, etc.) with the Golang results.
```
  # This assignment overwrites anything already in source_component["components"]
  # If we process the Golang components first, that's OK
  # But if we already processed pypi / npm / yarn components, they are overwritten
  (
      source_component["components"],
      remote_source.dependencies,
  ) = self._extract_golang(remote_source.dependencies, go_stdlib_version)```
```

Tests still pass after this without changes / never caught this bug to begin with, but I'm going to add new tests in a separate PR. The other bugs I'm trying to fix basically require a complete rewrite of these tests.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -509,6 +509,12 @@ class Brew:
     def _extract_remote_sources(
         cls, go_stdlib_version: str, remote_sources: dict[str, tuple[str, str]]
     ) -> list[dict[str, Any]]:
+        """Given a list of remote-source.json filenames / Cachito manifest names,
+        build and return a list of source component dicts, one for each manifest
+        Each source component has a "components" key with all the top-level
+        .packages in that Cachito manifest, and
+        Each top-level .package has a "components" key with all the child
+        .dependencies of that package (e.g. .packages[index].dependencies in JQ)"""
         source_components: list[dict[str, Any]] = []
         for build_loc, coords in remote_sources.items():
             remote_source = cls._get_remote_source(coords[0])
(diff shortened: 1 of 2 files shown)


corgi/collectors/brew.py
```diff
@@ -539,21 +545,20 @@ def _extract_remote_sources(
...
-                    (
-                        source_component["components"],
-                        remote_source.dependencies,
-                    ) = cls._extract_golang(remote_source.dependencies, go_stdlib_version)
+                    provides, remote_source.packages = cls._extract_golang(
+                        remote_source.packages, go_stdlib_version
+                    )
+                    provides, remote_source.dependencies = cls._extract_golang(
+                        remote_source.dependencies, go_stdlib_version
+                    )
```
juspence, line 554: This looks like a bug, and it is, in both the old and new code. In other words, I'm deliberately not fixing this yet and nothing has changed in this case. For Golang components only, we process the .packages first, then overwrite them with the .dependencies immediately after.

This is the other bug I'm working to fix, but it needs many changes to the test logic and code, so I'll handle this in a follow-up PR.

corgi/collectors/brew.py
```diff
@@ -539,21 +545,20 @@ def _extract_remote_sources(
...
                 else:
                     logger.warning("Found unsupported remote-source pkg_manager %s", pkg_type)
+                    continue
+                try:
+                    source_component["components"].extend(provides)
+                except KeyError:
+                    source_component["components"] = provides
```
juspence, line 561: Now we take the Golang components we discovered (stored into "provides" above) and add them to the existing list of non-Golang components, instead of always overwriting the non-Golang results with the Golang results. This is the easier bug to fix, so I submitted a PR just for this.

juspence:
Going to merge this as-is since I need to open the follow-up PR with more changes.

PR #499, opened 2023-06-15

juspence:
@RedHatProductSecurity/corgi-devs Please review. This is the follow-up PR I mentioned that fixes more bugs:

1. When processing Golang components in Cachito manifests, don't overwrite the .packages results with the .dependencies results. Process only the .packages key (which has both top-level components and child components). Ignore the .dependencies key (which has only child components, and is missing top-level components).

2. When processing top-level Golang components in a Cachito manifest's list of .packages, don't move go-package components underneath a Go module with a matching name. For example, don't make a top-level "golang.org/x/text/cases" go-package a .dependency of a top-level "golang.org/x/text" Go module.

Based on a discussion, the logic in 2. was added only to provide a nicer dependency tree for our users, and isn't needed for a particular feature. Because we overwrote .packages results (a tree) with .dependencies results (a flat list), unflattening the dependencies a bit was useful.

Now that we've fixed the bug in 1., we can just rely on Cachito's relationship tree and not do any extra processing. This simplifies the code quite a bit, and fixes several bugs like:
1. Overwriting existing Go module dependencies from Cachito with the list of go-packages we moved
2. Failing to remove a "vendor/" prefix from names and set a version in certain cases
3. Linking components together (module depends on package) when they aren't really related (X depends on module and Y depends on package, where X and Y are independent of each other).

If we still want to move go-packages underneath their matching / parent modules, I can drop the last commit in this PR. This will need some more work though, since modules can depend on other modules. We would need to recursively move packages underneath modules, instead of only moving top-level packages underneath top-level modules (now that we process a tree instead of a list).

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -247,6 +247,8 @@ class Brew:
                     "arch": rpm_info["arch"],
                     "source": ["koji.listRPMs", "koji.getRPMHeaders"],
                     "rpm_id": rpm_id,
+                    # Our custom "source" key conflicts with the "SOURCE" RPM header
+                    # So we call it "source_files" here instead
                     "source_files": headers["source"],
                 },
             }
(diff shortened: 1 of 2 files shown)


jasinner (approved):
LGTM!
Sorry for the horrible mess I created the first time I attempted this parsing.

juspence:
> LGTM! Sorry for the horrible mess I created the first time I attempted this parsing.

No worries, this is still kind of a mess. I've put my best guess at what I think is right, but there's probably new bugs here I'm not seeing. I'll add a bunch more tests in CORGI-691 so we can make sure we're getting the right results.

Going to merge as soon as the pipeline passes (after fixing token + rerunning).

PR #500, opened 2023-06-16
Graph PoC (not for merging!)

JimFuller-RedHat:
this is only a draft for discussion purposes

### DIFF
--- a/corgi/core/graph.py
+++ b/corgi/core/graph.py
@@ -15,9 +15,9 @@ def create_product_node(product_type, product_name, ofuri, object_id):
         cursor.execute(
             """
             SELECT * from cypher(%s, $$
-            MERGE (V:Product {name:%s, ofuri:%s, object_id:%s})
-            RETURN V
-            $$) as (V agtype);
+                MERGE (node:Product {name:%s, ofuri:%s, object_id:%s})
+                RETURN node
+            $$) as (node agtype);
             """,
             ["products", product_name, ofuri, str(object_id)],
         )
@@ -25,9 +25,9 @@ def create_product_node(product_type, product_name, ofuri, object_id):
         cursor.execute(
             """
             SELECT * from cypher(%s, $$
-            MERGE (V:ProductVersion {name:%s, ofuri:%s, object_id:%s})
-            RETURN V
-            $$) as (V agtype);
+                MERGE (node:ProductVersion {name:%s, ofuri:%s, object_id:%s})
+                RETURN node
+            $$) as (node agtype);
             """,
             ["products", product_name, ofuri, str(object_id)],
         )
@@ -35,9 +35,9 @@ def create_product_node(product_type, product_name, ofuri, object_id):
         cursor.execute(
             """
             SELECT * from cypher(%s, $$
-            MERGE (V:ProductStream {name:%s, ofuri:%s, object_id:%s})
-            RETURN V
-            $$) as (V agtype);
+                MERGE (node:ProductStream {name:%s, ofuri:%s, object_id:%s})
+                RETURN node
+            $$) as (node agtype);
             """,
             ["products", product_name, ofuri, str(object_id)],
         )
@@ -45,9 +45,9 @@ def create_product_node(product_type, product_name, ofuri, object_id):
         cursor.execute(
             """
             SELECT * from cypher(%s, $$
-            MERGE (V:ProductVariant {name:%s, ofuri:%s, object_id:%s})
-            RETURN V
-            $$) as (V agtype);
+                MERGE (node:ProductVariant {name:%s, ofuri:%s, object_id:%s})
+                RETURN node
+            $$) as (node agtype);
             """,
             ["products", product_name, ofuri, str(object_id)],
         )
@@ -55,9 +55,9 @@ def create_product_node(product_type, product_name, ofuri, object_id):
         cursor.execute(
             """
             SELECT * from cypher(%s, $$
-            MERGE (V:Channel {name:%s, ofuri:%s, object_id:%s})
-            RETURN V
-            $$) as (V agtype);
+                MERGE (node:Channel {name:%s, ofuri:%s, object_id:%s})
+                RETURN node
+            $$) as (node agtype);
             """,
             ["products", product_name, ofuri, str(object_id)],
         )
@@ -66,7 +66,6 @@ def create_product_node(product_type, product_name, ofuri, object_id):
 
 def create_product_node_edge(pk_parent, pk_child):
     """create product node edge between a parent and a child"""
-    logger.info(pk_parent)
     conn = connections["graph"]
     conn.ensure_connection()
     cursor = conn.connection.cursor()
@@ -75,9 +74,9 @@ def create_product_node_edge(pk_parent, pk_child):
         """
         SELECT *
         FROM cypher(%s, $$
-        MATCH (a {object_id:%s}), (b {object_id:%s})
-        CREATE (a)-[e:CHILD]->(b)
-        RETURN e
+            MATCH (node1 {object_id:%s}), (node2 {object_id:%s})
+            MERGE (node1)-[e:CHILD]->(node2)
+            RETURN e
         $$) as (e agtype)
         """,
         ["products", str(pk_parent), str(pk_child)],
@@ -85,7 +84,7 @@ def create_product_node_edge(pk_parent, pk_child):
     return None
 
 
-def create_component_node(component_name, purl, object_id, type, namespace, version, nevra):
+def create_component_node(component_name, purl, object_id, type, namespace, version, nevra, arch):
     """create component node"""
     conn = connections["graph"]
     conn.ensure_connection()
@@ -94,11 +93,11 @@ def create_component_node(component_name, purl, object_id, type, namespace, vers
     cursor.execute(
         """
         SELECT * from cypher(%s, $$
-        MERGE (V:Component {name:%s, purl:%s, object_id:%s,type:%s,namespace:%s,version:%s,nevra:%s})
-        RETURN V
-        $$) as (V agtype);
+            MERGE (node:Component {name:%s, purl:%s, object_id:%s, type:%s, namespace:%s, version:%s, nevra:%s, arch:%s}) 
+            RETURN node
+        $$) as (node agtype);
         """,
-        ["components", component_name, purl, str(object_id), type, namespace, version, nevra],
+        ["components", component_name, purl, str(object_id), type, namespace, version, nevra, arch],
     )
     return None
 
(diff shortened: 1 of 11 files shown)


docker-compose.yml
```diff
@@ -2,6 +2,37 @@ version: '3'
 
 services:
 
+  graph-console:
+    container_name: graph-console
+    hostname: graph-console
+    image: "localhost/age-viewer"
```
juspence, line 8: If I want to run this PoC locally, how do I get this image? Can you change this to use the fully-qualified image name instead, like docker.io/_/age-viewer? Or if there's no pre-built image, can you add build instructions to the docs?

corgi/core/graph.py
```diff
@@ -0,0 +1,272 @@
...
+    """return flat list of component ids
+    it is likely that we would want to return Component node data as
+    well as a tree version
+    """
+    conn = connections["graph"]
+    conn.ensure_connection()
+    cursor = conn.connection.cursor()
+    if flat:
+        cursor.execute('SET search_path = ag_catalog, "$user", public')
+        # TODO: Apache Age does not support edge OR condition
```
juspence, line 190: I have a few different concerns about this code. Some of it I think could be cleaned up and optimized further, but that's definitely not needed for PoC work. Just a heads-up / note in case we merge something similar to this after discussions.

The biggest issue is that it's hard for me to tell what the SQL here is doing, so reviewing this is tricky and I have to guess. Can you please add a brief comment here explaining in words what we're selecting? For example:
```
# MATCH all nodes in the "components" graph
# that have the purl provided by the user
# and are linked to a second node
# with either a PROVIDES relationship
# or (UNION) a PROVIDES_DEV relationship
# RETURN the second node's ID
```

Can you also explain what "does not support edge OR condition" means? We use UNION in the query below, and I think this works like an OR. But are we basically saying "SELECT with multiple conditions isn't supported, and needs workarounds"? That feels like a red flag in terms of relying on this in production...am I misunderstanding something?

corgi/core/graph.py
```diff
@@ -0,0 +1,272 @@
...
+            SELECT *
+            FROM cypher(%s, $$
+                MATCH (node1:Component {purl:%s})-[r:PROVIDES*]->(node2:Component)
+                RETURN node2.object_id
+                UNION
+                MATCH (node1:Component {purl:%s})-[r:PROVIDES_DEV*]->(node2:Component)
+                RETURN node2.object_id
+            $$) as (node2 agtype)
+            """,
+            ["components", purl, purl],
```
juspence: I think that we're protected from SQL injection based on this note in the docs:
https://docs.djangoproject.com/en/4.2/topics/db/sql/#executing-custom-sql-directly
To protect against SQL injection, you must not include quotes around the %s placeholders in the SQL string.

But having to wonder / worry about this doesn't feel like good design. I'm much happier writing queries using standard Django constructs and letting the ORM worry about escaping the query parameters.

Is there some kind of django-graphdb or django-age extension we can use, instead of having to write these queries manually?

corgi/core/graph.py
```diff
@@ -0,0 +1,272 @@
...
+    if flat:
+        cursor.execute('SET search_path = ag_catalog, "$user", public')
+        out = []
+        cursor.execute(
+            """
+            SELECT * from cypher(%s, $$
+                MATCH (node1:Component {purl:%s})<-[R:PROVIDES*]-(node2:Component {type:"RPM", arch:"src"}) 
+                RETURN node2.object_id
+                UNION
+                MATCH (node1:Component {purl:%s})<-[R:PROVIDES*]-(node2:Component {type:"OCI", arch:"noarch"})
```
juspence: It looks like we have two queries here joined with a UNION / OR. One shows the source RPMs linked with PROVIDES relationships, and one shows the index / noarch containers linked with PROVIDES relationships. Is that right?

Are we missing the logic for root components (whether source RPMs or noarch containers) linked with PROVIDES_DEV relationships? Does that require two more queries with two more UNION operators?

Can we correctly find the root components, even when they're not direct parents of the purl / component we're given? In other words, does this lookup work recursively to handle transitive dependencies? What about if there are multiple relationship types going back to the parent, like below:

root_component PROVIDES another_component PROVIDES_DEV child_component

If we're given child_component's purl, will we be able to find the root_component?

corgi/core/models.py
```diff
@@ -1501,15 +1538,41 @@ def _build_pypi_download_url(purl: str) -> str:
...
-        self.sources.set(self.get_sources_nodes(using="default"))
+        self.upstreams.set(
+            set([str(pk) for pk in self.get_upstreams_queryset().values_list("pk", flat=True)])
+        )
+        self.provides.set(
+            set([str(pk) for pk in self.get_provides_queryset().values_list("pk", flat=True)])
+        )
+        self.sources.set(
+            set([str(pk) for pk in self.get_sources_queryset().values_list("pk", flat=True)])
+        )
```
juspence: Instead of adding new methods and deleting the old ones, can we fix the existing methods? That helps avoid accidentally returning the wrong result / making a change we didn't intend in the get_x_queryset method, and makes the changes easier to review.

I also think there's a bug here because we're loading a full queryset into memory (when iterating over it). Before, we were using .iterator() to avoid this. Maybe we can get rid of .iterator() and the set(), and just use the queryset results directly without needing to fully load it into memory to dedupe it?

I think that's possible since we're making one DB call now, so there should be no duplicates. Before, we were iterating over cnodes and finding the sources of each, which involved multiple DB calls and could have duplicates if two cnodes shared any of their sources.

For example, imagine a component included in both an x86 container and an aarch64 container. There would be two child nodes (one for each container), but each node would ultimately have the same noarch container in their list of sources.

corgi/core/models.py
```diff
@@ -1501,15 +1538,41 @@ def _build_pypi_download_url(purl: str) -> str:
...
+
+    def get_root(self, using: str = "read_only"):
+        """TODO: should only ever return one "root" queryset using the read-only DB"""
+        pks = retrieve_component_root(self.purl)
+        return Component.objects.filter(pk__in=pks).using(using)
```
juspence, line 1555: What happens when a component participates in more than one tree / is included in multiple containers / etc.? Wouldn't we return multiple roots in that case?

corgi/core/models.py
```diff
@@ -1501,15 +1538,41 @@ def _build_pypi_download_url(purl: str) -> str:
...
+
+    def get_upstreams_queryset(self, using: str = "read_only"):
+        """Return the "upstreams" queryset using the read-only DB"""
+        pks = retrieve_component_upstream_relationships(self.purl)
+        return Component.objects.filter(pk__in=pks).using(using)
+
+    def get_sources_queryset(self, using: str = "read_only"):
+        """Return the "sources" queryset using the read-only DB"""
+        pks = retrieve_component_source_relationships(self.purl)
+        return Component.objects.filter(pk__in=pks).using(using)
```
juspence: If a component is very popular, the list of components we return as "sources" can be massive. I've also read that graph DB performance scales based on the percentage of the graph you're querying, so I'm worried about if we can handle these "supernodes" / highly-linked components.

You can find a good example of this in tests/test_performance.py::test_displaying_component_with_many_sources. The component we test there, "pkg:rpm/redhat/systemd-libs@250-12.el9_1?arch=aarch64" should have over 2,000 sources. There's also a Brotli component, "pkg:rpm/redhat/brotli@1.0.6-3.el8?arch=x86_64", that has 263701 sources in stage.

Could you run that performance test locally for both of the above components, with a fully-loaded graph DB / a complete set of data? Note that by default performance tests run against stage. To run them locally, you'll need to set the CORGI_DOMAIN variable in your venv activation script and .env file.

I'd really want to see what the hard numbers are for a fully-loaded graph DB with lots of data, before I'd feel comfortable switching. I know you said in your demo video that the graph DB would be an improvement. But I want to make sure it will handle edge cases like this one, and I'm curious to see what kind of speedup (10%, 15%, etc.) we'll get, if any.

juspence (commented):
Still very uncertain about this approach. My biggest concerns are:

1. Will it handle edge cases, like highly-connected components with many sources?
2. What kind of speedup are we looking at? What are the hard numbers in a graph DB vs. our current approach?
3. Is there any kind of Django plugin or third-party Cypher library, that lets us use a graph DB without needing to write raw SQL queries?

corgi/core/graph.py
JimFuller-RedHat: this PoC is not production code ... I would not want anyone to review graph.py like it would be inserted in our current code base ... I choose a list of functions as it was the quickest, simplest way to demonstrate cypher without all the django cruft ... going to ignore these kind of comments (and hoping you will understand).

The SQL is only handling invoking cypher function and return from that function ... the stuff inside is cypher query language which will take some time for everyone to learn.

As for the comment about 'does not support edge OR condition' ... this is mostly syntatic... Apache AGE in this instance does not allow OR conditions directly in the edge match that is we cannot directly state in -[edge:PROVIDES|PROVIDES_DEV]- edge to match either PROVIDES or PROVIDES_DEV but we can build up the same behaviour using UNION. In terms of execution both approaches boil down to the same internal query plan (hence why I say it is syntatic).

corgi/core/graph.py
JimFuller-RedHat: re SQL injection refer to my quote about production code ... also this PoC is not going to build out django abstraction (which I mentioned as limitation in ADR).

there is no django-apache-age ... and like with django-ltree we could consider doing it but it may turn out to be additional work, etc etc.

corgi/core/graph.py
JimFuller-RedHat: the logic for matching node2 is for matching root ... we return node2.object_id ... we are providing child component purl to match

corgi/core/models.py
JimFuller-RedHat: added new funcs to the PoC to be able to demonstrate diff between old and new easily with same instance

+1 to get rid of iterator and your commen ton duplicates .. I will attempt to scale up to full dataset on my local machine and prob will need to do this refinement

corgi/core/models.py
JimFuller-RedHat: yes ... I should remove this function  ... retrieve_component_root does that ... the TODO was a question for me ;)

corgi/core/models.py
JimFuller-RedHat: good point - in our case we do have a narrowing down of the graph eg. root components  ... thank you for the super node example - I will try to work up an example and yes have thought of this. 

FWIW we do not have to set any indexes for graph db but we still *CAN* set indexes if we want to for special situations (through the magic of postgres) and this maybe just a special scenario where we take advantage of that.

docker-compose.yml
JimFuller-RedHat: pull this repo https://github.com/apache/age-viewer and build dockerfile

JimFuller-RedHat:
> Still very uncertain about this approach. My biggest concerns are:
> 
> 1. Will it handle edge cases, like highly-connected components with many sources?

guessing not out of the box ... but we have a few techniques to apply for super nodes.

> 2. What kind of speedup are we looking at? What are the hard numbers in a graph DB vs. our current approach?

useful 'numbers' are for MP+ hence my ADR suggestion that we work up to a full demo and deploy to MP+ ... in the short term can demonstrate like for like using a local instance ...

* getting as good as using joins (1-2X better perf)
* all graph operations are performing well, .. on average seeing at least a magnitude better performance over django-mptt ... in some cases much better (eg. django-mptt is dog slow with inserts on an existing tree - a limitation pointed out in original ADR for component registry - in graph database these operations are quick) 
 
> 3. Is there any kind of Django plugin or third-party Cypher library, that lets us use a graph DB without needing to write raw SQL queries?

could be ... I thought it would be simpler to demonstrate using raw

corgi/core/graph.py
JimFuller-RedHat: will inject a lot more code comments ;)

PR #501, opened 2023-06-16
CORGI-647 skip released check in community mode when generating manifests

jasinner:
This allow generating of manifests in community mode.


juspence (approved):
LGTM, thanks for fixing this!

PR #503, opened 2023-06-18
Add PNC SBOM UMB listener and fetch task

kgrant-rh:
Add a UMB listener for SBOMer's "sbom available" messages and a task to fetch and parse them. The actual parsing will be implemented in CORGI-488.

### DIFF
--- a/corgi/tasks/pnc.py
+++ b/corgi/tasks/pnc.py
@@ -1,0 +1,40 @@
+import logging
+
+import requests
+from celery_singleton import Singleton
+from requests.exceptions import HTTPError
+
+from config.celery import app
+from corgi.collectors.pnc import parse_pnc_sbom
+from corgi.core.models import ProductVariant
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(base=Singleton)
+def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
+    logger.info("Fetching PNC SBOM %s for purl %s", sbom_data["id"], purl)
+    """Fetch a PNC SBOM from sbomer"""
+
+    # Validate the supplied product information
+    try:
+        ProductVariant.objects.get(name=product_data["productVariant"])
+    except ProductVariant.DoesNotExist:
+        logger.warning(
+            "PNC SBOM provided for nonexistant product variant: %s", product_data["productVariant"]
+        )
+        return
+
+    # Fetch and parse the SBOM
+    try:
+        r = requests.get(sbom_data["link"])
+        r.raise_for_status()
+    except HTTPError:
+        logger.warning("Failed to fetch PNC SBOM: %s", sbom_data["link"])
+        return
+
+    # TODO: Fetch any necessary PNC build info
+
+    logger.info(f"SBOM fetched for purl {purl}")
+
+    parse_pnc_sbom(r.json())
--- a/corgi/tasks/pnc.py
+++ b/corgi/tasks/pnc.py
@@ -2,39 +2,32 @@ import logging
 
 import requests
 from celery_singleton import Singleton
-from requests.exceptions import HTTPError
 
 from config.celery import app
+from corgi.collectors.models import CollectorErrataProductVariant
 from corgi.collectors.pnc import parse_pnc_sbom
-from corgi.core.models import ProductVariant
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
 logger = logging.getLogger(__name__)
 
 
-@app.task(base=Singleton)
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
     logger.info("Fetching PNC SBOM %s for purl %s", sbom_data["id"], purl)
     """Fetch a PNC SBOM from sbomer"""
 
     # Validate the supplied product information
     try:
-        ProductVariant.objects.get(name=product_data["productVariant"])
-    except ProductVariant.DoesNotExist:
+        CollectorErrataProductVariant.objects.get(name=product_data["productVariant"])
+    except CollectorErrataProductVariant.DoesNotExist:
         logger.warning(
             "PNC SBOM provided for nonexistant product variant: %s", product_data["productVariant"]
         )
         return
 
     # Fetch and parse the SBOM
-    try:
-        r = requests.get(sbom_data["link"])
-        r.raise_for_status()
-    except HTTPError:
-        logger.warning("Failed to fetch PNC SBOM: %s", sbom_data["link"])
-        return
-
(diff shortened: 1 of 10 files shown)


kgrant-rh:
The branch name is incorrect, this is actually related to CORGI-489.

juspence:
Tagging @juspence so I have an email reminder to review this (in my Github folder / dedicated PR queue, which is less busy than my inbox).

tests/data/pnc/sbom_complete.json
```diff
@@ -0,0 +1,59 @@
...
+  "topic": "/topic/VirtualTopic.eng.pnc.sbom.spike.complete",
+  "timestamp": 1686698102.0,
+  "certificate": null,
+  "signature": null,
+  "username": null,
+  "crypto": null,
+  "msg": {
+    "build": {
+      "id": "AZLCRNRPSHQAA",
+      "link": "https://orch.psi.redhat.com/pnc-rest/v2/builds/AZIE75KE6KIAA",
```
juspence: Let's remove internal URLs from our code / test data on Github. You can just replace the hostname with example.com.

Not a huge deal and we've got dozens of URLs like this in our commit history already. We have a separate task to clean all those up, so for now you can just amend this commit and force-push to this branch before merging.

tests/data/pnc/sbom_complete.json
```diff
@@ -0,0 +1,59 @@
...
+        "productName": "RHBQ",
+        "productVariant": "8Base-RHBQ-2.13",
+        "productVersion": "RHEL-8-RHBQ-2.13"
+      }
+    },
+    "purl": "pkg:maven/com.redhat.quarkus.platform/quarkus-bom@2.13.8.Final-redhat-00003?type=pom",
+    "sbom": {
+      "bom": {
+        "format": "cyclonedx",
+        "link": "https://sbomer.apps.ocp-c1.prod.psi.redhat.com/api/v1alpha1/sboms/456597260580114432/bom",
```
juspence: URL should be fixed, assuming it doesn't resolve when off-VPN. If SBOMer is publically available then it's fine to leave this.

tests/data/pnc/sbom_complete.json
```diff
@@ -0,0 +1,59 @@
...
+        "version": "1.4"
+      },
+      "id": "456597260580114432",
+      "link": "https://sbomer.apps.ocp-c1.prod.psi.redhat.com/api/v1alpha1/sboms/456597260580114432"
```
juspence: URL.

tests/data/pnc/pnc_sbom.json
juspence, line 1: Do we really need the whole SBOM for the tests? It's huge and has lots of internal URLs in it. Can we remove 99% of the contents, and just have e.g. 3 components with a few dependencies each and sanitized / removed internal URLs?

tests/test_monitor.py
```diff
@@ -70,8 +71,9 @@ def test_brew_umb_handler_setup():
 
     # The addresses and selectors of the dispatcher should match those of a handler
     handler = BrewUMBHandler()
-    assert dispatcher.virtual_topic_addresses.keys() == handler.virtual_topic_addresses.keys()
-    assert dispatcher.selectors == handler.selectors
+    assert set(dispatcher.virtual_topic_addresses).issuperset(
+        set(handler.virtual_topic_addresses))
+    assert set(dispatcher.selectors).issuperset(set(handler.selectors))
```
juspence: This makes sense, but should we add a similar test_sbomer_umb_handler_setup test here? In other words, do we need to test the PNC topic addresses and selectors we define for the new SBOMer class you added?

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,40 @@
...
+from requests.exceptions import HTTPError
+
+from config.celery import app
+from corgi.collectors.pnc import parse_pnc_sbom
+from corgi.core.models import ProductVariant
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(base=Singleton)
```
juspence: There are some standard kwargs we add to every Celery task, which control retries for known exceptions and some other fiddly bits. Let's add those here as well - you should be able to just copy the decorator from any other task.

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,40 @@
...
+    except ProductVariant.DoesNotExist:
+        logger.warning(
+            "PNC SBOM provided for nonexistant product variant: %s", product_data["productVariant"]
+        )
+        return
+
+    # Fetch and parse the SBOM
+    try:
+        r = requests.get(sbom_data["link"])
+        r.raise_for_status()
```
juspence: Let's avoid one-letter variable names:
```suggestion
        response = requests.get(sbom_data["link"])
        response.raise_for_status()
```

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,40 @@
...
+    except HTTPError:
+        logger.warning("Failed to fetch PNC SBOM: %s", sbom_data["link"])
+        return
```
juspence: I think we should let the task fail with an exception here, so we get an email about this error and can investigate. Or is this placeholder code because the link in the UMB message doesn't point to a real / working SBOM in PNC?

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,40 @@
...
+    """Fetch a PNC SBOM from sbomer"""
+
+    # Validate the supplied product information
+    try:
+        ProductVariant.objects.get(name=product_data["productVariant"])
+    except ProductVariant.DoesNotExist:
+        logger.warning(
+            "PNC SBOM provided for nonexistant product variant: %s", product_data["productVariant"]
+        )
+        return
```
juspence: I get that we don't want to be spammed with errors when SBOMer sends messages about variants / products we haven't loaded yet. But hiding the error here means we won't notice if every single message refers to a nonexistent variant.

I guess it's fine (or at least I can't think of a better solution), but let's keep an eye on things after we merge this code. That way we'll know if we process any messages correctly, or if e.g. the variant names in the message are always different than the variant names in ET / Corgi.

corgi/tasks/pnc.py
```diff
@@ -0,0 +1,40 @@
...
+
+from config.celery import app
+from corgi.collectors.pnc import parse_pnc_sbom
+from corgi.core.models import ProductVariant
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(base=Singleton)
+def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
```
juspence, line 15: Let's add a test for this task.

juspence (approved):

corgi/tasks/pnc.py
jasinner: We can probably ingest the Components without linking to a ProductVariant. Maybe a good idea here to be to make sure any product_data["productVariant"] existing in ET by checking CollectorErrataProductVariant exists by that name?

jasinner (approved):
LGTM. Just need to fix lint error from CI.

PR #504, opened 2023-06-19
CORGI-725 relations softwarebuild fk

jasinner:
Added SoftwareBuild ForeignKey to ProductComponentRelations table. We have to allow NULL on this field, and also keep the old build_id, and build_type fields because we often create the Relation before we create the SoftwareBuild. However as long as we update the new SoftwareBuild field each time we create a new SoftwareBuild object we can utilize it to query Relations by build.

On set CASCADE to SET_NULL so that deleting a SoftwareBuild object does not also delete the associated Relations. This will allow re-processing that relation to create the SoftwareBuild again. If you want to completely delete a SoftwareBuild, you'll need to delete the Relation as well.


jasinner:

corgi/core/migrations/0071_productcomponentrelation_sb_fk_data.py
```diff
@@ -0,0 +1,27 @@
...
+    for relation in ProductComponentRelation.objects.all():
+        try:
+            sb_pk = SoftwareBuild.objects.get(
+                build_id=relation.build_id, build_type=relation.build_type
+            )
+        except SoftwareBuild.DoesNotExist:
+            sb_pk = None
+        if sb_pk:
+            relation.software_build = sb_pk
+            relation.save()
```
juspence: Let's add a filter so that if the migration takes too long and has to be rerun manually, we'll continue where we left off instead of reprocessing all the relations again. We should also use .iterator() in migrations to avoid exhausting the init pod's memory:
```suggestion
    for relation in ProductComponentRelation.objects.filter(software_build__isnull=True).iterator():
        with transaction.atomic():
            sb_pk = SoftwareBuild.objects.filter(
                build_id=relation.build_id, build_type=relation.build_type
            ).values_list("pk", flat=True).first()
            if sb_pk:
                relation.software_build = sb_pk
                relation.save()
```

Trying to do everything in a single query instead of making one per relation should be faster, but I'm not sure if below will work. This avoids the risk of a long-running migration that will probably time out:
```suggestion
    relations_to_update = ProductComponentRelation.objects.filter(software_build__isnull=True)
    relation_build_ids = relations_to_update.values_list("build_id", flat=True)
    existing_build_ids = dict(SoftwareBuild.objects.filter(build_id__in=relation_build_ids).values_list("build_id", "pk"))
    for relation in relations_to_update.iterator():
        with transaction.atomic():
            # Get the UUID / PK of the build using the Brew build_id as mapping key
            existing_build_id = existing_build_ids.get(relation.build_id)
            if existing_build_id:
                relation.software_build = existing_build_id
                relation.save()
```

juspence (commented):
First of all, I'm really happy to see this change. I have one suggestion about the migration I wanted to mention ASAP, but feel free to ignore if it complicates things or you don't like the way it reads. Even if the current migration fails to run automatically, we can always run it manually if needed, it's just a little annoying to remember to do this during a prod release.

I'm submitting that comment only for now, since I won't have time to actually look at the rest of the code until later tomorrow. I have a bunch of other PRs in my queue to look at and I need to finish tests for CORGI-691. Sorry for the delay in getting to this one.

corgi/core/models.py
```diff
@@ -384,36 +382,8 @@ def components(self) -> "ComponentQuerySet":
...
-            return (
-                ProductComponentRelation.objects.filter(product_ref__in=product_refs)
-                .values_list("build_id", "build_type")
-                .distinct()
-                .using("read_only")
-            )
-        # Else no product variants or product streams - should never happen
-        return ProductComponentRelation.objects.none()
+        """Returns distinct software builds associated with this Product instance via components"""
+        return self.components.values_list("software_build", flat=True).distinct()
```
juspence: I wrote that TODO comment, but I think maybe Jim (or someone else?) said it wouldn't necessarily give the right results in all cases. I'm fine with the change if you've tested and it works, just wanted to give a heads-up.

I also can't remember if we define a default ordering on the Component model or not. If we do, clearing that ordering here will avoid some unnecessary sorting that can slow things down:
```suggestion
        return self.components.values_list("software_build", flat=True).order_by().distinct()
```

corgi/tasks/brew.py
```diff
@@ -576,35 +584,25 @@ def fetch_unprocessed_relations(
...
+            # It was done to avoid updating the collector models not to use build_id as
+            # a primary key. It's possible because the only product stream (openstack-rdo)
+            # stored in CENTOS koji doesn't use modules
+            slow_fetch_brew_build.delay(
+                relation.build_id, SoftwareBuild.Type.CENTOS, force_process=force_process
+            )
             processed_builds += 1
+            continue
+        slow_fetch_modular_build.delay(relation.build_id, force_process=force_process)
+        processed_builds += 1
```
juspence: Minor: I think if / else is a little clearer than if / continue, since it avoids duplicating the processed_builds counter:
```suggestion
        else:
            slow_fetch_modular_build.delay(relation.build_id, force_process=force_process)
        processed_builds += 1
```

corgi/tasks/brew.py
```diff
@@ -576,35 +584,25 @@ def fetch_unprocessed_relations(
         query &= Q(created_at__gte=created_since)
     relations_query = (
         ProductComponentRelation.objects.filter(query)
-        .values_list("build_id", "build_type")
-        .distinct()
+        .filter(software_build=None)
         .using("read_only")
```
juspence: Not related to this change, and I guess it's been working fine, but I'd be worried about using the read-only DB for tasks which need to process data / write results back. If there were a large amount of replication lag, we could get stale data from the read-only copy and write it back to the default DB, which could cause weird heisenbugs that are very hard to troubleshoot.

In this case, I think we'd fail to process some of the unprocessed relations, whenever they've been created in the default DB but haven't yet been replicated to the read-only copy.
```suggestion
```

corgi/tasks/management/commands/loadunprocessedrelations.py
```diff
@@ -14,15 +14,8 @@ def add_arguments(self, parser: CommandParser) -> None:
...
-            "--save_only",
-            action="store_true",
-            help="Call fetch anyway to save the product taxonomy if build already exists",
-        )
 
     def handle(self, *args, **options):
-        processed_builds = 0
         if options["product_ref"]:
             if (
                 not ProductComponentRelation.objects.db_manager("read_only")
```
juspence: Not related to this change, and I guess it's been working fine, but I'd be worried about using the read-only DB for tasks which need to process data / write results back. If there were a large amount of replication lag, we could get stale data from the read-only copy and write it back to the default DB, which could cause weird heisenbugs that are very hard to troubleshoot.

In this case, I think we'd fail to process some of the unprocessed relations, whenever they've been created in the default DB but haven't yet been replicated to the read-only copy.
```suggestion
                not ProductComponentRelation.objects
```

corgi/tasks/rhel_compose.py
```diff
@@ -63,16 +63,15 @@ def get_builds(
         relations_query = relations_query.filter(product_ref=stream_name)
 
     processed_builds = 0
-    for build_id, build_type in (
-        relations_query.values_list("build_id", "build_type").distinct().iterator()
+    for build_id in (
+        relations_query.filter(software_build=None)
+        .exclude(build_type=SoftwareBuild.Type.CENTOS)
```
juspence: This is good, but let's add a comment explaining why:
```suggestion
        # We don't support modular builds in composes for CENTOS
        .exclude(build_type=SoftwareBuild.Type.CENTOS)
```
We should also move up the .filter() and .exclude() logic to where relations_query is originally defined at the start of this task on line 57, just to keep things cleaner.

tests/test_manifests.py
```diff
@@ -436,14 +436,12 @@ def setup_products_and_components_upstreams():
     # The product_ref here is a variant name but below we use it's parent stream
     # to generate the manifest
     ProductComponentRelationFactory(
-        build_id=str(build.build_id),
-        build_type=build.build_type,
+        software_build=build,
```
juspence, line 439: Do we still need to set build_id and build_type manually using build.build_id and build.build_type? Otherwise the relation factory will probably generate them using a random value that's not related to the build, right? Or does it not matter in this case?

tests/test_model.py
```diff
@@ -284,8 +284,7 @@ def test_product_component_relations():
     ProductComponentRelation.objects.create(
         type=ProductComponentRelation.Type.COMPOSE,
         product_ref=rhel_7_1.name,
-        build_id=build_id,
-        build_type=sb.build_type,
+        software_build=sb,
```
juspence, line 287: As above, we might need to set build_id and build_type on the relation here. I guess maybe it's not needed if tests still pass so we aren't asserting on the value, but I can see this causing problems in the future.

tests/test_model.py
```diff
@@ -321,49 +321,58 @@ def test_product_stream_builds():
     rhel_8_2_base_build = SoftwareBuildFactory()
     rhel_7_1_build = SoftwareBuildFactory()
     rhel, rhel_7, rhel_7_1, _, rhel_8_1, rhel_8_2, rhel_8_2_base = create_product_hierarchy()
+    # ProductModel.builds now uses components to lookup builds, so we need to create Components
+    # and call save_product_taxonomy on the builds to ensure we have the connection from streams
+    # to builds via components
```
juspence, line 328: Aha, this was what Jim (or somebody) mentioned. Using ProductModel.components to find ProductModel.builds means that we have to have already processed the relations and saved their taxonomies. The old ProductModel.builds showed build IDs from relations even when they hadn't been processed yet.

Again, I'm fine with the change but I'm also not an expert on what the implications are. I guess let's just be careful we understand how .builds is used. Or if there are no usages and .builds is only needed to show stuff in the web API, then this is completely fine.

tests/test_model.py
```diff
@@ -321,49 +321,58 @@ def test_product_stream_builds():
...
+    rhel_8_2_build.save_product_taxonomy()
     ProductComponentRelation.objects.create(
         type=ProductComponentRelation.Type.CDN_REPO,
         # This is a product variant ref, and also a child of rhel_8_2 stream
         product_ref=rhel_8_2_base.name,
-        build_id=rhel_8_2_base_build.build_id,
+        software_build=rhel_8_2_base_build,
     )
+    c82base = ComponentFactory(software_build=rhel_8_2_base_build)
+    ComponentNode.objects.create(obj=c82base, purl=c82base.purl, parent=None)
```
juspence:
```suggestion
    rhel_8_2_base_component = ComponentFactory(software_build=rhel_8_2_base_build)
    ComponentNode.objects.create(obj=rhel_8_2_base_component, purl=rhel_8_2_base_component.purl, parent=None)
```

tests/test_model.py
```diff
@@ -321,49 +321,58 @@ def test_product_stream_builds():
...
+    rhel_8_2_base_build.save_product_taxonomy()
     ProductComponentRelation.objects.create(
         type=ProductComponentRelation.Type.BREW_TAG,
         # This is a product stream ref only
         product_ref=rhel_7_1.name,
-        build_id=rhel_7_1_build.build_id,
+        software_build=rhel_7_1_build,
     )
+    c71 = ComponentFactory(software_build=rhel_7_1_build)
+    ComponentNode.objects.create(obj=c71, purl=c71.purl, parent=None)
```
juspence:
```suggestion
    rhel_7_1_component = ComponentFactory(software_build=rhel_7_1_build)
    ComponentNode.objects.create(obj=rhel_7_1_component, purl=rhel_7_1_component.purl, parent=None)
```

tests/test_model.py
```diff
@@ -321,49 +321,58 @@ def test_product_stream_builds():
...
+    # to builds via components
     ProductComponentRelation.objects.create(
         type=ProductComponentRelation.Type.COMPOSE,
         # This is a product stream ref
         product_ref=rhel_8_2.name,
-        build_id=rhel_8_2_build.build_id,
+        software_build=rhel_8_2_build,
     )
+    c82 = ComponentFactory(software_build=rhel_8_2_build)
+    ComponentNode.objects.create(obj=c82, purl=c82.purl, parent=None)
```
juspence: Let's please not use cryptic variable names like this. rhel_8_2_component is much easier to understand.
```suggestion
    rhel_8_2_component = ComponentFactory(software_build=rhel_8_2_build)
    ComponentNode.objects.create(obj=rhel_8_2_component, purl=rhel_8_2_component.purl, parent=None)
```

corgi/tasks/brew.py
```diff
@@ -96,6 +95,16 @@ def slow_fetch_brew_build(
...
+    if created:
+        try:
+            relation = ProductComponentRelation.objects.get(
+                build_id=build_id, build_type=build_type
+            )
+        except ProductComponentRelation.DoesNotExist:
+            pass
+        else:
+            relation.software_build = softwarebuild
+            relation.save()
```
juspence: You can just move these inside the try block, there's no need for a separate else block here. And actually we should do that, then use a transaction or .update() to avoid race conditions where we overwrite some data:
```suggestion
        relation = ProductComponentRelation.objects.filter(
            build_id=build_id, build_type=build_type, software_build__isnull=True
        ).update(software_build=softwarebuild)
```
This might be a silly question, but isn't it possible we'll have more than one relation for the same build_id? We could relate the build to a stream with a Brew tag OR to a variant with an erratum, and hopefully never both if our data and process is sane.

But we could also have compose relations, Pulp or Yum CDN_REPO relations, and other types I'm forgetting, right? My updated version should handle this fine either way, but I want to confirm I'm understanding relations correctly.

corgi/tasks/brew.py
```diff
@@ -515,29 +532,21 @@ def load_brew_tags(build_id: str, brew_tags: list[str]) -> int:
...
+            # We do this inline instead of via create_relations function because
+            # it's the only time we call it where we have a software_build object created
+            _, created = ProductComponentRelation.objects.get_or_create(
+                external_system_id=tag,
+                product_ref=stream_name,
+                build_id=software_build.build_id,
+                build_type=build_type,
+                software_build=software_build,
+                defaults={"type": ProductComponentRelation.Type.BREW_TAG},
+            )
```
juspence: get_or_create arguments should only use fields that appear in a unique_together constraint, in order to guarantee it behaves atomically. In this case, I think that's external_system_id, product_ref, and build_id. Maybe build_type was added to the constraint as well.

If build_type isn't part of the unique_together constraint, it should be added, or we should move all the build_type values into the defaults dict.

software_build should also be part of the defaults dict, and we should probably switch this to use update_or_create instead.

Sorry for completely forgetting about this until just now. Can you please check the rest of your changes to make sure all other get_or_create / update_or_create usages are atomic?

```suggestion
            _, created = ProductComponentRelation.objects.update_or_create(
                external_system_id=tag,
                product_ref=stream_name,
                build_id=software_build.build_id,
                defaults={
                    "type": ProductComponentRelation.Type.BREW_TAG
                    "build_type": build_type,
                    "software_build": software_build,
                },
            )
```

juspence (approved):
I see some possible issues that need to be addressed, but once those are fixed LGTM.

corgi/core/migrations/0071_productcomponentrelation_sb_fk_data.py
jasinner: Good idea to add the ability to restart and using an iterator. I don't think the 2nd option will work well in community instance as we can have build_ids from KOJI/CENTOS which overlap, so we need to lookup the tuple of build_id/build_type.

corgi/tasks/brew.py
jasinner: Yes, there are often multiple relations for the same build. You question made me review the SoftwareBuild.get_product_details function used by SoftwareBuild.save_product_taxonomy and I think there is a bug there where if a build is associated with a Stream it will also be incorrectly be associated with all variants and channels in that stream family. We should stop doing that.
It's fine to assume that build to variant mapping also includes all parents in the hierarchy it's not correct to say that just because a build is associated with a stream, that it also can be associated with all the variant children of that stream. We should instead be relying on ERRATA_TOOL relations to determine if a build is associated with a variant or not. I think doing this will enable us to build SBOM per variant. SBOM per variant should allow EL8/9 versions of SBOM and probably other variant related features as well.

I filed CORGI-728 to address this.

corgi/core/models.py
jasinner: Querying for builds this way is really slow, now that we have a foreign key between SoftwareBuild and ProductComponentRelation I think it might actually be better to revert this to the old way, but just add a filter excluding relations where software_build is None.

This builds function is only used to get a build count in the API and tests today. However it's often a useful function to use in scripts, and probably more useful if it returned SoftwareBuild objects rather than Relations.

tests/test_model.py
jasinner: I removed these Component references.

tests/test_model.py
jasinner: I removed these Component references.

tests/test_model.py
jasinner: Since we only use the .builds function to get a count on the number of builds for a ProductModel I think the change here to exclude unprocessed builds from the count will have no negative impacts.

tests/test_model.py
jasinner: Ok, I added them back.

tests/test_manifests.py
jasinner: I'll add them back just in case.

corgi/tasks/rhel_compose.py
jasinner: Ok, good idea.

corgi/tasks/management/commands/loadunprocessedrelations.py
jasinner: It's possible I suppose, removing read_only here.

corgi/tasks/brew.py
jasinner: good idea.

corgi/tasks/brew.py
jasinner: ok

corgi/tasks/brew.py
jasinner: Build_type is part of the unique_together constraint. I moved software_build to defaults dict thought, thanks for noticing that.

This is the only place I added a *_create.

PR #505, opened 2023-06-20
CORGI-691: Add support for git-submodule component types in Cachito

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have time. We probably need more tests to handle other recent issues, but the tests + feature being added here should be relatively complete and standalone. So I'll open a follow-up PR for any final test enhancements I identify, and for now this should be ready for review.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -556,20 +556,31 @@ class Brew:
                 # The top-level .dependencies are duplicates
                 # of each top-level .package's child .dependencies
                 # We don't need to process them again
-                if pkg_type in ("npm", "pip", "yarn", "rubygems"):
+                if pkg_type in (
+                    *cls.CARGO_TYPE_MAPPING,
+                    *cls.GEM_TYPE_MAPPING,
+                    *cls.NPM_TYPE_MAPPING,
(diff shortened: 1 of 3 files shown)


corgi/collectors/brew.py
```diff
@@ -746,6 +772,90 @@ def _build_golang_component_dict_from_objs(
...
+                if remaining_dependencies:
+                    # We filtered out child .dependencies of this top-level .package
+                    # because they didn't have git-submodule type
+                    # Since they're not top-level .packages themselves,
+                    # they won't get processed anywhere else in our code
+                    # So in theory we'd be missing components
+                    # This shouldn't happen, but raise an error just in case
+                    raise ValueError(
+                        f"Top-level package {typed_pkg.version} had {len(remaining_dependencies)} "
+                        "child dependencies that were not Git submoudles"
```
jasinner:
```suggestion
                        "child dependencies that were not Git submodules"
```

jasinner (approved):
Looks good to me, thanks for adding this!

corgi/collectors/brew.py
juspence: Oops, fixed now and thank you for catching!

PR #506, opened 2023-06-22
CORGI-668: Add back indexes that we might need to delete ComponentNodes

juspence:

In order to fix the purls on all container image components / remove the "redhat/" namespace from the start of the purl, we have to migrate purl values on each container's nodes. This was attempted previously, but failed due to duplicate (type, parent, purl) values on certain nodes.

Given that the data is duplicated, I tried to just delete the duplicate ComponentNodes. There are also other duplicate nodes that need to be cleaned up as part of other tickets. However, currently the deletion times out and no data can be cleaned up.

I see there were some custom ComponentNode indexes added in migration number 37 to improve performance, and I also see one of these indexes was removed in migration number 59 because it looked unused. If I remember right, we enabled / reset Postgres statistics about index usage, waited 24 hours, then checked the statistics to see which indexes were unused.

I assume we never manually cleaned up any nodes / other data in that very small window, so the indexes we removed might be needed in order to delete things. This MR adds back the missing indexes, in the hope (no guarantee) that will let us delete these duplicate nodes. I also redefined the other indexes from 0037 on the ComponentNode model, so that our Django models remain in sync with our Postgres database.

In the future:
1. We should wait at least a week after resetting statistics, before checking to see which indexes are unused. This prevents accidentally removing an index which is needed by some infrequently-run query, e.g. in one of our weekly tasks.

2. We should always make changes to Django models (including indexes and constraints) in models.py, never by running custom SQL. This prevents the Django models from becoming desynchronized with the actual underlying schema in the database.

3. We should never delete indexes that Django creates, such as for a field's unique=True constraint, a UniqueConstraint on multiple fields, a ForeignKey field, a ManyToManyField, or a TextField. These indexes support checking uniqueness on insertion / updates, iterating over related models, and running SQL LIKE queries, respectively.

4. We should always ensure that get_or_create() and update_or_create() use keyword arguments that exactly match the list of fields in a UniqueConstraint. These functions are atomic and guaranteed to work correctly, only when the keyword arguments / fields given are guaranteed to be unique (so appear together in a UniqueConstraint), and have an index to support this uniqueness check.

### DIFF
--- a/corgi/core/migrations/0070_fix_indexes.py
+++ b/corgi/core/migrations/0070_fix_indexes.py
@@ -1,0 +1,58 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = (("core", "0069_auto_20230606_1542"),)
+
+    operations = (
+        # Below were originally added in 0037_custom_indexes.py
+        # These were manually created by us to improve performance
+        # If you want to add custom indexes, update models.py
+        # Do not run custom SQL, or Postgres and Django will disagree
+        # about the current state of our tables / models
+        migrations.RunSQL("DROP INDEX core_componentnode_tree_parent_lft_idx"),
+        migrations.RunSQL("DROP INDEX core_cn_tree_lft_purl_parent_idx"),
+        # Already removed in 0059
+        # migrations.RunSQL("DROP INDEX core_cn_lft_tree_idx"),
+        migrations.RunSQL("DROP INDEX core_cn_lft_rght_tree_idx"),
+        # Add them back so the model definition is in sync with the DB
+        migrations.AddIndex(
+            model_name="componentnode",
+            index=models.Index(
+                fields=("tree_id", "parent_id", "lft"), name="core_cn_tree_parent_lft_idx"
+            ),
+        ),
+        migrations.AddIndex(
+            model_name="componentnode",
+            index=models.Index(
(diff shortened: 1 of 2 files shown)


jasinner (approved):
Let's try it.

juspence:
@jasinner I just tried to delete a whole tree using ```ComponentNode.objects.filter(tree_id=tree_id).delete()```, but this still times out just like deleting the individual nodes. And it's slow even when there's only one node in the tree / nothing should need to be rebuilt.

I think it's just an issue with finding "which node" in a table that contains many rows. Without that index on lft child's ID, I think we can't efficiently query "given some node ID I want to delete, which parent nodes have this node as a child?"

In other words, the MPTT algorithm tries to compare node_to_delete.node_id against other node lft ID values. If node_to_delete.node_id is greater than another_node's lft ID and less than another_node's rght ID, node_to_delete is a descendant of another_node. Any another_node with lft IDs greater than node_to_delete.node_id will also need to have its tree rebuilt (lft and rght IDs updated) to avoid gaps in the tree's node_id numbering.

I guess MPTT isn't very clever and tries to run this "which trees should I rebuild" query using the lft ID even when deleting a whole tree, even when a node has no parents, and so on.

EDIT: Oops, Github didn't refresh for me so I just saw your comment. Ack, going to merge this and then try the cleanup in the morning. Recreating the indexes may take too long - if the migration times out and the deployment fails, I'll fix the indexes manually tomorrow morning.

PR #507, opened 2023-06-23
CORGI-726 keep stream variants up to date

jasinner:
Sometimes a variant is moved from one stream to another. This allows that move, and also ensures all the builds associated with that stream are updated to reflect the new stream/variant relationship. The variant will no longer be associated with the old stream, but any builds we've called save_product_taxonomy on before the variant was moved will still reflect both streams, but that's OK in my opinion.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -1,4 +1,5 @@
 import re
+from typing import Any, Optional
 
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
@@ -11,11 +12,13 @@ from corgi.collectors.models import (
 )
 from corgi.collectors.prod_defs import ProdDefs
 from corgi.core.models import (
+    Component,
     Product,
     ProductNode,
     ProductStream,
     ProductVariant,
     ProductVersion,
+    SoftwareBuild,
 )
 from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
@@ -23,6 +26,66 @@ logger = get_task_logger(__name__)
 # Find a substring that looks like a version (e.g. "3", "3.5", "3-5", "1.2.z") at the end of a
 # searched string.
 RE_VERSION_LIKE_STRING = re.compile(r"\d[\dz.-]*$|$")
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def slow_update_builds_for_variant(
+    variant_name: str,
+    updated_stream: tuple[str, str],
+    updated_version: Optional[tuple[str, str]] = None,
+    updated_product: Optional[tuple[str, str]] = None,
+):
+    """Ensures new stream parent of the passed in variant is reflected in all the variant's builds
+    and child components"""
+    logger.info(
+        f"Updating components related to {variant_name} with new stream details: {updated_stream}"
+    )
+    updated_stream_objs = (
+        ProductStream.objects.get(name=updated_stream[0]),
+        ProductStream.objects.get(name=updated_stream[1]),
+    )
+    updated_version_objs = None
+    updated_product_objs = None
+    if updated_version:
+        logger.info(f"Also updating product_version with {updated_version}")
+        updated_version_objs = (
+            ProductVersion.objects.get(name=updated_version[0]),
+            ProductVersion.objects.get(name=updated_version[1]),
+        )
+    if updated_product:
+        logger.info(f"Also updating product {updated_product}")
+        updated_product_objs = (
+            Product.objects.get(name=updated_product[0]),
+            Product.objects.get(name=updated_product[1]),
+        )
+    for software_build in (
+        SoftwareBuild.objects.filter(relations__product_ref=variant_name).distinct().iterator()
+    ):
+        component = software_build.components.get()
+        _update_product_streams(
+            component, updated_stream_objs, updated_version_objs, updated_product_objs
+        )
+        for cnode in component.cnodes.get_queryset().iterator():
+            for d in cnode.get_descendants().iterator():
+                _update_product_streams(
+                    d.obj, updated_stream_objs, updated_version_objs, updated_product_objs
+                )
+
+
+def _update_product_streams(
+    component: Component,
+    updated_stream: tuple[ProductStream, ProductStream],
+    updated_version: Optional[tuple[ProductVersion, ProductVersion]] = None,
+    updated_product: Optional[tuple[Product, Product]] = None,
+):
+    component.productstreams.add(updated_stream[0])
+    component.productstreams.remove(updated_stream[1])
+    if updated_version:
+        component.productversions.add(updated_version[0])
+        component.productversions.remove(updated_version[1])
+    if updated_product:
+        component.products.add(updated_product[0])
+        component.products.remove(updated_product[1])
 
 
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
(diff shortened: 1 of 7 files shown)


jasinner:

jasinner:
I added a comment to CORGI-726 comparing variants before and after running the updated update_products task on the stage data (copied to dev). Most things are an improvement except the issue with rhscl/dts stream/variants, which didn't get better, but arguably isn't worse.

scripts/dump_db.sh
```diff
@@ -3,9 +3,4 @@ set -e
...
 export PGPASSWORD=test
 
-if ! podman ps | grep 'corgi-db' &>/dev/null; then
-    echo 'error: corgi "corgi-db" container does not appear to be running.'
-    exit 1
-fi
-
-pg_dump -v -j 2 -F d -f corgi.db -U corgi-db-user -h localhost -p 5433 corgi-db
\ No newline at end of file
+pg_dump -v -j 2 -F d -f corgi.db -U corgi-db-user -h localhost -p 5433 corgi-db
```
juspence: It might be better to use the long options, or else add a comment explaining what this does, but that's minor.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +78,211 @@ def update_products() -> None:
...
-                    active = pd_product_stream.pop("active")
-                    errata_info = pd_product_stream.pop("errata_info", [])
-                    brew_tags = pd_product_stream.pop("brew_tags", [])
-                    yum_repos = pd_product_stream.pop("yum_repositories", [])
-                    composes = pd_product_stream.pop("composes", [])
 
-                    brew_tags_dict = {}
-                    for brew_tag in brew_tags:
-                        brew_tags_dict[brew_tag["tag"]] = brew_tag["inherit"]
+def parse_stream(pd_product_stream, product, product_version, product_version_node, version):
```
juspence: Let's add header comments and type annotations to all these methods, and keep the names consistent (parse_product_stream).

juspence:
> ensures all the builds associated with that stream are updated to reflect the new stream/variant relationship...any builds we've called save_product_taxonomy on before the variant was moved will still reflect both streams, but that's OK in my opinion.

I'm a little confused here. Should the builds only report the new variant and not the old one, or are you saying they should report both?

I think having correct links is really important for e.g. the manifests we publish. Maybe the solution is just to manually clear all the linked Variants on all the components in this build, before we run build.save_product_taxonomy() to link components to the new Variant. I might be confused, happy to discuss if so.

tests/conftest.py
```diff
@@ -27,9 +27,9 @@ def api_path(api_version):
 
 
 def setup_product():
-    product = ProductFactory()
-    version = ProductVersionFactory(products=product)
-    stream = ProductStreamFactory(products=product, productversions=version)
```
juspence: Do randomly-generated names cause some issue? Otherwise, can we keep them random to make our tests a little more robust / catch edge cases our code doesn't handle?

tests/data/proddefs-update-variants-moved.json
```diff
@@ -0,0 +1,39 @@
...
+        {
+          "tag": "test_tag",
+          "inherit": false
+        }
+      ]
+    },
+    "new_stream": {
+      "brew_tags": [
+        {
+          "tag": "test_tag",
```
juspence: I haven't looked at code for how this is used yet. But maybe having different brew_tags for different streams will make it easier to assert we got the right data in our tests.

EDIT: Nope, this is right but maybe we can make the data more clear. Using something like "two_streams_with_same_tag" would make it a little easier to understand why the value is duplicated without looking at code.

tests/test_products.py
```diff
@@ -76,8 +79,12 @@ def test_products(requests_mock):
     # Which should include the stream's CPE + all the child variant CPEs (if any)
     assert et_variant.cpe in rhacm24z.cpes
 
+    # Assert mock_update_builds was called for all new variants
+    assert not mock_update_builds.called
```
juspence: FYI you can also do mock_update_builds.assert_not_called(), assert_called_once_with(args), and some other useful helper methods.

Your comment says "assert the mock was called", but we actually assert not called.

tests/test_products.py
```diff
@@ -76,8 +79,12 @@ def test_products(requests_mock):
...
+
 
 @pytest.mark.django_db
+@patch("corgi.tasks.prod_defs.update_builds_for_variant.delay")
```
juspence: If we started patching this task, should we add the mock to the list of test arguments and assert_not_called() or assert_has_calls()?

juspence (commented):
Still looking at main code change, really sorry for the delay in getting to this.

corgi/tasks/prod_defs.py
```diff
@@ -160,12 +179,10 @@ def parse_variants_from_brew_tags(brew_tags, name, product, product_stream, prod
     # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
     # active. See CORGI-546 for more details.
     if (
-            len(brew_tags) > 0
-            and product_version.name != "quay-3"
-            and name
-            not in ["rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9"]
+        len(brew_tags) > 0
+        and product_version.name != "quay-3"
+        and name not in ["rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9"]
```
juspence: Minor nit: use tuples for constant iterables:
```suggestion
        and name not in ("rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9")
```

corgi/tasks/prod_defs.py
```diff
@@ -208,14 +225,22 @@ def parse_variants_from_brew_tags(brew_tags, name, product, product_stream, prod
...
-                    product_variant.save_product_taxonomy()
+                    if not variant_created and not node_created:
+                        # If there was an existing product variant, and it was updated
+                        # to now be associated with a different stream, we need to
+                        # update all the builds to reflect the new stream and variant
+                        # relationship
+                        update_builds_for_variant.delay(product_variant.name)
+                    # Don't call product_variant.save_product_taxonomy here to avoid a truncated
+                    # tree where 2 et_product_versions with matching brew_tags exist, see
+                    # tests/test_products.py::test_product_variant_index_out_of_range
```
juspence: Do we have lots of et_product_versions with matching brew_tags, or is that a rare edge case?

Also, I think all the calls to save_product_taxonomy() are no-ops, at least in the old code before your changes (haven't finished reviewing yet).

I left them in since I didn't want to break something by mistake, but I updated this task a while ago. We now directly link variants to their parent streams when we create the variant. So save_product_taxonomy() shouldn't find anything to link, overwrite anything or make any changes at all.

I also think all ProductModel subclasses (products, versions, streams, variants) enforce unique=True on the name field. So if two different et_product_versions have the same brew_tags, I think we end up creating only one thing (Variant / Stream / Version) in the DB (not sure which ProductModel matches with et_product_version).

The first thing would do update_or_create(), not find anything to update, and end up creating a new model instance. The second thing with the matching brew_tags / name would also do update_or_create, find the existing model instance, and do an update that overwrites the first thing's data.

Not sure how we need to handle this or if I've misunderstood something, but wanted to mention here in case we need to think about this more.

jasinner:
> I'm a little confused here. Should the builds only report the new variant and not the old one, or are you saying they should report both?
> 
> I think having correct links is really important for e.g. the manifests we publish. Maybe the solution is just to manually clear all the linked Variants on all the components in this build, before we run build.save_product_taxonomy() to link components to the new Variant. I might be confused, happy to discuss if so.

> > ensures all the builds associated with that stream are updated to reflect the new stream/variant relationship...any builds we've called save_product_taxonomy on before the variant was moved will still reflect both streams, but that's OK in my opinion.
> 
> I'm a little confused here. Should the builds only report the new variant and not the old one, or are you saying they should report both?
> 
> I think having correct links is really important for e.g. the manifests we publish. Maybe the solution is just to manually clear all the linked Variants on all the components in this build, before we run build.save_product_taxonomy() to link components to the new Variant. I might be confused, happy to discuss if so.

I think you are right, we should clear the variants on all the old stream's builds. It's more aligned with our changed proposed  for CORGI-728 as well. I'll update the code to do that.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+    )
+    product_version_node, _ = ProductNode.objects.get_or_create(
+        object_id=product_version.pk,
+        defaults={
+            "parent": product_node,
+            "obj": product_version,
+        },
+    )
+    for pd_product_stream in pd_product_streams:
+        version = parse_stream(pd_product_stream, product, product_version, product_version_node, version)
```
juspence: This returns a "version" variable, but we already have a "product_version" ProductVersion obj and a "version" string, which is confusing. Is this creating and returning a stream object, or a dict?

If it's just returning the stream name, do we need to return anything at all? Is there some reason this parse_stream method would change the name we already found?

I think it will be more clear if you add a header comment to the parse_stream method, explaining what it returns. You could also update the variables here like "product_version_obj", and "version_name" or "version_obj" or "version_dict".

Do we also need to return something from the other methods?

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
-                        brew_tags_dict[brew_tag["tag"]] = brew_tag["inherit"]
+def parse_stream(pd_product_stream, product, product_version, product_version_node, version):
+    active = pd_product_stream.pop("active")
+    errata_info = pd_product_stream.pop("errata_info", [])
+    brew_tags = pd_product_stream.pop("brew_tags", [])
+    yum_repos = pd_product_stream.pop("yum_repositories", [])
+    composes = pd_product_stream.pop("composes", [])
+    brew_tags_dict = {}
+    for brew_tag in brew_tags:
+        brew_tags_dict[brew_tag["tag"]] = brew_tag["inherit"]
```
juspence: Minor nit, let's just build the dict directly when it's simple to do so / we already know what it should contain. This also helps mypy automatically infer what the type should be without needing to manually add an annotation:
```suggestion
    brew_tags_dict = {
        brew_tag["tag"]: brew_tag["inherit"]
        for brew_tag in brew_tags
    }
```

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
-                product_version_node, _ = ProductNode.objects.get_or_create(
-                    object_id=product_version.pk,
-                    defaults={
-                        "parent": product_node,
-                        "obj": product_version,
-                    },
-                )
+def parse_product_version(pd_product_version, product, product_node):
+    pd_product_streams = pd_product_version.pop("product_streams", [])
+    name = pd_product_version.pop("id")
```
juspence, line 137: If we name below something like stream_name, maybe we should make this one version_name just to be clear.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+    composes = pd_product_stream.pop("composes", [])
+    brew_tags_dict = {}
+    for brew_tag in brew_tags:
+        brew_tags_dict[brew_tag["tag"]] = brew_tag["inherit"]
+    composes_dict = {}
+    for compose in composes:
+        composes_dict[compose["url"]] = compose["variants"]
+    name = pd_product_stream.pop("id")
+    if match_version := RE_VERSION_LIKE_STRING.search(name):
+        version = match_version.group()
```
juspence, line 187: We return this value but I can't quite tell what it is. Is it the name with some part removed, or just the numeric part of the string / a version number?

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+    parse_variants_from_brew_tags(brew_tags, name, product, product_stream, product_stream_node,
+                                  product_version)
+    parse_errata_info(errata_info, product, product_stream,
+                      product_stream_node, product_version)
+    product_stream.save()
+    # Save taxonomies for newly-created model at end of each loop iteration
+    # All child models + nodes have already been created and had taxonomies saved
+    # This should be a no-op since we link models directly upon creation above
+    # Keeping it here just to be safe, but we could remove in future
+    product_stream.save_product_taxonomy()
```
juspence: You might be able to remove this, but you also had a note about how saving the taxonomy too early caused a truncated tree / overwritten data. Maybe we should check our assumptions in this code (and thank you for making it simpler / easier to do that).

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+        defaults={
+            "parent": product_version_node,
+            "obj": product_stream,
+        },
+    )
+    parse_variants_from_brew_tags(brew_tags, name, product, product_stream, product_stream_node,
+                                  product_version)
+    parse_errata_info(errata_info, product, product_stream,
+                      product_stream_node, product_version)
+    product_stream.save()
```
juspence: Did we change something about the stream in one of the methods above, and if so, can you add a comment explaining why we need to call .save() here? Or is this just for safety?

EDIT: Looks like we should actually call .save() at the end of the parse_errata_info() method.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+    # here to ensure the rhn_satelite_6.10 variants only get linked with that stream
+    # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
+    # active. See CORGI-546 for more details.
+    if (
+            len(brew_tags) > 0
+            and product_version.name != "quay-3"
+            and name
+            not in ["rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9"]
+    ):
+        #    continue
```
juspence: This `# continue` isn't needed anymore, so let's remove it to avoid confusion. I think it's a remnant of when the if-condition was the other way around, so we would skip creating variants using the Brew tags for this stream if any of the conditions matched.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+    # with the RHEL-7-SATELLITE-6.10 ET Product Version. We skip them
+    # here to ensure the rhn_satelite_6.10 variants only get linked with that stream
+    # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
+    # active. See CORGI-546 for more details.
+    if (
+            len(brew_tags) > 0
+            and product_version.name != "quay-3"
+            and name
+            not in ["rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9"]
+    ):
```
juspence: Do we need to add satellite 6.10 or newer versions to this list? Or can we remove any of these / is e.g. satellite 6.7 EOL yet?

Also one FYI about style - you can avoid many levels of nesting like this, by writing the if-condition the other way around and returning early from it:
```suggestion
    if not brew_tags or product_version.name == "quay-3" or name in ("rhn_satellite_6.7", "rhn_satellite_6.8", "rhn_satellite_6.9")
        return
    # else do stuff as usual, but no need for an indented block / more levels of nesting
```
It makes it a little easier to read in very long / complex code, and SonarQube will probably rank the complexity of the code lower, but no need to switch if you prefer this form.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +63,203 @@ def update_products() -> None:
...
+                ProductNode.objects.get_or_create(
+                    object_id=product_variant.pk,
+                    defaults={
+                        "parent": product_stream_node,
+                        "obj": product_variant,
+                    },
+                )
+                product_variant.save_product_taxonomy()
+    # persist et_product_versions plucked from errata_info
+    product_stream.et_product_versions = sorted(et_product_versions_set)
```
juspence, line 407: Aha, I think this is why we're calling product_stream.save() above. Let's call it here instead.

juspence (approved):
I'm not fully understanding the test about the index out of range error. I guess I'm probably confused and this is fine to merge if you're happy with it.

But maybe there are some bugs here still - if you don't mind waiting one more day to merge I can check again in my morning and have a much better idea what's happening.

corgi/tasks/prod_defs.py
```diff
@@ -25,6 +27,19 @@
...
+def update_builds_for_variant(variant_name):
+    """Ensures variants in relations table are associated to builds/components"""
+    for build_id in (
+        # We can skip selecting by type here as variant relations don't exist in community
+        ProductComponentRelation.objects.filter(product_ref=variant_name)
+        .values_list("build_id", flat=True)
+        .distinct()
+        .iterator()
+    ):
+        slow_fetch_brew_build.delay(build_id=build_id)
```
juspence: We're just relying on save_product_taxonomy() here right? Can we make a separate Celery task, like slow_update_build_variants, that fixes the variant links? Like below:

1 Find all the components for some build
2 Call component.variants.remove(the_old_variant) for all the components discovered above
3A Call component.variants.add(the_new_variant) for all the components discovered above
3B OR call build.save_product_taxonomy() once for the whole build

scripts/dump_db.sh
jasinner: Done.

tests/data/proddefs-update-variants-moved.json
jasinner: Good idea, I'll update the tag name.

tests/test_products.py
jasinner: Thanks, I'll update the comment. The update_builds task should only get called if variants (linked via brew_tags) moved between streams.

tests/test_products.py
jasinner: I just patched the task so that the test doesn't fail trying to call Celery. The focus on this test is on other things.

tests/conftest.py
jasinner: The stream name has to match the one in the proddefs.json for a couple of the new tests. I'll add stream_name as an argument with a default value, and only pass it where this is required.

corgi/tasks/prod_defs.py
jasinner: Yes, I'll update the task so it also clears the old_variants.

corgi/tasks/prod_defs.py
jasinner: We already use the variable name product_version below for the new object. Also we already use the name pd_product_version as the dict object parse from the JSON, so we are running out of options for names. I think the current name is acceptable because the focus of the function is on parsing a product version, so that should be enough context to figure out what we are naming.

corgi/tasks/prod_defs.py
jasinner: ok.

corgi/tasks/prod_defs.py

corgi/tasks/prod_defs.py
jasinner: See the RE_VERSION_LIKE_STRING comment.

corgi/tasks/prod_defs.py
jasinner: ok, I'll move the save call there.

corgi/tasks/prod_defs.py
jasinner: ok

corgi/tasks/prod_defs.py
jasinner: I didn't get to the bottom of what was causing that, but it only happens in a scenario like 'test_product_variant_index_out_of_range'. I also removed the other save_product_taxonomy calls.

corgi/tasks/prod_defs.py

corgi/tasks/prod_defs.py
jasinner: This is a rare edge case. However I ran into this issue with the actual product_definitons hosted on the dashboard so it's an edge case which exists today. The current code handles it.

jasinner (commented):
I updated the update_builds_for_variant function and associated calling code quite a bit. Please give that bit another review @juspence

jasinner:
> > ensures all the builds associated with that stream are updated to reflect the new stream/variant relationship...any builds we've called save_product_taxonomy on before the variant was moved will still reflect both streams, but that's OK in my opinion.
> 
> I'm a little confused here. Should the builds only report the new variant and not the old one, or are you saying they should report both?
> 
> I think having correct links is really important for e.g. the manifests we publish. Maybe the solution is just to manually clear all the linked Variants on all the components in this build, before we run build.save_product_taxonomy() to link components to the new Variant. I might be confused, happy to discuss if so.

I updated the PR so that we clear old product details from all Components.

corgi/tasks/prod_defs.py
jasinner: No need to add satellite 6.10 or later as it uses a different way (errata_info I think) to define it's builds, not brew_tags.

corgi/tasks/prod_defs.py
jasinner: removed.

corgi/tasks/prod_defs.py
jasinner: Every method just persist things to the DB, including this one, so no need to return anything. I updated that now.

jasinner:
> I'm not fully understanding the test about the index out of range error. I guess I'm probably confused and this is fine to merge if you're happy with it.
> 
> But maybe there are some bugs here still - if you don't mind waiting one more day to merge I can check again in my morning and have a much better idea what's happening.

I am also slightly confused by that. When I had product_variant.save_product_taxonomy() call where that comment is the test was failing, but I fixed the test by removing that. I hit that issue when I was testing this code on the actual product-defintions data.

.env-community
juspence, line 1: Looks like this file got committed by mistake.

tests/test_products.py
```diff
@@ -109,3 +117,101 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
...
+    et_product = CollectorErrataProduct.objects.create(et_id=1, name="et_product_1")
+    # Sets up 2 CollectorErrataProductVersions with matching brew tags
+    # The first one has no matching stream in product_definitions.json
+    et_version_2 = CollectorErrataProductVersion.objects.create(
+        et_id=12, product=et_product, name="et_stream_2", brew_tags=["two_streams_with_same_tag"]
+    )
+    et_version = CollectorErrataProductVersion.objects.create(
+        et_id=11, product=et_product, name="et_stream_1", brew_tags=["two_streams_with_same_tag"]
+    )
+    # This is a new variant for et_stream_2, but should get associated with the original stream
```
juspence: If we're using the -moved test data, should we add ```mock_update_builds.assert_called_once()```?

Or does this test just check that new Variants we create from ET data (versus existing Variants we update, in other tests) have the correct links to begin with, without any update_builds() call being needed? In that case, let's assert ```mock_update_builds.assert_not_called()```.

I'm a little confused because the comment says the second variant is created for the new / second stream, and should get associated with the original / first stream. But at the end we assert that both variants are linked to the new_stream. Maybe a header comment explaining the test is good:
```
"""Given no existing data in Corgi,
   two streams with different names but the same brew tags in Product Definitions,
   and one variant for each stream,
   test that both variants are linked to the most recently created stream
   when the variants are created, without needing to call update_builds()"""
```

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
-                product_version.save_product_taxonomy()
-            product.save_product_taxonomy()
+                    else:
+                        slow_update_builds_for_variant.delay(
+                            product_variant.name,
+                            (
+                                product_stream.name,
+                                existing_stream_name,
+                            ),
+                        )
```
juspence: I don't think checking if version changed before elif product_changed is quite right, since if the version changes the product may or may not change, but if the product changes then the version definitely changed:
```
                    updated_product = None
                    updated_version = None
                    updated_stream = {product_stream.name, existing_stream_name)
                    
                    if existing_product_name != product.name:
                        updated_product = (
                            product.name,
                            existing_product_name,
                        )
                    if existing_version_name != product_version.name:
                        updated_version = (
                            product_version.name,
                            existing_version_name,
                        ),
                    slow_update_builds_for_variant.delay(
                        product_variant.name,
                        updated_stream,
                        updated_version,
                        updated_product,
                    )
```

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
+    # quay-3 Errata Tool product version Quay-3-RHEL-8 list too many brew tags
+    # Linking quay streams to the 8Base-Quay-3 variant here via brew tags leads
+    # to builds from later streams being included in earlier ones,
+    # see PROJQUAY-5312.
+    # The rhn_satellite_6 streams have brew tags, but those brew tags are associated
+    # with the RHEL-7-SATELLITE-6.10 ET Product Version. We skip them
+    # here to ensure the rhn_satelite_6.10 variants only get linked with that stream
+    # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
+    # active. See CORGI-546 for more details.
+    # Also skip brew_tag matching for dts ps_module which share Variants/Brew Tags with rhscl
```
juspence, line 234: We avoid trying to match the Brew tags here, so the Corgi Variant doesn't move from one stream to another or have bad links created. I think this makes sense. But if ps_modules share both ET Variants and Brew tags, do we need to do something in the parse_errata_info code below?

Sorry if this is a dumb question, I swear this variant matching / linking code will make sense to me eventually.

tests/test_products.py
```diff
@@ -109,3 +117,101 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
...
+            f"{settings.PRODSEC_DASHBOARD_URL}/product-definitions", text=(prod_defs.read())
+        )
+
+    update_products()
+
+    variant = ProductVariant.objects.get(name="1")
+    new_stream = ProductStream.objects.get(name="new_stream")
+    assert variant.productstreams == new_stream
+    # The variant is no longer associated with the original stream
+    assert not stream == variant.productstreams
```
juspence: You only need the first assert since a ForeignKey field can only point to one object. If you're trying to assert that new_stream really is new and not a different name for stream, you could ```assert new_stream != stream```, or ```assert stream.name == "stream"``` above. I don't think we need to assert new_stream's name is "new_stream" since we looked it up here.

tests/test_products.py
```diff
@@ -109,3 +117,101 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
...
+    with open("tests/data/proddefs-update-variants.json") as prod_defs:
+        requests_mock.get(
+            f"{settings.PRODSEC_DASHBOARD_URL}/product-definitions", text=(prod_defs.read())
+        )
+
+    update_products()
+
+    # Verify that a new ProductVariant called "2" was created
+    variant_2 = ProductVariant.objects.get(name="2")
+    assert variant_2
```
juspence: You don't need this assert, .get() will raise an error if no object (or more than one object) is returned.

corgi/tasks/prod_defs.py
```diff
@@ -25,6 +28,64 @@
...
+    if updated_product:
+        logger.info(f"Also updating product {updated_product}")
+        updated_product_objs = (
+            Product.objects.get(name=updated_product[0]),
+            Product.objects.get(name=updated_product[1]),
+        )
+    for software_build in (
+        SoftwareBuild.objects.filter(relations__product_ref=variant_name).distinct().iterator()
+    ):
+        for component in software_build.components.get_queryset().iterator():
```
juspence: I think this can be simplified now that only root components are linked to builds:
```suggestion
        root_component = software_build.components.get():
```

corgi/tasks/prod_defs.py
```diff
@@ -25,6 +28,64 @@
...
+            _update_product_streams(
+                component, updated_stream_objs, updated_version_objs, updated_product_objs
+            )
+            for cnode in component.cnodes.get_queryset().iterator():
+                for d in cnode.get_descendants().iterator():
+                    _update_product_streams(
+                        d.obj, updated_stream_objs, updated_version_objs, updated_product_objs
+                    )
```
juspence: Is there a reason we iterate over component.cnodes instead of component.provides and component.upstreams? Using the ForeignKey fields avoids having to make a separate query to find all the descendant nodes, then one query for each node to access its related .obj / component.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
+
+def parse_errata_info(
+    errata_info: list[dict],
+    product: Product,
+    product_stream: ProductStream,
+    product_stream_node: ProductNode,
+    product_version: ProductVersion,
+):
+    """Parse and create ProductVariants from errata_info in product-definitions.json"""
+    et_product_versions_set = set(product_stream.et_product_versions)
```
juspence, line 367: I guess there's a bug here where if some stream has an et_pv_name removed from its errata_info, we'll keep the old et_pv_name around forever instead of cleaning it up. Maybe just starting with an empty set() here is enough to fix that, but it's been a rough week for me and I don't know if that's right or will cause other issues.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
+
+            for variant in et_product_version["variants"]:
+                logger.debug("Creating or updating Product Variant: name=%s", variant)
+                et_variant_cpe = (
+                    CollectorErrataProductVariant.objects.filter(name=variant)
+                    .values_list("cpe", flat=True)
+                    .first()
+                )
+
+                product_variant, _ = ProductVariant.objects.update_or_create(
```
juspence, line 384: Here we take the ET Variant data, and use it to update the Corgi Variant if it exists. Below, we try to get the Corgi Variant's node, or create it if it doesn't exist.

What happens when we process a Brew tag and create a Corgi Variant first, then find an ET Variant with a matching name and update the same Corgi Variant? I guess this would only happen if a stream defined both brew_tags and errata_info, and the brew_tag's name without the "-released" suffix matched the ET Product Version name, right?

Would we overwrite some of the Brew tag Corgi Variant's data in that case? If we moved the Brew tag Corgi Variant from one stream to another, would this code end up moving it back?

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
+                        },
+                    },
+                )
+                ProductNode.objects.get_or_create(
+                    object_id=product_variant.pk,
+                    defaults={
+                        "parent": product_stream_node,
+                        "obj": product_variant,
+                    },
+                )
```
juspence, line 405: Above we do update_or_create for the Brew tag-based Corgi Variant's node, here we do get_or_create for the ET-based Corgi Variant's node. if the related Variants are the same (have the same PK), would this code update the Variant object's ForeignKey to be linked to a different product_stream?

If that happened, and we didn't also update the node, the node would still be pointed at the older / original stream. The next time we called save_product_taxonomy() on the Variant, we would overwrite the ForeignKey update we made here using the node data which we never updated.

Using update_or_create() for the node instead could also cause some issue I haven't thought of. I guess I have no concerns if we'll never process the same Variant in both branches (based on Brew tags and errata_info).

Maybe you can run this task locally (in an Ipython session) using real prod-defs data and an empty / freshly cleaned Corgi DB. Add loop counters before each ProductVariant object and ProductVariant node get / update / create call, then either return or print the final counter values after the loop ends and we've processed all the data.

if the number of ProductVariant objects / nodes in the Corgi DB doesn't match the number of calls from the counter, it means the same Variant or node was processed twice, once in both functions. Above is just a suggestion to help you check this case, it's not at all necessary if you're sure we won't process the same Variant twice using both Brew tags and errata info.

corgi/tasks/prod_defs.py
```diff
@@ -25,6 +28,64 @@
...
+                component, updated_stream_objs, updated_version_objs, updated_product_objs
+            )
+            for cnode in component.cnodes.get_queryset().iterator():
+                for d in cnode.get_descendants().iterator():
+                    _update_product_streams(
+                        d.obj, updated_stream_objs, updated_version_objs, updated_product_objs
+                    )
+
+
+def _update_product_streams(
```
juspence, line 75: This LGTM, and it looks nice and simple. But in practice maybe it will cause some issue for builds with a huge number of provides, or variants with a huge number of builds. If it does, I have no objections to just disabling this part and only updating the ProductVariants to be part of a new stream, without also relinking their builds and components.

I am still glad we added this, and I think it's worth trying, but we have enough other issues with our data due to 727 and 728 that I don't think removing this will make things much worse.

corgi/tasks/prod_defs.py
```diff
@@ -63,192 +124,280 @@ def update_products() -> None:
...
+                    existing_product_name = product_variant.products.name
+                    existing_version_name = product_variant.productversions.name
+                    existing_stream_name = product_variant.productstreams.name
+                    product_variant.products = product
+                    product_variant.productversions = product_version
+                    product_variant.productstreams = product_stream
+                    product_variant.save()
+
+                    if existing_version_name != product_version.name:
+                        slow_update_builds_for_variant.delay(
```
juspence: Because we're running this task in a transaction, the values in the DB haven't changed yet. In SDEngine, we use .apply_async() with a delay= keyword argument to not run the other task right away. That gives some time for this task to finish and the transaction to commit.

If we don't wait, a build could be created after the update_builds task finds the list of builds to process (so would be skipped). If that build was also created before this update_products() task finished and committed its transaction / updated taxonomy data for this variant, there would be some builds which refer to the old parent stream / have old taxonomy values and never got updated. Let's make sure this doesn't happen - don't run the update_builds task until maybe a minute later, so this task has definitely finished and committed.

juspence (commented):
I don't really understand the code, or the data, or the tests...this is absolutely not your fault at all, I've just had a crazy week and I feel my brain melting.

I was able to reproduce that index out of range error, but I still don't understand it. Your test reproduces the conditions when it happens, and only calling product_variant.save_product_taxonomy() causes it. Updating the product_variant ForeignKeys manually doesn't trigger the issue, so that seems like a good workaround I guess.

I guess I'm just nervous because I don't understand what that implies about our data and the product taxonomy / tree we're building. Maybe it has to do with GenericForeignKey queries inside a transaction or something.

I found one definite bug, and one probably rare race-condition related bug. Other than that I think this is fine to merge if you're comfortable it won't cause issues.

tests/test_products.py
jasinner: I think this test was confusing because Variant "1" was first created by the setup_product() call, and the new streams in the proddefs data shared a product with that function as well. Therefore variant "1" is moved twice, from the stream created in setup_product to "stream", and then again to "new_stream". Variant "2" is also moved once. I added some extra comments and asserts to better demonstrate this.

corgi/tasks/prod_defs.py
jasinner: Seems the version of the code you have reviewed there is out of date. I think I have the logic correct. But I'll add a unit test to confirm. If the product changes we include the version update in the slow_update_builds_for_variant call. If the product hasn't changed, then we compare versions to see if that has changed instead.

.env-community
jasinner: removed, thanks.

corgi/tasks/prod_defs.py
jasinner: That is not a situation we see in the data today, so I think we can ignore it for now.

tests/test_products.py
jasinner: ok, removed one.

tests/test_products.py
jasinner: ok, removed.

corgi/tasks/prod_defs.py
jasinner: ok, fixed.

corgi/tasks/prod_defs.py
jasinner: That's not something we see in the data, so ignore it for now.

corgi/tasks/prod_defs.py
jasinner: That's not a situation we see in the data, so ignoring it for now. I think it's kind of OK if we overwrite the Variant here though. I think errata_info should be authoritative.

corgi/tasks/prod_defs.py
jasinner: I think we should address errata_info in a separate PR. We're not seeing variants moved between streams when using errata_info though, so I don't think there is anything to address at this point. This is basically how the brew_tag matching logic worked before. The first time you create a stream with errata_info the taxonomy will be created and will not change. Which is not a bad status-quo in my opinion.

corgi/tasks/prod_defs.py
jasinner: Ok, I added a countdown to these of 300 sec, or 5 mins.

corgi/tasks/prod_defs.py
jasinner: This is the same code found in SoftwareBuild.save_product_taxonomy. Using the cnodes nodes avoid any issue where we didn't update the .sources() for example. We may add another desc type in future, so this is a little more future proof as well.

corgi/tasks/prod_defs.py
jasinner: Ok, agree. Let's see how it goes.

PR #508, opened 2023-06-23
CORGI-725: Try to avoid timeouts in migration / many slow DB queries

juspence:
@jasinner The migration to add these SoftwareBuild values timed out in stage even when I tried to run it manually. I looked over the code again and realized it would look up the same SoftwareBuild multiple times, and update all relations with that build ID one at a time instead of all at once. Sorry about the bad code I suggested :facepalm: 

I hope this version is slightly better, since it should make one query to look up all the SoftwareBuild objects we need, then update all the relations with that build ID together. It also uses .update() instead of transaction.atomic(), so we still avoid race conditions but more of the filtering logic can be pushed to the DB and there's less overhead.

I didn't want to run this myself / manually or mess around with the data too much, since I was worried something would get messed up. But I looked in community stage and don't see any overlapping build IDs like you mentioned:
```
>>> from corgi.core.models import *
>>> ProductComponentRelation.objects.order_by().values_list("build_type", flat=True).distinct()
<QuerySet ['KOJI', 'CENTOS']>

>>> ProductComponentRelation.objects.order_by().values_list("build_id", flat=True).distinct().count()
118027
>>> ProductComponentRelation.objects.filter(build_type="KOJI").order_by().values_list("build_id", flat=True).distinct().count()
117083
>>> ProductComponentRelation.objects.filter(build_type="CENTOS").order_by().values_list("build_id", flat=True).distinct().count()
944

>>> centos_build_ids = ProductComponentRelation.objects.filter(build_type="CENTOS").order_by().values_list("build_id", flat=True).distinct()
>>> ProductComponentRelation.objects.filter(build_type="KOJI", build_id__in=centos_build_ids).order_by().values_list("build_id", flat=True).distinct().count()
0
```
And of course in enterprise stage I only see BREW-type builds / relations. So I think it should be OK to ignore build_type, but I also figured out a way to include it here and I left that in for safety.


jasinner (commented):
Let's try it. Thanks.

PR #509, opened 2023-06-23
CORGI-730: Clean up old builds in Corgi when they're deleted in Brew

juspence:
@RedHatProductSecurity/corgi-devs Please review. This will reduce the size of our DB significantly:
```
>>> Component.objects.filter(products__isnull=True).count()
1308204
>>> Component.objects.filter(products__isnull=True).values_list("software_build_id", flat=True).order_by().distinct().count()
160410
>>> ComponentNode.objects.filter(component__products__isnull=True).count()
4880584
>>> unshipped_builds = Component.objects.filter(products__isnull=True).values_list("software_build__build_id", flat=True).order_by().distinct()
>>> ProductComponentRelation.objects.filter(build_id__in=unshipped_builds).order_by().distinct().count()
67472
```
Components and builds not linked to any product have not been shipped. Some of these are "not shipped yet", but may ship in the future. However, many of these builds and components (plus their relations and nodes) have been deleted in Brew but never cleaned up in Corgi.

This causes issues for e.g. OpenLCS, when they try to find some container build's remote-source archive file in Brew. When Corgi's download_url doesn't work, OpenLCS tries to download and unpack the archive file, in order to get component source code they can scan.

However, if the Brew build was deleted then their process will fail with an error. In the case when Corgi's download_url does work, they're still doing extra work they don't need to, since the component they're scanning was never shipped (and will never ship) in any product.

Finally, I have to reload all container builds for CORGI-685 / CORGI-691, but this spams our monitoring email with failure messages for builds that have been deleted. Firing a slow_delete_brew_build task instead of an error means that we can get rid of this unneeded data, reduce the size of our DB, and avoid these errors.

No guarantees this is right, I need to look this over again Monday and make sure I didn't do something silly. But this should be ready for review.


corgi/tasks/brew.py
```diff
@@ -701,3 +699,49 @@ def slow_refresh_brew_build_tags(build_id: int) -> None:
...
+
+    # When we call .delete() on a queryset, Django returns a tuple with 2 elements
+    # The first element is the number of models we deleted directly / the queryset length
+    # The second element is a dict. The keys are this model's name, plus all the related model names
+    # which have a ForeignKey to the deleted model that defines on_delete=models.CASCADE
+    # The values are the number of model instances we deleted either directly or due to a ForeignKey
+    with transaction.atomic():
+        deleted_info: tuple[int, dict[str, int]] = ProductComponentRelation.objects.filter(
+            build_id=build_id, build_type=SoftwareBuild.Type.BREW
+        ).delete()
```
mprpic: Just curious: why don't we have a cascade delete for ProductComponentRelation that would handle this? Is there a case where the relation should remain in place even if it links to a non-existing build?

corgi/tasks/brew.py
```diff
@@ -701,3 +699,49 @@ def slow_refresh_brew_build_tags(build_id: int) -> None:
...
+        relation_count, relation_links = deleted_info
+        logger.info(
+            f"Deleted {relation_count} relation(s) and their related models: {relation_links}"
+        )
+
+        deleted_info = SoftwareBuild.objects.filter(
+            build_id=build_id, build_type=SoftwareBuild.Type.BREW
+        ).delete()
```
mprpic: Does this also handle the deletion of the ComponentNode models linked to the Components of this build?

mprpic (approved):
LGTM, just couple questions for clarity.

corgi/tasks/brew.py
juspence: Yes, we create "unprocessed relations" which need to have their related build loaded into Corgi. Jason is working on a ForeignKey approach to make the "did we process the relation" check easier, but it will use on_delete=models.SET_NULL instead of CASCADE. Right now we have to manually check each build ID in the ProductComponentRelations table to see if a matching SoftwareBuild exists.

If we deleted just the build and skipped the relations, we would just end up loading the build again the next day, when we try to process all the unprocessed relations (load all the referenced but nonexistent build IDs).

corgi/tasks/brew.py
juspence: Yep, we have a GenericRelation from Components back to their linked ComponentNodes which defines a related_name, and that triggers the node deletion when a Component is deleted.

Although it occurs to me that we only delete root components like source RPMs, since they're the only ones linked to a SoftwareBuild. We wouldn't delete e.g. any of the binary RPMs with the current logic, so maybe this needs more thought.

I guess I could filter to find all the provides and upstreams components with exactly one source / downstream. That would avoid deleting binary RPMs that are children of this (unshipped) SRPM, but are also children of / shipped in some module or container(s).

corgi/tasks/brew.py
```diff
@@ -701,3 +699,82 @@ def slow_refresh_brew_build_tags(build_id: int) -> None:
...
+    if build_state != koji.BUILD_STATES["DELETED"]:
+        raise ValueError(
+            f"Invalid state for build {build_id}: "
+            f"expected {koji.BUILD_STATES['DELETED']}, received {build_state}"
+        )
+
+    deleted_count = 0
+    with transaction.atomic():
+        # Build might not exist, or might exist but have no / multiple root components
+        # TODO: Check this! Why do builds have multiple root components?
```
juspence: This is expected per Jason, we sometimes link binary RPMs directly to their SRPM build. We expect that:
1. Binary RPMs are not linked to container / module builds, only to SRPM builds. I need to check the code / data in stage to confirm.

2. SRPM builds that are not shipped will end up cascading deletes to all their child binary RPMs, even if those binary RPMs appear in another build. Those other builds should ALSO be unshipped / not part of any product. I need to check the data in stage to confirm this. AKA find SRPM builds with products__isnull=True, then find container builds that provide the same binary RPMs, and confirm the container builds are also not shipped.

3. Before deleting a SoftwareBuild with directly-linked components (has items in the root_components_list), we should check that these components are not provided by another build. If they are, relink them to the other build, so deleting the SoftwareBuild doesn't cascade to its children / delete components that are still being used elsewhere and cause data loss.

I need to add a test for "multiple components provide this component, but actually they're all part of the same build". i.e. noarch containers and arch-specific containers that both provide some remote-source component. We should, but currently do not, clean up the remote-source component too, even though it has multiple "sources" (because all the sources come from the same build - just look at sources__build_id instead).

tests/test_tasks.py
```diff
@@ -229,6 +247,217 @@ def test_slow_refresh_brew_build_tags():
...
+    # So it should get cleaned up
+    unshipped_binary_rpm = ComponentFactory(
+        type=source_rpm.type,
+        namespace=source_rpm.namespace,
+        name=source_rpm.name,
+        epoch=source_rpm.epoch,
+        version=source_rpm.version,
+        release=source_rpm.release,
+        arch="aarch64",
+        software_build=None,
```
jasinner, line 336: nitpick: In reality such an binary RPM should have a software_build foreign key set to the same as the SRPM. To make this test reflect reality better we could update the type of this Component to be PYPI, or GOLANG to simulate a provided component either:

- Detected by a Syft scan
- Declared as a bundled component in an RPM spec file

tests/test_tasks.py
```diff
@@ -229,6 +247,217 @@ def test_slow_refresh_brew_build_tags():
...
+        type=ComponentNode.ComponentNodeType.PROVIDES,
+        parent=srpm_node,
+        purl=unshipped_binary_rpm.purl,
+        obj=unshipped_binary_rpm,
+    )
+    unshipped_binary_rpm.sources.set((source_rpm,))
+
+    # The X86 binary RPM is provided by both the SRPM and some container
+    # So it shouldn't get cleaned up
+    shipped_binary_rpm = ComponentFactory(
```
jasinner, line 362: nitpick: this should also never happen. But I guess a good idea to check it. To better reflect reality this type should be changed to PYPI, or GOLANG as explained above.

If there really was a binary RPM it should have software_build FK set to the same as the SRPM only, never the Container build.

corgi/tasks/brew.py
```diff
@@ -698,3 +699,82 @@ def slow_refresh_brew_build_tags(build_id: int) -> None:
...
+                Component.objects.annotate(downstreams_count=Count("downstreams")).filter(
+                    downstreams=root_component_pk, downstreams_count=1
+                )
+            )
+            deleted_count += _delete_queryset(upstream_components, "upstream component")
+
+        # Relations without a linked (NULL) build are "unprocessed"
+        # We process these relations each day by loading their build ID
+        # Deleting the relation avoids reloading the build we are deleting
+        relations = ProductComponentRelation.objects.filter(
```
jasinner, line 776: Nitpick: Once we update the ERRATA relations to only be created for shipped errata (CORGI-723) we should never have a deleted build in the relations table. Perhaps it's a good idea to fail the task in this case so we can investigate why that is happening. 

I'm happy for this to be merged as is, and we can update the task to fail in this case once CORGI-723 is implemented.

jasinner (approved):
Looks good to me. I'm looking forward to reducing the amount of extraneous build data we have. 
I provided some suggestions for improvement, but they are just ideas, take them or leave them.

corgi/tasks/brew.py
juspence: Ack, I added a note to the ticket about this case.

tests/test_tasks.py
juspence: Going to reply below.

tests/test_tasks.py
juspence: Going to reply below.

corgi/core/migrations/0073_fix_component_upstreams_and_software_builds.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import migrations, models
+
+from corgi.core.constants import ROOT_COMPONENTS_CONDITION
+
+
+def unset_software_build_for_non_root_components(apps, schema_editor):
+    """Only set SoftwareBuild for root components
+    This makes deleting a build in Corgi (when deleted in Brew) simpler
+    We can avoid checking which child components are linked to the build
+    and if they are also provided by other builds"""
```
juspence, line 12: @jasinner I know in your review you said we should never have binary RPMs with a NULL / None SoftwareBuild ForeignKey. But linking the binary RPMs to the build causes many problems, like we discussed in Gchat.

When we delete the build, any linked binary RPM components will also be deleted, because the ForeignKey sets on_delete=models.CASCADE. That means binary RPMs provided by 2 buiilds, for example a source RPM and a container, would be deleted even though they're still used elsewhere.

If we link only root components to the build, we guarantee that deleting the build has no surprising side-effects, and the components we automatically delete (which we can't control) are guaranteed to only be part of one build.

This avoids lots of complicated, error-prone logic to check if any directly-linked components are part of some other build, then either relink those components to the other build, or check if that other build should also be deleted.

The components that are not directly linked to a build are deleted manually (not automatically as part of a CASCADE). In this case, it's much easier to check what we're deleting and exclude something if needed. There's a safety check in corgi.tasks.brew.slow_delete_brew_build that looks at each component's list of sources, and only deletes components if they only belong to this build and not any others.

So I fixed the code to only link root components to the SoftwareBuild, then added this migration to fix our existing data. I need to fix one other item in the slow_delete_brew_build task, but does this approach seem OK?

jasinner: As discussed, let's just check with @JimFuller-RedHat to see if Griffon uses binary RPM, or binary Container Image software_build link.

corgi/tasks/brew.py
juspence: Number 3 was going to be really complicated, so I added a migration to fix our data. Now we only set the SoftwareBuild ForeignKey on root components.

I tweaked the safety check slightly. Now instead of checking for "components provided by multiple sources", we check for "components provided by multiple builds". That way if two different components from the same build (like arch-specific and noarch containers) both provide a remote-source component, we'll still clean up the entire tree.

corgi/core/migrations/0073_fix_component_upstreams_and_software_builds.py
juspence: Jim confirmed in Gchat he doesn't think it will be an issue.

juspence:
Lots of typos and little mistakes here, all fixed now but it makes me worry....I'll keep an eye on stage and revert this ASAP if I see issues.

PR #511, opened 2023-06-26
CORGI-298 / CORGI-728: Try to link better variants to a stream when processing variant-type ProductComponentRelations

juspence:
@RedHatProductSecurity/corgi-devs This may not be the right fix, and is definitely not a complete fix. But this code makes my head spin a bit, so early review is appreciated. We have three different bugs / cases:

1. We link Components to Channels, even though not every Component ships to every Channel. I added this linking by mistake (thinking it was omitted) so we can just stop linking them / go back to the code we had before. No other code depends on Component -> Channel links existing as far as I'm aware.

2. When we have a ProductComponentRelation between a ProductVariant and a SoftwareBuild, we link the SoftwareBuild's Components to the Variant given in the relation (correct), its parent Stream (correct), and all child Variants of that Stream / siblings of that Variant (incorrect). This is also an easy fix.

3. When we have a ProductComponentRelation between a ProductStream and a SoftwareBuild, we link the SoftwareBuild's Components to the Stream given in the relation (correct) and all child Variants of that Stream (incorrect). I don't know how to fix this. I think some of our code assumes a Component always has at least 1 linked Variant, and will break if we stop linking any Variants and only give the Stream instead.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -858,7 +858,6 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
         "productversions": set(),
         "productstreams": set(),
         "productvariants": set(),
-        "channels": set(),
     }
 
     for pnode in ProductNode.objects.filter(
@@ -875,12 +874,25 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
         product_streams = ProductNode.get_product_streams(family)
         product_details["productstreams"].update(product_streams)
 
-        product_variants = ProductNode.get_product_variants(family)
-        product_details["productvariants"].update(product_variants)
+        if not variant_names:
+            # We don't know which variants are the right ones
+            # So we just link all of them
+            # TODO: What can we do instead?
+            product_variants = ProductNode.get_product_variants(family)
+            product_details["productvariants"].update(product_variants)
+        # else we have relations for particular Variants
+        # Link only those Variants (below)
 
-        channels = ProductNode.get_channels(family)
-        product_details["channels"].update(channels)
+    if variant_names:
+        for pnode in ProductNode.objects.filter(
+            level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"], productvariant__name__in=variant_names
+        ):
+            family = pnode.get_family()
+            product_variants = ProductNode.get_product_variants(family)
+            product_details["productvariants"].update(product_variants)
+
     # For some build, return a mapping of ProductModel type to all related ProductModel UUIDs
+    # Except channels, because we can't link these correctly (CORGI-298)
     return product_details
 
 
@@ -1628,7 +1640,11 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         self.productversions.add(*product_pks_dict["productversions"])
         self.productstreams.add(*product_pks_dict["productstreams"])
         self.productvariants.add(*product_pks_dict["productvariants"])
-        self.channels.add(*product_pks_dict["channels"])
+        # Don't link channels for all variants to this component (CORGI-728)
+        # Not every channel has the same content sets / ships this component
+        # We don't know which do and don't, so for now just stop linking
+        # TODO: I need to review description in ticket and other code
+        #  This fix may not be right / complete
         return None
 
     def get_roots(self, using: str = "read_only") -> list[ComponentNode]:


corgi/core/models.py
```diff
@@ -875,12 +874,25 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
         product_details["productstreams"].update(product_streams)
 
-        product_variants = ProductNode.get_product_variants(family)
-        product_details["productvariants"].update(product_variants)
+        if not variant_names:
+            # We don't know which variants are the right ones
+            # So we just link all of them
+            # TODO: What can we do instead?
+            product_variants = ProductNode.get_product_variants(family)
+            product_details["productvariants"].update(product_variants)
```
juspence, line 882: This is point number 3 in the description. I guess I can try to just remove this code, and see what starts breaking. But is there something else we can do insteadd?

corgi/core/models.py
```diff
@@ -875,12 +874,25 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+        # else we have relations for particular Variants
+        # Link only those Variants (below)
+
+    if variant_names:
+        for pnode in ProductNode.objects.filter(
+            level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"], productvariant__name__in=variant_names
+        ):
+            family = pnode.get_family()
+            product_variants = ProductNode.get_product_variants(family)
+            product_details["productvariants"].update(product_variants)
```
juspence, line 892: After this is merged, I'll need to fix data by reprocessing all the variant-type ProductComponentRelations, and clearing every Component's list of Channel links.

EDIT: I also need to find the builds mentioned in all stream-type relations, find their set of Components the same way we do in SoftwareBuild.save_product_taxonomy(), and then clear each Component's list of Variant links, if we stop linking all Variant children of a given Stream as described above.

I'm going to try to handle this in a migration.

corgi/core/models.py
```diff
@@ -1628,7 +1640,11 @@ def save_product_taxonomy(
         self.productversions.add(*product_pks_dict["productversions"])
         self.productstreams.add(*product_pks_dict["productstreams"])
         self.productvariants.add(*product_pks_dict["productvariants"])
```
juspence, line 1642: Does anyone know why we used .add() here instead of .set()? I recall a comment saying that we couldn't rely on only one build having all the relations.

I guess that means e.g. a binary RPM can be part of multiple builds, so we have to append to the relations / links instead of clearing them. Otherwise only the last build's relations will win, and overwrite the relations from the other builds. Is that right?

corgi/core/models.py
```diff
@@ -1628,7 +1640,11 @@ def save_product_taxonomy(
...
-        self.channels.add(*product_pks_dict["channels"])
+        # Don't link channels for all variants to this component (CORGI-728)
+        # Not every channel has the same content sets / ships this component
+        # We don't know which do and don't, so for now just stop linking
+        # TODO: I need to review description in ticket and other code
+        #  This fix may not be right / complete
```
juspence, line 1647: Reminder for me, I'll remove this before merging. Wanting to let this sit overight and come back to it tomorrow with fresh eyes.

corgi/core/models.py
jasinner: I think get_product_details needs to be rewritten something like this:

~~~
def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> dict[str, set[str]]:
    """
    Given stream / variant names from the PCR table, look up all related ProductModel subclasses

    In other words, builds relate to variants and streams through the PCR table
    Components relate to builds, and products / versions relate to variants / streams
    We want to link all build components to their ancestor products, versions, and streams
    """

    product_details: dict[str, set[str]] = {
        "products": set(),
        "productversions": set(),
        "productstreams": set(),
        "productvariants": set(),
        "channels": set(),
    }

    if not variant_names and not stream_names:
        return product_details

    query = ProductNode.objects.get_queryset()
    stream_query = Q(level=MODEL_NODE_LEVEL_MAPPING["ProductStream"], productstream__name__in=stream_names)
    variants_query = Q(level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"], productvariant__name__in=variant_names)
    if stream_names and variant_names:
        query = ProductNode.objects.filter(stream_query | variants_query)
    elif stream_names:
        query = ProductNode.objects.filter(stream_query)
    elif variant_names:
        query= ProductNode.objects.filter(variants_query)

    for pnode in query:
        ancestors = pnode.get_ancestors(include_self=True)

        products = ProductNode.get_products(ancestors)
        product_details["products"].update(products)

        product_versions = ProductNode.get_product_versions(ancestors)
        product_details["productversions"].update(product_versions)

        product_streams = ProductNode.get_product_streams(ancestors)
        product_details["productstreams"].update(product_streams)

        product_variants = ProductNode.get_product_variants(ancestors)
        product_details["productvariants"].update(product_variants)

    # For some build, return a mapping of ProductModel type to all related ProductModel UUIDs
    return product_details
~~~

I updated the query to directly lookup ProductNodes with a ProductVariant level and in the product_variant argument set, so I could removed the aggregation of variant_streams in to the stream_names set.

Also inside the loop I use get_ancestors() instead of get_family() so that we don't traverse leaf nodes from those returned in the query.

corgi/core/models.py
jasinner: That sounds good to me.

corgi/core/models.py
jasinner: The problem is with UPSTREAM components, because we call Component.save_product_taxonomy() via SoftwareBuild.save_product_taxonomy() we are adding product relationships for that build only. UPSTREAM components can and are used in multiple builds, if we replaced the relations of these components every time they would only reflect the last builds relations.

Take for example a GOLANG component (x/text) bundled by the "cri-o" RPM, but also build into the "etcd" Container Image. If we called the "cri-o" RPM save_product_taxonomy first the (x/text) component would reflect the products of the cri-o RPM. However later if we called the "etcd" Container Image build save_product_taxonomy function (x/text) would not only reflect the products of the "etc" Container Image whereas it should reflect both.

corgi/core/models.py
jasinner: I think this PR addresses CORGI-728 nicely, if we also run a script which cleans up relations and their builds as you mentioned earlier. However for CORGI-727 we will also need to remove the `setup_pulp_relations` task and associated data cleanup.

jasinner (changes requested):
Please change the get_product_details function as described.

corgi/core/models.py
juspence: We won't traverse leaf nodes (when we shouldn't) to begin with. We use get_family() to get all the nodes (products, versions, streams, and variants) in a single query. Then we report only nodes of a certain level in each dict key. For example, the product_details["productstreams"] set contains only ProductStream nodes with level=2, because we called ProductNode.get_product_streams() which returns only those nodes from the queryset it's given.

The problem that I fixed here is just point number 2 of the description. When we have a variant_ref, we also link sibling variants to this Component instead of only the variant given.

Your proposed change seems to be trying to address point number 3, for when we're given a stream_ref in the relations table instead of a variant_ref. In that case, what variants (if any) should get linked? I think you're saying we shouldn't link any variants, but I'm not sure and don't want to assume.

I also think this is going to break other code, which expects Components to always be linked to at least 1 variant.

corgi/core/models.py
juspence: OK, I think I was confused earlier - I mentioned cleaning up Channel-Component links thinking that would be enough to fix this issue. But actually we'd only leave the channels unlinked, and the CDN_REPO relations we create for every Channel-Variant pair would still get processed. So I guess we'd still link every Component in some Channel to all that Channel's Variants, even when that Component isn't really part of the Variant.

Sorry for the dumb questions and thanks so much for your patience. This code and this bug both make my head spin, but I think I'm finally starting to get what the issue is. I'll handle 727 in a follow-up PR.

juspence:
@jasinner I've addressed some feedback, and I won't ever merge without a thumbs-up / ignore your comments, but when you click "request changes" it blocks the PR from being merged until you later approve it. So that means I can't merge even if I make those changes you suggested. So going to close this PR and open a new one.

jasinner:
Ok, I'll stop using the request changes feature and just leave it unapproved next time.

corgi/core/models.py
jasinner: Yes, this addresses point 3. I don't think any code expect a Component has a variant. Where we're hoping to get to is that if a build has a variant, we can then determine it's CPE for an SBOM. And probably other things in the future.

PR #512, opened 2023-06-27
CORGI-727: Remove Pulp logic that links Components to the wrong Variants / Channels

juspence:
@jasinner Please review when you have time. This will remove the CDN_REPO relation type, the Pulp tasks which create these relations, the Celery logic to schedule those tasks, and finally tests and test data for all of these.

There's also a migration that finds builds with CDN_REPO relations, clears the product / version / stream / variant links on all their components, then calls build.save_product_taxonomy() again so we get only the correct ProductModel links, without the incorrect links that used the CDN_REPO relations.


corgi/tasks/errata_tool.py
```diff
@@ -152,35 +152,36 @@ def update_variant_repos() -> None:
...
                         "relative_url": rpm_repo.relative_url,
                         "meta_attr": {"content_set": rpm_repo.content_set},
                     },
                 )
                 if created:
-                    logger.info("Created new channel %s for variant %s", repo, pv.name)
+                    logger.info("Created new channel %s for variant %s", repo_obj, pv.name)
                 # TODO: investigate whether we need to delete CDN repos that were removed from a
                 #  Variant between two different runs of this task.
+                #  Seems likely - should we handle as part of CORGI-727?
```
juspence: Maybe we don't need to fix this now, since we aren't relating Pulp repos to builds and their components anymore. But did this cause one of the recent bugs, or make them worse? Should we fix this anyway to prevent problems in the future / in other areas?

jasinner: I'd say it's not urgent, but perhaps just leave the TODO as a future task.

jasinner (approved):
LGTM! great work

PR #514, opened 2023-06-29
CORGI-728: Stop linking channels to components, and try to link better variants

juspence:
Already reviewed by Jason, see https://github.com/RedHatProductSecurity/component-registry/pull/511. As discussed on a call, we completely stop linking variants to a component when we only know a stream name / are processing a stream-type relation. Linking all the variants for that stream to a component creates bad data / causes many issues for IR, since we report the component in variants that don't use it at all.

When we're processing a variant-type relation, we now report only the current variant, intead of that variant and all its siblings (all the child variants of some parent stream). After we removed CDN_REPO-type relations in CORGI-727 / https://github.com/RedHatProductSecurity/component-registry/pull/512, the only relations that refer to Variants are ERRATA-type relations, so it should be easier to find the right CPE for some component when e.g. building manifests.


corgi/core/migrations/0075_fix_channel_variant_component_links.py
```diff
@@ -0,0 +1,64 @@
...
+            for upstream_component in component.upstreams.get_queryset().iterator():
+                # Fix the upstream components
+                # They aren't ever linked to the build, so can't be handled above
+                upstream_component.productvariants.clear()
+                upstream_component.productstreams.clear()
+                upstream_component.productversions.clear()
+                upstream_component.products.clear()
+
+        # Add the correct Component-Variant links on all Components in this build
+        build.save_product_taxonomy()
```
juspence: Note to self: this needs to get fixed.

juspence: Punting on migrations since they take forever to update links on all these Components. I'll add notes to CORGI-566 with details of what to / how to clean up, and the migrations themselves are still in the commit history if needed.

PR #515, opened 2023-06-29
CORGI-727: Fix bad migration that fails to run build.save_product_taxonomy()

juspence:
@RedHatProductSecurity/corgi-devs FYI. I've manually edited the migration with these changes (using vi in a stage pod) and run it successfully.

EDIT: I wrote above referring to a typo that blocked the migration from running at all. I opened this PR once the migration got past the previous typo / failure point, without waiting for it to finish since I knew it would take a long time. There's still a timeout issue, but I hope if I merge this code now and run it overnight manually (e.g. with nohup), we'll be able to get past this stupid issue.

### DIFF
--- a/corgi/core/migrations/0074_remove_cdn_repo_relations.py
+++ b/corgi/core/migrations/0074_remove_cdn_repo_relations.py
@@ -4,61 +4,83 @@ from django.db import migrations, models
 
 # Non-historical model, i.e. if you change save_product_taxonomy() tomorrow
 # the migration will not run the same code as before (at the time the migration was written)
-from corgi.core.models import SoftwareBuild as CurrentSBModel
+from corgi.core.models import ROOT_COMPONENTS_CONDITION
+from corgi.tasks.brew import slow_fetch_brew_build
 
 
 def remove_cdn_repo_relations(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


corgi/core/migrations/0074_remove_cdn_repo_relations.py
```diff
@@ -2,57 +2,76 @@
...
         # Fix all components on builds that have CDN_REPO relations
-        for component in build.components.get_queryset().iterator():
-            # Fix the root components linked directly to this build
-            # Clear the Variant / Channel links since they're wrong
-            # We only append to links when saving taxonomies
-            # So to remove bad data and add the right data
-            # We need to clear first, and then reprocess afterwards
-            component.channels.clear()
-            component.productvariants.clear()
+        root_component = Component.objects.filter(ROOT_COMPONENTS_CONDITION).get(
```
juspence: This is wrong because it doesn't account for RPMMOD builds...will fix tomorrow.

juspence:
Abandoning the data cleanup effort and removing the migration so other migrations (if needed / added during my PTO) are not blocked from running. I will add a note to CORGI-566 about how to clean up data for this issue.

PR #517, opened 2023-06-30
CORGI-668: Fix OCI purls

juspence:
This was already reviewed, but the migration failed the first time it was merged due to duplicate data. Opening this as a draft so I don't forget this exists after PTO. Also needs coordination with OpenLCS.

### DIFF
--- a/corgi/core/migrations/0075_fix_oci_purls.py
+++ b/corgi/core/migrations/0075_fix_oci_purls.py
@@ -1,0 +1,36 @@
+from django.db import IntegrityError, migrations, transaction
+from django.db.models import Value, functions
+
+
+def fix_oci_purls(apps, schema_editor):
+    """Remove "redhat/" namespace from all "pkg:oci/" purls"""
+    Component = apps.get_model("core", "Component")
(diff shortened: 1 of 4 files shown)


juspence:
@RedHatProductSecurity/corgi-devs Please review. The only change from the previous attempt is to try and remove "redhat/" from a ComponentNode's purl, but to delete that node instead if a matching node already exists with the same node type and node parent, and has the same purl except for "redhat/" not being present.

There are about 16,000 build IDs with this condition in stage. I added the list to CORGI-566 so we remember to reload / rerun SCA on all of them. And since we're talking about container builds, we need to do that reloading anyway to get the missing Cachito data, so I don't think deleting the nodes will cause issues.

I also threw in two minor commits to fix a mypy warning, and to clean up old task results (after 30 days) to try and reduce our DB size. That may or may not help with the DB errors we're seeing - I haven't looked to see how large the django_celery_results.models.TaskResult table is, but I'm assuming it's significant and we haven't run any cleanup before.

jasinner (approved):

kgrant-rh (approved):

PR #518, opened 2023-06-30
CORGI-725: Fix an error with tuple unpacking in slow_save_errata_product_taxonomy

juspence:
@jasinner Could you please look this over and add a test when you get a chance? I guess I'll go ahead and merge now but please feel free to revert if I did something stupid here.


juspence:
Actually, not going to merge since CORGI-741 is also messing up our tasks and there's no reason for me to rush and break things on a Friday.

jasinner (approved):

PR #520, opened 2023-07-05
Add collector which reads SBOMer manifests

kgrant-rh:
There are some outstanding issues, namely that Brew information isn't yet included in SBOMer manifests, either for components built only there, or for components with both PNC and Brew build info.


kgrant-rh:
@RedHatProductSecurity/corgi-devs Tests for the task forthcoming, but please double-check my component and node creation logic.

corgi/tasks/pnc.py
```diff
@@ -30,4 +31,63 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            name=component["name"],
+            version=component["version"],
+            release="",
+            arch="noarch",
+            purl=component["purl"],
+            namespace=component["namespace"],
+            softwarebuild=builds[component["build_id"]],
+            related_url=component["related_url"],
+            defaults={"meta_attr": component["meta_attr"]},
+        )
```
jasinner: We need to make sure that we're only using the 'unique_together' fields as keyword argument, so that database fetch is always hitting an index. Also, if we want to be able to re-process an SBOMer SBOM we should use update_or_create here instead.

```suggestion
        components[component.bomref], _ = Component.objects.update_or_create(
            type=Component.Type.MAVEN,
            name=component["name"],
            version=component["version"],
            release="",
            arch="noarch",
            defaults={
                 "namespace": component["namespace"],
                 "softwarebuild": builds[component["build_id"]],
                 "related_url": component["related_url"],
                 "meta_attr": component["meta_attr"]
            },
        )
```

I also took purl out of the fields because I think we should generate this using our own logic we have in corgi/core/models.py::Component.get_purl() and _build_maven_purl(). We should however also be saving the 'declared' purl to meta_attr for future reference.

There are other fields of components which we are not persisting into our models. All of this can probably be saved to meta_attr in case we need to look it up at some later time. For example licenses and externalReferences.distribution. For other tasks/collectors, we've also been saving a 'source' meta_attr which indicates where this Component came from.

Component.pedigree seems like something we should definitely be saving. We should probably be creating a distinct Component for it with Type.GENERIC, or Type.GITHUB, as well as a new child ComponentNode of ComponentNodeType.SOURCE. Doing so will allow compliance reports such as pedigree report (GRIF-60). However if it's an internal Red Hat reference we probably only want to save those as SoftwareBuild.source for the root component, or meta_attr for non-root components.

corgi/tasks/pnc.py
```diff
@@ -30,4 +31,63 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+                defaults={"obj": components[parent]},
+            )
+
+        for dep in deps:
+            nodes[dep], _ = ComponentNode.get_or_create(
+                type=ComponentNode.ComponentNodeType.PROVIDES,
+                parent=nodes[parent],
+                purl=components[dep].purl,
+                defaults={"obj": components[dep]},
+            )
```
jasinner: Every component needs a ComponentNode so we should create a ComponentNode for the root component, and nest all the other components under it. The level of nesting can be determined by the dependencies section.

We can't use ComponentNode.get_or_create because that will search across the entire Node forest, including other SBOMs from SBOMer. What we need to do is create a new tree for each SBOM received. So I'm pretty sure we'll always be calling ComponentNode.create and using local variables of nodes already created (eg. for the root node) as the parent reference.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,76 @@
...
+                    ref["comment"]: ref["url"]
+                    for ref in component["externalReferences"]
+                    if ref["type"] == "build-system"
+                }
+
+                # Everything built within Red Hat should have *some* build info
+                if not build_urls:
+                    # FIXME: Current manifests don't always have build info
+                    raise ValueError("Component has no builds")
+                else:
```
jasinner: In our current data model only root components have software_build foreign keys. Continuing this pattern might make sense here also. For non-root components we can always save the the PNC build reference to the Component.meta_attr in case we need it in future, instead of creating a SoftwareBuild object.

jasinner (commented):
The code is nice and clean. However as you mentioned, needs more tests for pnc task. I provided quite a bit of feedback on model creation. I'm not opposed to breaking the current Component/SoftwareBuild/ComponentNode patterns where it makes sense. However I can't think of a good reason why we'd need to link SoftwareBuilds to every Component. Doing so would mean we need to adjust our approach to product linking for Components created by the pnc task.

I don't see any product references in the pnc_sbom.json test data. Do we have an example of an SBOM with a 'productVariant'? Before this goes live we should make sure we are creating ProductComponentRelations with a product_ref set to the 'productVariant', with a new type 'SBOMER' and PNC build reference for the root component. I think we'll also need to include errata_info in product-definitions for quarkus to be able to relate new ProductComponentRelations with the SoftwareBuilds/Components.

corgi/tasks/pnc.py
```diff
@@ -30,4 +31,61 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+
+    # Link dependencies
+    nodes = {}
+    # Create the root node from the SBOM's component
+    root_component = components.pop("root")
+    nodes["root"], _ = ComponentNode.objects.get_or_create(
+        type=ComponentNode.ComponentNodeType.SOURCE,
+        parent=None,
+        defaults={"obj": root_component},
+    )
```
jasinner: Let's also set the purl attribute of ComponentNode

```suggestion
    nodes["root"], _ = ComponentNode.objects.get_or_create(
        type=ComponentNode.ComponentNodeType.SOURCE,
        parent=None,
        purl=root_component.purl,
        defaults={"obj": root_component},
    )
```

jasinner (commented):
I think it would be a good idea to add more tests for the pnc task which verified the component hierarchy is created as expected using ComponentNodes. I think you are probably missing a call to root_component.save(), possibly need to save the nested components too, which adds the foreign keys links to provides, sources etc.

juspence:
@kgrant-rh I just saw this in my emails again after some recent push. I think I never replied here originally since Jason had already reviewed, but please let me know if you'd still like me to take a look as well.

One small FYI - it looks like you might have clicked Github's "rebase" button instead of doing this manually. If memory serves, that doesn't actually rebase but just merges the main branch into your feature branch, so that's why you have 63 commits in this PR now.

I think if you git pull locally, git rebase onto origin/main, then force push the rebased branch to Github, the "extra" commits will disappear / end up in the right part of the git history.

kgrant-rh:
@juspence Yeah, I'd appreciate your input if you have a chance. Tests are still outstanding, I'm working on complete coverage at the moment.

Thanks also for the info about rebasing, I'll clean that up.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+    to other middleware."""
+
+    def __init__(self, data: Mapping[str, Any]):
+        # Product information
+        root_meta = {
+            param["name"]: param["value"] for param in data["metadata"]["component"]["properties"]
+        }
+        self.product = root_meta["errata-tool-product-name"]
+        self.product_version = root_meta["errata-tool-product-version"]
+        self.product_variant = root_meta["errata-tool-product-variant"]
```
juspence, line 21: Do middleware teams / SBOMer manifests report what the product-stream / minor version is? Or is that kind of distinction something that only Product Security cares about?

If we don't know the stream directly from the manifest, can we determine the right value using the variant and version names? e,g, by looking for existing models in Corgi or ET or somewhere. Or maybe this is another middleware thing, and they don't define any streams?

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
-    Temporarily returns an int for testing purposes, to be
-    completed in CORGI-488"""
-    if "components" not in data:
-        raise ValueError("SBOM is missing component data")
+class SbomerSbom:
+    """Collector to parse SBOMs from PNC/Sbomer. See CORGI-488.
+    Future development should generalize this where applicable
+    to other middleware."""
+
+    def __init__(self, data: Mapping[str, Any]):
```
juspence: This is really minor, but why do we use Mapping in this context instead of just dict? Is this supporting something like `Union[dict, defaultdict]` where we don't know in advance exactly which dict-like type we'll receive?

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+        self.product_variant = root_meta["errata-tool-product-variant"]
+
+        if "components" not in data:
+            raise ValueError("SBOM is missing component data")
+
+        self.components = {c["bom-ref"]: c for c in data["components"]}
+        # The root component is listed separately in metadata
+        self.components["root"] = data["metadata"]["component"]
+        for bomref, component in self.components.items():
+            self.components[bomref]["meta_attr"] = {}
```
juspence: Since `component` already refers to `self.components[bomref]`, do you need to use .items() here? Can't we just do `for component in self.components.values()` to set `component["meta_attr"]` and other properties directly, instead of using an indirect `self.components[bomref]["meta_attr"]` reference?

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+            if "redhat" not in component["purl"]:
+                # Community components won't have build info
+                self.components[bomref]["namespace"] = Component.Namespace.UPSTREAM
+            else:
+                self.components[bomref]["namespace"] = Component.Namespace.REDHAT
```
juspence: This seems a little risky. In the past, we've only checked version strings for "redhat" and not the whole purl.

For e.g. maven components, we might have group identifiers like "com.redhat.thing". I guess any component like that should probably have its namespace set to REDHAT.

Another edge case might be some qualifier like `?download_url=redhat.com/stuff`. Again, maybe if it has a download URL from us, we can safely assume it really is a REDHAT component. But I'd be nervous about whatever edge-cases I haven't thought of here.

Is there any key in the manifest which has just the version info? Parsing only that might be safer. Or if not, maybe we can rely on purl syntax, like `pkg:type/name@version?qualifier`. If we split the purl on `@` at most once, and then split on `?` at most once, we can check only the version part (after the @ and before the ?).

Maybe that's also risky and not worth the added complexity - up to you.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+
+            # Declared purl
+            self.components[bomref]["meta_attr"]["purl_declared"] = component["purl"]
+
+            # Related URL
+            website = [
+                ref["url"] for ref in component["externalReferences"] if ref["type"] == "website"
+            ]
+            if website:
+                self.components[bomref]["related_url"] = website[0]
```
juspence: This code could technically have more than one result, but we only use the first one. Is that a side-effect of our API, since we only support one related URL?

And if we don't have any results at all, we could check the purl for a `?download_url=` qualifier and reuse that for the related_url as well.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+
+            # VCS URL
+            vcs = [ref["url"] for ref in component["externalReferences"] if ref["type"] == "vcs"]
+            if vcs:
+                # TODO: Should this be called something else? "remote_sources"?
+                self.components[bomref]["meta_attr"]["vcs"] = vcs[0]
```
juspence: vcs_ref_url or vcs_url_ref sounds a little better to me, but it's also just meta_attr so I wouldn't worry about it too much.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+
+            # Pedigree info when it's available
+            if "pedigree" in component:
```
juspence: These `if key in dict` checks have an annoying edge case, whenever the key is present but has a None value. In that case, we'll end up copying the None into our meta_attr dict. It's fine if that's what you wanted.

But if not, you can be a little more safe by doing `if component.get("pedigree")` instead. That will make sure the pedigree key in the meta_attr only exists if the value is truthy (not None or False or an empty string / list / etc.), and completely omit the key otherwise.

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
     response.raise_for_status()
     logger.info(f"SBOM fetched for purl {purl}")
 
-    parse_pnc_sbom(response.json())
+    # Make sure the SBOM is valid
+    sbom = SbomerSbom(response.json()["sbom"])
```
juspence: Minor nit: let's break this onto two lines so it's easier to see which one fails in a traceback:
```suggestion
    # Make sure the SBOM is valid
    sbom_json = response.json()["sbom"]
    sbom = SbomerSbom(sbom_json)
```

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+
+    # Create a build for the root component
+    root_build, created = SoftwareBuild.objects.get_or_create(
+        build_id=sbom.components["root"]["meta_attr"]["pnc_build_id"],
+        build_type=SoftwareBuild.Type.PNC,
+    )
```
juspence, line 48: Do we ever expect two different SBOMer messages to refer to the same PNC build ID? Or is it possible for a build to be "rerun" or rebuilt again at some later date?

I think there was some discussion about this in Gchat but I don't remember the answer. I guess it's not necessarily a problem either way, it's just an edge-case we might need to consider.

Or if we want to enforce PNC build IDs are unique and we only process them once, we could do a raw create() instead of get_or_create(). Then we'd fail if that assumption was violated and the same build ID already existed. This might be risky if requirements change in the future and some Middleware team starts rerunning builds.

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+
+    # Create ProductComponentRelation if it doesn't already exist
+    if created:
+        pcr, _ = ProductComponentRelation.objects.get_or_create(
+            build_id=root_build.build_id,
+            build_type=SoftwareBuild.Type.PNC,
+            software_build__isnull=root_build,
```
juspence: Looks like a typo here? The __isnull= lookup usually wants a True or False value, but you're passing the related build object itself. Did you mean to just use a software_build= kwarg / field, without the extra __isnull lookup?

I also don't fully recall how the relations table is set up, but to avoid issues, we need to do get_or_create only with certain arguments. We have to use all the fields in a unique_together constraint, OR only one field that has unique=True.

If we don't do this, I think get_or_create() behaves unexpectedly (creating duplicate components, or giving weird results / other errors) because it doesn't end up using a database index for the query (typing from memory).

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+                defaults={
+                    "obj": components[dep],
+                },
+            )
+
+    # Save product taxonomy
+    root_build.save_product_taxonomy()
+
+    # Save component taxonomy
+    root_component.save_component_taxonomy()
```
juspence: There's a Celery task which does exactly this (slow_save_taxonomy() I think), which avoids duplicating that logic here and reduces the risk of timeouts for very large builds with lots of dependent components.

corgi/collectors/pnc.py
```diff
@@ -2,14 +2,71 @@
...
+                self.components[bomref]["meta_attr"]["pedigree"] = component["pedigree"]
 
-    return len(data["components"])
+        self.dependencies = {d["ref"]: d["dependsOn"] for d in data["dependencies"]}
```
juspence: Maybe a quick comment about components vs. dependencies would be good. I guess the first is top-level components, direcrly beneath the root, and the second is transitive dependencies?

And what would happen if we had the same component both in sbom.components (as its own top-level dependency X) and in sbom.dependencies (as a transitive dependency X of a different top-level dependency Y)? Do we handle this case correctly, do we need to do something special here, or is this guaranteed to never happen?

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+        if bomref == "root":
+            defaults["software_build"] = root_build
+
+        components[bomref], _ = Component.objects.update_or_create(
+            type=Component.Type.MAVEN,
+            name=component["name"],
+            version=component["version"],
+            release="",
+            arch="noarch",
+            defaults=defaults,
```
juspence: Again, I don't recall exactly which kwargs we need to use here. You've definitely got most of them, but maybe "namespace" is also part of the unique_together constraint.

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+        # created as a dependency of another component
+        if parent not in nodes:
+            nodes[parent], _ = ComponentNode.objects.get_or_create(
+                type=ComponentNode.ComponentNodeType.PROVIDES,
+                parent=nodes["root"],
+                purl=components[parent].purl,
+                defaults={
+                    "obj": components[parent],
+                },
+            )
```
juspence, line 108: I might have misunderstood part of this, but I see a few possible cases:

1. Parent not in nodes, but ComponentNode object exists - we will get() it. Does this ever happen?
2. Parent not in nodes, and ComponentNode object doesn't exist - we will create() it. This seems fine.
3. Parent in nodes, so we must have created it earlier in this loop, as a dependency of another component. Right now we don't do anything in this case, but should we update something instead?

If case number 1 above isn't expected, we could do a raw create() instead of get_or_create(). I'm not sure if failing whenever we'd reuse a node / some node with the same parent and purl already exists is better code. though. Maybe reusing the existing node is fine.

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+        defaults={"obj": root_component},
+    )
+
+    # Create the relationships recorded in the manifest's dependencies
+    for parent, deps in sbom.dependencies.items():
+        # Only create a new node for parent if it wasn't already
+        # created as a dependency of another component
+        if parent not in nodes:
+            nodes[parent], _ = ComponentNode.objects.get_or_create(
+                type=ComponentNode.ComponentNodeType.PROVIDES,
```
juspence, line 102: We always assume the nodes have PROVIDES type. Is there any information in the manifest for dev dependencies, AKA should we create some nodes with PROVIDES_DEV type instead?

On a similar note, does the root component (which I assume is always a Red Hat component) have any of its upstream components listed in the manifest? If yes, those upstream components should have nodes created with SOURCE type instead (but still be underneath / children of the root component).

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            build_type=SoftwareBuild.Type.PNC,
+            software_build__isnull=root_build,
+            defaults={
+                "type": ProductComponentRelation.Type.SBOMER,
+            },
+        )
+
+    # Create components
+    components = {}
+    for bomref, component in sbom.components.items():
```
juspence, line 61: I see we create components for all sbom.components. I also see we create nodes for all sbom.components and sbom.dependencies. Did we forget to create components for the sbom.dependencies, or did I misunderstand something?

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+        defaults = {"namespace": component["namespace"], "meta_attr": component["meta_attr"]}
+
+        if "related_url" in component:
+            defaults["related_url"] = component["related_url"]
+        if bomref == "root":
+            defaults["software_build"] = root_build
+
+        components[bomref], _ = Component.objects.update_or_create(
+            type=Component.Type.MAVEN,
```
juspence: This is a really silly question, but could we ever have a middleware component that isn't Maven / Java type? I'm imagining some Java program with a dependency that's actually an RPM. I assume the RPM must be installed on a system before running the actual middleware product / Maven component.

Maybe no such dependencies exist, or maybe those dependencies never show up in the manifest, so it's a moot point either way.

corgi/tasks/pnc.py
```diff
@@ -30,4 +36,82 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            release="",
+            arch="noarch",
+            defaults=defaults,
+        )
+
+    # Link dependencies
+    nodes = {}
+    # Create the root node from the SBOM's component
+    root_component = components.pop("root")
+    nodes["root"], _ = ComponentNode.objects.get_or_create(
```
juspence: I might be misremembering, but did some of the Quarkus people ever talk about the same root component being used for different builds?

Every child component in the tree below should never be duplicated - the parent node / root_component will always be unique even if two child nodes have the same type and purl / related component.

But the root component is a little special, since it has a null parent. If two builds have the same root component, the type (SOURCE) and purl / related component would be the same. So in that case, I think we'd end up merging the child component data for the two builds into the same ComponentNode tree.

I hope we receive sane enough data so that this is guaranteed to never happen. But maybe instead of get_or_create(), we should do a raw create() for the root component, or the root component's node, or both. Then we'd fail if the root component / node already existed. Or maybe that's wrong - we won't be able to reprocess any SBOMer messages without get_or_create().

Or maybe I'm way too concerned about random edge cases and this is fine :smile:

juspence (approved):
I know almost nothing about middleware so there are lots of silly questions here. Feel free to ignore them if they're very silly or don't make sense.

My biggest concern is probably get_or_create() statements. Maybe a raw create() is better, or maybe we should be doing update_or_create() instead. I'm not really sure what the expectation is for the data we're receiving or what kind of guarantees we have.

Overall, this LGTM, and thank you for working on it!

corgi/collectors/pnc.py
kgrant-rh: There are streams defined in prod defs, but the values provided by sbomer are ET values. Since the components are really only ever related to products through builds via ProductComponentRelations, these values might not even be necessary.

corgi/collectors/pnc.py
kgrant-rh: Yep, since component registry only supports one URL, this code only looks for one. There are other candidates that might make sense in the absence of a "website" type reference though, like VCS, issue trackers, mailing lists, or download locations. None of the provided purls have download_urls, but I should definitely try to find _something_ to use if website isn't available.

kgrant-rh: Or, since I'm already storing the VCS url in meta where it's available, would it be better to save each of the types of external references in meta, and then also use one of those as `related_url` if there's no `website`?

corgi/tasks/pnc.py
kgrant-rh: Good call. Maybe we should promote that task out of the brew module since it looks like it's used beyond brew.

corgi/tasks/pnc.py
kgrant-rh: sbom.components is a list of all included components, and sbom.dependencies is just a list of those components' relationship to each other.

corgi/collectors/pnc.py
juspence: Yeah, let's save everything to the meta_attr so we can clean up / add heuristics later if needed.

corgi/collectors/pnc.py
juspence: I think today we only support manifesting a stream. This was a (bad) result of some other decisions we made hastily a long time ago, which we never had time to go back and fix.

I have some commits locally that fix this, so we could make that change and start manifesting the middleware Variants directly. Or we could try to find some MW stream in prod-defs, and link it to these components just so they can be manifested from the stream. Whatever you prefer is fine.

EDIT: Oops, I misread your comment above. Yes, if we created all the relations and our Corgi taxonomy is correct, manifesting should just work (fingers crossed) and we shouldn't need to worry about these. Maybe they're still useful for debugging purposes - up to you.

corgi/tasks/pnc.py
juspence: Actually this is fine - these five fields type, name, version, release, and arch are exactly what we want. namespace is not part of the UniqueConstraint, so no need for any chages here.

corgi/tasks/pnc.py
kgrant-rh: You were right, this was a typo. Since I'm assuming we'll only process each build once, I changed this to just create.

corgi/tasks/pnc.py
kgrant-rh: I think builds are immutable, and if the same product/commit/etc. is rebuilt, it'll still result in a new build. The only situation I can think of is that we might want to reprocess a build we've already processed, but in that case it should be easy enough to manually delete the duplicate build.

corgi/tasks/pnc.py
kgrant-rh: In the context of this collector, we'll only get manifests which document PNC builds. In the manifest for Quarkus 2.13.8, there are two upstream components (protoc and protoc-gen-rpc-java) that are executables and not Java artifacts. It might be worth writing some corgi-wide function to infer component type from the purl type parameter.

kgrant-rh: Created https://issues.redhat.com/browse/CORGI-793 to follow up on this later.

corgi/tasks/pnc.py
kgrant-rh: I believe every PNC build should have a unique version, even if built from the same source. Nevertheless, I replaced this code with just `create` to ensure we don't overwrite or munge existing components.

corgi/tasks/pnc.py
kgrant-rh: The reason for the logic here is that the CycloneDX manifests we get from SBOMer list a component separately (distinct only in having different bomrefs) for each time that component is referenced in the dependency mapping. For example, if `foo` is a dependency of `bar` and `baz`, there will be *two* component entries for `foo`, one for each of those relationships.

Case number 1 happens when we've already created a ComponentNode which was indexed under a different bomref. Having to call `get_or_create` here isn't ideal, but I think an alternative like trying to merge bomrefs would end up being more fragile.

corgi/tasks/pnc.py
kgrant-rh: There's currently no distinction between PROVIDES and PROVIDES_DEV in the CycloneDX manifest. It's tough to say whether a distinction could be made, given Quarkus' nature as a development framework itself, but I'll see if we can more accurately identify those components.

Regarding upstreams, that's _also_ a tough question, since the root components are essentially just metadata files themselves.

corgi/tasks/pnc.py
kgrant-rh: I have created CORGI-796 to track the pedigree part of this comment, which I will work on immediately after merging this PR.

corgi/tasks/pnc.py
```diff
@@ -80,6 +80,8 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
             defaults=defaults,
         )
 
+        set_license_declared_safely(components[bomref], ";".join(component["licenses"]))
+
```
juspence, line 84: @kgrant-rh I missed your comment in the ticket, and you even asked in Gchat about joining the strings with a ; semicolon but I missed that too. Sorry :facepalm: 

This is fine for now, but if we know from the MW team that:
1. Customers can pick any of the licenses, we should join them with " OR "
2. Customers must comply with all of the licenses, we should join them with " AND "

If we're not sure or we don't always know one of the above, this is probably the best approach, although I'm worried identifiers separated by semicolons will cause validation issues in the online SPDX validator: https://tools.spdx.org/app/validate/

PR #522, opened 2023-07-06
Add management command to load errata matching variants

jasinner:
This allows us to bulk load errata for a Product Stream with a new errata_info entries, such as the recent product_definition changes for streams using yum repositories.

### DIFF
--- a/corgi/tasks/management/commands/loaderratadata.py
+++ b/corgi/tasks/management/commands/loaderratadata.py
@@ -2,6 +2,8 @@ import sys
 
 from django.core.management.base import BaseCommand, CommandParser
 
+from corgi.collectors.errata_tool import ErrataTool
+from corgi.collectors.models import CollectorErrataProductVariant
 from corgi.tasks.errata_tool import slow_load_errata
 from corgi.tasks.pulp import update_cdn_repo_channels
 
@@ -29,6 +31,9 @@ class Command(BaseCommand):
             action="store_true",
             help="Call tasks inline, not in a celery task",
         )
+        parser.add_argument(
+            "-p", "--product_variants", nargs="+", help="A list of variants to load errata for"
+        )
 
     def handle(self, *args, **options) -> None:
         if options["errata_ids"]:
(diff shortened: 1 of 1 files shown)


corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -45,6 +48,50 @@ def handle(self, *args, **options) -> None:
...
+            product_ids = set()
+            for et_variant in CollectorErrataProductVariant.objects.filter(
+                name__in=options["product_variants"]
+            ):
+                product_ids.add(et_variant.product_version.product.et_id)
+            # make sure there is only 1 product id for the names
+            if len(product_ids) > 1:
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"Variants had more than 1 product id. " f"Found: {product_ids}"
```
mprpic: Nitpick: you can combine these two strings. Black messed up the formatting here.

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -45,6 +48,50 @@ def handle(self, *args, **options) -> None:
...
+            self.stdout.write(
+                self.style.SUCCESS(f"Searching shipped errata for product with id {product_id}")
+            )
+            product_errata = et.get_paged(
+                f"api/v1/erratum/search?show_state_SHIPPED_LIVE=1&product[]={product_id}",
+                page_data_attr="data",
+            )
+            all_errata = set()
+            for errata in product_errata:
+                self.stdout.write(self.style.SUCCESS(f"Found shipped errata {errata['id']}"))
```
mprpic: Nitpick: erratum

mprpic (approved):
It might be overkill to search for all shipped errata for a particular Product since that may include thousands of errata for larger products like RHEL. Search is also possible by Releases, which then link to Product Version and those to Variants; this could narrow down the set of errata you'd be looking at but probably involves a few more API queries. I also don't know if we store Releases in our product taxonomy.

Otherwise LGTM!

jasinner:
> It might be overkill to search for all shipped errata for a particular Product since that may include thousands of errata for larger products like RHEL. Search is also possible by Releases, which then link to Product Version and those to Variants; this could narrow down the set of errata you'd be looking at but probably involves a few more API queries. I also don't know if we store Releases in our product taxonomy.
> 
> Otherwise LGTM!

Good idea. I'm going to merge as is for now as I don't need to run this for RHEL for now, and we don't store releases in prod_defs. I think it would be possible to iterate all ET Releases periodically and cache them locally in CollectorErrata models, related to ET ProductVersions. Then we could use the ET ProductVersions from errata_info  to narrow down which ET Releases to search for here.

corgi/tasks/management/commands/loaderratadata.py
jasinner: updated.

corgi/tasks/management/commands/loaderratadata.py
jasinner: fixed.

PR #525, opened 2023-07-09
CORGI-257 persist errata modules

jasinner:
The errata tool builds list requires extra processing to capture modular builds. This change looks for modular builds in the errata builds list and prepares the modular data for processing by the existing Brew.persist_module function. It also switches slow_load_errata to use slow_fetch_modular_build instead of slow_fetch_brew_build.

### DIFF
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -7,11 +7,13 @@ import requests
 from django.conf import settings
 from requests_gssapi import HTTPSPNEGOAuth
 
+from corgi.collectors.brew import Brew
 from corgi.collectors.models import (
     CollectorErrataProduct,
     CollectorErrataProductVariant,
     CollectorErrataProductVersion,
 )
+from corgi.core.models import SoftwareBuild
 
 logger = logging.getLogger(__name__)
 
@@ -149,11 +151,17 @@ class ErrataTool:
           }
         }
         """
-        variant_to_component_map = defaultdict(list)
+        variant_to_component_map: typing.DefaultDict[str, list] = defaultdict(list)
         builds = self.get(f"api/v1/erratum/{erratum_id}/builds_list.json")
+        brew = Brew(SoftwareBuild.Type.BREW)
         for product_version in builds.values():
             for build in product_version["builds"]:
                 for build_nvr, build_data in build.items():
+                    if build_data["is_module"]:
+                        self._parse_module(
+                            build_nvr, build_data["variant_arch"], brew, variant_to_component_map
+                        )
+                        continue
                     build_id = build_data["id"]
                     for variant, components_by_arch in build_data["variant_arch"].items():
                         variant_components = []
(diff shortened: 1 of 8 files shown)


jasinner:

kgrant-rh (approved):

corgi/collectors/errata_tool.py
```diff
@@ -225,3 +233,30 @@ def get_builds_for_errata(self, errata_id: int) -> list[int]:
...
                 build_ids.update(component.keys())
         return list(build_ids)
+
+    def _parse_module(
+        self,
+        module_name: str,
+        module_data: dict[str, dict[str, list[str]]],
+        brew: Brew,
+        variant_to_component_map: typing.DefaultDict[str, list],
+    ):
```
juspence: If this is returning None, let's specify that:
```suggestion
    ) -> None:
```

corgi/collectors/errata_tool.py
```diff
@@ -225,3 +233,30 @@ def get_builds_for_errata(self, errata_id: int) -> list[int]:
...
+                # persist_modules expects the .rpm suffix stripped
+                stripped_component = [
+                    rpm[: -len(".rpm")] for rpm in components if rpm.endswith(".rpm")
+                ]
+                modular_rpms.extend(stripped_component)
+            build_ids = brew.persist_modules({module_name: modular_rpms})
+            for build_id in build_ids:
+                # persist_modules returns a flat list of build_ids, so its no longer possible to
+                # map modular_rpms to build_ids. In slow_save_errata these are only saved as
+                # meta_attr so, let's just save the entire list for each build_id
```
juspence, line 262: I really don't know much about these ET data structures. Will this cause similar issues as the Variant-to-Component linking, where some modular RPM gets linked to every build instead of only some builds?

I think the comment about meta_attr means that we don't actually use this data, but I'm not really familiar with slow_save_errata or persist_modules, so I don't know what's getting returned, how it's saved, if it's used later, etc.

corgi/collectors/errata_tool.py
```diff
@@ -225,3 +233,30 @@ def get_builds_for_errata(self, errata_id: int) -> list[int]:
...
+        for variant, components_by_arch in module_data.items():
+            modular_rpms = []
+            for arch, components in components_by_arch.items():
+                # SRPMS are looked up in persist_modules from Brew via RPMs
+                if arch == "SRPMS":
+                    continue
+                # persist_modules expects the .rpm suffix stripped
+                stripped_component = [
+                    rpm[: -len(".rpm")] for rpm in components if rpm.endswith(".rpm")
+                ]
```
juspence: We always know the length of ".rpm", so we can hardcode it instead of calling len() on each loop iteration:
```suggestion
                stripped_components = tuple(
                    rpm[:-4] for rpm in components if rpm.endswith(".rpm")
                )
```

corgi/collectors/errata_tool.py
```diff
@@ -225,3 +233,30 @@ def get_builds_for_errata(self, errata_id: int) -> list[int]:
...
+                modular_rpms.extend(stripped_component)
```
```suggestion
                modular_rpms.extend(stripped_components)
```

juspence (approved):
LGTM, question about linking is minor and can be ignored if I'm confused.

corgi/collectors/errata_tool.py
jasinner: When we call slow_fetch_modular build modular RPM get linked to their correct SRPMs by looking up the relationships from the Collector models. Actually I now realize we could store the correct relationships, but we'd have to lookup the details from the Collector models. It seems like overkill to do that just to save them in meta_attr though. We don't use that data anywhere yet. At least having some reference to every build nvr in meta_attr will give us a clue of finding the mapping in collector models if we need it later.

corgi/collectors/errata_tool.py
jasinner: ok.

corgi/collectors/errata_tool.py
jasinner: updated

corgi/collectors/errata_tool.py
jasinner: thanks, fixed

corgi/collectors/errata_tool.py
```diff
@@ -225,3 +233,31 @@ def get_builds_for_errata(self, errata_id: int) -> list[int]:
...
+            modular_rpms = []
+            for arch, components in components_by_arch.items():
+                # SRPMS are looked up in persist_modules from Brew via RPMs
+                if arch == "SRPMS":
+                    continue
+                # persist_modules expects the .rpm suffix stripped
+                rpm_suffix_len = ".rpm"
+                stripped_components = [
+                    rpm[: -len(rpm_suffix_len)] for rpm in components if rpm.endswith(".rpm")
+                ]
```
juspence, line 256: Did you mean to define this at the top of the function, and call len(".rpm")? Right now it's still calling len on every loop iteration. Not a big deal if you did it this way on purpose, but just wanted to make sure you didn't forget to push a commit before merging.

PR #528, opened 2023-07-10
CORGI-730: Fix slow query when deleting builds

juspence:
Just a draft for now, we need to fix a bug first before merging this. Some GENERIC components are created as root components and have no sources / upstreams. The new logic here will delete those components as well, which we don't want.

Once that data is fixed (all generic components are linked correctly to their root components), we can assume that all non-root / child components will have either non-null sources (are provided components) or non-null downstreams (are upstream components).

Any non-root / child components with both null sources and null upstreams, are either provided components or upstream components. They are not linked to their source / downstream root component anymore because that root component (and its build) just got deleted. We should be able to safely clean them up, but I need to confirm we don't have other edge cases in our data.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -6,12 +6,13 @@ from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
 from django.db import transaction
-from django.db.models import Count, Q, QuerySet
+from django.db.models import Q, QuerySet
 from django.utils import dateformat, dateparse, timezone
 from django.utils.timezone import make_aware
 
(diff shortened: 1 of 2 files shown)


juspence:
The new query I wrote here is also slow...because there are many components (usually from Syft, sometimes from the Brew collector) that don't have any sources (so not a provided component) and don't have any downstreams (so not an upstream component).

One component that I checked also didn't have any provides / upstreams of its own, and didn't even have any cnodes. I'm not sure how or why Syft is creating components that don't participate in any tree, but maybe that's a legacy bug we already fixed without any data cleanup.

PR #529, opened 2023-07-11
CORGI-746 - refactor latest component filter into pg stored proc

JimFuller-RedHat:


corgi/core/migrations/0075_install_stored_proc.py
```diff
@@ -0,0 +1,265 @@
...
+    latest_version text;
+    compare_versions int;
+  BEGIN
+    IF model_type = 'ProductStream' THEN
+      FOR COMPONENT IN select *, row_number() over () from core_component
+           INNER JOIN "core_component_productstreams"
+           ON ("core_component"."uuid" = "core_component_productstreams"."component_id")
+           INNER JOIN "core_productstream"
+           ON ("core_component_productstreams"."productstream_id" = "core_productstream"."uuid")
+           where core_component.namespace=component_ns and core_component.name=component_name and ( (core_component.type='RPM' and core_component.arch='src') or (core_component.type='OCI' and core_component.arch='noarch'))
```
jasinner, line 189: I think we might not need this part because that's already done as part of the root_components filter, which we are still using previous to using this stored procedure. That also gives us the flexibility to use this stored procedure to filter things which are not root components.

JimFuller-RedHat: sounds reasonable ;) will remove to make more generic

JimFuller-RedHat:
will need to adjust for type (as per https://github.com/RedHatProductSecurity/component-registry/pull/562)

corgi/core/migrations/0075_install_stored_proc.py
jasinner: Maybe it does make sense to include this condition in the stored procedure, perhaps adding RPMMOD as well. The current root component condition includes Github components for managed serviced, which are unlikely to have version matching RPM schematics.

JimFuller-RedHat: yes I need to refresh it as there has been some changes - will do

PR #532, opened 2023-07-12
fix a bug where we passed the wrong type to slow_load_errata

jasinner:
This change also makes the command slightly more efficient by loading erratum straight away, rather than waiting for the the check off all product errata to complete

### DIFF
--- a/corgi/tasks/management/commands/loaderratadata.py
+++ b/corgi/tasks/management/commands/loaderratadata.py
@@ -87,7 +87,7 @@ class Command(BaseCommand):
                                             f"Found variant {variant} in {erratum['id']}"
                                         )
                                     )
-                                    all_errata.add(erratum["id"])
+                                    all_errata.add(str(erratum["id"]))
 
             for erratum in all_errata:
                 self.stdout.write(self.style.SUCCESS(f"Loading erratum: {erratum}"))


jasinner:

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -73,26 +73,28 @@ def handle(self, *args, **options) -> None:
...
-            for erratum in all_errata:
-                self.stdout.write(self.style.SUCCESS(f"Loading erratum: {erratum}"))
-                slow_load_errata.delay(erratum)
+            self.load_errata_with_variants(et, options["product_variants"], product_errata)
 
         else:
             self.stderr.write(self.style.ERROR("No errata IDs or repo flag, or variants specified"))
             sys.exit(1)
+
+    def load_errata_with_variants(self, et, product_variants, product_errata):
```
juspence: When we add new functions, let's always add type hints (including -> None for the return type if needed) and a header comment.

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -73,26 +73,28 @@ def handle(self, *args, **options) -> None:
...
+                            if variant in product_variants:
+                                self.stdout.write(
+                                    self.style.SUCCESS(
+                                        f"Found variant {variant} in {erratum['id']}"
+                                    )
+                                )
+                                has_variants = True
+            if has_variants:
+                self.stdout.write(self.style.SUCCESS(f"Loading erratum: {erratum['id']}"))
+                slow_load_errata.delay(str(erratum["id"]))
```
juspence: If you're trying to make the command load the errata sooner, I think you need to break out of each loop after setting the flag:
```suggestion
        for erratum in product_errata:
            has_variants = False
            self.stdout.write(self.style.SUCCESS(f"Found shipped erratum {erratum['id']}"))
            builds = et.get(f"api/v1/erratum/{erratum['id']}/builds_list.json")
            for product_version in builds.values():
                for build in product_version["builds"]:
                    for _, build_data in build.items():
                        for variant in build_data["variant_arch"].keys():
                            if variant in product_variants:
                                self.stdout.write(
                                    self.style.SUCCESS(
                                        f"Found variant {variant} in {erratum['id']}"
                                    )
                                )
                                has_variants = True
                                break  # from build_data loop
                        if has_variants:
                            break  # from build.items() loop 
                    if has_variants:
                        break  # from product_version loop
                if has_variants:
                    break  # from builds.values() loop

            # don't break from erratum loop
            if has_variants:
                self.stdout.write(self.style.SUCCESS(f"Loading erratum: {erratum['id']}"))
                slow_load_errata.delay(str(erratum["id"]))
```
I also think adding the erratum ID to a set is a good idea, since it prevents us from scheduling a Celery task for the same errata twice. Even though we use base=Singleton on that task (I think), if the first task finishes very quickly then a second task for the same erratum could still be scheduled.

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -73,26 +73,28 @@ def handle(self, *args, **options) -> None:
...
             sys.exit(1)
+
+    def load_errata_with_variants(self, et, product_variants, product_errata):
+        for erratum in product_errata:
+            has_variants = False
+            self.stdout.write(self.style.SUCCESS(f"Found shipped erratum {erratum['id']}"))
+            builds = et.get(f"api/v1/erratum/{erratum['id']}/builds_list.json")
+            for product_version in builds.values():
+                for build in product_version["builds"]:
+                    for _, build_data in build.items():
```
juspence: If we don't need the key value here, let's use build.values() instead of build.items():
```suggestion
                    for build_data in build.values():
```

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -73,26 +73,28 @@ def handle(self, *args, **options) -> None:
...
+                        for variant in build_data["variant_arch"].keys():
+                            if variant in product_variants:
```
juspence: I guess based on the old code, product_variants AKA options["product_variants"] is a list the user passes. That means this "if item in list" check has to iterate over every single value in the list, which will be slow.

If you convert this variable to a set (at the top of this fuction, not inside the for loop), the "if item in set" check can avoid iterating due to Python implementation details. "if item in set" and "if key in dict" both return a result instantly, without needing to iterate over the collection and check every value. So using "if item in set" is much faster for large collections.

juspence (approved):

corgi/tasks/management/commands/loaderratadata.py
jasinner: I reverted to building a set and loading them all after the errata loop.

corgi/tasks/management/commands/loaderratadata.py
jasinner: This was existing code, leaving it for later.

corgi/tasks/management/commands/loaderratadata.py
jasinner: Reverted to the old way of building a set and loading them after the errata loop.

corgi/tasks/management/commands/loaderratadata.py
jasinner: I was doing this quickly and just made use of the pycharm extract method feature in order to keep sonarcube happy. Anyway reverted now.

PR #533, opened 2023-07-12
CORGI-747 - add exclude_components to product stream

JimFuller-RedHat:
Extracts from product definitions the exclude_components field which can then be used downstream (ex. Griffon).


corgi/core/models.py
```diff
@@ -586,6 +586,8 @@ class ProductStream(ProductModel):
     active = models.BooleanField(default=False)
     et_product_versions = fields.ArrayField(models.CharField(max_length=200), default=list)
 
+    exclude_components = models.JSONField(default=dict)
+
```
juspence: Not familiar with the real prod-defs data, but based on the example in your test, can we use a simpler ArrayField(TextField) here instead? Or if the real data is more complicated / full JSON, let's make the test data match.

tests/test_api.py
```diff
@@ -1457,3 +1457,13 @@ def test_srpm_component_provides_sources_upstreams(client, api_path):
...
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_product_streams_exclude_components(client, api_path):
+    stream = ProductStreamFactory(exclude_components=["golang"])
+
+    response = client.get(f"{api_path}/product_streams?name={stream.name}")
+    assert response.status_code == 200
+    assert len(response.json()["exclude_components"]) == 1
+    assert response.json()["exclude_components"] == ["golang"]
```
juspence: Let's do `response = response.json()` once and save the result. We also don't need to assert the len() of the value if we're asserting the exact contents of the value:
```suggestion
    assert response.status_code == 200
    # It's easier to understand any traceback / errors if the response.json() conversion is on its own line
    response = response.json()
    assert response["exclude_components"] == ["golang"]
```

juspence (approved):
LGTM!

EDIT: Ah, I see you changed this to a draft and removed the review request. Feel free to tag me again when this is ready to review.

corgi/core/models.py
JimFuller-RedHat: just following status quo

tests/test_api.py
JimFuller-RedHat: will do

corgi/core/models.py
juspence: Not sure I understand - do we expect to receive full JSON values / dicts with subkeys from prod-defs, like `{"some_component": ["subcomponent_1", "subcomponent_2", "etc"]}`? In this case, let's make the test data match the real values we get.

Or are we only receiving simple lists of string component names, like `["component_1", "component_2", "etc"]`? In this case, let's use an ArrayField instead:
```suggestion
    exclude_components = fields.ArrayField(models.CharField(max_length=200), default=list)

```

JimFuller-RedHat: thx - useful conversation - changed it to use fields.ArrayField as you proposed with the caveat that hopefully some day we will have a 'plan' for product definition.

jasinner (approved):
LGTM

PR #534, opened 2023-07-13
CORGI-738 match more variants via brew tags

jasinner:
This allows matches between variants and streams for extra streams including openshift-4.13.z as in the description of CORGI-738.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -186,9 +186,14 @@ def parse_variants_from_brew_tags(
         )
         for brew_tag in brew_tags.keys():
             # Also match brew tags in prod_defs with those from ET
-            trimmed_brew_tag = brew_tag.removesuffix("-released")
+            # Brew tags in ET
+            trimmed_brew_tag = brew_tag.removesuffix("-candidate")
+            trimmed_brew_tag = trimmed_brew_tag.removesuffix("-released")
+            # This is a special case for 'rhaos-4-*` brew tags which don't have the -container
+            # suffix in ET, but do have that suffix in product_definitions
+            sans_container_released = brew_tag.removesuffix("-container-released")
             et_pvs = CollectorErrataProductVersion.objects.filter(
-                brew_tags__contains=[trimmed_brew_tag]
+                brew_tags__overlap=[trimmed_brew_tag, sans_container_released]
             )
 
             for et_pv in et_pvs:
--- a/tests/test_products.py
+++ b/tests/test_products.py
@@ -111,21 +111,3 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
     assert len(sat_610_variants) == 2
     for variant in sat_610_variants:
         assert variant.name in ("7Server-Capsule610", "7Server-Satellite610")
-
-
-def _create_builds_and_components(stream):
-    stream_build = SoftwareBuildFactory()
-    stream_component = ComponentFactory(software_build=stream_build)
-    stream_component_node = ComponentNode.objects.create(
-        parent=None, obj=stream_component, type=ComponentNode.ComponentNodeType.SOURCE
-    )
-    stream_child_component = ComponentFactory()
-    ComponentNode.objects.create(
-        parent=stream_component_node,
-        obj=stream_child_component,
-        type=ComponentNode.ComponentNodeType.PROVIDES,
-    )
-    stream_component.productstreams.add(stream)
-    stream_component.productversions.add(stream.productversions)
-    stream_component.products.add(stream.products)
-    return stream_build, stream_component, stream_child_component
--- a/tests/test_products.py
+++ b/tests/test_products.py
@@ -8,9 +8,8 @@ from corgi.collectors.models import (
     CollectorErrataProductVariant,
     CollectorErrataProductVersion,
 )
-from corgi.core.models import ComponentNode, Product, ProductStream
+from corgi.core.models import Product, ProductStream
 from corgi.tasks.prod_defs import update_products
-from tests.factories import ComponentFactory, SoftwareBuildFactory
 
 pytestmark = pytest.mark.unit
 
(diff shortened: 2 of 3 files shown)


jasinner:
@RedHatProductSecurity/corgi-devs please review. I added some details of comparing stream and variants from supported_cpe_map to CORGI-738

corgi/tasks/prod_defs.py
```diff
@@ -186,10 +186,19 @@ def parse_variants_from_brew_tags(
         )
         for brew_tag in brew_tags.keys():
             # Also match brew tags in prod_defs with those from ET
-            trimmed_brew_tag = brew_tag.removesuffix("-released")
+            tag_no_candidate = brew_tag.removesuffix("-candidate")
+            tag_no_released = brew_tag.removesuffix("-released")
```
juspence: Can we just have one trimmed_brew_tag that we need to check for each original brew tag that we process? Like below:
```suggestion
            trimmed_brew_tag = brew_tag.removesuffix("-container-released")
            trimmed_brew_tag = trimmed_brew_tag.removesuffix("-candidate")
            trimmed_brew_tag = trimmed_brew_tag.removesuffix("-released")
            # And so on, for any other suffixes we want to remove
```
I'm assuming here that either a tag doesn't have multiple suffixes, like "-candidate-released" or "-released-candidate". Or if a tag does have multiple suffixes, we probably want to strip all of them, right?

corgi/tasks/prod_defs.py
```diff
@@ -186,10 +186,19 @@ def parse_variants_from_brew_tags(
...
+                brew_tags__overlap=[tag_no_released, tag_no_candidate]
             )
+            # Using icontains on an arrayfield combines the array values together before doing a
+            # substring match. Therefore, only do it as a last resort if we couldn't match with
+            # overlap above
+            if not et_pvs:
+                tag_no_container_released = brew_tag.removesuffix("-container-released")
+                et_pvs = CollectorErrataProductVersion.objects.filter(
+                    brew_tags__icontains=tag_no_container_released
+                )
```
juspence: I'm not sure why we need an icontains here to check all the arrayfield values together. Usually this is needed for substring matching.

For example, say you have an ArrayField like `["rhel", "ocp", "ansible"]`. We can easily check if some value is an exact match for one of the items in the ArrayField. This would use `arrayfield__contains="rhel" or "ocp" or "ansible"`.

But if you want to find a substring (not an exact match) of some item in the ArrayField, `arrayfield__contains="e"` won't work. You have to use `arrayfield__icontains="e"` instead, and this would return `["rhel", "ansible"]`.

So can't we just check if the whole set of brew_tags contains an exact match for tag_no_container_released, AKA a tag like "product-name" without its "-container-released" suffix?

Or are we also trying to match some other tag which looks like "product-name-special-case", where "product-name" without the "container-released" suffix is only a substring?

tests/test_products.py
```diff
@@ -113,19 +112,30 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
...
 
-def _create_builds_and_components(stream):
-    stream_build = SoftwareBuildFactory()
-    stream_component = ComponentFactory(software_build=stream_build)
-    stream_component_node = ComponentNode.objects.create(
-        parent=None, obj=stream_component, type=ComponentNode.ComponentNodeType.SOURCE
+brew_tag_streams = [
+    ("8Base-GitOps-1.7", "gitops-1.7-rhel-8", "gitops-1.7"),
+    ("OSE-4.10-RHEL-8", "rhaos-4.10-rhel-8-candidate", "openshift-4.10.z"),
+    ("RHACM-2.4", "rhacm-2.4-rhel-7-container-candidate", "rhacm-2.4.z"),
```
juspence: I don't see any tests for the "-container-released" case, but I see you do have a test for a stream with a "rhel-7-container-candidate" tag. Right now it looks like we only strip "-container-released" and not "-container-candidate".

So for a tag like "rhel-7-container-candidate", we end up stripping only "-candidate" and checking values like "rhel-7-container" with no "-candidate" suffix.

But for a tag like "rhel-7-container-released", we'd strip both "-released" and "-container-released". Then we'd check values like both "rhel-7-container" with no "-released" suffix, and "rhel-7" with no "-container-released" suffix.

If we need to start stripping the extra "-container" bit, let's do that consistently for both "-container-released" and "-container-candidate".

Or if there's some reason why stripping "-container-released" is OK, but stripping "-container-candidate" will cause problems, let's document that and add a test.

corgi/tasks/prod_defs.py
jasinner: You are correct, we don't need icontains here. I'm going to build 2 tags for matching with overlap instead. One without `-candidate` or `-released`, and another without `-container-released`. While we could also strip `-container` from the `trimmed_brew_tag` and it wouldn't cause any issues with matching today (as far as I am aware), I would rather make it more future proof by only allowing the special case of `rhaos-4.*` now.

corgi/tasks/prod_defs.py
jasinner: The special case is `rhaos-4.10-rhel-8` which has the brew tag 'rhaos-4.10-rhel-8-candidate' in ET product version `{ET_URL}/products/79/product_versions/1625`, but we want to match it against builds with the tag 'rhaos-4.10-rhel-8-container-released' (without the whole `container-released` suffix.)

If we always strip `container-released` first however that won't match the `rhacm-2.4-rhel7` case which we need to strip the `-released` suffix only to match what we have in Collector models `rhacm-2.4-rhel7-container`.

tests/test_products.py
jasinner: We strip '-candidate' from ET Product Version brew tags before persisting them. So I should have stripped that suffix here also. In this test the brew_tag parameter is used to build a CollectorErrataProductVersion. I'll fix that now.

The current test for rhacm-2.4.z stream is a match for `-container-candidate` suffix. If we change update_products to always strip the `-container` suffix that test will fail.

corgi/tasks/prod_defs.py
juspence: Ack, thanks for giving that example! This makes sense and LGTM. It's a shame our data is this complicated but oh well :shrug:

juspence (approved):
LGTM, thanks for explaining. I am having a crazy Monday so apologies in advance if I missed something.

PR #536, opened 2023-07-13
CORGI-727 / CORGI-728: Remove Component-Channel links that cause issues for IR

juspence:
A long time ago, I wrote code to link Components to their Channels since we linked Components to other models, e.g. ProductVariants. But not creating the Channel links was actually the intended behavior and not an omission. We can't identify which Channel (there can be many) actually ships the Component.

Previously we linked all the Channels for all the Variants to the Component. We also had a related bug where all sibling Variants of one Variant (for a Variant-type relation), or all child Variants of one Stream (for a Stream-type relation) were linked to the Component. We should have linked the Component to either one Variant or no Variants at all. So all of these bugs together caused us to create many Channel links at the bottom of this hierarchy.

This is now fixed and documented in code. We no longer create these links for new builds, and we have CORGI-298 to find a more accurate way to link Channels to Components. Trying to clear the links for existing builds / components was also attempted previously, but timed out due to slowness.

Deleting and readding the ManyToManyField should drop and recreate the Component_channels through table it relies on. I hope this will be faster then trying to delete all 293 million rows in the table. Again, this won't cause issues (it will solve them) because we don't need / want this data anyway.


juspence:
@RedHatProductSecurity/corgi-devs FYI, going to just merge this since we've already discussed / approved related changes. I'll back out the migration if it's too slow or causes other issues (again :cry:).

PR #538, opened 2023-07-13
CORGI-748 revert buggy code for moving variant between streams

jasinner:
Seems this code has caused corruption of the ProductNode tree for at least the `7Client-RH7-RHOS-TOOLS-13.0` variant. We're also seeing many variants being moved back and forth between the same streams, and in other cases added and removed to the same stream.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -1,5 +1,5 @@
 import re
-from typing import Any, Optional
+from typing import Any
 
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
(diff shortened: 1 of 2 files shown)


corgi/tasks/prod_defs.py
```diff
@@ -268,84 +208,28 @@ def parse_variants_from_brew_tags(
...
                                 "et_product": variant_product_version.product.name,
                                 "et_product_version": variant_product_version.name,
                             },
-                        )
-                        variant_created = True
-
-                    _, node_created = ProductNode.objects.update_or_create(
+                        },
+                    )
+                    ProductNode.objects.get_or_create(
```
juspence, line 224: Looks good to me, but let's leave the `product_variant.save_product_taxonomy()` call that was here, just in case. I think it should be a no-op, but :shrug:

juspence (approved):

corgi/tasks/prod_defs.py
jasinner: ok, adding it back.

PR #539, opened 2023-07-14
CORGI-727 / CORGI-728: Fix component variant links

juspence:
@RedHatProductSecurity/corgi-devs Please review - I don't know if this is the right solution, but I do know this will work (with some side effects).

Some old builds had their relations processed, and incorrect component-to-variant links created, before my earlier bugfix in 727 / 728. To fix this data, we tried to clear all the links (including incorrect ones) for all components in some build, then call build.save_product_taxonomy() to add back only the correct links for that build's components.

Selectively fixing the data like this just takes too much time, especially given the slowness in our taxonomy-saving methods. This is true even after deleting all the Component-Channel links in a different migration and updating this one to do less work. There are 153,000 distinct builds with ERRATA-type relations, and that number will only grow over time. In 8 hours we only processed about 2500.

Instead, I could apply the same solution for Component-Variant links as I did for Component-Channel links - drop the entire table and recreate it. These two cases are a little different, since we don't want any Channel links / all of them were incorrect, but we do need some Variant links / only a few are incorrect.

For stream-type relations like BREW_TAG or YUM_REPO, we don't want any Component-Variant links for some build. For variant-type relations (only ERRATA type), we only want Component-Variant links for that Variant, and not all its sibling Variants that share a parent Stream. But identifying which builds / components / etc. need to be linked to which variants is too slow.

If we clear all Component-Variant links, then all builds with ERRATA-type relations will need to have build.save_product_taxonomy() called again. Until the taxonomy is saved again, all the components in that build will be missing links to that Variant. I don't know if this is OK, or if missing data will cause even more problems for IR than wrong data.


jasinner (approved):
LGTM.

I think if we called save_product_taxonomy on the latest builds in streams with ERRATA-type relation relations and do it over the weekend that should reduce the time we have missing builds in those streams to an acceptable level.

If we just try to call save_product_taxonomy all ERRATA-type relations that might take too long in production for IR I think. IR usually only search the latest builds in a stream for a target component and don't usually care about non-latest builds.

juspence:
Ack, I'm going to wait to merge this until https://github.com/RedHatProductSecurity/component-registry/pull/540 is also merged. Then calling save_product_taxonomy() for the latest builds should be faster.

EDIT: That other change didn't work, so I'll merge this later once Keith is finished debugging the UMB issue and we've processed all the backlogged messages. Once it's deployed, I'll trigger save_product_taxonomy() for all the builds that need it either manually (probably on Friday) or automatically with slow_fetch_brew_build() and force_process=False.

PR #540, opened 2023-07-14
CORGI-739: Try to make save_product_taxonomy() and save_component_taxonomy() faster

juspence:
@RedHatProductSecurity/corgi-devs Please review. Right now we see timeouts every day when loading builds, because the save_product_taxonomy() and save_component_taxonomy() code is very slow. Some of this is due to data size, when there are many descendant nodes / related components for one build / root component.

But some of the slowness may also be due to GenericForeignKey, which requires two DB queries when accessing each linked object. This is because you have to look up the content_type / "which table" first, then access the referenced object_id in the table you looked up.

Since ComponentNodes always refer to Components, we can just use a regular ForeignKey instead which avoids the content_type lookup and the related overhead (we always know which table the ForeignKey is pointing to). Other advantages include:

1. Being able to access the .cnodes property of a Component in a migration. We can't do this with GenericForeignKey because (I think) the content_type tables aren't available during migrations.

2. Being able to control the on_delete= behavior. Right now we're forced to use on_delete=models.CASCADE because we define a GenericRelation from Component back to ComponentNode. This GenericRelation is a hack to work around some GenericForeignKey behavior, and forces the cascade deletion behavior.

3. Being able to use both ComponentNode.objects.filter(component__prop=value) and ComponentNode.objects.exclude(component__prop=value). GenericForeignKey doesn't support either of these by default, and adding a GenericRelation only allows using .filter() and still doesn't work with .exclude().

4. Being able to use .select_related() instead of .prefetch_related(). These methods let you avoid making 1 DB query per node when accessing each node's linked Component. prefetch_related() relies on prefetching the list of objects and doing a join in Python, instead of querying for each related Component individually. This can cause performance problems in the DB (SELECT ... IN ... queries with a large list of nodes) or in Python (risking out-of-memory errors / pod restarts) when the taxonomy / number of nodes is very large.

select_related() selects all the related Components and JOINs them to the nodes in pure SQL / the DB, which should be faster and definitely avoids the OoM risk in our Python process. But I don't know how it will perform in practice. Currently we stopped using prefetch_related() due to the issues mentioned above, so iterating over a large tree with 2000 nodes requires 4001 DB calls: 1 to get the overall list of nodes, 2000 to get each node's content type, and 2000 to get each node's linked object based on the content type.

After this change, we should be making 2001 DB calls: 1 to get the overall list of nodes, 2000 to get each node's linked object which we know is always a Component. If we also add select_related(), we can get the whole tree (all nodes and their related objects) in one query.

5. Not having to create custom indexes for the reverse relation / the set of all nodes linked to some Component. Foreign keys automatically have the needed indexes created by default, but GenericForeignKeys don't and require us to manually create these. If I messed up when adding these indexes manually, that might explain why iterating over a taxonomy slows down in some cases.

6. Fixing some stupid mypy errors. We never create nodes without Components, but as a side-effect of all of the above, mypy can't determine that a node's related Component exists or not. It assigns the "Optional[Any]" type to the linked node.obj and can't do any useful type-checking. 

The first commit here is just general cleanup of some nits in tests and code. The last two commits add the ForeignKey from ComponentNode to Component, then remove all usages of the GenericForeignKey (object_id and obj properties) in favor of the regular ForeignKey (component_id and component properties). I'd like to deploy this to stage Monday and see if it reduces the number of failures when loading large builds.

If this causes issues, I can revert these changes by just dropping the ForeignKey and switching usages back to the GenericForeignKey instead. Or if it works well, I could try to speed things up by adding select_related() to our taxonomy-parsing code, then drop the old GenericForeignKey now that it's unused. I already have commits to do this locally, I just need to confirm the regular ForeignKey doesn't cause issues by itself in stage before I tweak it further.


corgi/tasks/managed_services.py
```diff
@@ -93,7 +93,6 @@ def cpu_manifest_service(product_stream_id: str, service_components: list):
             root_node = ComponentNode.objects.create(
                 type=ComponentNode.ComponentNodeType.SOURCE,
                 parent=None,
-                purl=root_component.purl,
```
jasinner, line 96: We should probably remove the setting of purl attribute in other places too to keep thing consistent. I don't think we need it since ComponentNode.save() always set it the the component.purl anyway.

jasinner (approved):
LGTM, let's try it!

corgi/tasks/managed_services.py
juspence: I don't see any other usages of a purl= keyword argument when creating an object. I do see a purl= keyword argument when doing get_or_create(), which is still needed to make get_or_create behave atomically.

juspence:
I see the BackendStoreErrors have all disappeared after we removed the Brew build deleting code, which used up all temporary storage space that other queries needed to run successfully.

But we're still having issues with UMB consumers not receiving any messages, so I'm going to wait to merge this until that issue is resolved.

PR #545, opened 2023-07-17
Revert UMBDispatcher

kgrant-rh:
Per CORGI-753, there seems to be a problem with the UMBDispatcher class not handling messages. This PR reverts to the previous method of handling UMB messages.


jasinner (approved):
LGTM

PR #547, opened 2023-07-18
CORGI-739: Add nullable Component ForeignKey on ComponentNode

juspence:
@jasinner This will add just the nullable ForeignKey field, and not try to set the links for all existing ComponentNode / Component pairs.

That will definitely be much faster, but we won't be able to start using the field until we fix all the data, or delete all the data and prepare to reload.


jasinner:
As discussed. Let's revisit updating the FK from GFK at a later time so we can focus on other more important issues. The save_*_taxonomy functions seem to be fast enough for now.

PR #548, opened 2023-07-19
Add authentication to AppInterface request

mprpic:
See also ops repo MR 174.

Resolves: CORGI-734

### DIFF
--- a/corgi/collectors/app_interface.py
+++ b/corgi/collectors/app_interface.py
@@ -42,6 +42,7 @@ class AppInterface:
         """
         response = requests.post(
             f"{settings.APP_INTERFACE_URL}/graphql",
+            auth=(settings.APP_INTERFACE_USERNAME, settings.APP_INTERFACE_PASSWORD),
             json={"query": repo_query},
         )
         response.raise_for_status()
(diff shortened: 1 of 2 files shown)


mprpic:

corgi/collectors/app_interface.py
```diff
@@ -42,6 +42,7 @@ def fetch_service_metadata(
         """
         response = requests.post(
             f"{settings.APP_INTERFACE_URL}/graphql",
+            auth=(settings.APP_INTERFACE_USERNAME, settings.APP_INTERFACE_PASSWORD),
```
juspence, line 45: Does app interface support any kind of cert-based auth or auth with Kerberos tickets? Lots of other RH services seem to be deprecating simple username / password-based auth interfaces to meet the ESS requirements.

It's a pain to have to switch these later or watch stuff break, so if possible I'd rather just jump straight to an auth option that's (hopefully) going to hang around for a while.

juspence (approved):
LGTM!

corgi/collectors/app_interface.py
mprpic: No other option exists.

PR #549, opened 2023-07-19
Combine all UMB handling into one listener and one handler

kgrant-rh:
Combines address specification and handlers into two classes: one which specifies addresses, and one which contains handler methods.

### DIFF
--- a/corgi/monitor/consumer.py
+++ b/corgi/monitor/consumer.py
@@ -1,6 +1,5 @@
 import json
 import logging
-from typing import Optional
 
 from django.conf import settings
 from proton import Event, SSLDomain
@@ -9,6 +8,7 @@ from proton.reactor import Container, Selector
 
 from corgi.tasks.brew import slow_fetch_brew_build, slow_update_brew_tags
 from corgi.tasks.errata_tool import slow_handle_shipped_errata
+from corgi.tasks.pnc import slow_fetch_pnc_sbom
 
 logger = logging.getLogger(__name__)
 
@@ -83,11 +83,10 @@ class UMBReceiverHandler(MessagingHandler):
                 f"UMB event {event.message.id} had unrecognized address: {event.message.address}"
             )
 
-
-class BrewUMBReceiverHandler(UMBReceiverHandler):
-    """Handle messages about completed Brew builds, tagged builds, and untagged builds"""
-
-    def handle_builds(self, event: Event) -> None:
+    ##########################
+    # Message Handlers: Brew #
+    ##########################
+    def brew_builds(self, event: Event) -> None:
         """Handle messages about completed Brew builds"""
         logger.info("Handling UMB event for completed builds: %s", event.message.id)
         message = json.loads(event.message.body)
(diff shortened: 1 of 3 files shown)


kgrant-rh:
@RedHatProductSecurity/corgi-devs Per our discussion yesterday, this change makes it easier to add new topics and handlers to our UMB handlers without drastic changes to the relationship of the classes involved.

corgi/monitor/consumer.py
```diff
@@ -136,72 +165,52 @@ def handle_shipped_errata(self, event: Event) -> None:
...
     @classmethod
     def consume(cls):
         """Run a single message handler, which can listen to multiple virtual topic addresses"""
-        if not cls.handler_class or not cls.virtual_topic_addresses:
-            raise NotImplementedError(
-                "Subclass must define handler class and virtual topic address(es)"
-            )
-
         logger.info("Starting consumer for virtual topic(s): %s", cls.virtual_topic_addresses)
         handler = cls.handler_class(
```
juspence: Since there are no subclasses anymore, you could delete the handler_class attribute that's defined above and just call UMBReceiverHandler() directly here.

corgi/monitor/consumer.py
```diff
@@ -108,7 +107,37 @@ def handle_builds(self, event: Event) -> None:
...
+                "Failed to schedule slow_update_brew_tags task for build ID %s: %s",
+                build_id,
+                str(exc),
+            )
+            # Release message back to the queue but report back that it was delivered. The
+            # message will be re-delivered to any available client again.
+            self.release(event.delivery, delivered=True)
+        else:
+            # Accept the delivered message to remove it from the queue.
+            self.accept(event.delivery)
```
juspence: No need to change this code if you prefer this form. But you could copy some of the logic from your other PR, and return a bool from these handler functions:
```suggestion
        except Exception as exc:
            logger.error(
                "Failed to schedule slow_update_brew_tags task for build ID %s: %s",
                build_id,
                str(exc),
            )
            return False
        else:
            return True
```
Getting rid of self.accept and self.release here would let us make all the handler functions into staticmethods. Then we could add these methods directly to the UMBListener class's dict of virtual_topic_addresses, like `{"brew.build.tag": UMBReceiverHandler.brew_tags}`.

Then on_message() above wouldn't need to convert the string to a callable with getattr() (see lines 72 - 84 above). We could just call the function directly and check the return value, like you had before:
```
        callback_function = self.virtual_topic_addresses.get(address)

        if not address:
            raise ValueError(f"UMB event {event.message.id} had no address!")
        elif callback_function:
            success = callback_function(event)
        else:
            raise ValueError(
                f"UMB event {event.message.id} had unrecognized address: {event.message.address}"

        if success:
            self.accept(event.delivery)
        else:
            self.release(event.delivery, delivered=True)
```
That also avoids a lot of duplication in the code, which reduces the risk of typos when we add new handler functions and makes testing them much simpler.

juspence (approved):
LGTM!

PR #550, opened 2023-07-19
CORGI-729: Fix warning about unpatched SSL module

juspence:
@RedHatProductSecurity/corgi-devs This fixes a few minor issues, but probably not the root cause of this bug. The biggest change fixes a Gunicorn warning which may cause "silent failures" and other heisenbugs.

Since we use the gevent worker type in gunicorn, the standard library is automatically monkeypatched to support greenthreads logic during gunicorn's setup. This means that an application-managed greenthread can cooperatively multitask with other greenthreads, whenever code calls time.sleep(), socket.recv(), etc.

The greenthread will automatically be suspended whenever it would block while waiting for I/O, and another greenthread will automatically start running instead. This works well for I/O bound workloads, since it has much lower overhead than spawning a new process, and avoids contention when an OS-managed thread needs to acquire Python's Global Interpreter Lock.

But if you import a module too soon, before gunicorn has finished setting up, you will get the unpatched version of that module. Importing a module "too soon" can also happen when you do "from package.module import func", if package has some \_\_init\_\_.py module that itself imports other code.

Then any calls to time.sleep() or socket.recv() would end up blocking as normal instead of cooperatively multitasking / yielding control to another greenthread. At best your other greenthreads will starve and never execute, but at worst this causes heisenbugs like RecursionError or "silent failures", according to the text in the warning.

But if we monkeypatch too early, before gunicorn has forked its workers, monkeypatching may cause issuses with signaling between the parent process and all the workers it spawns, or the OS, or something. There are only vague mentions of this in some Github issues I read, so I don't have any more details. I think this change should be OK, since the warning is only about / we only patch the SSL module early. Gunicorn probably only needs an unpatched socket module to communicate between the parent and worker processes.

In our case we're calling django.core.wsgi.get_wsgi_application(), which loads our Django app. Currently we preload the app once in the parent process, so that we don't load the Django app each time a new worker process is spawned. This reduces the memory used by each worker, which hopefully reduces how often the Gunicorn worker runs out of memory and dies (bad gateway error described in 729).

To fix this warning about an unpatched SSL module, we need to either monkeypatch before the app is preloaded by the parent process (what this PR does), or load the app once per worker without preloading. Given how magical gevent is, I'm not sure if I should merge this, just give up on preloading, try some other fix like a different worker class (not Gevent or eventlet which both use greenthreads), or just give up completely and CANTFIX this ticket.

### DIFF
--- a/config/settings/prod.py
+++ b/config/settings/prod.py
@@ -9,7 +9,7 @@ ALLOWED_HOSTS = [
     # Allow local host's IP address and hostname for health probes
     socket.gethostname(),
     socket.gethostbyname(socket.gethostname()),
-    CORGI_DOMAIN_BASE,  # noqa: F405
+    CORGI_DOMAIN,  # noqa: F405
 ]
 
 CSP_UPGRADE_INSECURE_REQUESTS = True
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -404,11 +404,6 @@ UMB_BROKER_URL = os.getenv("CORGI_UMB_BROKER_URL")
 # https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool
 UMB_BREW_MONITOR_ENABLED = strtobool(os.getenv("CORGI_UMB_BREW_MONITOR_ENABLED", "true"))
 
-if CORGI_DOMAIN.endswith(".fedoraproject.org"):
-    CORGI_DOMAIN_BASE = ".fedoraproject.org"
-else:
-    CORGI_DOMAIN_BASE = ".prodsec.redhat.com"
-
 # Brew
 BREW_URL = os.getenv("CORGI_BREW_URL")
 BREW_WEB_URL = os.getenv("CORGI_BREW_WEB_URL")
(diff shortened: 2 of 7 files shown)


config/settings/prod.py
```diff
@@ -9,7 +9,7 @@
     # Allow local host's IP address and hostname for health probes
     socket.gethostname(),
     socket.gethostbyname(socket.gethostname()),
-    CORGI_DOMAIN_BASE,  # noqa: F405
+    CORGI_DOMAIN,  # noqa: F405
```
jasinner, line 12: I'm a little concerned this won't work in community prod, but I guess we'll know pretty quick and be able to back it out.

jasinner (approved):

config/settings/prod.py
juspence: I wondered the same thing :smile: When I looked at the CORGI_DOMAIN var in the ops repo, I see it's now set correctly in our playbooks for all four envs - enterprise stage and prod as well as community stage and prod. Previously, community stage didn't have a cname of its own, so we couldn't rely on the CORGI_DOMAIN var.

But now that community stage is accessible by that cname, we can restrict the Host: headers we accept. Currently enteprise prod and stage, as well as community stage, will all accept each other's host headers (as well as any other subdomain controlled by prodsec). This is a little dangerous since Django relies on the Host: header for some of its security checks:
https://docs.djangoproject.com/en/4.2/topics/security/#host-header-validation

EDIT: Deploy just finished, everything seems to be still working. Definitely ping me if you see issues though.

PR #551, opened 2023-07-20
CORGI-355 match cpe to stream using pattern

jasinner:
Use cpe patterns from product definitions ps_modules to build a list of cpes per stream. This works well to fill in the gaps where we can't match streams to variants in other ways. When combining the results of aggregating variant cpes, and cpes matched via ps_module patterns we get a list very close the current supported_cpe_map. 

I updated the supported_cpe_map to remove all the streams which have good matches in the list below. Some are still unable to be matched, and for those, we override the manifest cpe values. See the comments in core.core.fixups for details of why the remaining streams cannot be matched.

```
In [1]: import corgi.core.models as m
   ...: from corgi.core.fixups import supported_stream_cpes
   ...: from copy import deepcopy
   ...: 
   ...: 
   ...: def compare_cpes(stream, current, pattern_matches, variant_matches):
   ...:     future = deepcopy(pattern_matches)
   ...:     future.update(variant_matches)
   ...:     if current != future:
   ...:         print(f"mismatched stream: {stream}")
   ...:         print(f"supported_cpes_set: {current}")
   ...:         print(f"future: {future}")
   ...:         print(f"pattern matches: {pattern_matches}")
   ...:         print(f"variant_matches: {variant_matches}")
   ...:         print("=======================")
   ...:     else:
   ...:         print(f"Got match for {stream}")
   ...: 
   ...: 
   ...: for stream, supported_cpes in supported_stream_cpes.items():
   ...:     skip = False
   ...:     for ignore_prefix in ["certificate_system", "convert2rhel", "directory_server"]:
   ...:         if stream.startswith(ignore_prefix):
   ...:             skip = True
   ...:             break
   ...:     if skip:
   ...:         continue
   ...: 
   ...:     supported_cpes_set = set(supported_cpes)
   ...:     ps = m.ProductStream.objects.get(name=stream)
   ...:     cpe_matches_set = set(ps.cpes_matching_patterns)
   ...:     compare_cpes(stream, supported_cpes_set, cpe_matches_set, set(ps.cpes))
   ...: 
   ...: 
   ...: 
mismatched stream: dts-11.1.z
supported_cpes_set: {'cpe:/a:redhat:rhel_software_collections:3::el7'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: dts-12.0.z
supported_cpes_set: {'cpe:/a:redhat:rhel_software_collections:3::el7'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: dts-12.1
supported_cpes_set: {'cpe:/a:redhat:rhel_software_collections:3::el7'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: mtr-1.0
supported_cpes_set: {'cpe:/a:redhat:migration_toolkit_runtimes:1.0::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================

mismatched stream: rhes-3.5
supported_cpes_set: {'cpe:/a:redhat:storage:3.5:server:el7', 'cpe:/a:redhat:storage:3.5:nfs:el7', 'cpe:/a:redhat:storage:3.5:samba:el7', 'cpe:/a:redhat:storage:3.5:na:el7', 'cpe:/a:redhat:storage:3.5:nfs:el8', 'cpe:/a:redhat:storage:3.5:samba:el8', 'cpe:/a:redhat:storage:3.5:wa:el8', 'cpe:/a:redhat:storage:3.5:server:el8', 'cpe:/a:redhat:storage:3.5:wa:el7', 'cpe:/a:redhat:storage:3.5:na:el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: openshift-enterprise-3.11.z
supported_cpes_set: {'cpe:/a:redhat:openshift:3.11::el7'}
future: {'cpe:/a:redhat:openshift:3.11::el7', 'cpe:/a:redhat:rhosemc:1.0::el7'}
pattern matches: {'cpe:/a:redhat:openshift:3.11::el7'}
variant_matches: {'cpe:/a:redhat:openshift:3.11::el7', 'cpe:/a:redhat:rhosemc:1.0::el7'}
=======================
mismatched stream: openshift-4.12.z
supported_cpes_set: {'cpe:/a:redhat:openshift:4.12::el8', 'cpe:/a:redhat:openshift_ironic:4.12::el9', 'cpe:/a:redhat:openshift:4.12::el9'}
future: {'cpe:/a:redhat:openshift_ironic:4.12::el9', 'cpe:/a:redhat:openshift:4.12::el9', 'cpe:/a:redhat:openshift:4.12::el8', 'cpe:/a:redhat:openshift_security_profiles_operator_stable:::el8', 'cpe:/a:redhat:openshift_file_integrity_operator:1.0::el8', 'cpe:/a:redhat:openshift_compliance_operator:1::el8'}
pattern matches: {'cpe:/a:redhat:openshift:4.12::el8', 'cpe:/a:redhat:openshift_ironic:4.12::el9', 'cpe:/a:redhat:openshift:4.12::el9'}
variant_matches: {'cpe:/a:redhat:openshift:4.12::el9', 'cpe:/a:redhat:openshift:4.12::el8', 'cpe:/a:redhat:openshift_security_profiles_operator_stable:::el8', 'cpe:/a:redhat:openshift_file_integrity_operator:1.0::el8', 'cpe:/a:redhat:openshift_compliance_operator:1::el8'}
=======================
mismatched stream: openstack-13-optools
supported_cpes_set: {'cpe:/a:redhat:openstack-optools:13::el7'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: pipelines-1.7
supported_cpes_set: {'cpe:/a:redhat:openshift_pipelines:1.7::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: pipelines-1.8
supported_cpes_set: {'cpe:/a:redhat:openshift_pipelines:1.8::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: quay-3.6
supported_cpes_set: {'cpe:/a:redhat:quay:3::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: quay-3.7
supported_cpes_set: {'cpe:/a:redhat:quay:3::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: quay-3.8
supported_cpes_set: {'cpe:/a:redhat:quay:3::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: rhel-8.8.0
supported_cpes_set: {'cpe:/a:redhat:enterprise_linux:8::sap', 'cpe:/o:redhat:enterprise_linux:8::fastdatapath', 'cpe:/o:redhat:enterprise_linux:8::baseos', 'cpe:/a:redhat:enterprise_linux:8::highavailability', 'cpe:/a:redhat:enterprise_linux:8::supplementary', 'cpe:/a:redhat:enterprise_linux:8::nfv', 'cpe:/o:redhat:enterprise_linux:8::hypervisor', 'cpe:/a:redhat:enterprise_linux:8::appstream', 'cpe:/a:redhat:enterprise_linux:8::resilientstorage', 'cpe:/a:redhat:enterprise_linux:8::crb', 'cpe:/a:redhat:enterprise_linux:8::sap_hana', 'cpe:/a:redhat:enterprise_linux:8::realtime'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: rhel-9.2.0
supported_cpes_set: {'cpe:/a:redhat:enterprise_linux:9::nfv', 'cpe:/a:redhat:enterprise_linux:9::sap_hana', 'cpe:/a:redhat:enterprise_linux:9::sap', 'cpe:/o:redhat:enterprise_linux:9::baseos', 'cpe:/o:redhat:enterprise_linux:9::fastdatapath', 'cpe:/a:redhat:enterprise_linux:9::crb', 'cpe:/a:redhat:enterprise_linux:9::realtime', 'cpe:/a:redhat:enterprise_linux:9::highavailability', 'cpe:/o:redhat:enterprise_linux:9::hypervisor', 'cpe:/a:redhat:enterprise_linux:9::appstream', 'cpe:/a:redhat:enterprise_linux:9::resilientstorage', 'cpe:/a:redhat:enterprise_linux:9::supplementary'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: rhn_satellite_6.10
supported_cpes_set: {'cpe:/a:redhat:satellite:6.10::el7', 'cpe:/a:redhat:satellite_capsule:6.10::el7'}
future: {'cpe:/a:redhat:rhel_satellite_tools:6.10::el7', 'cpe:/a:redhat:satellite_capsule:6.10::el7', 'cpe:/a:redhat:rhel_satellite_tools:6.10::el8', 'cpe:/a:redhat:satellite:6.10::el7', 'cpe:/a:redhat:rhel_satellite_tools:6.10::el6'}
pattern matches: {'cpe:/a:redhat:rhel_satellite_tools:6.10::el7', 'cpe:/a:redhat:satellite_capsule:6.10::el7', 'cpe:/a:redhat:rhel_satellite_tools:6.10::el8', 'cpe:/a:redhat:satellite:6.10::el7', 'cpe:/a:redhat:rhel_satellite_tools:6.10::el6'}
variant_matches: {'cpe:/a:redhat:satellite:6.10::el7', 'cpe:/a:redhat:satellite_capsule:6.10::el7'}
=======================
mismatched stream: rhosc-1
supported_cpes_set: {'cpe:/a:redhat:openshift_sandboxed_containers:1.3.0::el8'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
mismatched stream: rhscl-3.8.z
supported_cpes_set: {'cpe:/a:redhat:rhel_software_collections:3::el7', 'cpe:/a:redhat:rhel_software_collections:3::el6'}
future: set()
pattern matches: set()
variant_matches: set()
=======================
```


jasinner:
To allow the rhscl variants to move to the dts stream we'll need to first delete any variants current attached to the rhscl streams before running update_products (which will create the variants again with the correct stream parent). We will then have to call save_product_taxonomy on all builds with those variant product_refs in the relations table.

jasinner:

jasinner:
After this is merged we should run corgi.tasks.et::load_et_products() to update the variant live field before running corgi.tasks.prod_defs.py::update_products()

corgi/api/serializers.py
```diff
@@ -675,7 +675,6 @@ class Meta(ProductModelSerializer.Meta):
         model = ProductStream
         fields = [
             *ProductModelSerializer.Meta.fields,
-            "cpe",
```
juspence, line 678: Will this impact IR?

tests/factories.py
```diff
@@ -83,7 +83,6 @@ class Meta:
     name = factory.Faker("word")
     version = "8.2.0.z"
     description = factory.Faker("word")
-    cpe = "cpe:/o:redhat:enterprise_linux:8"
```
juspence, line 86: Should we generate a random string instead of removing this property entirely?

corgi/collectors/errata_tool.py
```diff
@@ -261,3 +261,11 @@ def _parse_module(
                 # map modular_rpms to build_ids. In slow_save_errata these are only saved as
                 # meta_attr so, let's just save the entire list for each build_id
                 variant_to_component_map[variant].append({build_id: modular_rpms})
+
+    def load_cpes(self):
+        cpe_list = self.get("api/v1/security/cpes")
+        for cpe in cpe_list:
+            for cpe_variant in cpe["variants"]:
+                CollectorErrataProductVariant.objects.filter(name=cpe_variant["name"]).update(
+                    live=cpe_variant["live"]
```
juspence: Should we create the variant if it doesn't exist? I'm not familiar with the format of this data or if we have enough information to do that here.

The method name is also a little confusing. I'd expect it to either load and return e.g. a dict mapping variant names to CPEs, or to save a CPE onto each CollectorVariant object. Should we call this method "mark_variants_live" instead?

corgi/core/files.py
```diff
@@ -65,21 +65,23 @@ class ProductManifestFile(ManifestFile):
     # We can subclass this to handle different Product subclasses with different properties
     # Or to handle different ways of generating manifest properties from Product properties
 
-    def render_content(self, cpe_mapping=True) -> str:
+    def render_content(self):
```
juspence: Let's keep the return type hint:
```suggestion
    def render_content(self) -> str:
```

corgi/core/fixups.py
```diff
@@ -261,73 +87,25 @@
...
-    "rhcertification-7": ["cpe:/a:redhat:certifications:1::el7"],
-    "rhcertification-8": ["cpe:/a:redhat:certifications:1::el8"],
-    "rhcertification-9": ["cpe:/a:redhat:certifications:1::el9"],
 }
 
 
-def cpe_lookup(product_stream_name: str) -> list[str]:
-    """Temporary cpe fixup"""
-    return supported_stream_cpes.get(product_stream_name, [])
+def cpe_lookup(product_stream_name: str) -> Optional[set[str]]:
```
juspence: This isn't Optional / returning a None value, it's always returning a set:
```suggestion
def cpe_lookup(product_stream_name: str) -> set[str]:
```

corgi/core/fixups.py
```diff
@@ -261,73 +87,25 @@
...
+    """Manual CPE overrides for streams which cannot be matched with variants"""
+    if product_stream_name in supported_stream_cpes:
+        return set(supported_stream_cpes.get(product_stream_name, []))
+    else:
+        return set()
```
juspence: You don't need to do .get() here because you've already checked that this stream_name is in the dict of supported_streams. You can do a one-liner that matches the old code:
```suggestion
    return set(supported_stream_cpes.get(product_stream_name, []))
```
This will aways return either a set of the CPEs, or an empty set. Or you can do an if / else without .get():
```suggestion
    if product_stream_name in supported_stream_cpes:
        return set(supported_stream_cpes[product_stream_name])
    else:
        return set()
```

corgi/core/fixups.py
```diff
@@ -140,99 +38,26 @@
...
     ],
-    "omr-1": ["cpe:/a:redhat:mirror_registry:1.1.0::el8"],
-    "openshift-4.10.z": ["cpe:/a:redhat:openshift:4.10::el8", "cpe:/a:redhat:openshift:4.10::el7"],
-    "openshift-4.11.z": ["cpe:/a:redhat:openshift:4.11::el7", "cpe:/a:redhat:openshift:4.11::el8"],
+    # variant with cpe 'cpe:/a:redhat:rhosemc:1.0::el7' is matched with brew_tags
+    "openshift-enterprise-3.11.z": ["cpe:/a:redhat:openshift:3.11::el7"],
+    # investigate why brew tag matching gives us variants with cpes:
+    # 'cpe:/a:redhat:openshift_security_profiles_operator_stable:::el8'
+    # 'cpe:/a:redhat:openshift_file_integrity_operator:1.0::el8'
+    # 'cpe:/a:redhat:openshift_compliance_operator:1::el8'
```
juspence: I think before we make changes here, we should check to make sure we don't have any odd results and that we handle all the edge cases like this.

corgi/core/migrations/0078_adjust_product_cpe_fields.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.18 on 2023-07-17 04:27
+
+import django.contrib.postgres.fields
```
juspence: You're only using ArrayField so the code is a little more clear if you import that directly:
```suggestion
from django.contrib.postgres.fields import ArrayField
# or "from django.contrib.postgres import fields"
```

corgi/core/migrations/0078_adjust_product_cpe_fields.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name="productstream",
+            name="cpe",
+        ),
+        migrations.AddField(
+            model_name="productversion",
+            name="cpe_patterns",
+            field=django.contrib.postgres.fields.ArrayField(
```
juspence:
```suggestion
            field=ArrayField(
            # or "fields=fields.ArrayField" if you use the second import form above
```

corgi/core/migrations/0078_adjust_product_cpe_fields.py
```diff
@@ -0,0 +1,25 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0077_clear_component_channels"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="productstream",
+            name="cpe",
```
juspence: Is this always unused / unset in all our code and data? I'd be worried that something will break if we just get rid of it.

corgi/core/migrations/0078_adjust_product_cpe_fields.py
```diff
@@ -0,0 +1,25 @@
...
+        ),
+        migrations.AddField(
+            model_name="productversion",
+            name="cpe_patterns",
```
juspence: I thought prod-defs defined CPE patterns on streams, and that our productstream model was what matched the prod-defs streams, not productversion.

Does prod-defs only define cpe_patterns on versions, never on streams? Are there ever cpe_patterns on both?

corgi/core/migrations/0080_productstream_cpes_matching_patterns.py
juspence, line 1: It's a little confusing to have all these migrations in separate files. Could you combine at least the two "modelname_cpes_matching_patterns" (number 79 and number 80) into one, and maybe even combine that with number 78?

corgi/core/models.py
```diff
@@ -408,22 +408,17 @@ def builds(self) -> QuerySet:
...
         variant_cpes = (
-            descendants.filter(level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"])
+            self.pnodes.get_queryset()
+            .get_descendants()
+            .using("read_only")
+            .filter(level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"])
             .values_list("productvariant__cpe", flat=True)
             .distinct()
         )
+        # include_self=True so we don't have to override this property for streams
```
juspence: We only report CPEs for Variants now and not for streams, which isn't a problem if that's what you wanted. In that case, this comment should go away. Or if you still want to report CPEs / CPE patterns for streams, you need different filtering logic here.

corgi/core/models.py
```diff
@@ -530,6 +525,9 @@ class ProductTag(Tag):
 
 
 class ProductVersion(ProductModel):
+    cpe_patterns = fields.ArrayField(models.CharField(max_length=1000), default=list)
+    cpes_matching_patterns = fields.ArrayField(models.CharField(max_length=1000), default=list)
```
juspence: Can you please add a comment describing what these fields are used for and where the data comes from / how the fields are populated?

corgi/core/models.py
```diff
@@ -580,6 +576,8 @@ class ProductStream(ProductModel):
 
     exclude_components = fields.ArrayField(models.CharField(max_length=200), default=list)
 
+    cpes_matching_patterns = fields.ArrayField(models.CharField(max_length=1000), default=list)
```
juspence, line 583: Why do we have only cpes_matching_patterns here, but both cpes_matching_patterns and cpe_patterns above?

If we're trying to find "all the CPEs on all descendants which match some CPE pattern", maybe we should compute that dynamically (as a property) instead of storing it. That way we'll stop reporting old / stale data, as soon as a Variant moves between streams / versions / products, or if a Variant becomes inactive and we want to stop reporting it, and so on.

If we store the list of cpe patterns, I'd be worried that we would forget to update this when something changes in our taxonomy, in ET, or in the prod-defs data.

corgi/tasks/prod_defs.py
```diff
@@ -101,6 +190,8 @@ def parse_product_version(
         parse_product_stream(
             pd_product_stream, product, product_version, product_version_node, version
         )
+    if cpes_matching_patterns:
+        match_stream_cpes(product_version)
 
 
 def parse_product_stream(
```
juspence, line 200: You added logic to parse_product_version that (I think) loads the CPE patterns from prod-defs, finds the matching CPEs on any descendants, and then sets that list of descendant CPEs on the version.

Do we also need to set the list of descendant CPEs when we parse_product_stream?

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,89 @@ def update_products() -> None:
...
+        _split_cpe_version(cpe_part, versions_by_cpe)
+    elif len(variant_split) == 1:
+        _split_cpe_version(cpe, versions_by_cpe)
+    else:
+        raise ValueError(f"More than one el qualifier '::' in {cpe}")
+
+
+def match_stream_cpes(product_version: ProductVersion) -> None:
+    """Given a Product Version with Product Stream foreign keys and cpes_matching_patterns
+    This will match those cpes to the stream children using the stream versions"""
```
juspence: The function name and comment implies that it should just search data and match values. My expectation before even looking at this method was that it would return all the matched CPEs or stream names or something.

If we want to update the CPE value on the matching streams, let's make that clear in the function name and header comment.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,89 @@ def update_products() -> None:
...
+        for cpe_version in versions:
+            cpe = cpe_name + ":" + cpe_version
+            _match_stream_version_to_cpe_version(
+                version_streams, cpe_version, cpe, el_by_cpe, cpes_by_stream
+            )
+
+    for stream, cpes in cpes_by_stream.items():
+        product_stream = ProductStream.objects.get(name=stream)
+        product_stream.cpes_matching_patterns = cpes
+        product_stream.save()
```
juspence: I think the code that calls this function is already running in a transaction, but let's use an inner transaction or .update() here in case we ever start calling this function from somewhere else. .update() is also a little faster than calling .get() + making a change + calling .save():
```suggestion
    for stream, cpes in cpes_by_stream.items():
        ProductStream.objects.filter(name=stream).update(cpes_matching_patterns=cpes)
```
Or if we never expect to call this code anywhere else, let's add an _ underscore in front of the method name.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,89 @@ def update_products() -> None:
...
+        _split_cpe_version(cpe, versions_by_cpe)
+    else:
+        raise ValueError(f"More than one el qualifier '::' in {cpe}")
+
+
+def match_stream_cpes(product_version: ProductVersion) -> None:
+    """Given a Product Version with Product Stream foreign keys and cpes_matching_patterns
+    This will match those cpes to the stream children using the stream versions"""
+    versions_by_cpe: defaultdict[str, set] = defaultdict(set)
+    el_by_cpe: defaultdict[str, list] = defaultdict(list)
```
juspence: Maybe el_suffix_by_cpe is a better name:
```suggestion
    el_suffix_by_cpe: defaultdict[str, list] = defaultdict(list)
```

juspence (commented):
Sorry for the nitpicking, but there are a lot of changes here and it's a little hard to review. Can we split this out into separate, smaller PRs?

Maybe we can start by only adding fields or properties to our model, without changing any of the existing fields / properties, the API, or the tests.

Once that runs for a few days, we can validate the data we loaded looks like whatever we expect, and that we don't have edge cases we didn't think of. Then we can either merge the new data with the old data, or switch the model to rely on only the new fields and get rid of the old ones.

Right now (if I'm not misunderstanding) I think we have a "cpes" property that generates all the descendant CPEs, but you changed this to only report Variant CPEs and not stream CPEs. Then we also have a list of descendant CPEs on a stream, and a list of descendant CPEs on a version, but these are hardcoded instead of generated.

Or if I'm confused, maybe we can discuss this on Monday when we meet about other stuff.

jasinner:
> Sorry for the nitpicking, but there are a lot of changes here and it's a little hard to review. Can we split this out into separate, smaller PRs?
> 
> Maybe we can start by only adding fields or properties to our model, without changing any of the existing fields / properties, the API, or the tests.
> 
> Once that runs for a few days, we can validate the data we loaded looks like whatever we expect, and that we don't have edge cases we didn't think of. Then we can either merge the new data with the old data, or switch the model to rely on only the new fields and get rid of the old ones.
> 
> Right now (if I'm not misunderstanding) I think we have a "cpes" property that generates all the descendant CPEs, but you changed this to only report Variant CPEs and not stream CPEs. Then we also have a list of descendant CPEs on a stream, and a list of descendant CPEs on a version, but these are hardcoded instead of generated.
> 
> Or if I'm confused, maybe we can discuss this on Monday when we meet about other stuff.

I split the changes into commits, so perhaps it would be helpful to review the changes commit by commit. I think all the changes here only make sense as a single PR, because the ultimate goal of all these changes is to remove the supported_cpe_map. The next step could be to also match Variant to Streams based on the cpe pattern matching, but I left that for a subsequent PR. I think that's something we should consider carefully before changes as it will affect the product taxonomy of many components.

The current cpe attribute on ProductStreams doesn't make sense because that property is always defined at the variant level. At the ProductStream level it's always an aggregation of ProductVariant cpes. The cpes_matching_patterns ArrayList attribute is calculated by matching cpes from Errata Tool with the cpe_patterns from product_defintions. This attribute is recalculated every day when we refresh the data from those 2 systems. You only need to run update_et_products and update_products tasks in dev to discover all the matching, so I don't see the point of letting this run in stage for a few days. Also I've provided all the data which was matched or not in the description of this PR.

corgi/api/serializers.py
jasinner: The attribute doesn't make sense because is should be an aggregate of the variant children. The attribute has not been populated with any data.

corgi/collectors/errata_tool.py
jasinner: There's not enough information to create a variant here. I updated the name of the method.

corgi/core/migrations/0078_adjust_product_cpe_fields.py
jasinner: It's currently unused, and the field doesn't make sense.

corgi/core/migrations/0078_adjust_product_cpe_fields.py
jasinner: There are only cpe_patterns on versions, and never on streams.

corgi/core/migrations/0080_productstream_cpes_matching_patterns.py
jasinner: The migrations match the commits. I could of course combine the migrations, but wouldn't that make it harder to review the changes? Also, I think it would make it harder to revert commits, and rollback the migrations.

corgi/core/models.py
jasinner: We only report CPEs for Variants. I'll remove the comment.

corgi/core/fixups.py
jasinner: The variants that are matched to the openshift-4.12.z stream are correct. We could actually remove the override for openshift-4.12.z and let the additional CPEs be included in the SBOM in case we release a security update for any of these components with these variants/CPEs. The change of the CPEs in the SBOM might be a little controversial internally though, which is why I decided to leave the override in for this stream. The chances we will actually have a security update for those operators in this stream is unlikely.

corgi/core/files.py
jasinner: ok

corgi/core/fixups.py
jasinner: ok, fixed thanks.

corgi/core/fixups.py
jasinner: Using the first suggestion, thanks.

tests/factories.py
jasinner: The field was removed from the model.

corgi/core/models.py
jasinner: Done.

corgi/core/models.py
jasinner: The cpe_patterns on ProductVersion are directly read from product_definitions ps_module 'cpe' field. There is no such field on ps_update_stream model in product_definitions. I don't think we should compute them dynamically so as not to add a dependency from core models to collector models, it's better that it's computing by the prod_defs's update_products task, and only saved to the core models here.

It is replaced every day when we run the update_products task.

corgi/tasks/prod_defs.py
jasinner: I'll update the function name and comment.

corgi/tasks/prod_defs.py
jasinner: Good idea, I'll use .update() function to make it atomic.  I'll also add an underscore to the method name.

corgi/tasks/prod_defs.py
jasinner: No need. The match_stream_cpes function save cpes_matching_patterns on the streams.

corgi/core/migrations/0078_adjust_product_cpe_fields.py
jasinner: Updated.

corgi/tasks/prod_defs.py
jasinner: updated

corgi/core/fixups.py
```diff
@@ -261,73 +86,22 @@
...
-    "rhcertification-9": ["cpe:/a:redhat:certifications:1::el9"],
 }
 
 
-def cpe_lookup(product_stream_name: str) -> list[str]:
-    """Temporary cpe fixup"""
-    return supported_stream_cpes.get(product_stream_name, [])
+def cpe_lookup(product_stream_name: str) -> set[str]:
+    """Manual CPE overrides for streams which cannot be matched with variants"""
+    return set(supported_stream_cpes.get(product_stream_name, []))
```
juspence, line 96: FYI you can also use set literal syntax in the dict above, instead of having to convert the values to a set at the end. Any way you prefer is fine:

```suggestion
def cpe_lookup(product_stream_name: str) -> set[str]:
    """Manual CPE overrides for streams which cannot be matched with variants"""
    return supported_stream_cpes.get(product_stream_name, set())
```

And then above would be something like:
```
supported_stream_cpes = {
    # curly brace above creates a dict, because it has colons / keys and values
    # curly brace below creates a set, because it has no colons / keys, only values
    "stream_A": {"value_1", "value_2"},
    # Can also use set() constructor syntax
    "stream_B": set("value_3", "value_4"),
}
```

corgi/core/models.py
```diff
@@ -408,22 +408,16 @@ def builds(self) -> QuerySet:
     @property
     def cpes(self) -> tuple[str, ...]:
         """Return CPEs for child streams / variants."""
```
juspence:
```suggestion
        """Return CPEs of all descendant variants."""
```

corgi/core/models.py
```diff
@@ -408,22 +408,16 @@ def builds(self) -> QuerySet:
...
-        )
-        stream_cpes = (
-            descendants.filter(level=MODEL_NODE_LEVEL_MAPPING["ProductStream"])
-            .values_list("productstream__cpe", flat=True)
-            .distinct()
-        )
         variant_cpes = (
-            descendants.filter(level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"])
+            self.pnodes.get_queryset()
+            .get_descendants()
```
juspence: Let's make sure that if we access `some_variant.cpes`, we'll still get a tuple with one entry like `(some_variant.cpe,)` and not an empty () tuple:
```suggestion
            .get_descendants(include_self=True)
```

corgi/api/serializers.py
juspence: In that case, should we report the "cpes" property we generate instead? We could add it here for just ProductStreams, or to the ProductModel serializer's list of fields above. Then we'd report CPEs for descendant Variants on all Products, Versions, etc.

tests/test_manifests.py
```diff
@@ -276,8 +281,10 @@ def test_product_manifest_properties():
 
     assert product_data["SPDXID"] == f"SPDXRef-{stream.uuid}"
     assert product_data["name"] == stream.name
-    for index, cpe in enumerate(stream.cpes):
-        assert product_data["externalRefs"][index]["referenceLocator"] == cpe
+    cpes_in_manifest = [ref["referenceLocator"] for ref in product_data["externalRefs"]]
+    stream_cpes = set(stream.cpes)
+    stream_cpes.update(stream.cpes_matching_patterns)
+    assert set(cpes_in_manifest) == stream_cpes
```
juspence: You can just build the set directly, you don't need to convert it later:
```suggestion
    cpes_in_manifest = set(ref["referenceLocator"] for ref in product_data["externalRefs"])
    stream_cpes = set(stream.cpes)
    stream_cpes.update(stream.cpes_matching_patterns)
    assert cpes_in_manifest == stream_cpes
```

tests/test_model.py
```diff
@@ -58,11 +57,11 @@ def test_cpes():
     ps2 = ProductStreamFactory(name="Ansercaerulescens", products=p1, productversions=pv2)
     ps3 = ProductStreamFactory(
         name="Brantabernicla",
-        cpe="cpe:/o:redhat:Brantabernicla:8",
         description="The brant or brent goose is a small goose of the genus Branta.",
         products=p1,
         productversions=pv2,
     )
+    pvariant1 = ProductVariantFactory(cpe="cpe:/o:redhat:Brantabernicla:8")
```
juspence: Let's link the variant to the parent product 1 / version 2 / stream 3 for RHEL 8, then create a second variant with a different CPE, and link it to the parent product 1 / version 1 / stream 1 for RHEL 7:
```suggestion
    pvariant1 = ProductVariantFactory(cpe="cpe:/o:redhat:enterprise_linux:7", products=p1, productversions=pv1, productstreams=ps1)
    pvariant2 = ProductVariantFactory(cpe="cpe:/o:redhat:Brantabernicla:8", products=p1, productversions=pv2, productstreams=ps3)
```
If we do that, I think the old asserts will keep passing.

tests/test_model.py
```diff
@@ -76,12 +75,9 @@ def test_cpes():
...
     ps3node = ProductNode.objects.create(parent=pv2node, obj=ps3)
     assert ps3node
+    ProductNode.objects.create(parent=ps1node, obj=pvariant1)
 
-    assert pv1.cpes == ("cpe:/o:redhat:enterprise_linux:8",)
-    assert sorted(p1.cpes) == [
-        "cpe:/o:redhat:Brantabernicla:8",
-        "cpe:/o:redhat:enterprise_linux:8",
-    ]
+    assert p1.cpes == ("cpe:/o:redhat:Brantabernicla:8",)
```
juspence: I think the old asserts were mostly correct, but the new variant needs to be linked into the taxonomy correctly. Then we can check that Variant CPEs are discovered correctly for a product with two different versions that each have their own variant:
```suggestion
    # Version 1 for RHEL 7 has 1 child variant, so 1 CPE is reported
    assert pv1.cpes == ("cpe:/o:redhat:enterprise_linux:7")
    # Product 1 for RHEL has 2 child variants, one for each RHEL version
    # So 1 CPE for RHEL 7 and 1 CPE for RHEL 8 is reported
    assert sorted(p1.cpes) == [
        "cpe:/o:redhat:Brantabernicla:8",
        "cpe:/o:redhat:enterprise_linux:7",
    ]
```

tests/test_model.py
```diff
@@ -76,12 +75,9 @@ def test_cpes():
     assert ps2node
     ps3node = ProductNode.objects.create(parent=pv2node, obj=ps3)
     assert ps3node
+    ProductNode.objects.create(parent=ps1node, obj=pvariant1)
```
juspence: I think we're just checking that node creation doesn't fail due to some unique constaint. It might not be needed but let's add that anyway for both variants.
```suggestion
    pvariant1node = ProductNode.objects.create(parent=ps1node, obj=pvariant1)
    assert pvariant1node
    pvariant2node = ProductNode.objects.create(parent=ps3node, obj=pvariant2)
    assert pvariant2node
```

tests/test_errata_data.py
```diff
@@ -393,3 +393,28 @@ def test_parse_module_and_normal_errata_components_mixed_variants(
...
+
+    with open("tests/data/errata/et_cpes.json", "r") as remote_source_data:
+        requests_mock.get(
+            f"{settings.ERRATA_TOOL_URL}/api/v1/security/cpes",
+            text=remote_source_data.read(),
+        )
+    et = ErrataTool()
+    et.mark_variants_live()
+
+    variant = CollectorErrataProductVariant.objects.get(name="6Server-Supplementary-6.0.Z")
```
juspence: We test that this function works correctly, but maybe we should also test that it gets called during load_et_products().

Also FYI you can just do variant.refresh_from_db() instead of having to write out the whole query again.

corgi/tasks/prod_defs.py
```diff
@@ -76,6 +159,8 @@ def parse_product_version(
     if match_version := RE_VERSION_LIKE_STRING.search(name):
         version = match_version.group()
     description = pd_product_version.pop("public_description", [])
+    cpes = pd_product_version.pop("cpe", [])
```
juspence: Let's call this cpe_patterns so the variable has the same name as the field we're storing it into.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
                 parse_product_version(pd_product_version, product, product_node)
 
 
+def _find_by_cpe(cpe_patterns: list[str]) -> list[str]:
+    """Given a list of CPE patterns find all the CPEs from ET Variants which match it"""
+    if not cpe_patterns:
+        return []
+    regex_query = Q()
+    for pattern in cpe_patterns:
+        regex = pattern.replace(".", "\\.").replace("*", ".*")
```
juspence, line 79: I guess based on the logic here, cpe_patterns are always glob expressions and not proper regular expressions. If that's ever not the case, the matching code here will either break or give surprising results.

I looked briefly at fnmatch, which can translate a glob to a regex for us and might be slightly more reliable. But I don't see any way to check if something really is a glob expression first.

I guess we could use `re.compile()` first, catch any exception if it's really a glob and not a valid regex, then call `fnmatch.translate()` instead. But that also seems error-prone since a glob expression like "some.*.pattern" might look like a regex, even though it's really a glob.

Feel free to implement whatever you think is best, I'm just nervous we have to rely on the prod-defs patterns always being correct without any validation.
```suggestion
        regex = fnmatch.translate(pattern)
```

corgi/tasks/prod_defs.py
```diff
@@ -208,6 +297,7 @@ def parse_variants_from_brew_tags(
                         et_variant.name,
                         product_stream.name,
                     )
+                    # TODO also ensure the Variant is live before linking
```
juspence: Do we need to handle this now, before merging, or is it OK if we link inactive Variants?

Or maybe we should keep linking the inactive Variants, but just hide them from display using the ET live flag. That would match what we do for inactive Streams, using the prod-defs active flag.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+                for el in el_by_cpe[cpe]:
+                    cpes_by_stream[version_stream.name].append(cpe + "::" + el)
+            else:
+                cpes_by_stream[version_stream.name].append(cpe)
+
+
+def _split_cpe_version(cpe_part: str, versions_by_cpe: defaultdict[str, set]):
+    version_split = cpe_part.rsplit(":", 1)
+    cpe_name = version_split[0]
+    cpe_version = version_split[-1]
```
juspence: Using `version_split[-1]` avoids an error when the length of version_split is only 1 / when a : colon wasn't found in the cpe_part. Is that on purpose, or do we need to raise an error in this case?

Currently both cpe_name and cpe_version will both be the same in this case, so the dict key would be "something_without_a_colon" and we'd also append "something_without_a_colon" to the list of values.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+    """Given a Product Version with Product Stream foreign keys and cpes_matching_patterns
+    This will match those cpes to the stream children using the stream versions, and save the cpes
+    to the cpes_matching_patterns attribute of the streams."""
+    versions_by_cpe: defaultdict[str, set] = defaultdict(set)
+    el_suffix_by_cpe: defaultdict[str, list] = defaultdict(list)
+    for cpe in product_version.cpes_matching_patterns:
+        _split_cpe(cpe, versions_by_cpe, el_suffix_by_cpe)
+
+    cpes_by_stream: defaultdict[str, list] = defaultdict(list)
+    version_streams = product_version.productstreams.all()
```
juspence:
```suggestion
    version_streams = product_version.productstreams.get_queryset()
```

What happens if a stream is deleted in prod-defs? I guess maybe it never happens. But if it did, we'd keep the CPE for that stream's Variants on both the stream (which should be deleted) and the version (which should stop referencing the Variant CPEs from that deleted stream).

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+    version_streams: QuerySet[ProductStream],
+    cpe_version: str,
+    cpe: str,
+    el_by_cpe: defaultdict[str, list],
+    cpes_by_stream: defaultdict[str, list],
+):
+    """Iterate product stream versions, looking for matches to the cpe_version passed in.
+    If a match is found, store it in cpes_by_streams dict along with any el suffixes."""
+    for version_stream in version_streams:
+        if version_stream.version.removesuffix(".z") == cpe_version:
```
juspence: Instead of iterating over the entire queryset, can we just filter it and iterate over only the matching values, like below?
```suggestion
    cpe_version_with_z = f"{cpe_version}.z"
    for version_stream_name in version_streams.filter(models.Q(version=cpe_version) | models.Q(version=cpe_version_with_z)).values_list("name", flat=True):
```

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+            if cpe in el_by_cpe:
+                for el in el_by_cpe[cpe]:
+                    cpes_by_stream[version_stream.name].append(cpe + "::" + el)
+            else:
+                cpes_by_stream[version_stream.name].append(cpe)
```
juspence:
```suggestion
                    cpes_by_stream[version_stream_name].append(cpe + "::" + el)
            else:
                cpes_by_stream[version_stream_name].append(cpe)
```

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+    """Given a list of CPE patterns find all the CPEs from ET Variants which match it"""
+    if not cpe_patterns:
+        return []
+    regex_query = Q()
+    for pattern in cpe_patterns:
+        regex = pattern.replace(".", "\\.").replace("*", ".*")
+        regex_query |= Q(cpe__regex=regex)
+    return list(
+        CollectorErrataProductVariant.objects.filter(live=True)
+        .filter(regex_query)
```
juspence: Are Variants in ET ever deleted, instead of just being marked with live=False? If that ever does happen, I think this will find some Collector model in Corgi that doesn't exist in ET anymore. So we'd end up reporting a Variant CPE on both the parent Version and parent Stream, but that CPE should just disappear instead.

juspence (approved):
I didn't look at the new tests in test_products.py, but overall I think this looks pretty good. Comments are all very minor, feel free to ignore if you'd like and thanks for taking the time to explain this to me.

I think this should be pretty safe since we always run errata_tool.load_et_products() before prod_defs.update_products(), so all the ET data needed by the prod-defs collector should already be loaded.

We also update / overwrite the cpes and cpes_matching_patterns fields on the versions and the streams, and we never append to them. So if CPE patterns change or are removed, I think we'll handle correcting all the related data with the update_or_create.

The only issue I see is when something is deleted in prod-defs or ET. In that case, let's make sure we clean up those models in Corgi, and we don't keep reporting their CPE values.

We probably do need that other PR to handle streams that move between versions - the ForeignKey links are updated but not the nodes. Definitely a separate issue though.

corgi/tasks/prod_defs.py
```diff
@@ -67,6 +69,87 @@ def update_products() -> None:
...
+    version_streams = product_version.productstreams.all()
+    for cpe_name, versions in versions_by_cpe.items():
+        for cpe_version in versions:
+            cpe = cpe_name + ":" + cpe_version
+            _match_stream_version_to_cpe_version(
+                version_streams, cpe_version, cpe, el_suffix_by_cpe, cpes_by_stream
+            )
+
+    for stream, cpes in cpes_by_stream.items():
+        ProductStream.objects.filter(name=stream).update(cpes_matching_patterns=cpes)
```
juspence, line 155: One more edge case I thought of: this only updates the streams that we match. If the Version's cpe_patterns change in prod-defs, we might stop matching streams that we matched before. Those streams would keep reporting old values for this data and never get updated.

The same issue might happen if a stream moves from one version to another version, but the new version's CPE patterns don't match the stream.

To avoid this, let's update cpes_matching_patterns to an empty [] list each time we update_or_create() the stream, in line 217 of parse_product_stream below. We should still set the cpes_matching_patterns to the correct value after processing the stream, when we finish processing the version. But this should give us a little extra safety when some stream no longer matches the patterns of its parent version.

corgi/api/serializers.py
jasinner: Alright, I added cpes property to ProductStream serializer.

corgi/core/migrations/0080_productstream_cpes_matching_patterns.py
jasinner: I merged 78 and 79.

corgi/tasks/prod_defs.py

corgi/tasks/prod_defs.py
jasinner: I took this straight from SDEngine PSModule.find_by_cpe. I use the values a little differently however. In SDEngine we're searching from an Errata CPE, matching back to modules and trying to find the 'best' match. In Corgi we are searching from PSModule, matching back to Errata CPEs. We want to find all matches.

We store all the matches on the Product Version (which matches to a ps_module in prod_defs), then later we try to match those to ProductStreams using their versions.

corgi/tasks/prod_defs.py
jasinner: It's a good idea. I updated it so it only links live variants. Have to delete any variants which aren't live. 

Live means they have had at least one build shipped to them. No point listing variants if they aren't live yet.

corgi/tasks/prod_defs.py
jasinner: I disagree. We replace the cpes_matching_patterns each time we run parse_product_version, so we wouldn't keep that stream's CPES.

corgi/tasks/prod_defs.py
jasinner: good idea.

corgi/tasks/prod_defs.py
jasinner: fixed.

corgi/tasks/prod_defs.py
jasinner: Nice catch, will do.

corgi/tasks/prod_defs.py
juspence: Oops, I think you're right about the version. Sorry for the confusion, I've been rather tired lately. But don't streams that disappear in prod-defs hang around in Corgi?

I guess at the end of the update_products() task, we need to find all the products, versions, streams, and variants which weren't updated (or created) by the task. We could add each object's PK / name to a set, then iterate over all Product / ProductVersion / ProductStream / ProductVariant.objects.exclude(name__in=set_of_processed_names) if any exist.

Any results in that queryset mean we have an orphaned model in Corgi, which doesn't exist anymore in prod-defs. This would handle deleted streams as well as (I think) deleted Brew tag Variants. Then we could log a warning, delete the orphaned model, or something else.

Maybe that needs to be a separate ticket, but I think we need to add this logic ASAP given how orphaned Brew tag Variants in Corgi caused issues, and how in the future orphaned streams (or CPEs of orphaned Variants) might get published to the Customer Portal (or end up in a published manifest) by mistake.

jasinner:
We need to delete all ProductVariants in the data which were never live in order to avoid matching those using Brew Tags, and getting the CPEs:

```
dead_variants = cm.CollectorErrataProductVariant.objects.filter(live=False).values_list("name",flat=True)
m.ProductVariant.objects.filter(name__in=dead_variants).delete()

```

corgi/tasks/prod_defs.py
jasinner: I don't think so. Also they only ever go from live False to True, not True to False. Therefore the logic we have the PR which only ever adds a variant where it is live is correct.

corgi/tasks/prod_defs.py
jasinner: Deleting a ps_update_stream does happen, there is a proposed change now for ossm (product-definitions/-/merge_requests/2234/). I updated CORGI-726 with your good idea here.

corgi/tasks/prod_defs.py
jasinner: Not having a ':' in the cpe_part should never happen. But I updated the code here to check for that log a warning, and skip updating the versions_by_cpe map for that cpe.

PR #552, opened 2023-07-20
CORGI-740: Document custom db config

juspence:
Documents custom settings we've applied to our managed database. This doesn't include any default settings, or config file changes the ITDB team made based on a request / ticket from us. It only tracks changes we explicitly set ourselves by running custom Postgres commands.

@JimFuller-RedHat Does this look right, or do you know of some other changes we've made that I didn't capture here?


juspence:
@RedHatProductSecurity/corgi-devs FYI, going to merge this.

PR #553, opened 2023-07-20
Add logic to find managed services repos, which are also root components

juspence:
@RedHatProductSecurity/corgi-devs Please review this bugfix I found while working on other tickets. Ignore the SonarQube failure, it's complaining because a migration looks "duplicated" compared to a previous one. They share many common lines of code but I don't see a way to mark this as a false-positive.

We're about to start creating Github components for repos that are used by some managed service. These are created as root components, but our existing root component filter won't find them.

The fix for that here is very small, and probably incomplete. We probably need to discuss what should happen if:
1. The SCA task runs on a managed services build / a Github repo root component. Code should never do this, but a human might try to run SCA on every root component / build with e.g. `Component.objects.filter(ROOT_COMPONENTS_CONDITION).values_list("software_build__build_id", flat=True)`. This will now include some Github repo root components (newly included by this change) as well as index container root components (already included today). In that case, what should we do?

2. The manifest-generating task runs on a ProductStream that has only managed service components. We try to find all the root components in the stream, then find only released builds (which have tags for a released erratum), then take the latest versions of whatever's released. I think this shouldn't ever find any managed services builds, since we create a fake build with no tags, so all the service builds should look "unreleased". Maybe nothing else is needed here?

3. Any other usages of ROOT_COMPONENTS_CONDITION we should pay attention to?


mprpic:
> * The SCA task runs on a managed services build / a Github repo root component. Code should never do this, but a human might try to run SCA on every root component / build with e.g. `Component.objects.filter(ROOT_COMPONENTS_CONDITION).values_list("software_build__build_id", flat=True)`. This will now include some Github repo root components (newly included by this change) as well as index container root components (already included today). In that case, what should we do?
> 

I'm not sure I understand the question here. What's the issue with including Managed Services root components in that queryset?

> * The manifest-generating task runs on a ProductStream that has only managed service components. We try to find all the root components in the stream, then find only released builds (which have tags for a released erratum), then take the latest versions of whatever's released. I think this shouldn't ever find any managed services builds, since we create a fake build with no tags, so all the service builds should look "unreleased". Maybe nothing else is needed here?
> 

We don't have a definition of what a released managed services is yet so we only care about latest stuff. The daily task ditches all previous builds and creates new ones. Whether it appears as released or unreleased is probably irrelevant right now.

> * Any other usages of ROOT_COMPONENTS_CONDITION we should pay attention to?

We'll see! :-)

mprpic (approved):

juspence:
> I'm not sure I understand the question here. What's the issue with including Managed Services root components in that queryset?
> 
Maybe nothing, I just wanted to call out that we try to run SCA only on root components, so this is a change to "what we run SCA on". SCA may also fail depending on how it tries to find the source code for the component (don't remember, haven't looked at code recently). That's not really a problem, but does lead to annoying errors in our monitoring email, so we could special-case these Github root components just like we do for RPMMOD root components.

> We don't have a definition of what a released managed services is yet so we only care about latest stuff. The daily task ditches all previous builds and creates new ones. Whether it appears as released or unreleased is probably irrelevant right now.
> 
The concern isn't latest or not, it's whether we generate a manifest at all. I assumed we don't want to manifest managed services, since that's Red Hat internal stuff that we don't list as affected on CVE pages.

Right now the managed service builds have no released_errata_tags in their meta_attr, so we won't generate a manifest for any managed service streams (they have no released builds / components). If we do want to manifest the managed service streams, then we need more changes here.

> We'll see! :-)
>
Heh, I looked and didn't see anything, so I got nervous and assumed I must have forgotten something. I said in the description that ROOT_COMPONENTS_CONDITION is used in lots of places, but it actually looks like only a few places use it.

PR #554, opened 2023-07-20
CORGI-739: Call save_product_taxonomy() in another task to reduce risk of timeouts

juspence:
@RedHatProductSecurity/corgi-devs Please review another "fix" that won't actually fix anything. This doesn't make the taxonomy-saving code any faster or fix the root cause of this bug, it just tries to avoid it in some circumstances.

If this looks good I'll merge this tomorrow morning, then merge #539 to delete the bad Component-Variant links, then trigger save_product_taxonomy() for only the latest build IDs with ERRATA-type relations. Saving the taxonomy again will add back the missing Variant links, and then the data should be linked correctly for IR.

There are other usages of this taxonomy-saving code, but I will update those in a future PR. The usages I update here should be the only ones we hit when reingesting a build.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -41,26 +41,19 @@ def slow_fetch_brew_build(
 ):
     logger.info("Fetch brew build called with build id: %s", build_id)
 
-    try:
-        softwarebuild = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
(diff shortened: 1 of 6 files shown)


jasinner (approved):
Looks good to me. Some ideas for improvements might be to move the save_taxonomy task to the cpu queue and workers which have more resources and seem to be idle more often.

We might also consider increasing the timeout for this task to the maximum.

PR #555, opened 2023-07-26
CORGI-729: Fix Gunicorn worker out-of-memory issues

juspence:
This will move the sources, provides, and upstreams data to separate (paginated) endpoints. These fields will now show a link to this endpoint for the current component, instead of displaying all the component's sources / provides / upstreams data inline.

This is needed to avoid OoM issues in Gunicorn which trigger "bad gateway" errors in Nginx. These errors happen somewhat randomly for different queries, whenever the amount of data returned / the number of sources, provides, and upstreams on one page is very large. The error is consistent for a particular query as long as the list of components doesn't change. 

This does require a breaking API change, but is the smallest change I could make that actually solves the issue. We can't paginate these fields currently because the top-level request is already paginated. Using ?exclude_fields=sources doesn't work for all usecases, since e.g. Griffon sometimes needs to see this data.

@JimFuller-RedHat is on PTO I think, but let's discuss this change when you're back to make sure it won't cause more issues for Griffon.

I also need to add separate tests for the new endpoints, but I have lots of other stuff to do and there's no point working on tests before we figure out if this is OK for Griffon. Leaving as a draft for now.

### DIFF
--- a/corgi/api/serializers.py
+++ b/corgi/api/serializers.py
@@ -351,6 +351,20 @@ class SoftwareBuildSummarySerializer(IncludeExcludeFieldsSerializer):
 
 
 class ProductTaxonomySerializer(IncludeExcludeFieldsSerializer):
+    tags = TagSerializer(many=True, read_only=True)
+    link = serializers.SerializerMethodField()
+
+    products = serializers.SerializerMethodField()
+    product_versions = serializers.SerializerMethodField()
+    product_streams = serializers.SerializerMethodField()
+    product_variants = serializers.SerializerMethodField()
+    channels = serializers.SerializerMethodField()
+
+    @staticmethod
+    @abstractmethod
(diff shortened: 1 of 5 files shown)


jasinner (approved):
LGTM, thanks!

JimFuller-RedHat (approved):
LGTM and it is fine to break griffon on stage for now - we will use stage to fix and coordinate a release with corgi (when it pushes to prod)

JimFuller-RedHat:
recent conversations remind me of previous attempt at /provides (will send example link separately)

that includes a node_type (ex. PROVIDES) ... unclear how we might include in a normal paginated list of components but wanted to raise here that we might have to think about it eg. a component entity cannot provide this as it might participate in several 'trees' so the value has to be derived in context.

PR #556, opened 2023-07-26
CORGI-693: Set namespace correctly for Maven components, and fix old data

juspence:
@RedHatProductSecurity/corgi-devs Please review. If this looks right I'll add a test tomorrow before I merge, but I wanted to double-check first:

1. Is it true that all Maven components built by PNC / middleware teams will always have ".redhat" or "-redhat" in their version string? Are there any components we build which don't have these indicators? Or components which have it in e.g. the release instead of the version?

2. Are there any other magic values which indicate a component is built at Red Hat? Right now we assume all RPMs are built at Red Hat, and this PR adds the check for Maven component versions. Everything else is assumed to be upstream.

3. Should we set namespace to REDHAT for e.g. Github components under the RedHat, RedHatProductSecurity, etc. orgs?

4. What about Golang components with "redhat/" or "redhat.com" somewhere in their name / URL?

I guess @kgrant-rh is our middleware expert. Please let me know if I've missed something, or if you think I should double-check with those teams directly.

EDIT: I also added some fixes for errors in this morning's monitoring email. I think we don't need a new test, since we already test SCA creates Maven components with the correct purl. I added "redhat/" to that value and the test began passing again.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -167,7 +167,7 @@ class Brew:
             c = component
         else:
             return ""
-        # Strip right parens
+        # Strip right parens, even if we didn't strip prefix
         return c.replace(")", "")
 
     @staticmethod
@@ -292,9 +292,9 @@ class Brew:
         # TODO: list all components used as build requirements
         return srpm_component
 
-    @staticmethod
+    @classmethod
     def _parse_bundled_provides(
-        bundled_provides: list[tuple[Component.Type, str, str]], rpm_info: dict[str, str]
+        cls, bundled_provides: list[tuple[Component.Type, str, str]], rpm_info: dict[str, str]
     ) -> list[dict[str, Union[str, dict[str, Union[str, list[str]]]]]]:
         """Parse a list of (type, name, version) tuples, build a list of bundled component dicts"""
         id_counter = 0
@@ -310,7 +310,7 @@ class Brew:
 
             bundled_component: dict[str, Union[str, dict[str, Union[str, list[str]]]]] = {
                 "type": component_type,
-                "namespace": Component.Namespace.UPSTREAM,
+                "namespace": cls.check_red_hat_namespace(component_type, version),
                 "meta": bundled_component_meta,
             }
             # We can't set go_component_type here for Golang components
@@ -318,6 +318,19 @@ class Brew:
             # There's no easy way for us to tell which type this component is
             parsed_provides.append(bundled_component)
         return parsed_provides
+
+    @staticmethod
+    def check_red_hat_namespace(component_type, version) -> Component.Namespace:
+        """Given a component type and version, return the correct namespace"""
+        if component_type == Component.Type.RPM:
+            # RPMs are always built at Red Hat
+            return Component.Namespace.REDHAT
+        elif component_type == Component.Type.MAVEN and "redhat" in version:
+            # .redhat or -redhat in the version string indicate this Maven component
+            # was built in a Red Hat build system (e.g. by PNC / for a Middleware product)
+            return Component.Namespace.REDHAT
+        else:
+            return Component.Namespace.UPSTREAM
 
     @staticmethod
     def _build_archive_dl_url(filename: str, build_info: dict[str, str]) -> str:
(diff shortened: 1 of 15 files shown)


corgi/collectors/brew.py
```diff
@@ -168,6 +168,7 @@ def _bundled_or_golang(component: str) -> str:
         else:
             return ""
         # Strip right parens
+        # TODO: We strip right parens from golang( too
```
juspence: This function is supposed to process only components like "bundled(name)" - we strip the bundled( prefix. It's supposed to ignore components like "golang(name)" and not strip the golang( prefix, based on the header comment.

But for both types, we strip the right ) parenthesis, even if we didn't strip the prefix. Am I misunderstanding or is that a bug?

corgi/collectors/brew.py
```diff
@@ -868,6 +884,7 @@ def _filter_by_type(
                 remaining_deps.remove(dep)
         return filtered, remaining_deps
 
+    # TODO: Why is this method unused?
     @staticmethod
     def get_maven_build_data(
         build_info: dict[str, Any], build_type_info: dict[str, Any]
```
juspence: This is all Maven-related stuff, but it's not being used. I guess this is the logic for processing Maven wrapper / placeholder builds in Brew. Should we remove this code if we're not going to use it?

corgi/collectors/brew.py
jasinner: It does look weird, but if you look at the associated test data it seems it's expected we'll have mismatched parenthesis.

corgi/collectors/brew.py
jasinner: Yeah, let's remove it.

corgi/core/migrations/0081_fix_maven_namespaces.py
```diff
@@ -0,0 +1,35 @@
...
+    """Set namespace to REDHAT for Maven components with "redhat" in their version strings"""
+    Component = apps.get_model("core", "Component")
+    ComponentNode = apps.get_model("core", "ComponentNode")
+
+    # Values in current data are always -redhat or .redhat
+    maven_components = Component.objects.filter(
+        type="MAVEN", namespace="UPSTREAM", version__contains="redhat"
+    )
+    upstream_maven = Value("pkg:maven/")
+    redhat_maven = Value("pkg:maven/redhat/")
```
jasinner: Instead of modifying the namespace with a `redhat/` suffix let's use the repository_url qualifier ie:

`pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?repository_url=repo.spring.io%2Frelease`

From https://github.com/package-url/purl-spec/blob/master/README.rst#some-purl-examples

In our case the repository_url for Red Hat Maven repository is:

`maven.repository.redhat.com/ga/`

From https://access.redhat.com/maven-repository

jasinner (commented):
Looks good. Although please check the use of 'redhat' in Maven purl namespace.

corgi/core/migrations/0081_fix_maven_namespaces.py
juspence: Oops, thanks for the review and thanks for catching this. I agree we should use the repository_url qualifier for the namespace, instead of a redhat/ prefix at the start of the purl.

But on that page you linked, it says that there are actually two repos - one for GA releases and one for Early Access releases. I guess in general, Early Access components will eventually become GA, and we don't want purls to change over time. So I think it makes sense to use the GA repo's URL even if the component isn't actually GA / available there yet.

But I wonder if this will cause confusion for our users (and customers using our manifests). They might try to find some Early Access component in the GA repo, then fail because it doesn't exist there yet. I guess we can't really fix this, since we don't know ourselves whether something is GA yet or still Early Access. There's also a case where some Early Access component never makes it to GA, for example if a product goes end-of-life without promoting all the Early Access components in its final release.

I'll go ahead and add the repository_url to the purl, assuming that any Red Hat components are always in the GA repo. But please let me know if you think there's a better approach.

jasinner: Using the GA repository seems like the only reasonable think to do, since we won't be able to tell which repository a component is destined for at build time. I'm confident most of the time the components will be destined for the GA repository though.

PR #557, opened 2023-08-01
CORGI-554: Improve component type detection 

kgrant-rh:
Some components are labeled generic when their type can probably be better inferred.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -178,6 +178,19 @@ class Brew:
             return Component.Type.MAVEN, component.replace(":", "/"), version
         return None
 
+    @staticmethod
+    def _check_npm_component(
+        component: str, version: str
+    ) -> Optional[tuple[Component.Type, str, str]]:
+        # NPM bundled packages appear to follow the pattern of
+        # "rh-nodejsXX-component", where component is something
+        # other than "nodejs"
+        if "nodejs" in component:
+            component_match = re.match(r"^rh-nodejs\d+(?!-nodejs)-(.*)", component)
+            if component_match:
+                return Component.Type.NPM, component_match.group(1), version
+        return None
+
     @classmethod
     def _get_bundled_component_type(
         cls, component_type: str, component: str
@@ -206,6 +219,10 @@ class Brew:
             if not component:
                 continue
             bundled_component = cls._check_maven_component(component, version)
+            if bundled_component:
+                bundled_components.append(bundled_component)
+                continue
+            bundled_component = cls._check_npm_component(component, version)
             if bundled_component:
                 bundled_components.append(bundled_component)
                 continue
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -180,14 +180,22 @@ class Brew:
             return Component.Type.MAVEN, component, version
         elif component.startswith("apache-commons"):
             return Component.Type.MAVEN, component, version
+        elif component.startswith("java-"):
+            return Component.Type.MAVEN, component, version
         return None
 
     @staticmethod
     def _check_npm_component(
         component: str, version: str
     ) -> Optional[tuple[Component.Type, str, str]]:
-        if "nodejs" in component:
-            component_match = re.match(r"^nodejs\d*-(.*)", component)
+        if component.startswith("js-"):
+            return Component.Type.NPM, component[len("js-") :], version
+        elif component.startswith("npm-"):
+            return Component.Type.NPM, component[len("npm-") :], version
+        elif component.startswith("nodejs-"):
+            return Component.Type.NPM, component[len("nodejs-") :], version
+        else:
+            component_match = re.match(r"^nodejs\d+-(.*)", component)
             if component_match:
                 return Component.Type.NPM, component, version
         return None
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -176,19 +176,20 @@ class Brew:
     ) -> Optional[tuple[Component.Type, str, str]]:
         if ":" in component:
             return Component.Type.MAVEN, component.replace(":", "/"), version
+        elif component.startswith("maven"):
+            return Component.Type.MAVEN, component, version
+        elif component.startswith("apache-commons"):
+            return Component.Type.MAVEN, component, version
         return None
 
     @staticmethod
     def _check_npm_component(
         component: str, version: str
     ) -> Optional[tuple[Component.Type, str, str]]:
-        # NPM bundled packages appear to follow the pattern of
-        # "rh-nodejsXX-component", where component is something
-        # other than "nodejs"
         if "nodejs" in component:
-            component_match = re.match(r"^rh-nodejs\d+(?!-nodejs)-(.*)", component)
+            component_match = re.match(r"^nodejs\d*-(.*)", component)
             if component_match:
-                return Component.Type.NPM, component_match.group(1), version
+                return Component.Type.NPM, component, version
         return None
 
     @classmethod
@@ -218,6 +219,8 @@ class Brew:
             component = cls._bundled_or_golang(component)
             if not component:
                 continue
+            if component.startswith("rh-"):
+                component = component[3:]
             bundled_component = cls._check_maven_component(component, version)
             if bundled_component:
                 bundled_components.append(bundled_component)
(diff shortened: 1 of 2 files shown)


corgi/collectors/brew.py
```diff
@@ -206,6 +219,10 @@ def _extract_bundled_provides(
             if not component:
                 continue
             bundled_component = cls._check_maven_component(component, version)
+            if bundled_component:
+                bundled_components.append(bundled_component)
+                continue
+            bundled_component = cls._check_npm_component(component, version)
```
juspence, line 236: I think the real bug is around line 242 below. At that point, we've stripped a prefix like "bundled(", and should have something like `component_type == "rh"`, and `component == "nodejs12-zlib"`, with a dash - as our separator.

We'd want component_type to be "nodejsXXX" or "rh-nodejsXXX" instead, and then we'd add a case to _get_bundled_component_type(). If it finds either an exact match like "nodejs" (requires more parsing), or a component_type starting with "rh-nodejs" or "nodejs", then we can set the bundled_component_type to Component.Type.NPM instead of GENERIC.

I'd rather not special-case things too much since this code is already a little complex. Or if you see a way to simplify it, feel free to do that refactoring.

We should also add some tests for this - I can dig up a few more examples of this bug if that will help.

Fianlly, I see in the CACHITO_PKG_TYPE_MAPPING, we handle many different types but not "maven" or "java". It's fine if you want to do one PR for just the node.js issue first, but I think we should also handle those cases. I mentioned them in the description but didn't give an example, sorry.

kgrant-rh: You're right about the situation where after splitting on dash, we end up with `component_type == "rh"` and `component == "nodejs12-nodejs-zlib"`. The issue I encountered trying to solve that is that there are many components with similarly structured names but perhaps different semantics, for example:

```
['rh-eclipse-guice-assistedinject', 'rh-nodejs6-nodejs-preserve', 'rh-eclipse-ed25519-java', 'rh-eclipse46-eclipse-oprofile', 'rh-postgresql12-container', 'rh-ror41-rubygem-rack', 'rh-java-common-netty', 'rh-postgresql10-postgresql-plpython', 'rh-eclipse-xmlgraphics-commons', 'rh-php73-runtime', 'rh-maven35-rhino', 'rh-maven35-osgi-compendium', 'rh-git218-git-instaweb', 'rh-eclipse46-eclipse-mylyn', 'rh-postgresql96-postgresql-static', 'rh-nodejs8-nodejs-yargs', 'rh-php71-php-mbstring', 'rh-nodejs6-nodejs-realize-package-specifier', 'rh-perl526-perl-Compress-Bzip2', 'rh-ruby24-ruby-libs']
```

It seems like these follow the general pattern of `rh-{framework}-{package}`, but even the randomly selected subset above illustrates the problems with trying to handle these generally. Some specify a specific framework/bundle release (`rh-eclipse46-eclipse-oprofile`) while some don't (`rh-eclipse-xmlgraphics-commons`).

My first inclination would be to just lstrip "rh-", since I don't think that'll ever be a meaningful component type, but the meaning of `3` on [line 231](https://github.com/RedHatProductSecurity/component-registry/pull/557/files#diff-35566a508ba49d2e035c1d2f80337b84909bf7628df7e9082563766cd96b3927R231) is not immediately clear to me, and I wouldn't want to break whatever case that's handling.

jasinner: I suspect the meaning of 3 of line 231 is just that we aren't going to try and detect the type where there are more than 2 dashes. If you think you can do a better job of detecting the type by lstrip- "rh-" by all means, give it a go.

juspence: Sorry for the delay in responding here, and thanks for providing the details about all these edge cases. I agree we can't (and shouldn't try to) parse all of these successfully. So I think you're right, and stripping `rh-` will get us better data in most cases without being too much work.

We call re.split() with maxsplit=1, so 1 dash, 2 dashes, or 3+ dashes doesn't affect the result. We should always end up with a component_type (prefix), a separator that's either a left ( parenthesis or a - dash, and a component (package name), after we unpack the re.split() result tuple on line 235.

The "!= 3" case handles values we want to split, but that don't have any matching separator. In that case we assume the Component is GENERIC type. I think this code is correct and shouldn't need any changes.

I hope if I'm understanding the code correctly, stripping "rh-" before calling re.split() will give us `component_type`s like "eclipse", "nodejs6", "postgresql12", "ror41", and so on instead of "rh". `component` names would still have a lot of extra stuff, like "nodejs-realize-package-specifier" or "common-netty".

But there's nothing we can do to get better names, given how inconsistent this is. We'd move the "nodejs6" or "java" to the component_type instead of the component (name), and that's good enough to detect the right component type in Corgi. Having inconsistent names is an engineering problem.

Then we could add some logic to map only the most common component_types. For example, values starting with nodejs become NPM components, values starting with java / maven / eclipse become MAVEN components, and so on. Unmapped values will still default to GENERIC type, and we can handle these edge cases depending on priority / frequency.

We'd check what the component_type starts with, instead of an exact match, to handle both components that report a specific framework/bundle release (`eclipse46`) and components that don't (`eclipse`) like you mention above.

kgrant-rh: In this latest change, I've kept the separate handling for each component type because without doing so, we lose information. For example, given a bundled component named `bundled(rh-nodejs14-icu)`, if we strip the leading `rh-` and then split on `-`, end up with a component_type of `nodejs14` and a component of `icu`. We can determine that this is an NPM package by matching `nodejs*` in the component_type, but then we lose any distinction between say `nodejs14-icu` and `nodejs16-icu`, because the component_type is split.

For that reason, I think it's better to detect these cases before splitting the component into component_type and component, but please let me know if there's anything I'm missing about that approach.

corgi/collectors/brew.py
```diff
@@ -176,6 +176,20 @@ def _check_maven_component(
     ) -> Optional[tuple[Component.Type, str, str]]:
         if ":" in component:
             return Component.Type.MAVEN, component.replace(":", "/"), version
+        elif component.startswith("maven"):
+            return Component.Type.MAVEN, component, version
```
juspence, line 180: Could we also return a MAVEN component_type for anything starting with "java(" or "java-" or whatever form those dependencies use? I'm not suggesting "java" since that might also catch "javascript(", which we obviously don't want.

Or maybe there are no "java(" components. I thought I saw a few but might be misremembering.

corgi/collectors/brew.py
juspence: Oops, good point. You're correct, that distinction is important and your approach above sounds good. We do want to treat `nodejs14-icu` and `nodejs16-icu` as separate components with different nodejs framework versions, even though they have the same "icu" base component name.

tests/test_brew_collector.py
```diff
@@ -873,18 +885,27 @@ def test_parsing_bundled_provides():
         (Component.Type.PYPI, "django-stubs"),
         (Component.Type.PYPI, "selectors2"),
         (Component.Type.NPM, "@babel/code-frame"),
-        (Component.Type.NPM, "yargs-parser"),
+        (Component.Type.NPM, "nodejs-yargs-parser"),
```
juspence: This looks like a regression. If we have just `nodejs-thing`, without any version number, is it possible to keep stripping the "nodejs-" prefix since that won't lose version information? I'm worried if we stop stripping this "nodejs-" prefix, we'll create duplicate components.

In the case where there is a version number, like "rh-nodejs12-zlib" below, stripping just "rh-" and keeping "nodejs12-" is fine, since that preserves the version information.

corgi/collectors/brew.py
```diff
@@ -176,6 +176,20 @@ def _check_maven_component(
...
+        return None
+
+    @staticmethod
+    def _check_npm_component(
+        component: str, version: str
+    ) -> Optional[tuple[Component.Type, str, str]]:
+        if "nodejs" in component:
+            component_match = re.match(r"^nodejs\d*-(.*)", component)
+            if component_match:
+                return Component.Type.NPM, component, version
```
juspence: I think in our Cachito type mapping, we handle other forms of nodejs components, like "npm", "js", and "javascript". I'm not sure if those forms are also used for bundled dependencies in specfiles, but maybe it's better to be thorough and add a case for them anyway? Up to you I guess.

I think there's also no need for the "if nodejs in component", since you can rely only on the re.match() logic.

juspence (approved):
LGTM, thanks for working on this!

PR #558, opened 2023-08-02
CORGI-726 keep products updated

jasinner:
Another attempt at fixing CORGI-726. I've added many tests this time and pushed the many-to-many remove calls down into the component level, which should make the changes more robust.

This is not very efficient where brew_tags are moved between streams because all of the builds related to the brew_tag will have their product taxonomy updated, and all relations deleted. The brew_tag will then be processed as if new in the new stream.

I need to add similar code for yum_repositories, errata_info, and composes. This only addresses the moving of brew_tags between streams, or removal from a stream.

I also added support from removal of a product_stream, product_version, or product from product_definitions.


jasinner:
I pushed another commit which makes use of the new remove_product_from_build task to clean up the CDN relations.

kgrant-rh (approved):
Looks good. I wonder if we could implement `disassociate_with_product` in the managers of Product, ProductVersion, etc. to handle the recursion there, but that's just me bikeshedding, and it's definitely easier to understand what updates are being made having it all in one method.

corgi/core/constants.py
```diff
@@ -28,6 +28,12 @@
     value.title().replace("_", ""): key for key, value in NODE_LEVEL_MODEL_MAPPING.items()
 }
 
+# Map node levels defined above to component many-to-many attribute names as defined in models.py
+# "product_version" -> "productversions"
+NODE_LEVEL_MODEL_MAPPING = {
```
juspence: This is redefining a name that we already used above. I think it should be named NODE_LEVEL_ATTRIBUTE_MAPPING instead.

juspence:
I'm not sure this is the right approach, but probably I just don't understand it. The code we had before moved Variants from one Stream to another, and also updated the parent Versions and Products if needed / if they changed.

I think the problem with that code was the bug with calling variant.save_product_taxonomy() which caused things to break. And we had a test that asserted "we don't save the taxonomy / this error is not seen", but we weren't really avoiding the problem. We were just hiding it until some other code did end up saving the taxonomy.

Can we start from the code we had before, and just fix that bug in the taxonomy? I think this was a side-effect of Variants moving from Stream A to Stream B, then back to Stream A. More robust checks in our data could prevent us from moving Variants back and forth, so that we don't link things incorrectly in this case. Or we could allow streams to move back and forth, but just make sure the taxonomy at the end is unchanged.

I also think it would be helpful, before changing any code, to write a script which parses the current prod-defs, Errata Tool, and Corgi data. That would help us understand:

1. How many prod-defs models are there overall?
2. How many Errata Tool variants are there overall?
3. How many Products / Versions / Streams are still defined in Corgi, but have been removed from prod-defs?
4. Do any Streams have linked Variants (from Brew tags or ET Variant names) in Corgi, which don't match brew_tags or errata_info in prod-defs? In other words, do any Corgi streams still have a reference to some Variant, even though the Variant shouldn't exist due to updated prod-defs data?

5. How many streams have only brew_tags, how many have errata_info, and how many have both?
6. Do any streams in prod-defs reuse a Brew tag that's already defined on another stream? This would cause Variants (which Corgi creates from the Brew tag) to move back and forth between Streams.
7. In the prod-defs data, do any Brew tags for stream X match any ET variant names for stream Y? This would also cause Corgi Variants to move back and forth between Corgi Streams.


To explain number 7 a little more, imagine we create a new Variant when parsing stream X's brew_tags in prod-defs. Then when parsing stream Y's errata_info in prod-defs, we'd find and update the existing Variant with the same name, unlinking it from Corgi's Stream X and relinking it to Corgi's Stream Y.

The next time we processed prod-defs, we'd move the Corgi Variant to Stream X while processing brew_tags, then move it back to Stream Y while processing errata_info. This back-and-forth movement works OK today (we think). But the earlier code change to let variants "drift" to another parent stream (when prod-defs is manually updated) seemed to conflict a bit with this edge-case.

I think it's more helpful to understand what the bug in the old code was, and what the edge cases in our data are, before we try to make more changes. My big concern is that we write some complex linking code we don't really understand, which might trigger bugs or create bad data due to edge cases we aren't aware of.

corgi/tasks/prod_defs.py
```diff
@@ -342,8 +422,8 @@ def parse_errata_info(
     """Parse and create ProductVariants from errata_info in product-definitions.json"""
     et_product_versions_set = set(product_stream.et_product_versions)
     for et_product in errata_info:
-        et_product_name = et_product.pop("product_name")
-        et_product_versions = et_product.pop("product_versions")
+        et_product_name = et_product.get("product_name", "")
+        et_product_versions = et_product.get("product_versions", [])
```
juspence: The old code makes sure that we stop processing with an error if the data in prod-defs is formatted incorrectly. The new code you have here will silently ignore any missing product_name or product_versions keys in some stream's errata_info. That seems surprising to me...is there a reason we need this change?

corgi/tasks/prod_defs.py
```diff
@@ -236,7 +312,7 @@ def parse_product_stream(
             "cpes_matching_patterns": [],
         },
     )
-    product_stream_node, _ = ProductNode.objects.get_or_create(
+    product_stream_node, _ = ProductNode.objects.update_or_create(
```
juspence: If we're updating the stream's node instead of just getting it, we'll change the parent node to a different ProductVersion, whenever the stream moves in prod-defs.

I think it's better to be explicit, and only support these types of changes / data movement in a separate method, like you had before. That would let us reprocess all the relations / builds that reference this stream, and update their Product, Version, etc. links whenever the Stream moves to a different parent Version.

In the current code, we ignore the second "created" value that update_or_create returns, so we won't know when the stream was updated to a different parent and won't be able to make any other updates needed.

corgi/tasks/prod_defs.py
```diff
@@ -159,12 +197,12 @@ def parse_product_version(
     pd_product_version: dict[str, Any], product: Product, product_node: ProductNode
 ):
     """Parse the product versions from ps_modules in product-definitions.json"""
-    pd_product_streams = pd_product_version.pop("product_streams", [])
-    name = pd_product_version.pop("id")
+    pd_product_streams = pd_product_version.get("product_streams", [])
```
juspence: Doing .get() here silently hides an error due to missing keys. See other comment for details, but let's fix this one too, unless we need this for some reason. If this really is needed / some products don't define the product_streams key, I'd rather fix the prod-defs data to be valid, and let Corgi fail with an error when it's not.

corgi/tasks/prod_defs.py
```diff
@@ -205,20 +277,24 @@ def parse_product_stream(
...
+    exclude_components = pd_product_stream.get("exclude_components", [])
     composes_dict = {}
     for compose in composes:
         composes_dict[compose["url"]] = compose["variants"]
-    name = pd_product_stream.pop("id")
+    name = pd_product_stream["id"]
     if match_version := RE_VERSION_LIKE_STRING.search(name):
         version = match_version.group()
     logger.debug("Creating or updating Product Stream: name=%s", name)
+    _clean_existing_stream_brew_tags(name, brew_tags_dict.keys())
```
juspence: Shouldn't we do this cleanup after we create the Stream in Corgi?

corgi/tasks/prod_defs.py
jasinner: The reason is that we reuse the dict again at the end of update_product function to reconcile what's in the Corgi DB with what we just read from product_definitions. If we use pop then all the keys are missing from the original dictionary.

jasinner:
> * How many prod-defs models are there overall?

new_product_names = 110
new_version_names = 349
new_stream_names = 1276

> * How many Errata Tool variants are there overall?

cm.CollectorErrataProductVariant.objects.get_queryset().count() = 1935

> * How many Products / Versions / Streams are still defined in Corgi, but have been removed from prod-defs?

None, but that is on the horizon with prod-defs `merge_requests/2234`

> * Do any Streams have linked Variants (from Brew tags or ET Variant names) in Corgi, which don't match brew_tags or errata_info in prod-defs? In other words, do any Corgi streams still have a reference to some Variant, even though the Variant shouldn't exist due to updated prod-defs data?

In this commit [1] rhaos-middleware-rhel-8-container was incorrectly added to the cyrostat-2.0 ps_update_stream. I suspect that was cleaned up by the [changes merged](https://github.com/RedHatProductSecurity/component-registry/pull/507) previously for CORGI-726.
The same situation would occur once prod-defs `merge_requests/2234` was merged.

[1] baba3db65f560b801f1b68d8c5ff5c933cdd132c

> * How many streams have only brew_tags, how many have errata_info, and how many have both?

m.ProductStream.objects.exclude(et_product_versions=[]).count() = 46
m.ProductStream.objects.exclude(brew_tags={}).count() = 294
rhel-8.8.0.z is the only one which has both errata_info and brew_tags.

> * Do any streams in prod-defs reuse a Brew tag that's already defined on another stream? This would cause Variants (which Corgi creates from the Brew tag) to move back and forth between Streams.

This seems to be the main cause of the remaining data issues not resolved by [this PR](https://github.com/RedHatProductSecurity/component-registry/pull/507). Judging by the differences remaining between the Corgi models and prod-defs data generated by compare_variants.py script attached to CORGI-726.

example variant_differences:
```
{
    # Brew tag shared with another stream rhacm-2.7.z
    'rhacm-2.7': {'8Base-RHACM-2.7'},
    # Brew tag shared with another stream fdp-el7-default
    'fdp-el7': {'7Server-RH7-Fast-Datapath'},
    # Brew tag shared with multiple streams including jaeger-1.20.4
    'jaeger-1.20.3': {'8Base-JAEGER-1.20'},
    # Brew tag shared by multiple streams including openjdk-11
    'openjdk-17': {'7Server-OpenJDK-TextOnly'},
    # Brew tag shared with another stream devworkspace_operator-0
    'codeready_ws-2-default': {'8Base-DEVWORKSPACE-1.0'},
    # Brew tag shared with another stream rhacm-2.5.z
    'rhacm-2.5': {'8Base-RHACM-2.5'},
    # Brew tag shared by quarkus-2
    'quarkus-2.13': {'8Base-MW-RHBQ-2.13'},
    # Brew tag shared by pipelines-1.7.1
    'pipelines-1.7': {'8Base-PIPELINES-1.7'},
    # Brew tag shared by gitops-1.2.z
    'gitops-1.2': {'8Base-GitOps-1.2'},
    # Brew tag shared by openstack-13-els
    'openstack-13': {'7Server-RH7-RHOS-DEVTOOLS-13.0-ELS', '7Workstation-RH7-RHOS-TOOLS-13.0-ELS',
                  '7Server-RH7-RHOS-13.0-ELS', '7Workstation-RH7-RHOS-TOOLS-13.0', '7Server-RH7-RHOS-13.0',
                  '7Server-RH7-RHOS-TOOLS-13.0-ELS', '7Server-RH7-RHOS-DEVTOOLS-13.0',
                  '7Client-RH7-RHOS-TOOLS-13.0-ELS', '7Server-7-7.6.EUS-RH7-RHOS-13.0',
                  '7Server-RH7-RHOS-DEPLOYMENT-TOOLS-13.0', '7Client-RH7-RHOS-TOOLS-13.0',
                  '7Server-RH7-RHOS-DEPLOYMENT-TOOLS-13.0-ELS', '7Server-RH7-RHOS-TOOLS-13.0'},
   ....
}
```

> * In the prod-defs data, do any Brew tags for stream X match any ET variant names for stream Y? This would also cause Corgi Variants to move back and forth between Corgi Streams.

I don't think so, except for rhel-8.8.0.z which has both brew_tags and errata_info which overlap. The values in the Corgi data seem to come from the brew_tag variants, not errata_info. 

There are some variants currently linked to dts-10.1.z, and dts-12.1.z which shouldn't be. Those should be skipped during parse_variants_from_brew_tags. That was also a change introduced in [this PR](https://github.com/RedHatProductSecurity/component-registry/pull/507). Which needs to be cleaned up.

corgi/tasks/prod_defs.py
jasinner: Cleaning up before-hand allows any moved variants to be recreated in the new stream. Anyway, I think we should probably remove this change from the PR.

corgi/tasks/prod_defs.py
jasinner: We use product_streams value again at the end of the look to check for deleted streams.

corgi/tasks/prod_defs.py
jasinner: You are right, I'll remove most of this commit from the PR.

corgi/core/constants.py
jasinner: Ok, I'll fix this.

jasinner:
The changes remaining here seems to do a good job of cleaning up the remaining CDN relations from components. I ran this last week in dev using product data and the deptopia-reconciliation script is now showing that we have less extraneous packages remaining. The ones that are remaining seem to be caused by CORGI-778

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
         """Return only the purls from the set of all upstream nodes"""
         return set(node.purl for node in self.get_upstreams_nodes(using=using))
 
+    def disassociate_with_product(self, product_ref: ProductModel):
+        """Disassociate this component with the passed in ProductModel and any child ProductModels
+         in that product's hierarchy. This is the reverse of what happens in save_product_taxonomy
+        and is used when a product is removed from product_definitions, or moved to a new place
+        in the product hierarchy, which also requires re-processing those products from scratch
+        Removing proudctvariants is a no-op where a stream does not have errata_info. It's easier to
+        just remove productvariants everytime then check for errata_info in a stream.
```
juspence: This comment isn't right...we link Variants from brew_tags as well as Variants from errata_info. I might have lost track of this code, should we meet and discuss these changes? Sorry for constantly being confused by all this.

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+        """
+        if isinstance(product_ref, Product):
+            self.productvariants.remove(*product_ref.productvariants.get_queryset())
+            self.productstreams.remove(*product_ref.productstreams.get_queryset())
+            self.productversions.remove(*product_ref.productversions.get_queryset())
```
juspence, line 2008: Minor: Since Components define many-to-many relationships with ProductModels, you can use `*product_ref.productmodelfield.values_list("pk", flat=True)` instead.

This avoids fetching all data for every related object from the DB. Only fetching the primary keys might avoid possible performance issues. I'm not really sure how this will perform with data volume in stage. It might be OK even without this change.

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+            self.productstreams.remove(*product_ref.productstreams.get_queryset())
+            self.productversions.remove(product_ref)
+            self._check_and_remove_orphaned_product_refs(product_ref, "Product")
+        elif isinstance(product_ref, ProductStream):
+            self.productvariants.remove(*product_ref.productvariants.get_queryset())
+            self.productstreams.remove(product_ref)
+            self._check_and_remove_orphaned_product_refs(product_ref, "ProductVersion")
+            self._check_and_remove_orphaned_product_refs(product_ref, "Product")
+        elif isinstance(product_ref, ProductVariant):
+            self.productvariants.remove(product_ref.productvariants)
```
juspence: Should this just be `product_ref`, or `*product_ref.productvariants.get_queryset()`?
```suggestion
            self.productvariants.remove(product_ref)
```

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+            self._check_and_remove_orphaned_product_refs(product_ref, "ProductStream")
+            self._check_and_remove_orphaned_product_refs(product_ref, "ProductVersion")
+            self._check_and_remove_orphaned_product_refs(product_ref, "Product")
+        else:
+            raise NotImplementedError(
+                f"Add class {type(product_ref)} to Component.disassociate_with_product()"
+            )
+
+    def _check_and_remove_orphaned_product_refs(self, product_ref, product_model_name):
```
juspence: Let's add type hints for this method if possible.

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
         """Return only the purls from the set of all upstream nodes"""
         return set(node.purl for node in self.get_upstreams_nodes(using=using))
 
+    def disassociate_with_product(self, product_ref: ProductModel):
```
juspence: Can you please add a `-> None` return type annotation (or `-> Whatever_Type`, as needed) to all the new methods you've added?

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+            raise NotImplementedError(
+                f"Add class {type(product_ref)} to Component.disassociate_with_product()"
+            )
+
+    def _check_and_remove_orphaned_product_refs(self, product_ref, product_model_name):
+        """Remove product_models from this component where there are no remaining product children
+        or the remaining product children don't share this product_model as a parent."""
+        product_node_level = MODEL_NODE_LEVEL_MAPPING[product_model_name]
+        product_attribute = NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level]
+        query = getattr(self, NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level + 1]).get_queryset()
```
juspence: This is a bit confusing...if I understood below correctly, can we split this out into multiple statements or add a comment?
```suggestion
        child_product_attribute = NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level + 1]
        # e.g. this_component.product_versions.get_queryset()
        child_products_query = getattr(self, child_product_attribute).get_queryset()
```

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+        remaining_product_child_count = query.count()
+        parents_of_remaining_children = query.values_list(product_attribute, flat=True)
```
juspence: You can avoid making an extra query to the DB if we don't need `.distinct()` here:
```suggestion
        parents_of_remaining_children = child_products_query.values_list(product_attribute, flat=True)
        remaining_product_child_count = len(parents_of_remaining_children)
```
Or do we want a distinct() list of the parents?

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+        parent_of_product_ref = getattr(product_ref, product_attribute).pk
```
juspence: product_ref is a ProductModel instance, e.g. some_product_version. parent_of_product_ref should be some_product's primary key value in that case.

product_attribute is a string, based on a mapping of product_node_level to an attribute name, like "1" to "productversions".

product_node_level is a string, based on a mapping of product_model_name to a node level, like "ProductVersion" to "1".

So in this case, aren't we doing `getattr(some_product_version, "productversions")`.pk here? AKA we're returning some_product_version's pk instead of some_product's pk.

Do we need to use a parent_product_attribute instead of this current product_ref's product_attribute? For example:
`parent_product_attribute = NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level - 1]`

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+        if (
+            remaining_product_child_count == 0
+            or parent_of_product_ref not in parents_of_remaining_children
+        ):
+            getattr(self, product_attribute).remove(getattr(product_ref, product_attribute))
```
juspence: I'm not sure this makes sense. We get an attribute like "productversions" from a Component instance. Then we call that attribute's remove() method to get rid of a ProductVersion instance's "productversions" attribute. Can't the second getattr just be replaced with `product_ref.pk`?

Instead of doing all this stuff with getattr which is very indirect, could / should we just pass in the attribute(s) and / or PK values we want to work with?

For example, we could pass self.product_versions instead of product_attribute, self.products instead of parent_product_attribute, and self.product_streams instead of child_product_attribute, using self / this_component as defined in disassociate_with_product.

corgi/core/models.py
```diff
@@ -1986,6 +1999,54 @@ def get_upstreams_purls(self, using: str = "read_only") -> set[str]:
...
+        or the remaining product children don't share this product_model as a parent."""
+        product_node_level = MODEL_NODE_LEVEL_MAPPING[product_model_name]
+        product_attribute = NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level]
+        query = getattr(self, NODE_LEVEL_ATTRIBUTE_MAPPING[product_node_level + 1]).get_queryset()
+        remaining_product_child_count = query.count()
+        parents_of_remaining_children = query.values_list(product_attribute, flat=True)
+        parent_of_product_ref = getattr(product_ref, product_attribute).pk
+        if (
+            remaining_product_child_count == 0
+            or parent_of_product_ref not in parents_of_remaining_children
```
juspence: By definition, parent_of_product_ref will not be in parents_of_remaining_children, if parents_of_remaining_children is an empty queryset / remaining_product_child_count is 0.

So I think this is redundant and only the second condition is needed.

corgi/core/models.py
```diff
@@ -322,6 +324,17 @@ def save_product_taxonomy(self):
...
+    def disassociate_with_product(self, product_model_name: str, product_pk: str):
+        """Remove the product references from all components associated with this build."""
+        product_model = apps.get_model("core", product_model_name)
+        product = product_model.objects.get(pk=product_pk)
+
+        for component in self.components.iterator():
+            component.disassociate_with_product(product)
+            for cnode in component.cnodes.iterator():
+                for descendant in cnode.get_descendants().iterator():
+                    descendant.obj.disassociate_with_product(product)
```
juspence, line 336: Is there a reason we iterate over cnodes? This requires two lookups, one to find the cnode's linked object's content_type, then another to find the cnode's linked object itself (which is always a Component). We can just iterate over component.provides and component.upstreams instead, right?
```suggestion
        for root_component in self.components.iterator():
            root_component.disassociate_with_product(product)
            for provided_component in root_component.provides.iterator():
                    provided_component.disassociate_with_product(product)
            for upstream_component in root_component.upstreams.iterator():
                    upstream_component.disassociate_with_product(product)
```

corgi/tasks/prod_defs.py
juspence: Maybe we should leave this change for a separate MR. But I think it's better to keep the .pop() we had in the old code, and fail in Corgi / this task if some value is missing in prod-defs.

We could create a new dict (instead of reusing this one for multiple purposes) to keep track of what we read from prod-defs, and use that at the end to check for orphaned objects. But I also don't think a dict is needed.

We do update_or_create for each ProductModel. After the update_or_create succeeds, we can add the created / updated ProductModel's primary key to a set for that ProductModel type.

At the end of the update_products() task, we can do a query to find all the existing ProductModel objects of a certain type, which are not in the set / were not updated or created. That means they don't exist in prod-defs anymore, so we can just .delete() that queryset. For example:

```
# At the very start of the task
prod_defs_products = set()
prod_defs_product_versions = set()
# and so on
...
# When creating e.g. a ProductVersion
product_version, created = ProductVersion.objects.update_or_create(
    # omitted for clarity
)
# We add the PK to the set, because we processed it
# So this version must still be present in prod-defs data
prod_defs_product_versions.add(product_version.pk)
...
# At the very end of the task
# Any ProductVersions in Corgi, which were not created / updated / processed by this task at all,
# must no longer be defined in prod-defs, so we can delete them in Corgi
orphanded_corgi_product_versions = ProductVersion.objects.exclude(pk__in=prod_defs_product_versions)
orphaned_corgi_product_versions.delete()
```

tests/data/prod_defs/proddefs-update-variants-moved.json
```diff
@@ -3,15 +3,15 @@
     "stream": {
       "brew_tags": [
         {
-          "tag": "two_streams_with_same_tag",
+          "tag": "new_tag",
```
juspence: What effect does this have on the existing tests? I thought we wanted to test two streams that have the same tag. I think the logic is:

The first stream creates a ProductVariant from this Brew tag. The second stream finds the existing ProductVariant and doesn't create a new one. The ProductVariant is then updated to move from the first stream to the second stream.

juspence (commented):
Sorry for not "getting" this. I think the logic to disassociate_with_product is mostly OK once some bugs are fixed. I haven't looked at the migration or tests.

The logic to remove orphaned products / versions / etc. has some risks in it, since it changes how we parse the prod-defs data. I think we could find these orphaned ProductModels without any changes to how we parse prod-defs data. But we could also do this in a separate MR if you'd prefer.

corgi/tasks/prod_defs.py
juspence: Ack, see my reply in the other comment.

corgi/core/migrations/0081_clean_cdn_relations.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    variant_names_pk = ProductVariant.objects.filter(name__in=cdn_relation_variants).values(
+        "name", "pk"
+    )
+    variant_names_dict = {npk["name"]: str(npk["pk"]) for npk in variant_names_pk}
+
+    for cdn_relation_build_to_product in (
+        ProductComponentRelation.objects.filter(type="CDN_REPO", software_build_id__isnull=False)
+        .values_list("software_build__pk", "product_ref")
```
juspence: You can do `software_build_id` here, which just reuses the build's PK that's already stored in the relations table instead of having to look up the related model and get the PK from the SoftwareBuild table.
```suggestion
        .values_list("software_build_id", "product_ref")
```
Otherwise this migration looks good. Thank you so much for finding a way to clean up these old CDN repo relations in a migration / script without having to reload all the data!

corgi/core/migrations/0081_clean_cdn_relations.py
```diff
@@ -0,0 +1,56 @@
...
+        if product_ref not in variant_names_dict:
+            continue
+
+        current_remove_product_from_build.delay(
+            str(cdn_relation_build_to_product[0]),
+            "ProductVariant",
+            variant_names_dict[product_ref],
+        )
+
+    ProductComponentRelation.objects.filter(type="CDN_REPO").delete()
```
juspence: Instead of deleting all the CDN repo relations at once, should we delete each relation individually after we schedule the task?

That way if this migration times out in stage, we can run it manually on the first e.g. ~10K relations, let it time out, then run it again on the next ~10K relations. If we wait until the end to delete the relations, rerunning the migration means it will get stuck in a loop processing the same relations over and over.

On the other hand, if the task fails or has some other issue, being able to rerun the migration (and therefore the task) on the same relations / build IDs / variant refs would be useful. Whatever you prefer is fine.

corgi/core/migrations/0081_clean_cdn_relations.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0080_adjust_product_cpe_fields"),
```
juspence: FYI I merged new migrations earlier, so you'll have to rebase this on either number 81 or number 82.

corgi/tasks/prod_defs.py
```diff
@@ -41,33 +48,63 @@ def update_products() -> None:
...
+    for product in products:
+        new_version_names.update([v.get("id") for v in product["product_versions"]])
+        for version in product["product_versions"]:
+            new_stream_names.update(s.get("id") for s in version["product_streams"])
+    orphaned_products = existing_product_names - new_product_names
+    orphaned_versions = existing_version_names - new_version_names
+    orphaned_streams = existing_stream_names - new_stream_names
+    ProductComponentRelation.objects.filter(
+        product_ref__in=orphaned_streams, type__in=ProductComponentRelation.STREAM_TYPES
+    ).delete()
```
juspence: I think we should hold off on cleaning up orphaned products for now. We can do this in a separate MR, but there are a few cases we need to handle.

When deleting a stream or version or product, we'll also delete all the child variants too. So I think this code needs to handle ERRATA-type relations where product_ref is a Variant name. These are now the only VARIANT_TYPE relations after we cleaned up the CDN_REPO-type relations.

Since we're deleting orphaned_streams, we also need to delete relations that point to children of these orphaned streams, like below:
```
variant_children_of_orphaned_streams = ProductVariant.objects.filter(productstreams__name__in=orphaned_streams)
# And the same for variant_children_of_orphaned_versions, orphaned_products, etc.
# And the same for stream_children_of_orphaned_versions, orphaned_products, etc.
```

There's a TODO note on line 61 about deleting Variants that are orphaned. I think we should do that too, and then we also need to _check_and_remove_orphaned_product_refs(). For example, we might delete some orphaned Variant linked by an ERRATUM relation to certain builds, or we might delete some orphaned stream linked by BREW_TAG or COMPOSE relations to certain builds.

We'd need to check all the builds / components that were (formerly) linked to the orphaned model we just deleted. Then we'd remove any parents of those deleted models from the build / component taxonomies, only if they aren't referenced by some other child model that's still present.

We have the logic already written, so it's not a lot of extra work (I think) when deleting a Variant or Stream. There is some complexity if we're deleting a Product or Version - we'll auto-delete all the child streams. We need to check if there any relations for those streams, then clean up the orphaned refs / fix the taxonomies. And the same for deleting a Product, Version, or Stream - check if there are any relations for the child variants that are being auto-deleted, then clean up the orphaned refs / fix the taxonomies.

All of that seems pretty complex, so I think it's less risky to hold off for now and just handle moving things around. We can always do cleanup of orphaned stuff later, but it's harder to undo cleanup if we get it wrong or miss an edge-case.

corgi/tasks/prod_defs.py
jasinner: Ok, I'll remove this clean up code for now, and create a separate issue to follow up with this after.

corgi/core/models.py
jasinner: The comment is no longer accurate. I just kept the first part.

corgi/core/models.py
jasinner: Changing as you suggested results in a mypy error:

```
error: Argument 1 to "remove" of "RelatedManager" has incompatible type "*_QuerySet[ProductVariant, UUID]"; expected "Union[ProductVariant, int]"  [arg-type]
```

corgi/core/models.py
jasinner: ok

corgi/core/models.py
jasinner: ok, good idea.

corgi/core/models.py
jasinner: You are correct, it should be product_ref.

corgi/core/models.py
jasinner: Good idea. Probably want distinct actually, to avoid removing the same parent more than once.

corgi/core/models.py
jasinner: Actually it's the name which is was confusing. I renamed that variable to `ancestor_of_product_ref`. We call `_check_and_remove_orphaned_product_refs` with each of the ancestor model names which are higher (more senior) than product_ref's model in the hierarchy so while the variable does represent some parent of the product_ref, we already know which level we want to check (it's the value of the passed in `str` attribute. This could probably be optimized somewhat by calling of the ancestors in the hierarchy directly in `_check_and_remove_orphaned_product_refs`, which would decrease coupling between these 2 functions but I feel the current way is probably easier to understand, and is already well tested by me in dev.

corgi/core/models.py
jasinner: This hopefully makes more sense after I renamed some of the attributes here. I didn't change any logic.

I originally wrote 3 functions which directly worked with the attribute names instead of `getattr`, but I refactored it to use `getattr` because the logic was shared by the 3 different functions. We can probably split it out to 3 different functions if you think it will be easier to maintain.

corgi/core/models.py
jasinner: You are right, removing the 0 check.

corgi/core/models.py
jasinner: I copied this code from save_product_taxonomy models.py 314-323. I hesitate to deviate from what's used there as this new function is designed to undo the changes made in that function.

tests/data/prod_defs/proddefs-update-variants-moved.json
jasinner: Changes to that test file are no longer necessary, reverting it.

corgi/core/migrations/0081_clean_cdn_relations.py
jasinner: I successfully tested the migration runs to completion in dev, so I think it's not going to timeout. At least not if run manually, and not during pod init.

corgi/tasks/prod_defs.py
jasinner: ok, I removed that commit.

jasinner: I created CORGI-784 to follow up on that.

corgi/core/migrations/0081_clean_cdn_relations.py
jasinner: ok, fixed.

corgi/core/migrations/0081_clean_cdn_relations.py
jasinner: updated, thanks.

corgi/tasks/prod_defs.py
jasinner: Created CORGI-784

corgi/core/models.py
jasinner: ok

jasinner:
replaced by pull/567

corgi/core/models.py
juspence: After thinking about it more, I think you're right and the getattr will be simpler in the end. Thanks for talking through this with me.

PR #560, opened 2023-08-04
CORGI-552: Fix duplicate purl error due to small Syft / Brew NEVRA differences

juspence:
@RedHatProductSecurity/corgi-devs This is honestly pretty terrible code, but there's no point in me staring at it any longer. I hope this will fix the errors we're seeing in our monitoring email, but I'm not confident the fix is right.

Maybe we should just deploy this and see if it helps, then clean it up if the errors go away. Or if there are still other edge cases, I can handle those in a follow-up PR.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -269,7 +269,7 @@
         "filename": "tests/test_sca.py",
         "hashed_secret": "b6bed62bf61a2710e03218decfb836a10a7bce0a",
         "is_verified": true,
-        "line_number": 171,
+        "line_number": 172,
         "is_secret": false
       },
       {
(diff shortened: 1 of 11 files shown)


JimFuller-RedHat (approved):
ya, its hard to know how to handle all this.

FWIW 

https://github.com/package-url/purl-spec/issues/136

shows that case sensitivity heuristics will probably have to be derived from each package ecosystem ... I agree we should try it out in stage (and cross our fingers).

juspence:
Re: above, case sensitivity isn't one of the issues I'm worried about. That's a farily simple fix, and only seems to affect Github and PyPI components anyway (based on testing I did when we first saw this issue).

The bigger problem is purl creation in general. I know that Github and PyPI purls are always lowercased, and I know that PyPI purls have _ underscores converted to - dashes, but I don't know what I don't know. There might be other special rules that cause two different components to end up with the same purl.

Most of the errors in our daily monitoring email are for PyPI components, so I deployed this to stage for testing to see if it would prevent errors for at least some of those. But it actually caused a deadlock in the DB, so I'll need to look more at this next week after I finish other tickets. Putting back into draft for now.

juspence:
I tweaked this slightly and deployed again, then didn't see any more deadlock issues. Those might be a rare problem, or a one-off issue when deploying this change while other tasks were running, or a real bug in the code that's now fixed.

I'm going to merge this as-is since it does seem to help reduce the IntegrityErrors we're seeing. Will open follow-up PRs or back this out as needed.

PR #561, opened 2023-08-08
CORGI-772: Fix miscellaneous SCA and prod-defs bugs

juspence:
@RedHatProductSecurity/corgi-devs Please review. I'm splitting this out from the other PR to fix slow_fetch_brew_build() IntegrityErrors since that's getting a little large. This will:

1. Make sure that existing source URLs on older builds, which use git:// and git+https:// and so on, continue to work after August 20th. OpenLCS relies on these source URLs to get artifacts, but the dist-git service will stop allowing this git:// protocol access when cloning repos.

2. Handle a 404 error in the SCA task correctly. I misread the traceback and added a fix in the wrong place. We need to remove an extra ".git" in the URL sometimes, but this logic is needed in the _download_source() method and not in the _clone_source() method.
EDIT: Actually we need this logic in both places...I'm working on tests now.

3. Handle a different 404 error when running SCA on newer builds, because the source URL has "https://" but not "git/" in it, so cloning the repo using this https:// protocol access fails.
EDIT: Martin suggested we needed "cgit/", but both "cgit/" and regular "git/" forms work. I'm adding regular "git/" to all the source URLs, since that form is already present on several builds, and it's what dist-git seems to use by default when generating links to use with "git clone".

4. Fix dependabot warnings.

5. Fix some miscellaenous prod-defs bugs I found while reviewing other PRs.


corgi/core/migrations/0081_fix_build_sources.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 3.2.18 on 2023-06-19 02:18
+
+import os
+
+from django.db import migrations
+from django.db.models import Value, functions
+
+
+def fix_build_source_urls(apps, schema_editor):
```
JimFuller-RedHat, line 9: perhaps I am missing something - I was expecting an invoke clean_source_of_build() at least as a final check here or is the data migration itself require a different code path ?

juspence: Yeah, it's a little confusing since the code is almost the same in both places. In the migration, we want to clean up existing git:// links in the DB. I also used .update() in the migration to directly manipulate the values in the DB, so that the migration is faster / less likely to time out. This avoids needing to open a transaction, get the build, fix the source URL, and then save the build in multiple steps.

In the Brew collector's clean_source_of_build() method, we want to convert git:// links when reprocessing an older build, so so we don't end up recreating git:// links in the future that the migration removed today. But we're manipulating a Python object that hasn't been saved / created in the DB yet, so we can't use .update() here.

I could reuse the same code for both if the migration wasn't using .update(), but trying to update each object individually has caused migrations to time out in the past.

JimFuller-RedHat: thx for explanation, LGTM!

corgi/tasks/prod_defs.py
```diff
@@ -202,22 +205,26 @@ def parse_product_stream(
...
     composes = pd_product_stream.pop("composes", [])
-    brew_tags_dict = {brew_tag["tag"]: brew_tag["inherit"] for brew_tag in brew_tags}
     exclude_components = pd_product_stream.pop("exclude_components", [])
+
     composes_dict = {}
     for compose in composes:
         composes_dict[compose["url"]] = compose["variants"]
+
     name = pd_product_stream.pop("id")
+    version = ""
```
juspence: This change sets the stream's version to an empty "" string if we can't find a match for RE_VERSION_LIKE_STRING in the stream's name. I thought passing in the parent ProductVersion's "version" was a bug in the old code.

But looking at this again, maybe that's on purpose? We would try to match RE_VERSION_LIKE_STRING against the parent PV's name, and default to that value for the stream as well, if the stream doesn't have a version of its own. I'll remove this one change.

corgi/tasks/prod_defs.py
```diff
@@ -298,6 +307,7 @@ def parse_variants_from_brew_tags(
                     et_pv.name,
                     brew_tag,
                 )
+                et_product_versions_set.add(et_pv.name)
```
juspence: @jasinner I'm not sure about this change anymore, based on your comment in CORGI-778. We add all the ET ProductVersion names to the set when processing a stream's errata_info stanza, but not when processing the stream's brew_tags stanza. I thought that was a bug, so I added the same logic for brew_tags here.

I can merge the SCA fixes without the prod-defs fixes if you'd like to discuss this more, or I can just take it out if you know this change is incorrect. Please just let me know either way - thanks.

jasinner: Thanks for asking. I didn't realize we had this field. If we don't set it here with variants from Brew tags we can use it in save_product_taxonomy to limit which streams we associate with builds having the variant. Also, the variants from Brew tags are only really useful for generating the CPE list, so better not to display it on the API anywhere. I didn't check if we expose this field on the API though.

corgi/tasks/prod_defs.py
jasinner: I think it makes sense to leave it as it was, and pass through the version from ProductVersion's "version". Sometimes there is only one stream for a product version and in that case the value will be correct. I haven't check whether this actually happens in today's  prod_defs data though.

juspence: Yes, I'm going to remove it. I just left this comment as a reminder. I think everything works today - we would see NameErrors about an undefined variable in the update_products() task if not. This is just fixing a warning in the code, and making the task more future-proof if some future stream / version doesn't have a version-like string in its name.

corgi/tasks/prod_defs.py
juspence: Ack, will leave this unchanged as well then. We can easily make this change in the future if needed.

PR #562, opened 2023-08-10
CORGI-764: Don't hide PyPI packages / upstream RPMs with same NEVRA in latest filter

juspence:
@RedHatProductSecurity/corgi-devs Small fix for a hopefully-rare edge case. NEVRAs like "PyYAML-1.2.3.noarch" usually refer to a component with RPM type, UPSTREAM namespace, no release, and no architecture. But these NEVRAs can also refer to e.g. components with PYPI type, UPSTREAM namespace, no release, and no architecture.

So the latest filter needs to consider component type when grouping components, in addition to namespace / name / architecture. The latest filter will then report the latest versons of both the PYPI package and the upstream RPM, instead of only one or the other. After this fix, both will be reported even if they have the same NEVRA and can't be distinguished by namespace / name / architecture alone.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -903,8 +903,8 @@ class ComponentQuerySet(models.QuerySet):
         component_versions = cls._versions_by_name(queryset)
         latest_components = Q()
         # We need to build tuples of the nevras for comparison with the rpm.labelCompare function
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat (approved):
LGTM

PR #563, opened 2023-08-14
CORGI-773: Make all regex filters use case-insensitive lookup

juspence:
@RedHatProductSecurity/corgi-devs Please review this small change to our API. Some IR people requested this as part of GRIF-107. But I'm not sure if this will cause issues for other IR users who expect regexes to do case-sensitive matching by default.

I also don't know if other Corgi users (e.g. OpenLCS, the Open Source Program Office's Community Health team, others?) will be affected by this change. I've left a note in the Component Registry Gchat just in case.


kgrant-rh (approved):
Looks good!

JimFuller-RedHat (approved):
as you might imagine ... LGTM

PR #565, opened 2023-08-15
CORGI-772: Fix a very silly typo that added an extra "/git/" to some source URLs

juspence:
@RedHatProductSecurity/corgi-devs Please review. My last PR had an f-string that joined a prefix like "https://hostname/" with a trailing slash, to a path like "/git/" with a leading slash.

We ended up with a final value like "https://hostname//git/" that has two // slashes, which obviously didn't match anything in the database. This caused the code to fall through to an else case. We changed "https://hostname/" with a trailing slash but no "git/" in the path to "https://hostname/git/", in order to add the "git/" value to the path since it was "missing".

Now 1393 builds in stage have source URLs with a doubled-up "/git/git/". I don't know any legitimate reason why a source URL would have this value, so I'm assuming all of these are a side-effect of the last migration and can be switched back to just one "/git/" value.

### DIFF
--- a/corgi/collectors/brew.py
+++ b/corgi/collectors/brew.py
@@ -145,7 +145,7 @@ class Brew:
         dist_git_url = f"https://{dist_git_hostname}/git/"
 
         for git_url in gitlike_urls:
-            git_url_with_path = f"{git_url}/git/"
+            git_url_with_path = f"{git_url}git/"
 
             # If "/git/" is already present, just fix the scheme
             if source_url.startswith(git_url_with_path):
--- a/corgi/core/migrations/0081_fix_build_sources.py
+++ b/corgi/core/migrations/0081_fix_build_sources.py
@@ -29,7 +29,7 @@ def fix_build_source_urls(apps, schema_editor):
     for git_url in gitlike_urls:
         # Always use https:// instead of git:// or similar
         builds_to_update = SoftwareBuild.objects.filter(source__startswith=git_url)
-        git_url_with_path = f"{git_url}/git/"
+        git_url_with_path = f"{git_url}git/"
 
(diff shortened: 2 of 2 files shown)


corgi/core/migrations/0082_fix_build_sources_typo.py
```diff
@@ -0,0 +1,31 @@
...
+    SoftwareBuild = apps.get_model("core", "SoftwareBuild")
+
+    # Find any build where the source URL has two "/git/git/"
+    # Replace it with just one "/git/"
+    one_git_value = Value("/git/")
+    two_git_path = "/git/git/"
+    two_git_value = Value(two_git_path)
+
+    builds_to_update = SoftwareBuild.objects.filter(source__contains=two_git_path)
+    builds_to_update.update(source=functions.Replace("source", two_git_value, one_git_value))
```
juspence: Actually, instead of making this a migration, maybe I should just manually run this cleanup code in stage. That would avoid running the migration in prod, which might cause more issues.

The typoed migration (number 81) will be fixed by this PR, so prod will deploy and run a good migration, there will not be any bad data there, and there's no need to run this migration to clean anything up. Running this migration anyway might cause side-effects or other surprises.

kgrant-rh: I concur that this kind of change is probably better handled manually than with a migration. It's a really specific circumstance that we ourselves created; it doesn't originate in data that we're getting from a third party. That said, since the area of effect of this migration is so small, I think the likelihood of side effects is low.

kgrant-rh (approved):
Looks good to me. If you don't want to update the affected builds manually, I think it makes sense to let the migration handle everything automatically.

PR #571, opened 2023-08-21
CORGI-778 stop creating variants from brew tags

jasinner:
Stop creating and linking Variants to Streams based on brew_tags. Only populate stream CPEs from variants matching brew tags instead. This allows us to automate CPE mapping for streams while avoiding the possibility builds in those variants will also have components matched to the stream. 

After this change we will only create Variants for streams with errata_info.

There is one stream 'rhel-8.8.0.z' which has both brew_tags and errata_info it will have all of it's variants removed, but that's OK, because on the next run the stream will get it's variants from errara_info information.


jasinner:
@mprpic after this change we'll be writing a manifest for every stream, not just those in support_cpe_map. Will that be a problem for SDEngine?

corgi/core/models.py
```diff
@@ -594,6 +594,7 @@ class ProductStream(ProductModel):
     exclude_components = fields.ArrayField(models.CharField(max_length=200), default=list)
 
     cpes_matching_patterns = fields.ArrayField(models.CharField(max_length=1000), default=list)
```
juspence, line 597: Do we also need to clean up or reset cpes_matching_patterns, for any stream which had a child brew_tag Variant deleted? I think right now this field still includes the CPEs from the child Variants which were Brew tags.

corgi/tasks/manifest.py
```diff
@@ -15,20 +15,20 @@
     autoretry_for=RETRYABLE_ERRORS,
     retry_kwargs=RETRY_KWARGS,
 )
-def update_manifests(fixup=True):
+def update_manifests():
     # Note - temporarily setting fixup=True
```
juspence: Let's remove the comment here as well.

corgi/api/serializers.py
```diff
@@ -549,9 +548,7 @@ def get_build_count(instance: ProductModel) -> int:
 
     @staticmethod
     def get_manifest(instance: ProductStream) -> str:
-        if (not instance.components.manifest_components(quick=True).exists()) or (
-            instance.name not in supported_stream_cpes
-        ):
```
juspence: I might have lost track of details...do we need to make this change at the same time as the others? Even if we stop creating Variants from Brew tags, can we still rely on our mapping of supported_stream_cpes for just a little while longer?

jasinner:
> @mprpic after this change we'll be writing a manifest for every stream, not just those in support_cpe_map. Will that be a problem for SDEngine?

Martin replied out of band and I created CORGI-791 to address that.

corgi/core/files.py
```diff
@@ -72,10 +72,9 @@ def render_content(self) -> str:
         distinct_provides = self.obj.provides_queryset  # type: ignore[attr-defined]
         distinct_upstreams = self.obj.upstreams_queryset  # type: ignore[attr-defined]
         cpes = cpe_lookup(self.obj.name)  # type: ignore[attr-defined]
-        # If the stream was not found in the supported_cpe_map, and it doesn't have any
-        # et_product_versions from errata_info in ps_update_stream, also add cpes from
-        # pattern matching
-        if not cpes and not self.obj.et_product_versions:  # type: ignore[attr-defined]
```
juspence, line 78: I'm not sure I really get this. We want to stop creating Variants using prod_defs brew_tags for a stream. But we still want to create Variants using prod_defs errata_info for a stream. Each errata_info Variant still adds its ET PV's name to the parent stream's list of et_product_versions. This code in the prod-defs collector / update_products task hasn't changed at all.

In the old code, our manifests would report:
1. The CPE from the supported_cpe_map, if present
2. Else nothing, if the stream had no cpe in the mapping but did have et_product_versions (so the stream had errata_info in prod-defs and some of these ET Variants matched)
3. Else the cpes_matching_patterns, which I think includes child Variant CPEs from both of brew_tag and errata_info Variants. But in this case, only the brew_tag variants have CPEs defined (case number 2 was skipped, so the stream has no errata_info or none of the ET Variants matched)

In the new code, our manifests will report:
1. The CPE from the supported_cpe_map, if present
2. Else the cpes_matching_patterns, which could include child Variant CPEs from both of brew_tag and errata_info Variants. In this case, we no longer skip streams with et_product_versions defined / streams which have errata_info in prod-defs and which matched at least one ET Variant.

I can follow the code poorly enough, I just don't really get the implication of the change. Maybe it's better to just change the brew_tag creating / linking code, and keep all the manifest-publishing / CPE lookup logic the same? Whatever you think is best is probably fine.

tests/test_products.py
```diff
@@ -253,27 +251,37 @@ def test_skip_brew_tag_linking_for_buggy_products(requests_mock):
...
+        "cpe:/a:redhat:openshift_gitops:1.7::el8",
+        "gitops-1.7-rhel-8",
+        "gitops-1.7",
+    ),
     # In the actual ET data the brew tag is rhaos-4.10-rhel-8-candidate
     # This matches ps_update_stream openshift-4.10.z, which has a brew tag
     # rhaos-4.10-rhel-8-container-released
-    ("OSE-4.10-RHEL-8", "rhaos-4.10-rhel-8", "openshift-4.10.z"),
+    (
+        "8Base-RHOSE-4.10",
```
juspence, line 264: In the other example, the first value stayed the same, like `8Base-GitOps-1.7`. But here we went from `OSE-4.10-RHEL-8` to `8Base-RHOSE-4.10`. Is that expected, or maybe a typo?

tests/test_products.py
```diff
@@ -77,10 +77,8 @@ def test_products(requests_mock):
     assert len(openshift410z.yum_repositories) == 5
 
     rhacm24z = ProductStream.objects.get(name="rhacm-2.4.z")
-    assert et_variant.name in rhacm24z.productvariants.values_list("name", flat=True)
-    # Stream "cpes" is a dynamically-generated property, different than the "cpe" field
-    # Which should include the stream's CPE + all the child variant CPEs (if any)
-    assert et_variant.cpe in rhacm24z.cpes
+    assert et_variant.name not in rhacm24z.productvariants.values_list("name", flat=True)
+    assert et_variant.cpe in rhacm24z.cpes_from_brew_tag_variants
```
juspence: If there's no longer a child Variant for this Brew tag, let's `assert et_variant.cpe not in rhacm24z.cpes`, so that we know the "cpes" property is only including CPEs from linked Variants, and not including cpes from any of the fields like "cpes_matching_patterns" or "cpes_from_brew_tag_variants".
```suggestion
    assert et_variant.cpe not in rhacm24z.cpes
    assert et_variant.cpe in rhacm24z.cpes_from_brew_tag_variants
```

It's also a little minor, but maybe we should call the new field "cpes_from_brew_tag_matches" or just "cpes_from_brew_tags", since the brew_tag_variants don't really exist anymore.

juspence (approved):
I'm still not really sure I can predict all the implications of this. I'll reread the ticket and PR description in my morning to see if I can grasp it better with a fresh start. For now this LGTM, comments are minor.

corgi/core/models.py
jasinner: No, those values are not generated from Variants.

corgi/tasks/manifest.py
jasinner: ok

corgi/core/files.py
jasinner: I tried to fix a bug where where no CPEs were being reported for stream with et_product_versions. I added a few tests here now and modified the code in this commit ready for also adding cpes_from_brew_tags.

The scenarios are:

1. Stream in supported_cpe_map - use those CPEs
2. Stream is not in supported_cpe_map and has productvariants - use the stream.cpes property
3. Stream is not in supported_cpe_map and has not productvariants - combine the cpe_matching_patterns and cpes_from_brew_tags.

tests/test_products.py
jasinner: The Variant was wrong in the original version. I updated it to the correct Variant name 8Base-RHOSE-4.10.

corgi/api/serializers.py
jasinner: Ok, I'll remove this commit and we can handle this as part of CORGI-791

tests/test_products.py
jasinner: Good idea.

jasinner:
In order for this to completely fix CORGI-778 we need to call `ProductVariant.objects.exclude(productstreams__brew_tags={}).delete()` after all the `slow_remove_product_from_build` tasks are complete. I tried to do that as part of the migration, but it causes those tasks to fail.

PR #572, opened 2023-08-22
CORGI-787 always update relation softwarebuild foreign key 

jasinner:
If fetch_unprocessed_relations is called with a relation with a SoftwareBuild fk set to None and that SoftwareBuild already exists we don't update the fk for the relation to the existing SoftwareBuild. That causes `save_product_taxonomy` function of the build to skip the relation, since it filters by relations with the fk set.

The migration here should fix the current data.

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -42,11 +42,17 @@ def slow_fetch_brew_build(
 ) -> bool:
     logger.info("Fetch brew build called with build id: %s", build_id)
 
-    if SoftwareBuild.objects.filter(build_id=build_id, build_type=build_type).exists():
+    try:
+        softwarebuild = SoftwareBuild.objects.get(build_id=build_id, build_type=build_type)
+    except SoftwareBuild.DoesNotExist:
+        # Process the build below
+        pass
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat (approved):
LGTM

corgi/tasks/brew.py
```diff
@@ -42,14 +42,20 @@ def slow_fetch_brew_build(
...
+        pass
+    else:
         logger.info("Already processed build_id %s", build_id)
 
         # Build exists, and we don't want to reload it
         if not force_process:
             # But we want to save the taxonomy again
             if save_product:
                 logger.info("Only saving product taxonomy for build_id %s", build_id)
+                update_relation_software_build_fk(build_id, build_type, softwarebuild)
```
juspence: Maybe we should move this to line 55 above. Then we'll always fix the relations, even when save_product is False.

juspence (approved):
LGTM!

corgi/tasks/brew.py
jasinner: good idea. Will do.

PR #573, opened 2023-08-22
CORGI-757: Add GitHub repo authentication

juspence:
@mprpic Please review. I've tested both URL formats with the "git clone" command in a stage pod, and this fixup seems to work / force our SSH key to be used for authentication.

This should be the last code change needed from us / in our code. Once this is merged, I'll work with the various repo owners / org admins to make sure we have access to any repo used by a managed service.

That's also waiting on SecArchs to update the list of components for each service. Rght now we're trying to clone some older repos that are no longer used, according to the service teams.

Tests are going to fail due to known / ongoing data issues in stage. Please ignore that for now.

### DIFF
--- a/corgi/collectors/app_interface.py
+++ b/corgi/collectors/app_interface.py
@@ -59,11 +59,10 @@ class AppInterface:
                     # Ignore container images in other registries since we assume all images are
                     # in Quay.io.
                     if org_repos["org"]["instance"]["url"] != "quay.io":
-                        logger.warning(
+                        raise ValueError(
                             f"Found non-quay.io container images for {component_name} in "
                             f"app-interface: {org_repos['org']['instance']}"
                         )
-                        continue
                     for repo in org_repos["items"]:
                         repo_name = f"{org_repos['org']['name']}/{repo['name']}"
                         subcomponent_data[repo["name"]]["quay_repo_name"] = repo_name
--- a/corgi/collectors/syft.py
+++ b/corgi/collectors/syft.py
@@ -92,6 +92,15 @@ class Syft:
         An optional target ref can be specified that represents a valid committish in the Git
         repo being scanned.
         """
+        # Convert unauthenticated / web-based URLs to SSH-based URLs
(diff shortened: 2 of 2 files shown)


corgi/collectors/app_interface.py
```diff
@@ -59,11 +59,10 @@ def fetch_service_metadata(
                     # Ignore container images in other registries since we assume all images are
                     # in Quay.io.
                     if org_repos["org"]["instance"]["url"] != "quay.io":
-                        logger.warning(
+                        raise ValueError(
```
juspence, line 62: This guarantees our manifest has all the components. If we just log a warning, we might never notice that we didn't manifest certain components, and then IR might think the manifest is complete when it's not.

I tested locally and never saw this error get raised, so it's fine to add for now (doesn't break anything) but will make sure we notice problems in the future.

corgi/collectors/syft.py
```diff
@@ -92,6 +92,15 @@ def scan_git_repo(
...
         """
+        # Convert unauthenticated / web-based URLs to SSH-based URLs
+        # so that "git clone" uses our existing SSH key and doesn't prompt
+        # for a username/password combo if the repo is private
+        if target_url.startswith("https://github.com/"):
+            target_url = target_url.replace("https://github.com/", "git@github.com:", 1)
+        elif target_url.startswith("http://github.com/"):
+            target_url = target_url.replace("http://github.com/", "git@github.com:", 1)
+        # Else it could be an internal Gitlab server - we allow these to fail
+        # They should be made into public repos that work without authentication
```
juspence, line 103: I could replace `https://` with `git@` and e.g. `.com/` with `.com:` if you think it's better to just use the SSH key everywhere. But I'd prefer to only use the SSH key for GitHub, then work with internal repo owners to open up the access more.

mprpic: Alternatively: `target_url = re.sub(r'https?://github.com/', "git@github.com:", target_url)`

mprpic (approved):
LGTM!

PR #575, opened 2023-08-23
CORGI-780: Fix mangen task errors when components already exist

juspence:
@mprpic Please review what's hopefully the last bugfix. I need to look over the tracebacks in our monitoring email again, and think more about the TODO notes, but my basic understanding is:

1. Services can reuse components
2. We currently delete and recreate components for service X, then for service Y, etc.
3. This doesn't clean up services / components which no longer exist in prod-defs
4. This may cause issues if tasks for service X and service Y are running at the same time and processing the same component

If this fixes the errors in the monitoring email, I'll do a final PR to add comprehensive tests for the mangen logic.

### DIFF
--- a/corgi/tasks/managed_services.py
+++ b/corgi/tasks/managed_services.py
@@ -74,27 +74,66 @@ def cpu_manifest_service(product_stream_id: str, service_components: list) -> No
                 build.delete()
 
             build_id = now.strftime("%Y%m%d%H%M%S%f")
-            build = SoftwareBuild.objects.create(
-                name=service_component["name"],
-                build_type=SoftwareBuild.Type.APP_INTERFACE,
-                build_id=build_id,
-                completion_time=now,
-                meta_attr={"service": service.name},
-            )
+            try:
+                # get_or_create requires all keyword arguments to appear in
+                # a unique_together constraint, like build_id and build_type
+                # Otherwise it will not be atomic and can behave incorrectly
+                # Since we don't have a database index for name + build_type,
+                # we try to find an APP_INTERFACE build with a matching name first
+                # And update the list of services it belongs to
+                # If that fails, we'll create a new build
+                # So Corgi reuses a component when services reuse the component
+                build = SoftwareBuild.objects.get(
+                    name=service_component["name"],
+                    build_type=SoftwareBuild.Type.APP_INTERFACE,
+                )
+                build.meta_attr["services"].append(service.name)
+                build.save()
+            except SoftwareBuild.DoesNotExist:
+                build = SoftwareBuild.objects.create(
+                    name=service_component["name"],
+                    build_type=SoftwareBuild.Type.APP_INTERFACE,
+                    build_id=build_id,
+                    completion_time=now,
+                    meta_attr={"services": [service.name]},
+                )
 
-            root_component = Component.objects.create(
-                type=Component.Type.CONTAINER_IMAGE if quay_repo else Component.Type.GITHUB,
-                name=service_component["name"],
-                version=component_version,
-                release="",
-                arch="noarch",
-                namespace=Component.Namespace.REDHAT,
-                software_build=build,
-            )
-            root_node = ComponentNode.objects.create(
+            # Root components can only be linked to one build
+            # if we use the same build / component for two different services,
+            # we will analyze the same component twice
+            # this should be OK and the taxonomy should be the same.
+            # Or if not, merging the data is probably what we want.
+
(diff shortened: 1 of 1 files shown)


corgi/tasks/managed_services.py
```diff
@@ -59,42 +79,60 @@ def cpu_manifest_service(product_stream_id: str, service_components: list) -> No
...
+            try:
+                # get_or_create requires all keyword arguments to appear in
+                # a unique_together constraint, like build_id and build_type
+                # Otherwise it will not be atomic and can behave incorrectly
+                # Since we don't have a database index for name + build_type,
+                # we try to find an APP_INTERFACE build with a matching name first
+                # And update the list of services it belongs to
+                # If that fails, we'll create a new build
+                # So Corgi reuses a component when services reuse the component
+                build = SoftwareBuild.objects.get(
```
mprpic, line 92: In what scenario will this actually return a build when all builds are deleted before these tasks run? Are there really services that share the same component (Quay image / GitHub repo)?

corgi/tasks/managed_services.py
```diff
@@ -59,42 +79,60 @@ def cpu_manifest_service(product_stream_id: str, service_components: list) -> No
...
+            # Use only indexed / unique_together kwargs to find the object
+            root_component, _ = Component.objects.update_or_create(
                 type=Component.Type.CONTAINER_IMAGE if quay_repo else Component.Type.GITHUB,
                 name=service_component["name"],
                 version=component_version,
                 release="",
                 arch="noarch",
-                namespace=Component.Namespace.REDHAT,
-                software_build=build,
+                # TODO: What if component exists but isn't in REDHAT namespace / linked to a build?
```
mprpic: The root component here is either a Red Hat-built image or a Red Hat-owned Git repo, so I think it should never be anything else than a Red Hat-namespaced component, no?

corgi/tasks/managed_services.py
```diff
@@ -23,6 +23,26 @@ def refresh_service_manifests() -> None:
...
+    # Services can reuse components, so this avoids deleting a build / component for one service
+    # while trying to create the same build / component for another service in another task.
+    # This also cleans up any old builds / components which were removed from product-definitions.
+    # We should recreate builds / components that are still present in prod-defs
+    # when we process the list of components for that service below.
+    SoftwareBuild.objects.filter(build_type=SoftwareBuild.Type.APP_INTERFACE).delete()
+    ProductComponentRelation.objects.filter(
+        type=ProductComponentRelation.Type.APP_INTERFACE
+    ).delete()
+    # TODO: Deleting the build deletes the linked root component
```
mprpic, line 42: I seem to remember that when I was writing this I tested whether it could delete the whole tree, and it did due to the cascading deletion on the parent attribute of the related components.

mprpic (approved):

corgi/tasks/managed_services.py
juspence: Yes, I saw in our monitoring email that two different services had the same quay_org/repo_name in their list of components. So at least some of the root components are shared between services.

corgi/tasks/managed_services.py
juspence: Yes, but if any of those GitHub repos are also used outside of a managed service, they might already exist in Corgi but be in the UPSTREAM namespace instead of the REDHAT namespace.

I thought about this over my lunch break, and if we update that component, we'd end up changing the purl (add "redhat/" just after "pkg:type/") but not updating the purl on all the nodes.

In that case, our get_or_create lookup for the node wouldn't find the existing node and would create a duplicate one. So I'm going to switch this back to just get_or_create for now, to keep things simpler and saner.

corgi/tasks/managed_services.py
juspence: It will delete all the ComponentNodes, but the actual child Components themselves don't get cleaned up, I think. I'll test this locally in a bit.

juspence: Yep, we delete the build, and (in theory) the root component + all its nodes are CASCADE-deleted. Since we delete the root component, that also deletes any ManyToMany relations that target the root component, like "provides" and "upstreams". We only delete the relations though, not the actual related components themselves.

And of course there's a bug where this cascade deletion from the SoftwareBuild to the root component is broken :rage:

EDIT: I haven't thought of any good solution for this, and there are probably other places that need this cleanup code too. So I'm just going to punt on this for now, and leave the real work for a future ticket.

PR #576, opened 2023-08-23
Fix performance test

juspence:
@jasinner FYI. In the future, let's please disable failing tests individually with pytest.mark.xfail or similar, instead of changing the Gitlab pipeline to skip an entire category of tests.


jasinner:
Ok, can do. Sorry, I think you have asked me do use that feature previously but I forgot. Thanks for fixing this!

PR #577, opened 2023-08-23
CORGI-790: Tweak links for provides, sources, and upstreams

juspence:
@RedHatProductSecurity/corgi-devs Please review this change thoroughly. This is our last chance to make changes here before our next release, and I really don't want to go back to OpenLCS and tell them we're changing things for a third time.

If this looks OK, I'll deploy this to stage and confirm we still don't have any "Bad Gateway" errors, then merge this.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -73,10 +73,13 @@ class ComponentFilter(FilterSet):
     sources = CharFilter(lookup_expr="purl")
     provides = CharFilter(lookup_expr="purl")
     upstreams = CharFilter(lookup_expr="purl")
+    downstreams = CharFilter(lookup_expr="purl")
+
     # otherwise use regex to match provides,sources or upstreams purls
     re_sources = CharFilter(field_name="sources", lookup_expr="purl__iregex")
     re_provides = CharFilter(field_name="provides", lookup_expr="purl__iregex")
     re_upstreams = CharFilter(field_name="upstreams", lookup_expr="purl__iregex")
+    re_downstreams = CharFilter(field_name="downstreams", lookup_expr="purl__iregex")
 
     el_match = CharFilter(label="RHEL version for layered products", lookup_expr="icontains")
     released_components = BooleanFilter(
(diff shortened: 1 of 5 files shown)


jasinner:
I tried this out in Dev but clicking through the links on the Web UI returns no results for me, eg:

[sources=pkg:golang/github.com/hashicorp/aws-cloudformation-resource-schema-sdk-go@v0.18.0](http://localhost:8080/api/v1/components?sources=pkg:golang/github.com/hashicorp/aws-cloudformation-resource-schema-sdk-go@v0.18.0)

@JimFuller-RedHat Is that the format of the URL you were expecting?

The same request for the old style links returns 1000+ results:

`https://{CORGI_DOMAIN}/api/v1/components/b558b38d-273d-48a9-a6f6-a5d479a37947/sources`

JimFuller-RedHat:
I think this is reversed eg. to find sources your would use ?provides and if u want provides use ?sources

jasinner:
Yes, you are correct, this returned the results I was expecting:

'api/v1/components?provides=pkg:golang/github.com/hashicorp/aws-cloudformation-resource-schema-sdk-go@v0.18.0'

jasinner (approved):
LGTM, just need to swtich provides/sources and upstreams/downstreams.

juspence:
It's not backwards, it's just really confusing. You can look at the tests or the header comments on the new methods for an explanation of the logic.

Basically, the "provides" field shows all the **provided** components of this component. It's expected that a Golang component will not have any provides of its own.

The ?provides= filter is different, and shows all the **source** components which have this component in their list of provided components (have this component in their "provides" field).

The ?provides= filter is therefore the same as the "sources" field. The "sources" field shows all the **source** components of this component, so the Golang example above should have at least 1 result, like a source RPM, RPM module, or container.

The ?sources= filter only shows all the **provided** components which have this component in their list of source components (have this component in their "sources" field).

Upstreams and downstreams follow the same logic.

jasinner:
Ah you are right. The links are correct! Sorry for the confusion.

juspence:
No problem, thanks for reviewing!

PR #579, opened 2023-08-24
Fix reference to nonexistent DB column that makes CASCADE deletion fail

juspence:
@RedHatProductSecurity/corgi-devs With a properly-linked set of components, like below:

1 SoftwareBuild
3 Components - 1 root, 1 provides, 1 upstream
3 ComponentNodes - 1 root, 1 provides, 1 upstream

Deleting the SoftwareBuild should also delete the root component (due to on_delete=models.CASCADE on the ForeignKey). This should also delete the root component's node, which should then delete all the other component nodes (again, due to cascade deletion).

Currently there's a bug which prevents the SoftwareBuild deletion from cascading to the root component. Somehow, all the ComponentNodes are deleted anyway, but the root component gets left behind and isn't deleted.

Before this change, I see buggy behavior in a shell, but tests pass both locally and in the CI pipeline. After this change, the root Component is deleted like I expect in all three environments. I'm not sure what's going on here, but maybe this is worth fixing just for safety?

Or if someone can confirm this is user error / a Heisenbug that only affects me, I'll just close this.


jasinner:
I can't reproduce the issue in a shell. But I'm happy to make the change anyway as I don't see how it will adversely affect anything.

```
import corgi.core.models as m

sb = m.SoftwareBuild.objects.get(build_id="1689053")

sb.components.first().purl
'pkg:rpm/redhat/CUnit@2.1.3-25.el9?arch=src'

sb.delete()

(57,
 {'core.Component_upstreams': 2,
  'core.Component_sources': 20,
  'core.Component_products': 1,
  'core.Component_productversions': 2,
  'core.Component_productstreams': 8,
  'core.SoftwareBuild': 1,
  'core.ComponentNode': 23})

c = m.Component.objects.get(purl="pkg:rpm/redhat/CUnit@2.1.3-25.el9?arch=src")
DoesNotExist: Component matching query does not exist.
```

jasinner (approved):

juspence:
Thanks for confirming! I'm going to let this sit while I'm on PTO. If I can still reproduce the issue when I get back, then I'll merge the fix.

PR #581, opened 2023-08-25
CORGI-798: Save container provides / upstreams in a separate task to avoid timeouts

juspence:
@RedHatProductSecurity/corgi-devs Please review. This should fix some of the SoftTimeLimitExceeded errors we see in each day's monitoring email, which I think are happening only for large container builds. Based on the tracebacks I've seen, some appear to time out during upstream creation, and some appear to time out during provide / Cachito dependency creation.

I think this change should be safe and we shouldn't have any more timeouts, miss any information, or fail to save the whole taxonomy. But even if we do, that's probably an improvement over not loading the child components at all (due to current timeouts).

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -113,7 +113,7 @@ def slow_fetch_brew_build(
     if component["type"] == Component.Type.RPM:
         root_node, root_created = save_srpm(softwarebuild, component)
     elif component["type"] == Component.Type.CONTAINER_IMAGE:
-        root_node, root_created = save_container(softwarebuild, component)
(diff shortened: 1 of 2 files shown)


juspence:
@jasinner I really think this will fix CORGI-798, and I looked over the related code but don't see any other possible bugs.

Even if this doesn't fix the issue, it will fix a separate issue in our monitoring email. Those timeout errors affect parity / reconciliation, so I'd like to get this in if possible. Right now we just fail to fully load all components for certain large container builds.

I think it should be safe, but please let me know if you think it's risky or might cause unexpected errors / side-effects.

jasinner (approved):
LGTM

PR #582, opened 2023-08-28
CORGI-802: Add RPMMODs back to list of root components to fix missing query results

juspence:
@RedHatProductSecurity/corgi-devs Please review. I've reviewed all usages of the ROOT_COMPONENTS_CONDITION which this PR directly changes, as well as the root_components() queryset which is indirectly changed.

I think the new code is mostly safe, but there are three three side effects: we now run SCA even on RPMMODs, the RPMMODs will start appearing in our manifests, and (of course) the RPMMOD components may appear in the list of search results (depending on the API parameters / queries the user gives).

For the first, I can back out the latest commit here to keep the SCA task the same as before.

For the second, I'm not sure if this will be a problem or very surprising for our manifest users. I don't recall off the top of my head if the manifest template I wrote will handle this case correctly. I think it should be generic enough, so we could merge this to try it and see.

Or I could add special handling to exclude the RPMMODs from the manifests, just like we exclude source containers. I don't think there will be duplicate data for certain binary RPMs (which might be a child of both an SRPM and an RPMMOD), since we dedupe the list of root component children / "provides" components before reporting that data in the manifest.

And of course the third is what IR needs to see the correct results in certain cases. This does change how the root components filter / query parameter in the API works, so e.g. OpenLCS may be surprised if they start seeing RPMMOD components in a list of search results for some query. Let's check with all our other users, besides just IR, to make sure this won't cause issues for them.


corgi/core/constants.py
```diff
@@ -49,11 +49,15 @@
     "ProductVariant": "product_variants",
 }
 
-# Filter on "root components": SRPMs, index container images, or Github repos for managed services
+# Filter on "root components":
+# SRPMs, modules, index container images, or Github repos for managed services
 SRPM_CONDITION = Q(type="RPM", arch="src")
+RPMMOD_CONDITION = Q(type="RPMMOD")
```
JimFuller-RedHat, line 55: should we not also include arch="noarch" ?

juspence: An RPMMOD is always an RPMMOD regardless of architecture. In other words, all RPMMODs are always considered root components - this is just adding back what we had before.

JimFuller-RedHat (approved):
apart from one comment, LGTM  - it is a good idea to cross check impact with other users (as  eg. manifest, other latest filter usages, etc).

corgi/core/migrations/0087_add_rpmmod_to_root_component_types.py
```diff
@@ -0,0 +1,67 @@
...
+                    models.Q(("arch", "src"), ("type", "RPM")),
+                    ("type", "RPMMOD"),
+                    models.Q(("arch", "noarch"), ("type", "OCI")),
+                    models.Q(("arch", "noarch"), ("namespace", "REDHAT"), ("type", "GITHUB")),
+                    _connector="OR",
+                ),
+                fields=["name", "type", "arch"],
+                name="compon_latest_type_name_idx",
+            ),
+        ),
```
jasinner, line 52: Is this a duplicate of the index above?

corgi/tasks/sca.py
```diff
@@ -156,28 +156,8 @@ def cpu_software_composition_analysis(build_uuid, force_process: bool = False):
...
-        # Else we have exactly one module component
-        # which is no longer considered a "root" component
-        # so now we skip doing SCA on the module instead of failing
-        logger.info(
-            f"Build {build_uuid} had only one module, no other root components. Skipping SCA"
-        )
-        return
-
+    # Get root component for this build.
+    root_component = Component.objects.filter(software_build=software_build).root_components().get()
```
jasinner: I think we should continue to skip RPMMOD when doing SCA. There is no real source code as far as I know, instead RPMMOD reuses content from another SRPM. We scan all the included RPMs separately when processing their respective SRPMs.

corgi/core/migrations/0087_add_rpmmod_to_root_component_types.py
juspence: It's just Django being a little silly...it expects us to delete and recreate the index whenever the condition / list of components it applies to changes. In this case we've added the `("type", "RPMMOD")` condition on line 44. It would be nice if there was an ALTER_INDEX operation Django could use instead.

corgi/tasks/sca.py
juspence: OK, I'll back out the second commit here and just use the first. That should continue to skip RPMMODs only when doing SCA.

corgi/core/constants.py
JimFuller-RedHat: I can see that logic (and prev status quo)  - though today we have a closed set of constraints (OCI-noarch, RPM-src, etc) on other root components and things can always change in the future ... when I see 2 equal choices I go with the more constrained choice as it is safer (and more contained).

jasinner (approved):

PR #583, opened 2023-08-29
CORGI-798: Fix more soft time limit errors

juspence:
@RedHatProductSecurity/corgi-devs Please review. The tasks are still timing out in some cases, so I've split it further to save the index container / root node and all the arch-specific child containers / nodes together in the first / main task.

Then we save all the upstream go modules, upstream container sources, and upstream nodes together in the second task.

Finally (in the new code), we spawn a third task for each upstream source to save its provides / Cachito components. This means we might spawn many tasks just to load one container build, but it should avoid the risk of timeouts (fingers crossed).

### DIFF
--- a/corgi/tasks/brew.py
+++ b/corgi/tasks/brew.py
@@ -618,7 +618,7 @@ def slow_save_container_children(
     logger.info(f"Saving upstreams for {build_type} container build {build_id}")
     root_node = ComponentNode.objects.get(pk=root_node_pk)
     logger.info(f"{build_type} container build {build_id} had root node with purl {root_node.purl}")
-    any_go_module_created = any_source_created = any_cachito_created = False
+    any_go_module_created = any_source_created = False
 
     meta_attr = {"go_component_type": "gomod", "source": ["collectors/brew"]}
     for module in build_data["meta"].get("upstream_go_modules", ()):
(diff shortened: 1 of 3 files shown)


corgi/tasks/brew.py
```diff
@@ -652,10 +652,39 @@ def slow_save_container_children(
             source["type"], component_name, component_version, source["meta"], extra, root_node
         )
         any_source_created |= temp_created
+        slow_save_container_provides.delay(
+            build_id, build_type, source, upstream_node.pk, save_product
+        )
+
+    if save_product and not build_data.get("sources"):
+        # Save the taxonomy either here in this task (if only upstreams created),
+        # or in the task below (if both upstreams and provides created)
```
kgrant-rh, line 661: My mistake, this comment is clear to me now.

kgrant-rh (approved):
Looks good.

corgi/tasks/brew.py
juspence: Ah, sorry. It's the slow_save_container_provides() task, right below this slow_save_container_upstreams() task. We always call the upstream-saving task, but we only call the provides-saving task if the container has any "sources" key in its build data.

So if there are any provides, we need to call the slow_save_taxonomy() task at the very end, after saving the provides. This is needed since another task will be spawned to save the provides, and we don't want to save the taxonomy here after creating only the upstreams. The taxonomy is still incomplete here so it's too early to save it.

If there are not any provides to save, we need to call the slow_save_taxonomy() task after saving only the upstreams. The current task here is the last task that runs if there are no provides to save.

PR #586, opened 2023-08-29
CORGI-693: Fix logic when building purl for SCA duplicate component detection

juspence:
@RedHatProductSecurity/corgi-devs This fixes a very silly typo in a previous MR for CORGI-693. We want Maven components with "redhat" in their version strings to be part of the REDHAT component namespace (they were built in one of our build systems).

For Maven purls, the purl namespace should always be a group ID like com.redhat.thing, so we can't just prepend "redhat/" to the purl. Instead, we SHOULD add a `?repository_url=https://maven.repository.redhat.com/ga` qualifier to the end of the purl.

But right now we're just adding `?https://maven.repository.redhat.com/ga` without the actual `repository_url=` part. I checked stage and don't see any purls with this value, so there should be no need for a migration. But I definitely want to fix this before we release CORGI-693 to production, to avoid creating any bad data in the future.

### DIFF
--- a/corgi/tasks/sca.py
+++ b/corgi/tasks/sca.py
@@ -46,6 +46,8 @@ def find_duplicate_component(meta_name: str, syft_purl: str) -> Component:
     # This probably isn't a bug in Syft, but we need names to match Brew's case
     # Sometimes Brew has lowercase and sometimes it's mixed-case
     # We can't just lowercase all names / NEVRAs because e.g. search apps need the original casing
+    # TODO: This is wrong - purl is not guaranteed to be lowercase either
+    #  Only certain types like GitHub or PyPI have their purls lowercased
     syft_purl = syft_purl.lower()
     possible_dupe = Component.objects.get(purl=syft_purl)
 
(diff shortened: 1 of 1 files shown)


corgi/tasks/sca.py
```diff
@@ -82,27 +84,53 @@ def save_component(
 ) -> bool:
     meta = component.get("meta", {})
     if component["type"] not in Component.Type.values:
-        logger.warning("Tried to save component with unknown type: %s", component["type"])
-        return False
+        raise ValueError("Tried to save component with unknown type: %s", component["type"])
```
juspence, line 87: I checked logs and didn't see this warning at all in the past 30 days. Let's make it an error so we'll notice if it happens in the future.

corgi/tasks/sca.py
```diff
@@ -82,27 +84,53 @@ def save_component(
...
+        arch = meta.get("arch")
+        if arch:
+            qualifiers["arch"] = arch
+
         syft_purl = PackageURL(
-            type=component["type"], name=meta.get("name", ""), version=meta.get("version", "")
+            type=component_type,
+            name=meta.get("name", ""),
+            version=version,
+            qualifiers=qualifiers,
```
juspence, line 110: If we weren't given a Syft purl in the meta dict, we build it ourselves. We then use it later to find any duplicate component. We look up the same purl, then try to figure out if:
1. The name / NEVRA is the same but with a different case
2. The name / NEVRA is the same but with dashes converted to underscores, or vice-versa
3. Some combination of the above

Syft also appears to report "version" and "release" info in the same "version" field (since purls have no concept of a "release"). This requires special handling in our code but we aren't always taking care of this correctly. Hence the TODO note - I need to look a lot more at this code sometime soon.

I think this code isn't getting hit right now and we always have a purl from Syft in the meta dict. So we haven't seen any problems here, and changes here shouldn't cause issues, but might prevent problems from happening in the future.

corgi/tasks/sca.py
```diff
@@ -82,27 +84,53 @@ def save_component(
...
         ).to_string()
 
     created = False
     name = meta.pop("name", "")
     version = meta.pop("version", "")
+    release = meta.pop("release", "")
+    arch = meta.pop("arch", "noarch")
```
juspence, line 117: Preparing for future state. We need to spli Syft's version into a Corgi version / release pair, then store both version and release separately in the meta dict. Right now this never happens, so this should be a no-op that gives the same result as before - both version and release are reported together in the "version" key.

The same logic applies to arch - we should use the value given if present, otherwise default to "noarch" like before. Right now nothing is supplying a custom arch value, as far as I know, so we'll always use the default "noarch" value like before.

corgi/tasks/sca.py
```diff
@@ -82,27 +84,53 @@ def save_component(
...
-        suffix = "&" if "?" in syft_purl else "?"
-        syft_purl = f"{syft_purl}{suffix}{RED_HAT_MAVEN_REPOSITORY}"
+        if component["type"] == Component.Type.MAVEN:
+            # Don't prepend the namespace, Maven-type purls don't allow this
+            # Add an extra &repository_url= qualifier to the end
+            # if any other qualifiers are already present
+            # Otherwise add the first (only) qualifier, ?repository_url=
+            # TODO: Check Syft isn't encoding these weirdly
+            suffix = "&repository_url=" if "?" in syft_purl else "?repository_url="
+            syft_purl = f"{syft_purl}{suffix}{RED_HAT_MAVEN_REPOSITORY}"
```
juspence, line 128: The stupid typo I mentioned in the description. I forgot to add the "repository_url=" part to the suffix, so purls always ended with `?https://maven.repository.redhat.com/ga` or `&https://maven.repository.redhat.com/ga`.

I also was adding the qualifier to all purls instead of only to Maven-type purls. This didn't cause issues (value not saved in stage anywhere) because it's only used for a lookup (find a duplicate component with the same purl).

corgi/tasks/sca.py
```diff
@@ -82,27 +84,53 @@ def save_component(
...
+        elif component["type"] not in Component.CUSTOM_NAMESPACE_TYPES:
+            # Add the REDHAT namespace to the beginning
+            syft_purl = syft_purl.replace(
```
juspence, line 131: We need the Syft purl to match the purls we generate in Corgi, so we can find a duplicate component (if any). We also probably need a lot more logic here, to make the duplicate component lookup work.

Right now we're seeing a lot of warnings about "Saved (Corgi) purl does not match (generated) Syft purl", which means this logic is incomplete.

juspence (commented):
Let's see if reviewing my own code helps me catch more dumb mistakes. I'm sure a few are lurking in here somewhere.

jasinner (approved):
LGTM

PR #587, opened 2023-08-30
CORGI-797 support removal of brew_tag from stream

jasinner:
The actual updating of brew_tag attribute of a ProductStream is handled by the code already. However if a brew_tag was removed from a stream we need some way to cleaning up all the existing component product taxonomies, and also relations so that the next time we call save_product_taxonomy on the build it doesn't get associated with the old brew_tag.

If a brew_tag is moved from one stream to another this code should handle that case as well. Next time load_stream_brew_tags runs relations will be created for the new stream. And the normal process of relating builds to streams will occur in fetch_unprocess_brew_tag relations.


jasinner:

corgi/tasks/prod_defs.py
```diff
@@ -263,10 +265,34 @@ def parse_product_stream(
...
+    for tag in removed_tags:
+        logger.info(f"Removing tag {tag} from stream {name}")
+        for brew_tag_relation in ProductComponentRelation.objects.filter(
+            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name, external_system_id=tag
+        ).iterator():
+            if brew_tag_relation.software_build_id:
+                slow_remove_product_from_build.delay(
+                    str(brew_tag_relation.software_build_id),
+                    "ProductStream",
+                    str(product_stream_pk),
```
juspence: These are already strings, right?
```suggestion
                    brew_tag_relation.software_build_id,
                    "ProductStream",
                    product_stream_pk,
```
And we're unlinking one build from one stream, if some Brew tag which refers to that build was removed from that stream in prod-defs.

What if there's also a different Brew tag that should keep this build linked to this stream? Maybe it's similar to what you mention in the description - we'll relink the build to the stream the next time we save the taxonomy. At the end of the slow_remove_product_from_build() task, we could schedule a follow-up slow_save_taxonomy() task to handle this if we don't already.

Or you could add logic to see if there are any other BREW_TAG relations with a different external_system_id (exclude the current brew_tag's ID) for the same software_build_id and product_ref. If so, we can just delete this relation and skip calling slow_remove_product_from_build. I think it would be good to add a test for this case if you have time.

What happens if a Brew tag is renamed? Maybe that never happens, maybe it's OK. I guess we'll delete the relations and unlink builds for the "old" tag, then create relations for the "new" tag and process them / link all the builds during our next daily task run. This might be another good case to add a test for.

corgi/tasks/prod_defs.py
```diff
@@ -263,10 +265,34 @@ def parse_product_stream(
...
+        .values_list("external_system_id", flat=True)
+        .distinct()
+    )
+    removed_tags = set(existing_stream_relation_tags) - brew_tags
+    for tag in removed_tags:
+        logger.info(f"Removing tag {tag} from stream {name}")
+        for brew_tag_relation in ProductComponentRelation.objects.filter(
+            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name, external_system_id=tag
+        ).iterator():
+            if brew_tag_relation.software_build_id:
```
juspence: I think you should be able to just filter the queryset directly, instead of fetching all the relations from the DB and checking them using an if statement:
```suggestion
        for brew_tag_relation in ProductComponentRelation.objects.filter(
            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name, external_system_id=tag, software_build_id__isnull=False
        ).iterator():
```

corgi/tasks/prod_defs.py
```diff
@@ -263,10 +265,34 @@ def parse_product_stream(
             "obj": product_stream,
         },
     )
+    _clean_orphaned_brew_tags(set(brew_tags_dict.keys()), name, str(product_stream.pk))
     parse_errata_info(errata_info, product, product_stream, product_stream_node, product_version)
     product_stream.save_product_taxonomy()
 
 
+def _clean_orphaned_brew_tags(brew_tags: set[str], name: str, product_stream_pk: str):
```
juspence:
```suggestion
def _clean_orphaned_brew_tags(brew_tags: set[str], name: str, product_stream_pk: str) -> None:
```

corgi/tasks/prod_defs.py
```diff
@@ -263,10 +265,34 @@ def parse_product_stream(
...
+        )
+        .values_list("external_system_id", flat=True)
+        .distinct()
+    )
+    removed_tags = set(existing_stream_relation_tags) - brew_tags
+    for tag in removed_tags:
+        logger.info(f"Removing tag {tag} from stream {name}")
+        for brew_tag_relation in ProductComponentRelation.objects.filter(
+            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name, external_system_id=tag
+        ).iterator():
```
juspence: Actually this can also be simplified, and then the second loop below isn't needed:
```suggestion
    removed_tags = (
        ProductComponentRelation.objects.filter(
            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name, software_build_id__isnull=False
        ).exclude(external_system_id__in=brew_tags)
        .iterator()
    )
    for brew_tag_relation in removed_tags:
        # schedule slow_remove_product_from_build, then delete the relation
```

juspence (approved):
I think it's OK, but there are a few edge cases we could add tests for. Maybe that's not needed if those edge cases never happen, or if I've misunderstood something.

corgi/tasks/prod_defs.py
jasinner: The problem with that is we also want to delete any unprocessed relations, where software_build_id__isnull=True. But if the SoftwareBuild isn't linked as a foreign key that means it's been processed and we want to remove it from the build. I'll add a comment.

corgi/tasks/prod_defs.py
jasinner: Whoops, not that again! I use the refactor feature in PyCharm to split this out and never adds a return type! Need to continue checking that.

corgi/tasks/prod_defs.py
jasinner: Good idea. Updated.

corgi/tasks/prod_defs.py
jasinner: Removing cast to string.

Added a test for existing relation with same product_ref/build_id combination. In addition to the situation you described with 2 brew_tags with the same build_id and stream I also added a test/check for existing relations with child variants of the stream.

corgi/tasks/prod_defs.py
```diff
@@ -263,10 +265,47 @@ def parse_product_stream(
...
+        )
+    )
+    for relation_to_remove in relations_to_remove:
+        # Make sure there is no other relation linking this build to the stream or it's child
+        # variants before removing
+        existing_stream_relation = (
+            ProductComponentRelation.objects.filter(
+                build_id=relation_to_remove.build_id, product_ref__in=related_product_refs
+            )
+            .exclude(pk=relation_to_remove.pk)
```
JimFuller-RedHat, line 295: FWIW .exclude maybe slower then negated Q (~Q)

juspence: Do you have a reference or more details about this? I looked at the generated SQL for a (very simple, not representative) query and they both look the same:
```
>>> ProductComponentRelation.objects.filter().exclude(pk=123).explain()
"Seq Scan on core_productcomponentrelation  (cost=0.00..589216.96 rows=8669655 width=141)\n  Filter: (uuid <> '00000000-0000-0000-0000-00000000007b'::uuid)"

>>> ProductComponentRelation.objects.filter().filter(~Q(pk=123)).explain()
"Seq Scan on core_productcomponentrelation  (cost=0.00..589216.96 rows=8669655 width=141)\n  Filter: (uuid <> '00000000-0000-0000-0000-00000000007b'::uuid)"
```

PR #588, opened 2023-08-30
Remove name and arch from versionInfo

kgrant-rh:
Removes name and arch from versionInfo fields.


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review. This is just a change to the spdx templates.

juspence (approved):
LGTM! Only minor note is that it might be good to make this a property on the Component model, instead of having the same logic repeated in the templates:
```
@property
def spdx_version() -> str:
    spdx_version = self.version
    if self.epoch:
        spdx_version = f"{self.epoch}-{spdx_version}"
    if self.release:
        spdx_version = f"{spdx_version}-{self.release}"
    return spdx_version
```

PR #590, opened 2023-08-31
CORGI-805 add back cdn relations

jasinner:
This essentially reverts the changes made for CORGI-727. I kept some of the refactoring, and cleaned up the code a little bit. This should get coverage for RHEL Z-Streams back up to parity with Deptopia.

To execute the code after merging, without waiting for the weekly task, manually run `setup_pulp_relations` task, and then `loadpulpdata` management command for the desire stream(s).


jasinner:

corgi/tasks/yum.py
```diff
@@ -9,14 +9,15 @@
...
-from corgi.tasks.brew import fetch_unprocessed_relations, slow_fetch_modular_build
+from corgi.tasks.brew import slow_fetch_modular_build
 from corgi.tasks.common import (
     BUILD_TYPE,
     RETRY_KWARGS,
     RETRYABLE_ERRORS,
     create_relations,
     get_last_success_for_task,
 )
+from corgi.tasks.pulp import fetch_unprocessed_relations
```
juspence: You can leave off this change - it was actually a typo where I imported fetch_unprocessed_relations from the corgi.tasks.pulp module instead of from the corgi.tasks.brew module.

It worked because pulp also imported fetch_unprocessed_relations from brew, but then it broke when we removed that import from the pulp module. Let's continue importing the module from where it really is defined, so we avoid confusion and possible breakage in the future.

corgi/tasks/pulp.py
```diff
@@ -1,14 +1,89 @@
...
+        # Channels created by Yum repo collector sometimes have no Variant
+        .exclude(productvariants__isnull=True)
+        .values_list("name", "productvariants__name")
+    ):
+        slow_setup_pulp_rpm_relations.delay(channel_name, variant_name)
+        slow_setup_pulp_module_relations.delay(channel_name, variant_name)
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def slow_setup_pulp_rpm_relations(channel, variant):
```
juspence, line 58: It's not related to this change, but maybe we could add type hints here while we're making changes. By default, mypy will ignore and not check any functions without type hits, so even something as simple as adding a `-> None` return type hint improves our mypy coverage.

corgi/tasks/pulp.py
```diff
@@ -1,14 +1,89 @@
...
+        variant,
+        ProductComponentRelation.Type.CDN_REPO,
+        slow_fetch_modular_build,
+    )
+    if no_of_relations > 0:
+        logger.info("Created %s new relations for SRPMs in %s", no_of_relations, channel)
+
+
+@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+def slow_setup_pulp_module_relations(channel, variant):
```
juspence, line 73: I'm not sure if we know the types or if mypy will start throwing some complicated error. If it's a problem we can leave off type hints for now.

corgi/tasks/management/commands/loadpulpdata.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    def handle(self, *args, **options) -> None:
+        if options["stream_names"]:
+            for stream_name in options["stream_names"]:
+                self.stdout.write(self.style.NOTICE(f"Fetching builds for stream: {stream_name}"))
+                self.get_builds_by_cdn_repo(stream_name=stream_name, force_process=options["force"])
+        elif options["all"]:
+            self.stdout.write(self.style.NOTICE("Fetching all unprocessed pulp relations"))
+            fetch_unprocessed_cdn_relations(force_process=options["force"])
```
juspence: Is it expected that the command runs this Celery task inline? Should we use .delay() to schedule it for later instead?

It's fine to leave for now, I just ask since this is the only time I know I'll be looking at this code. I don't want to accidentally add back a bug (?) we might have had without noticing before.

juspence (approved):
This code looks...familiar :grin:  I don't fully remember what I removed, but this looks like enough of it, and I assume you're looking at the diff so haven't missed anything.

I also don't remember what the side-effects were when we removed this the first time. Do we need to change something about the taxonomy-saving code to handle the CDN_REPO relations, or do we need to write some migration for something?

Sorry if that's really vague, I just remember the removal being more complicated so I thought adding it back wouldn't be this simple. I guess we can merge this, run the tasks, and see if all the relations get created / components get linked to products correctly now. Thanks for fixing this!

corgi/tasks/yum.py
jasinner: ok

corgi/tasks/pulp.py
jasinner: The create_relation function is a bit tricky to figure out the correct types, so I'm going to skip that for now.

corgi/tasks/pulp.py
jasinner: leaving off for now.

corgi/tasks/management/commands/loadpulpdata.py
jasinner: good idea. Adding .delay() call to this, and the fetch_modular_builds call below.

jasinner:
> This code looks...familiar 😁 I don't fully remember what I removed, but this looks like enough of it, and I assume you're looking at the diff so haven't missed anything.
> 
> I also don't remember what the side-effects were when we removed this the first time. Do we need to change something about the taxonomy-saving code to handle the CDN_REPO relations, or do we need to write some migration for something?
> 
> Sorry if that's really vague, I just remember the removal being more complicated so I thought adding it back wouldn't be this simple. I guess we can merge this, run the tasks, and see if all the relations get created / components get linked to products correctly now. Thanks for fixing this!

There was a related change CORGI-778 which adjusted the save_product_taxonomy function. That's probably the only change we really needed, not this one.

PR #591, opened 2023-08-31
CORGI-798: Fix missing upstreams

juspence:
Opening just as a draft so I don't forget this exists after PTO. I found a few possible bugs and some things that are confusing, which you could review and leave feedback on while I'm out if you have time.

Even if this is correct as written, I still need to add a lot more tests and confirm this fixes the bug we're seeing. That will have to wait until after PTO.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1554,8 +1554,8 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
     def save_component_taxonomy(self):
         """Link related components together using foreign keys. Avoids repeated MPTT tree lookups"""
         self.upstreams.set(self.get_upstreams_pks(using="default"))
-        self.provides.set(self.get_provides_nodes(using="default"))
-        self.sources.set(self.get_sources_nodes(using="default"))
+        self.provides.set(self.get_provides_pks(using="default"))
+        self.sources.set(self.get_sources_pks(using="default"))
 
     @property
     def provides_queryset(self, using: str = "read_only") -> Iterator["Component"]:
@@ -1913,9 +1913,8 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         """Return an SPDX-style manifest in JSON format."""
         return ComponentManifestFile(self).render_content()
 
-    def get_provides_nodes(self, include_dev: bool = True, using: str = "read_only") -> set[str]:
-        """return a set of descendant ids with PROVIDES ComponentNode type"""
-        # Used in taxonomies. Returns only PKs
+    def get_provides_pks(self, include_dev: bool = True, using: str = "read_only") -> set[str]:
+        """Return Component PKs which are PROVIDES descendants of this Component, for taxonomies"""
         provides_set = set()
 
         type_list: tuple[ComponentNode.ComponentNodeType, ...] = (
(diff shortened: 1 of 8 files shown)


juspence:
@RedHatProductSecurity/corgi-devs Please review this mess, or shout at me if it needs to be split out further. The list of changes here is rather long, but essentially there seem to be multiple bugs in our upstream-finding logic. Or I'm very confused - please let me know if I've misunderstood something or missed an edge case.

Before, we'd look for upstreams only if a Component was linked to a SoftwareBuild. This ignores binary RPMs (which used to be linked to builds, but are not anymore for good reasons) and arch-specific containers (which I don't think were ever linked to builds).

Now we look for upstreams only if a Component is in the REDHAT namespace. Components in the UPSTREAM namespace are assumed to have no further upstreams of their own.

Before, index / root containers would not report any upstreams in some cases (depending on number of upstreams - more than 1 caused issues), and I think the child / arch-specific containers wouldn't ever report any upstreams regardless of count.

Now, both types of containers report the same set:  "all SOURCE-type descendants of the root container" as upstreams. This should just be the upstream Go modules and Brew "sources" components (if any - there can be none).

TODO: I need to discuss the Brew "sources" case more. Code, tests, and comments implied containers could be Brew "sources", but this never happens in current data. All containers have PROVIDES-type nodes and none have SOURCE-type nodes.

Before, we tried to make Brew "sources" containers the upstreams of remote-source components (which were found in some Cachito manifest). But as described above, this code never did anything because 1) we never reported upstreams for remote-source components since they're never linked to a build. And 2) there are no Brew "sources" containers. All non-root containers in the DB use the "PROVIDES" node type. Finally, 3) there were bugs in this code that might have given incorrect results if it ever actually got hit.

Now, 1) still applies since remote-source components are always in the UPSTREAM namespace and never the REDHAT namespace. So we won't ever hit the code that tries to set upstreams for a remote-source component. 2) still applies since there are still no Brew "sources" containers (with SOURCE-type nodes instead of PROVIDES-type nodes). I can fix 3) very easily but it seems pointless to do so,

The current code also seems to be working correctly. The arch-specific containers with PROVIDES-type nodes each have a Cachito manifest. We report all the remote-source components in that manifest as "provides" of the arch-specific container that bundles / ships them. The container isn't really an "upstream" that a remote-source component, e.g. some PyPI package, depends on.

So I've removed this unused Cachito nesting code completely, and simplified some of the other code in the same function. Tests were passing before because the code was correct for the test we wrote, but the test itself doesn't seem to match existing data in the DB or other code / how we actually create components. 

I've rewritten the test to match current DB data, and it's still passing. I've also added more thorough logic to test most of the above edge cases, but please let me know if I've missed something.

jasinner (approved):
The logic makes sense to me. I don't see a whole heap of actual changes here, more of a refactoring, with lots of tests and comments to help explain the decisions embedded in the application logic. 

I think perhaps the problem described in CORGI-798 was probably just a side affect of the changes made to the brew collector/task that changed the structure of container graphs with multiple sources. Therefore I see this change as an extension / completion of that refactoring.

I don't see any downside to these changes if they fix CORGI-798.

corgi/core/constants.py
```diff
@@ -55,8 +55,17 @@
 RPMMOD_CONDITION = Q(type="RPMMOD")
 INDEX_CONTAINER_CONDITION = Q(type="OCI", arch="noarch")
 SERVICE_REPO_CONDITION = Q(type="GITHUB", namespace="REDHAT", arch="noarch")
+# TODO: Adding this means we do SCA on MAVEN components, probably not what we want
+#  add logic to SCA task to skip like we do for RPMMODs
+#  We would also include these RH Maven components in our API / manifests as "root components"
+#  even if they're not part of a Middleware product. We should think about this more
```
juspence: @kgrant-rh Any thoughts here? I'm probably going to drop this commit for now, since we definitely don't want to treat Red Hat Maven components as "root components" in e.g. a RHEL manifest, outside of a middleware product's manifest.

Maybe (in a separate ticket, not a quick fix here) we need to clean up our data so that all SoftwareBuilds have exactly one root component, and each root component has exactly one node with parent=None. Then we might not need to inspect the Component type, and we could just ask for "the linked component on all nodes with no parent" in order to get the set of all root components.

juspence:
@jasinner FYI I'm going to merge this before I forget about it (again). I dropped one commit I need to confirm with Keith, and which probably needs to be handled in a second ticket anyway. Apologies in advance if this causes issues over the weekend - I will clean up any issues first thing on Monday morning.

PR #592, opened 2023-09-01
Minor: ensure return distinct queryset when using tags, add basic component tag django command

JimFuller-RedHat:
Ensure that we return distinct component queryset when filtering by tags. This addresses the case where a component may have several tags with the same name and different values (ex. mytag:value1, mytag:value2).

While I was in the area - added  basic component tag management mainly for testing as griffon will eventually have _full flavoured_  commands.

**Add tags to a set of components** (Ex. named 'curl' though you can add on specific component with --purl)
```
./manage.py tagcomponent --tag-name fedramp --tag-value approved --name curl --operation add
```
**Retrieve components with this tag**
```
./manage.py tagcomponent --tag-name fedramp --tag-value approved --name curl 
```
**Remove tags from components**
```
./manage.py tagcomponent --tag-name fedramp --tag-value approved --name curl --operation remove
```

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -42,7 +42,7 @@ class TagFilter(Filter):
                 )
             else:
                 queryset = queryset.filter(tags__name__icontains=tag)
-        return queryset
+        return queryset.distinct()
 
 
 class ComponentFilter(FilterSet):
--- a/corgi/tasks/management/commands/tagcomponent.py
+++ b/corgi/tasks/management/commands/tagcomponent.py
@@ -1,0 +1,115 @@
+from django.core.management.base import BaseCommand, CommandParser
+
+from corgi.core.models import Component, ComponentTag
+
+
+class Command(BaseCommand):
+
+    help = "Component tag management."
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument(
+            "--purl",
+            dest="component_purl",
+            help="Select component to tag using component PURL",
+        )
+        parser.add_argument(
+            "--name",
+            dest="component_name",
+            help="Select component(s) to tag using component name",
+        )
+        parser.add_argument(
+            "--tag-name",
+            dest="tag_name",
+            help="Set tag name",
+        )
+        parser.add_argument(
+            "--tag-value",
+            dest="tag_value",
+            help="Set tag value",
+        )
+        parser.add_argument(
+            "--operation",
+            dest="operation",
+            help="Set tag operations (get,add,remove)",
+        )
+
+    def handle(self, *args, **options):
+        operation = "get"
+        if options["operation"]:
+            operation = options["operation"]
+        component_name = None
+        if options["component_name"]:
+            component_name = options["component_name"]
+        component_purl = None
+        if options["component_purl"]:
+            component_purl = options["component_purl"]
+        tag_name = None
+        if options["tag_name"]:
+            tag_name = options["tag_name"]
+        tag_value = ""
+        if options["tag_value"]:
+            tag_value = options["tag_value"]
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"operation: {operation}",
+            )
+        )
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"component: {component_name}{component_purl}",
+            )
+        )
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"tag: {tag_name}:{tag_value}",
+            )
+        )
+
+        tag_params = {}
+        if tag_name:
+            tag_params["name"] = tag_name
+        if tag_value:
+            tag_params["value"] = tag_value
+
+        component_params = {}
+        if component_purl:
+            component_params["purl"] = component_purl
+        if component_name:
+            component_params["name"] = component_name
+
+        components = Component.objects.filter(**component_params)
+
+        if operation == "get":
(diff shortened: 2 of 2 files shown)


corgi/api/filters.py
```diff
@@ -42,7 +42,7 @@ def filter(self, queryset: QuerySet, value: str) -> QuerySet:
                 )
             else:
                 queryset = queryset.filter(tags__name__icontains=tag)
-        return queryset
+        return queryset.distinct()
```
juspence, line 45: Let's add the .distinct() to the exact part of the queryset that causes issues / multiple results to appear:
```suggestion
                queryset = queryset.filter(tags__name__icontains=tag).distinct()
        return queryset
```
Let's also add a test that confirms this fixes the issue. I think based on the PR description, you're saying that the `__name__icontains` filter returns multiple results for the same component, right? And this happens when any component has multiple tag values with the same tag name, like `{"tag_name": "value_one", "tag_name": "value_two"}`?

corgi/tasks/management/commands/tagcomponent.py
```diff
@@ -0,0 +1,115 @@
...
+            self.style.SUCCESS(
+                f"tag: {tag_name}:{tag_value}",
+            )
+        )
+
+        tag_params = {}
+        if tag_name:
+            tag_params["name"] = tag_name
+        if tag_value:
+            tag_params["value"] = tag_value
```
juspence, line 73: See below comment for explanation:
```suggestion
        if tag_name:
            tag_params["tags__name"] = tag_name
        if tag_value:
            tag_params["tags__value"] = tag_value
```

corgi/tasks/management/commands/tagcomponent.py
```diff
@@ -0,0 +1,115 @@
...
+        if operation == "get":
+            component_ids = ComponentTag.objects.filter(**tag_params).values_list(
+                "tagged_model_id", flat=True
+            )
+            for c in Component.objects.filter(uuid__in=component_ids):
+                self.stdout.write(
+                    self.style.SUCCESS(
+                        f"{c.name} {c.nvr} {c.purl}",
+                    )
+                )
```
juspence, line 92: You can use Django's related field __lookups, like tags__name and tags__value, to iterate over the Components directly instead of needing to make separate queries:
```suggestion
            for c in Component.objects.filter(**tag_params):
                self.stdout.write(
                    self.style.SUCCESS(
                        f"{c.name} {c.nvr} {c.purl}",
                    )
                )
```

corgi/tasks/management/commands/tagcomponent.py
```diff
@@ -0,0 +1,115 @@
...
+                    ComponentTag.objects.filter(**tag_params, tagged_model_id=c.uuid).delete()
+                self.style.SUCCESS(
+                    f"{operation} tags successful on {components.count()} components.",
+                )
+            else:
+                componenttags = ComponentTag.objects.filter(**tag_params)
+                self.style.SUCCESS(
+                    f"{operation} tags successful on {componenttags.count()} components.",
+                )
+                componenttags.delete()
```
juspence, line 106:
```suggestion
                componenttags = ComponentTag.objects.filter(**tag_params)
            if components:
                componenttags == componenttags.filter(tagged_model_id__in=components)
            deleted_count, deleted_dict = componenttags.delete()
            self.style.SUCCESS(
                f"{operation} tags successful on (deleted_count} models: {deleted_dict}",
            )
```

corgi/tasks/management/commands/tagcomponent.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if operation == "add":
+            for c in components:
+                tag = ComponentTag.objects.get_or_create(**tag_params, tagged_model_id=c.uuid)
```
juspence, line 110: I just realized our current tag model will lead to duplicate data. We'll create one tag with the same name and value for each component which uses that tag.

Maybe that's expected, but I thought we wanted to support having one tag, like "support": "ooss" for example, that could be shared across multiple components. That way we could update or delete tags in bulk across all components that use them.

We could change the links between tags and components to use a ManyToMany field instead, and then above could become something like:
```suggestion
            for c in components:
                tag = ComponentTag.objects.get_or_create(**tag_params)
                # Directly setting either side of a many-to-many relation is prohibited, must use .set(), .add(), etc.
                tag.tagged_models.add(*components)
```

corgi/api/filters.py
JimFuller-RedHat: thx will change distinct location, great advice!

corgi/tasks/management/commands/tagcomponent.py
JimFuller-RedHat: ya I brute forced this command ... as a test tool - griffon should quickly supersede anything we do here but will change on your good advice. thx!

corgi/tasks/management/commands/tagcomponent.py
JimFuller-RedHat: yes, much more efficient representation - nice spot ! we should change this ... maybe as a different issue ? or can raise an issue and bundle fix here.

juspence: Whatever you'd prefer is fine :smile: I just mentioned here since I saw the code and wasn't sure if it was expected. It's fine to go back and fix later as part of another ticket.

JimFuller-RedHat:
going to close this for now

PR #593, opened 2023-09-01
nginx and gunicorn should have matching keepalive settings

JimFuller-RedHat:
will be a matching component-registry-ops PR for setting in stage/prod

### DIFF
--- a/etc/nginx/proxy.conf
+++ b/etc/nginx/proxy.conf
@@ -34,4 +34,7 @@ location / {
     proxy_read_timeout 300s;
     # Not used in production, but required in Dev to get correct links in the /api/v1 menu
     proxy_set_header   Host localhost:8080;
+
+    keepalive_timeout 5s;
+    keepalive_requests 1000;
 }
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -18,6 +18,8 @@ forwarded_allow_ips = "*"
 
 timeout = 300
 
+keepalive = 5
+
 if not running_dev():
     # Saves memory in the worker process, but breaks --reload
     preload_app = True


etc/nginx/proxy.conf
```diff
@@ -34,4 +34,7 @@ location / {
     proxy_read_timeout 300s;
     # Not used in production, but required in Dev to get correct links in the /api/v1 menu
     proxy_set_header   Host localhost:8080;
+
+    keepalive_timeout 5s;
+    keepalive_requests 1000;
```
juspence, line 39: I'm not sure we can define these settings here, since the context says "upstream" in Nginx docs. This config file only controls proxy settings, where the context says "proxy" in Nginx docs.

The only keepalive setting I saw for proxied connections seemed to have reasonable defaults:
https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_socket_keepalive

For these settings, it also looks like `keepalive_requests 1000;` is already the default:
https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive_requests

And I don't think these settings will have any affect anyway, since they seem to be for upstream servers (direct connections?) and not proxied servers (what we're using Nginx for):
https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

gunicorn_config.py
```diff
@@ -18,6 +18,8 @@
 
 timeout = 300
 
+keepalive = 5
+
```
juspence, line 22: If Nginx is using keepalives at all, it seems to default to a 60-second timeout. Since I don't think we can easily override those settings, we could change just the Gunicorn timeout to match (which would also avoid needing an ops repo PR):
https://docs.gunicorn.org/en/stable/settings.html#keepalive

JimFuller-RedHat: currently doing an analysis of performance across web (client) ---> nginx --> gunicorn --> django --> pg

keepalive is one thing we should set with nginx (and as nginx supports HTTP/2 we will eventually adjust bindings to make HTTP/2 calls ... its a shame gunicorn/django are stuck 25 years in the past with HTTP/1.1 ...)

60s is far too long for most of our queries (esp with pagination) ... and with async qeuries we keep on hanging onto too many conns for too long - I tested with 5s and it is nice. I still need to set this for stage/prod nginx (hence component-registory-ops PR).

In theory we should not have to set keepalive in gunicorn at all but local testing suggests (as well as some other ppl's observations) matching them is best ... I was going to do a test deploy to stage (after all the other processing finishes) to test out the theory - so the gunicorn setting might drop in this PR.

PR #595, opened 2023-09-05
CORGI-808 clean remove yum repos

jasinner:
Remove yum_repo product references from Components when those yum_repos are removed from a ps_update_stream in product_definitions.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -265,18 +265,29 @@ def parse_product_stream(
             "obj": product_stream,
         },
     )
-    if not stream_created:
-        _clean_orphaned_brew_tags(set(brew_tags_dict.keys()), name, str(product_stream.pk))
     parse_errata_info(errata_info, product, product_stream, product_stream_node, product_version)
+    if not stream_created:
+        _clean_orphaned_relations_and_builds(
+            set(brew_tags_dict.keys()),
+            name,
+            str(product_stream.pk),
+            ProductComponentRelation.Type.BREW_TAG,
+        )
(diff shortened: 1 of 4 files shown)


corgi/tasks/prod_defs.py
```diff
@@ -265,18 +265,29 @@ def parse_product_stream(
             "obj": product_stream,
         },
     )
-    if not stream_created:
-        _clean_orphaned_brew_tags(set(brew_tags_dict.keys()), name, str(product_stream.pk))
     parse_errata_info(errata_info, product, product_stream, product_stream_node, product_version)
+    if not stream_created:
```
JimFuller-RedHat, line 269: might be worth injecting some comments on rationale eg. this is always much faster then full deletion and reingest to make clear to our future selves.

corgi/tasks/prod_defs.py
```diff
@@ -265,18 +265,29 @@ def parse_product_stream(
...
+    product_stream_pk: str,
+    relation_type: ProductComponentRelation.Type,
+) -> None:
     relations_to_remove = (
-        ProductComponentRelation.objects.filter(
-            type=ProductComponentRelation.Type.BREW_TAG, product_ref=name
-        )
-        .exclude(external_system_id__in=brew_tags)
+        ProductComponentRelation.objects.filter(type=relation_type, product_ref=name)
+        .exclude(external_system_id__in=new_external_system_ids)
```
JimFuller-RedHat, line 290: FWIW I seem to recall django .exclude being slower then using negated Q

JimFuller-RedHat (approved):
apart from checks failing - LGTM

PR #596, opened 2023-09-05
move performance test to a scheduled task

jasinner:
I'll created a new scheduled pipeline in CI where we can run the performance test daily. This is how the reconciliation scripts run in the ops repository. It decouples the most recently code changes from the test which is actually testing the previous changes, not the current change.


JimFuller-RedHat (approved):
good idea and LGTM

PR #604, opened 2023-09-07
CORGI-65 - Task and collector for pyxis manifests

ralphbean:
For https://issues.redhat.com/browse/CORGI-65


JimFuller-RedHat:
very cool - we will review (and fix reason why CI did not run for your PR!).

jasinner:
Thanks Ralph! Looks like all the pieces are there well done. We will review it more detail in the coming days.

ralphbean:
Thanks guys. Maybe CI failed because this was submitted from my fork? (Untrusted!)

JimFuller-RedHat:
> Thanks guys. Maybe CI failed because this was submitted from my fork? (Untrusted!)

yes that will be it ;)

tests/test_pyxis_task.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.django_db
+@patch("corgi.tasks.sca.cpu_software_composition_analysis.delay")
+@patch("corgi.collectors.pyxis.session.post")
+def test_slow_fetch_pyxis_manifest(post, sca):
+    with open("tests/data/pyxis/manifest.json", "r") as data:
+        manifest = json.loads(data.read())
+    post.return_value.json.return_value = {"data": {"get_content_manifest": {"data": manifest}}}
+
+    image_id = "64dccc5b6d82013739c4f7b8"
+    slow_fetch_pyxis_manifest(image_id, SoftwareBuild.Type.BREW)
```
kgrant-rh: The reason behind the second argument here isn't totally clear to me.

ralphbean: That's an error! Good catch.

ralphbean: Resolved in 442b443.

ralphbean:
I have a mypy error that I'm struggling to understand:

```
corgi/collectors/pyxis.py:127: error: Argument "cert" to "post" of "Session" has incompatible type "Tuple[Optional[str], Optional[str]]"; expected "Optional[Union[str, Tuple[str, str]]]"  [arg-type]
```

**edit** nevermind, @kgrant-rh solved it. :)

config/settings/base.py
```diff
@@ -470,6 +470,11 @@
 
 LOOKASIDE_CACHE_BASE_URL = os.getenv("CORGI_LOOKASIDE_CACHE_URL")
 
+# Pyxis details
+PYXIS_GRAPHQL_URL = os.getenv("CORGI_PYXIS_GRAPHQL_URL")
+PYXIS_CERT = os.getenv("CORGI_PYXIS_CERT")
```
kgrant-rh: `os.getenv` has a mypy type of `Optional[str]` because it can return `None` if there's no matching variable, I think that's probably the cause of the error you encountered. If you add a default value like `""`, mypy should be happy.

ralphbean: Good find! Fixed in 8bc8a96.

tests/test_pyxis_task.py
```diff
@@ -0,0 +1,41 @@
...
+
+    image_index = Component.objects.get(
+        name="quay.io/redhat-appstudio/image-controller", type=Component.Type.CONTAINER_IMAGE
+    )
+
+    software_build = SoftwareBuild.objects.get(
+        build_id=image_id, build_type=SoftwareBuild.Type.PYXIS
+    )
+
+    components = {}
```
jasinner, line 32: This seems like an unusual way to verify the component taxonomy. Perhaps we are missing a call to image_index.save_component_taxonomy which populates the components attribute of software_build.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+def _slow_fetch_pyxis_manifest_for_repository(
+    oid: str,
+    image_id: str,
+    repository: Mapping[str, Any],
+    manifest: Mapping[str, Any],
+    save_product: bool = True,
+    force_process: bool = False,
+) -> bool:
+
+    component_name = f"{repository['registry']}/{repository['repository']}"
```
jasinner: According to the PURL spec, the 'name' part should be the last fragment of the repository url, 'image-controller' in the case of 'quay.io/redhat-appstudio/image-controller'. Assuming that the place where the container will be published to.

For the Brew collector we are storing the container component names internally as the Brew package name, and then only when rendering the Package URL do we replace it with the last fragment of the repository url.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+    else:
+        # This really shouldn't happen
+        raise ValueError(f"No completion_time for build {image_id}")
+
+    softwarebuild, build_created = SoftwareBuild.objects.get_or_create(
+        build_id=image_id,
+        build_type=SoftwareBuild.Type.PYXIS,
+        defaults={
+            "name": component_name,
+            "source": manifest["image"].get("source", ""),
```
jasinner, line 80: I don't see the 'source' property of 'image' in the test data. Is this something which is yet to be done in RHTAP/snitch? This should be populated with the build source code repository.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+            build_id=image_id, build_type=SoftwareBuild.Type.PYXIS, software_build__isnull=True
+        ).update(software_build=softwarebuild)
+
+    if not force_process and not build_created:
+        # If another task starts while this task is downloading data this can result in processing
+        # the same build twice, let's just bail out here to save cpu
+        logger.warning("SoftwareBuild with build_id %s already existed, not reprocessing", image_id)
+        return False
+
+    component_version = (
```
jasinner: These also look like they are not populated in the test data. In addition to the image sha256 it would be great to have a semantic version or similar which could be used to more easily compare the order of container image releases with each other.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+        repository["manifest_list_digest"] or repository["manifest_schema2_digest"] or ""
+    )
+    root_node, root_created = save_container(
+        softwarebuild, component_name, component_version, manifest
+    )
+
```
jasinner, line 117: This is a good place to call `slow_save_taxonomy.delay(build_id, build_type)` which persist the node structure onto the components/provides/sources/upstreams attributes of each component.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+) -> tuple[ComponentNode, bool]:
+    obj, root_created = Component.objects.update_or_create(
+        type=Component.Type.CONTAINER_IMAGE,
+        name=component_name,
+        version=component_version,
+        release="",
+        arch="noarch",
+        defaults={
+            "namespace": Component.Namespace.REDHAT,
+            "related_url": component_name,
```
jasinner: This is good, let's continue to save the repository_url to the related_url field.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+        build_created,
+        root_created,
+        image_id,
+        SoftwareBuild.Type.PYXIS,
+    )
+    logger.info("Finished fetching pyxis image: (%s, %s)", image_id, SoftwareBuild.Type.PYXIS)
+    return build_created or root_created
+
+
+def save_container(
```
jasinner, line 138: In additional to the internal source code repository, it would be ideal to also persist any identifiable upstream source repository for this image as a distinct component of in the 'UPSTREAM' namespace, with a node type of 'SOURCE'. See the 'save_upstream' function of corgi.tasks.brew.

corgi/tasks/pyxis.py
```diff
@@ -0,0 +1,261 @@
...
+
+    if component_type in mapping:
+        component_type = mapping[component_type]
+    else:
+        raise ValueError(f"Tried to create component with invalid component_type: {component_type}")
+
+    # A related url can be only one of any number of externalReferences provided. Pick the first.
+    related_url = ""
+    for reference in component["external_references"] or []:
+        related_url = reference["url"]
```
jasinner, line 207: Similar to the comment above we could create a new Components in the 'UPSTREAM' namespace with a relationship to this component of type ComponentNode.Type.SOURCE. We should only do this if the external reference can be parsed into the one of the PURL types, eg. a github reference, or other external code repository (use GENERIC purl type). 
External Reference is hack for historical RPM specfile Source0/URL fields.

jasinner (commented):
Great job Ralph. Thanks very much for this Pull Request. I think most of the feedback here about how to persist the data could probably be taken up by me, or someone from our dedicated team.

Probably the most important piece is just get a semantic version for the container image, and a repository location, ideally an upstream location for the container source as well if available. As long as we can get that data from RHTAP/snitch/Pyxis we should be able to handle the rest.

corgi/tasks/pyxis.py
ralphbean: Yes, I asked Keith G to file a RFE for that just on Friday. We don't provide it in the manifests coming out of RHTAP yet.

kgrant-rh: Here is that RFE: https://issues.redhat.com/browse/RHTAP-1570

tests/data/pyxis/umb/manifest_create.json
```diff
@@ -0,0 +1,71 @@
...
+    "JMS_AMQP_MESSAGE_FORMAT": "0",
+    "JMS_AMQP_NATIVE": "false",
+    "amq6100_destination": "queue://Consumer.datanommer.psi-ocp-bm-stage.VirtualTopic.eng.>",
+    "amq6100_originalDestination": "topic://VirtualTopic.eng.snitch.contentmanifest.create",
+    "correlation-id": "0a1b9095-d4c8-48e7-84a6-43106153f875",
+    "database": "data",
+    "destination": "/topic/VirtualTopic.eng.snitch.contentmanifest.create",
+    "entityName": "contentManifest",
+    "expires": "0",
+    "message-id": "ID:umb-stage-2.umb-001.preprod.us-east-1.aws.redhat.com-42991-1691783981453-34:61834:0:0:1",
```
juspence: FYI we prefer not to have internal URLs in our repo, but there are already plenty lurking in the Git history so don't worry too much about it.

tests/data/pyxis/README.md
```diff
@@ -0,0 +1,77 @@
+# Pyxis test data
+
+All changes in pyxis database are published to the messagebus by a sub-service called "snitch" ([datagrepper query](https://datagrepper.engineering.redhat.com/raw?category=snitch&delta=127800&rows_per_page=10&contains=manifest)).
+
+Specific message used here is:
+[datagrepper direct link](https://datagrepper.stage.engineering.redhat.com/id?id=ID:umb-stage-2.umb-001.preprod.us-east-1.aws.redhat.com-42991-1691783981453-34:61834:0:0:1&is_raw=true&size=extra-large)
```
juspence: Is it OK for these URLs to be publically available? If not, we have a corgi-ops repo in Gitlab that's private - we usually throw internal docs like this there. Or you could just rewrite the doc to mention "the datagrepper service" and the message ID / topic names.

And the same applies to the Pyxis URL below - I'm fine with including these if others are, but I wanted to mention them just in case.

juspence (commented):
Others have looked over the code so I won't bother duplicating too much of their feedback, I just wanted to raise one point about internal URLs.

Except I will duplicate the "thank you for working on this!" bits :smile:

corgi/tasks/pyxis.py
ralphbean: Fix is underway at https://github.com/redhat-appstudio/release-service-utils/pull/58 to populate those digests in pyxis.

ralphbean: For the semantic version, I think we don't have it at the moment, but I've filed https://issues.redhat.com/browse/RHTAP-1590 to make sure its in queue.

tests/data/pyxis/README.md
ralphbean: To be honest, I'm fine with it - but I scrubbed them in 80342d8.

tests/data/pyxis/umb/manifest_create.json
ralphbean: Scrubbed in 80342d8. (History preserved, though.)

juspence: Ack, we can squash them before merging.

corgi/tasks/pyxis.py
ralphbean: :+1: Updated in a6d9c14.

corgi/tasks/pyxis.py
ralphbean: Added in 4f9e7a3.

corgi/tasks/pyxis.py
ralphbean: Ack, I don't think I have a way to plumb that through pyxis yet.

tests/test_pyxis_task.py
ralphbean: Is this addressed by https://github.com/RedHatProductSecurity/component-registry/commit/4f9e7a369fbe586aae95e15dddcb5e1dfcd3fa01?

corgi/tasks/pyxis.py
ralphbean: I think doing that should wait for https://issues.redhat.com/browse/RHTAPBUGS-781

tests/test_pyxis_task.py
jasinner: Yup, that should do the trick, thanks.

corgi/tasks/pyxis.py
jasinner: Looks good.

corgi/tasks/pyxis.py
jasinner: yup, great.

corgi/tasks/pyxis.py
jasinner: Ok, I added a comment there as it mentioned the concept of streams which I don't fully understand. Perhaps a different option to tagging the repository is for us to compare the images in a "stream" using the creation_date, or last_updated_date instead if you don't to construction the tags from that anyway.

ralphbean: I see! We can use creation_date for now as a hack in the absence of RHTAP-1590. We'll only be doing single-stream repos for a while (until RHTAP-990 is done) - we just have to remember to return to this and start using that tag instead of creation_date.

ralphbean: Updated in 1c69dfd.

ralphbean:
Squashed, rebased on main, and signed my commit.

jasinner (approved):
I'm OK to merge this. Before it will work in stage we'll need to populate the Pyxis endpoint URL and CERT/KEY.

jasinner:
@ralphbean Can you please resolve the merge conflict with corgi/monitor/consumer.py so I can merge?

ralphbean:
@jasinner rebased and resolved!

PR #607, opened 2023-09-11
CORGI-812: Fix scorecard url

juspence:
@RedHatProductSecurity/corgi-devs Please review. This should fix an error when connecting to the scorecard domain instead of the regular Corgi domain. I'll open an ops repo PR shortly to add the new environment variable this needs.

This also "fixes" an issue with failing performance tests due to a slightly-different number of results. It's not clear to me why the count changed, but it's only off by a bit so I assume this was related to recent linking changes / data cleanup.

But the last time this test started failing, we changed the test without doing more research. That caused us to ignore missing links for RHEL z-streams, and we eventually had to revert CORGI-729 / removing all the CDN_REPO relations to fix this.

Do we need to discuss or research this performance test failure more, instead of just fixing the tests?


.gitlab-ci.yml
```diff
@@ -142,8 +139,6 @@ test-performance:
     - $DNF_WITH_OPTIONS install $RPM_REQUIREMENTS
     - python3.9 -m pip install tox
     - tox -e corgi -- -m performance --no-cov
-  rules:
```
jasinner, line 145: I'd rather keep the performance test as a scheduled task as it doesn't test the current branch, but the main branch itself. It's annoying to have the CI/CD fail for something not related to your current changes.

jasinner:
I'd say reducing the number of expected results here is the correct course of action. We use the deptopia-reconcilation script to measure the accuracy of the data, this test should remain focused on performance.

.gitlab-ci.yml
juspence: OK, I've put it back. Personally I find it useful to run the tests sooner, since waiting until the next day means I have to remember which PRs I did yesterday and which of those might have broken things. Running it after every PR means we get feedback sooner if something isn't behaving the way we expect.

I guess the right way is to run the tests after the deploy-stage step, but that requires some larger refactoring to check the deployment status and wait for all pods to roll out before running the tests. If we don't have that, the tests will randomly fail whenever the web pods are down (e.g. migration still running) or are still on the older version.

juspence:
OK, I'm going to merge this in that case (assert in test is fixed and still runs only once per day). I don't like relying on the reconciliation script alone - it does a good job but these tests seem to catch issues sooner / are more noticeable since they're part of the pipeline.

For example, we removed all the CDN_REPO relations, and the perf tests immediately noticed the issue with missing RHEL Z-stream components. I'm just worried we're ignoring a useful signal about some recent PR that didn't link data correctly or something.

PR #608, opened 2023-09-12
CORGI-746 - refactor latest component filter into pg storedproc

JimFuller-RedHat:
By migrating rpm compare function from python (rpmCompare) to postgres function we get the following benefits:

* _avoid out of memory_ errors (eg. Bad Gateway) in gunicorn as processing is now done in db instead of slow python
* faster raw performance of rpmvercmp and rpmvercmp_epoch in db versus looping through sql query and processing one at a time
* horizontal scalable query eg. when we add read replicas we get 1X performance increase for each read replica
* avoid 'cross-join' fun which ORM's seem to get wrong 

This initial commit focuses on parity with exiting so we ensure we have correct reconciliation - there are a few more performance enhancements we can consider if needed.

While in the area;
* fixed local dev pg config eg. we now set 300 MAX CONNs in django db and we should be equal or higher then that in pg itself. 
* updated compon_latest_idx index as we use 'namespace' in ROOT_COMPONENTS condition

**Note** - that get_latest_component postgres function will probably retire once we move to graphdb.


JimFuller-RedHat:
note tests are now passing, but CI is failing due to an underlying mirror failing (for now).

tests/test_api.py
```diff
@@ -427,7 +435,7 @@ def test_latest_components_by_streams_filter(client, api_path):
     # Even though it's "not the latest" for the RHEL8 stream
     # Report newest_component as the latest for the RHEL8 stream
     # Even though both have the same name / only one is latest overall
-    assert response["results"][0]["nevra"] == newer_component.nevra
+    print(response["results"])
```
jasinner: I think this should probably be an assert, perhaps keep the old one?

tests/test_model.py
```diff
@@ -1107,39 +1129,47 @@ def test_latest_components_queryset(client, api_path):
...
+    )
+    assert len(non_latest_non_root_components) == 18
     # Red and blue components for 2 arches each have 1 non-latest
     # Red and blue components for upstream (non-root) each have 1 non-latest
     # Red and blue components for non-RPMs (non-root) each have 1 non-latest
-    for component in non_latest_non_root_components:
-        assert (
-            component.purl == components[(component.type, component.name, component.arch)][0].purl
-        )
+    # for component in non_latest_non_root_components:
```
jasinner, line 1272: Can we restore this part of the test?

jasinner (approved):
Reconciliation is looking much better now. I did get a read timeout 300s from python requests running the reconciliation script however.

tests/test_model.py
JimFuller-RedHat: yes will do

JimFuller-RedHat:
oh I think I did something stupid that causes read timeout - will fix

JimFuller-RedHat:
FWIW @jasinner ready for final review ... there is a minor test failure (unrelated to ordering) which I will fix ...

jasinner (approved):
Seems to maintain data quality in respect to deptopia reconciliation.

PR #609, opened 2023-09-12
CORGI-811 stream multi variants

jasinner:
This addressed CORGI-811 by only attaching variants from errata_info to active streams.

It also removes some logic from brew_tag parsing which skipped some streams. This is no longer required as we don't associate variants with these streams any more.

### DIFF
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -22,6 +22,7 @@ from corgi.core.models import (
     ProductVersion,
     SoftwareBuild,
 )
+from corgi.tasks.brew import slow_save_taxonomy
 from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
 logger = get_task_logger(__name__)
@@ -322,7 +323,8 @@ def _parse_cpes_from_brew_tags(
     stream_name: str,
     version_name: str,
 ) -> list[str]:
-    """Match streams using brew_tags to Errata Tool Product Versions and their Variants"""
+    """Match streams using brew_tags to Errata Tool Product Versions and their Variants
+    Return a list of CPEs for the matched Variants"""
     if len(brew_tags) > 0:
         logger.debug(f"Found brew tags {brew_tags} in product stream: {stream_name}")
         cpes: set[str] = set()
@@ -352,7 +354,8 @@ def parse_errata_info(
     product_version: ProductVersion,
 ):
     """Parse and create ProductVariants from errata_info in product-definitions.json"""
-    et_product_versions_set = set(product_stream.et_product_versions)
+    # Reset the stream's et_product_versions_set so we keep it up to date with what's in prod_defs.
+    et_product_versions_set = set()
     for et_product in errata_info:
         et_product_name = et_product.pop("product_name")
         et_product_versions = et_product.pop("product_versions")
@@ -360,6 +363,11 @@ def parse_errata_info(
         for et_product_version in et_product_versions:
             et_pv_name = et_product_version["name"]
             et_product_versions_set.add(et_pv_name)
+
+            # This is a workaround for CORGI-811 which attaches variants from errata_info only to
+            # active streams
+            if not product_stream.active:
+                continue
 
             for variant in et_product_version["variants"]:
                 logger.debug("Creating or updating Product Variant: name=%s", variant)
@@ -384,13 +392,32 @@ def parse_errata_info(
                         },
                     },
                 )
-                ProductNode.objects.get_or_create(
+                # If multiple active streams share the same errata_info,
+                # use update to relink the node to a different parent product_stream_node
+                # the last matching stream always wins / gets the link
+                node, node_created = ProductNode.objects.get_or_create(
                     object_id=product_variant.pk,
                     defaults={
                         "parent": product_stream_node,
                         "obj": product_variant,
                     },
                 )
+                if node.parent != product_stream_node:
+                    node.parent = product_stream_node
+                    node.save()
+                    # save the existing builds to adjust the product_stream
+                    # we don't remove the old product_stream from the builds because are related via
+                    # the same variant. This a workaround for CORGI-811 stream to variant should be
+                    # many-to-many relationship
+                    builds_to_update = (
+                        ProductComponentRelation.objects.filter(
+                            type__in=ProductComponentRelation.VARIANT_TYPES, product_ref=variant
+                        )
+                        .values_list("build_id", "build_type")
+                        .distinct()
+                    )
+                    for build_id, build_type in builds_to_update:
+                        slow_save_taxonomy.delay(build_id, build_type)
                 product_variant.save_product_taxonomy()
     # persist et_product_versions plucked from errata_info
     product_stream.et_product_versions = sorted(et_product_versions_set)
--- a/corgi/tasks/prod_defs.py
+++ b/corgi/tasks/prod_defs.py
@@ -323,22 +323,7 @@ def _parse_cpes_from_brew_tags(
     version_name: str,
 ) -> list[str]:
     """Match streams using brew_tags to Errata Tool Product Versions and their Variants"""
-    # quay-3 streams brew_tags all share a single variant, and we have a one-to-many
-    # cardinality between streams and variants. quay-3 streams should probably be consolidated
-    # into a single stream in prod_defs. See also PROJQUAY-5312.
-    # The rhn_satellite_6 streams have brew tags, but those brew tags are associated
-    # with the RHEL-7-SATELLITE-6.10 ET Product Version. We skip them
-    # here to ensure the rhn_satelite_6.10 variants only get linked with that stream
-    # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
-    # active. See CORGI-546 for more details.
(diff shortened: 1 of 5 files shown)


corgi/tasks/prod_defs.py
```diff
@@ -323,22 +323,7 @@ def _parse_cpes_from_brew_tags(
     version_name: str,
 ) -> list[str]:
     """Match streams using brew_tags to Errata Tool Product Versions and their Variants"""
```
juspence: Maybe a small update to the comment is helpful:
```suggestion
    """Match streams using brew_tags to Errata Tool Product Versions and their Variants
       Return a list of CPEs for the matched Variants"""
```

corgi/tasks/prod_defs.py
```diff
@@ -323,22 +323,7 @@ def _parse_cpes_from_brew_tags(
...
-    # quay-3 streams brew_tags all share a single variant, and we have a one-to-many
-    # cardinality between streams and variants. quay-3 streams should probably be consolidated
-    # into a single stream in prod_defs. See also PROJQUAY-5312.
-    # The rhn_satellite_6 streams have brew tags, but those brew tags are associated
-    # with the RHEL-7-SATELLITE-6.10 ET Product Version. We skip them
-    # here to ensure the rhn_satelite_6.10 variants only get linked with that stream
-    # I haven't filed a product bug to get them fixed since 6.7 - 6.9 are no longer
-    # active. See CORGI-546 for more details.
-    # Also skip brew_tag matching for dts ps_module which share Variants/Brew Tags with rhscl
-    # See CORGI-726
```
juspence, line 335: For Satellite, the older streams are no longer active so I think this change doesn't matter. But if we stop skipping the DTS stream, I think we'll end up reporting whatever CPEs we find here in the manifest for that stream. Will that give the correct results, or do we need to keep skipping DTS here?

corgi/tasks/prod_defs.py
```diff
@@ -399,7 +388,7 @@ def parse_errata_info(
                         },
                     },
                 )
-                ProductNode.objects.get_or_create(
+                ProductNode.objects.update_or_create(
```
juspence: Maybe a short comment is helpful to remind us in the future:
```suggestion
                # If multiple active streams share the same errata_info,
                # use update to relink the node to a different parent product_stream_node
                # the last matching stream always wins / gets the link
                ProductNode.objects.update_or_create(
```
This feels a little risky - moving the nodes around caused some bugs before, and I still don't really understand why.

Can we check to see how many variants / errata_info are shared by different streams? I hope that the case you found is the only example. Or at least it's always one active z-stream / one inactive base stream that share variants.

But maybe there's some weird edge case where multiple active streams share the same variant. I think in that case, Variant-type ProductComponentRelations will get processed and components may get linked to the wrong variant, or to both variants if we saved their taxonomies before this change.

Finally, if we linked the variant to stream A before this change, it's possible this new code will relink the variant to stream B instead. So it may trigger some unwanted updates the next time the task runs.

corgi/tasks/prod_defs.py
```diff
@@ -369,6 +354,10 @@ def parse_errata_info(
     """Parse and create ProductVariants from errata_info in product-definitions.json"""
     et_product_versions_set = set(product_stream.et_product_versions)
     for et_product in errata_info:
+        # This is a workaround for CORGI-811 which attaches variants from errata_info only to active
+        # stream
+        if not product_stream.active:
+            continue
```
juspence: I think this should be either:

1. an early return outside the for-loop, since the check will always be the same inside the loop, and we won't do anything inside it if we skip processing this early.

2. OR a continue atter line 366 / `et_product_versions_set.add(et_pv_name)`, so we continue to update et_product_versions even for inactive streams, but without creating and linking any variants.

Unrelated to your changes, is there a small bug on line 355? We always append to this stream's list of et_product_versions. So I think Corgi will still report any old PVs even if they were removed (when data in prod-defs changes).

We could start processing with an empty set, and then add any current PVs to this stream only if they're still present in prod-defs (as of the day this task runs).

tests/test_products.py
```diff
@@ -534,3 +484,77 @@ def test_brew_tag_matching(variant_name, cpe, brew_tag, stream_name, requests_mo
...
+        requests_mock.get(f"{settings.PRODSEC_DASHBOARD_URL}/product-definitions", text=text)
+
+    update_products()
+
+    ga_stream.refresh_from_db()
+    assert ga_stream.productvariants.get_queryset().count() == 0
+    z_stream.refresh_from_db()
+    assert z_stream.productvariants.get_queryset().count() == 1
+    assert not ProductComponentRelation.objects.filter(software_build=sb).exists()
+    assert mock_remove.called_with(
```
juspence: Maybe below is better, or maybe it's expected that the mock is called multiple times?
```suggestion
    assert mock_remove.called_once_with(
```

tests/test_products.py
```diff
@@ -534,3 +484,77 @@ def test_brew_tag_matching(variant_name, cpe, brew_tag, stream_name, requests_mo
...
+            z_stream.pk,
+        )
+    )
+
+
+@pytest.mark.django_db
+def test_multi_variant_active_streams(requests_mock):
+    """Tests that multiple runs of update_products with multiple active streams sharing errata_info
+    does not cause the variants to move back and forth between the streams"""
+    with open("tests/data/prod_defs/proddefs-update-multi-stream-variant-active.json") as prod_defs:
```
juspence, line 565: Thanks for adding this! Should we also add a similar test, where the first call to update_products uses a different test file than the second?

This would simulate a prod-defs change, where some stream starts using the same errata_info as another stream (when it wasn't before).

tests/test_products.py
```diff
@@ -534,3 +484,77 @@ def test_brew_tag_matching(variant_name, cpe, brew_tag, stream_name, requests_mo
...
+
+    update_products()
+
+    ga_stream = ProductStream.objects.get(name="ga_stream")
+    assert ga_stream.productvariants.get_queryset().count() == 0
+    z_stream = ProductStream.objects.get(name="z_stream")
+    assert z_stream.productvariants.get_queryset().count() == 1
+
+    with open("tests/data/prod_defs/proddefs-update-multi-stream-variant-active.json") as prod_defs:
+        text = prod_defs.read()
```
juspence: I think you can just reuse the values stored in the text variable, you should't need to open the same file and read the same data again. We just need to add the second request to the mock.

tests/test_products.py
```diff
@@ -436,56 +436,6 @@ def test_find_by_cpe():
     assert cpe in results
 
 
-@pytest.mark.django_db
-def test_find_by_cpe_substring():
```
juspence: I get why we don't need the test for skipping buggy linking, but should this test be kept? I don't know where or how we use the _find_by_cpe method it's testing.

juspence (approved):
I have some concerns about rare edge-cases (as always), but maybe they never happen and it's fine as-is. Overall LGTM.

corgi/tasks/prod_defs.py
jasinner: Good idea I missed the fact that we are adding to the et_product_versions_set on line 366, thanks.

Good idea bout change 355 to an empty set.

corgi/tasks/prod_defs.py
jasinner: The cpes_from_brew_tags property now has the correct CPEs for dts streams.

corgi/tasks/prod_defs.py
jasinner: I wrote a little skip to check prod_defs data and found that the only variant which shares streams is the one I previously mentioned `8Base-CertSys-10.4`. I have a test for multiple active streams, and verified the variant didn't move when multiple update_product tasks were run. The only issue that could happen is if the order of the streams was changed in product_definitions. But I think the likelyhood of that happening is so low we should probably ignore it for now.

```
streams_by_variant = defaultdict(list)
for product in ProdDefs.load_product_definitions():
    for version in product["product_versions"]:
        for stream in version["product_streams"]:
            if "errata_info" in stream:
                for errata_product in stream["errata_info"]:
                    for errata_version in errata_product["product_versions"]:
                        for errata_variant in errata_version["variants"]:
                            streams_by_variant[errata_variant].append(stream["id"])
```

tests/test_products.py
jasinner: Good catch. Sorry I removed that by accident. Adding it back.

tests/test_products.py
jasinner: fixed.

tests/test_products.py
jasinner: That's what the other test `test_multi_variant_streams` does.

tests/test_products.py
jasinner: Good idea, fixed.

corgi/tasks/prod_defs.py
jasinner: updated

tests/test_products.py
juspence: That test is using one inactive stream and one active stream. I was thinking here we should add a test for two different streams which are both active. But you confirmed above this doesn't happen today, so I think it's fine.

If it does start happening, I think we'll see churn with Variants constantly moving back and forth between streams. Fingers crossed this won't be a problem...

PR #610, opened 2023-09-14
CORGI-723 create relations after errata shipped

jasinner:
These changes ensure that we only ever have errata relations for shipped errata. This allows us to remove our reliance on errata brew tags to determine if build has been released or not (CORGI-666). 

In CORGI-723 and CORGI-666 there is mention of "refresh all tags on a build" needing changes, but I couldn't find any related code. I suspect it was maybe removed when we added tag added/removed functionality via UMB monitor.


tests/test_tasks.py
```diff
@@ -86,7 +86,6 @@ def test_slow_update_brew_tags_added(tag_added, is_errata, is_released):
     if is_errata:
         tag_without_suffix = tag_added.rsplit("-", maxsplit=1)[0]
         assert ADVISORY_REGEX.match(tag_added).group() == tag_without_suffix
-        mock_load_errata.assert_called_once_with(tag_without_suffix)
```
juspence, line 89: I guess we're removing this because we're no longer calling "mock_load_errata" for an unshipped erratum. Maybe it's better to use `mock_load_errata.assert_not_called()` instead.

corgi/collectors/errata_tool.py
```diff
@@ -209,23 +208,16 @@ def get_variant_cdn_repo_mapping() -> dict[str, list[str]]:
         }
         return variants_with_repos
 
-    def normalize_erratum_id(self, name: str) -> int:
-        if name.isdigit():
-            return int(name)
+    def get_errata_key_details(self, name: str) -> typing.Tuple[int, bool]:
```
juspence: You don't need to import typing.Tuple, you can just use Python's built-in types instead: `tuple[int, bool]`

corgi/tasks/errata_tool.py
```diff
@@ -48,10 +48,9 @@ def slow_save_errata_product_taxonomy(erratum_id: int) -> None:
 @app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
 def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
     et = ErrataTool()
-    if not erratum_name.isdigit():
-        erratum_id = et.normalize_erratum_id(erratum_name)
-    else:
-        erratum_id = int(erratum_name)
+    erratum_id, shipped_live = et.get_errata_key_details(erratum_name)
+    if not shipped_live:
+        raise ValueError(f"Called slow_load_errata with non-shipped errata {erratum_name}")
```
juspence, line 53: This is good, but I don't recall how all the usages of slow_load_errata work. I think we probably need more checks in other tasks to avoid lots of slow_load_errata failures in our monitoring email.

corgi/tasks/brew.py
```diff
@@ -168,6 +130,70 @@ def slow_fetch_brew_build(
...
+
+
+def _save_children(child_components: list[dict], root_node: ComponentNode) -> bool:
+    any_child_created = False
+    for child_component in child_components:
+        any_child_created |= save_component(child_component, root_node)
+    return any_child_created
+
+
+def _get_completion_time(build_id: str, completion_time):
```
juspence: Thanks for pulling these functions out! They just need a few type hint fixes I think, so mypy can figure out the variable types in the other function, instead of defaulting to Any:
```suggestion
def _get_completion_time(build_id: str, completion_time: str) -> datetime:
```

corgi/tasks/brew.py
```diff
@@ -897,8 +923,10 @@ def slow_update_brew_tags(build_id: str, tag_added: str = "", tag_removed: str =
             build.meta_attr["tags"] = sorted(tags)
             errata_tag = ADVISORY_REGEX.match(tag_added)
             if errata_tag:
-                # Below should automatically create new relations for this build / erratum
-                slow_load_errata.delay(errata_tag.group())
+                advisory_ids = Brew(BUILD_TYPE).parse_advisory_ids([errata_tag.group()])
```
juspence: It's a static method, so you don't need to construct an actual Brew instance:
```suggestion
                advisory_ids = Brew.parse_advisory_ids([errata_tag.group()])
                # If there was a result / this erratum was released:
```

corgi/tasks/brew.py
```diff
@@ -168,6 +130,70 @@ def slow_fetch_brew_build(
...
+    else:
+        # Build has no timestamp even though it's in COMPLETE state?
+        # This really shouldn't happen
+        raise ValueError(f"No completion_time for build {build_id}")
+    return completion_dt
+
+
+def _check_and_save_type(
+    component: dict[str, Any], softwarebuild: SoftwareBuild, save_product: bool, build_id: str
+) -> Optional[Tuple[ComponentNode, bool]]:
```
juspence:
```suggestion
) -> tuple[Optional[ComponentNode], bool]:
```
It might be a little cleaner to always return a tuple below:
```
    else:
        logger.warning(f"Build {build_id} type is not supported: {component['type']}")
        return (None, False)
```
Then above can always unpack the function's return value directly:
```
    root_node, root_created = _check_and_save_type(component, softwarebuild, save_product, build_id)
    if root_node is None:
        return False
```

corgi/tasks/brew.py
```diff
@@ -168,6 +130,70 @@ def slow_fetch_brew_build(
...
+            raise ValueError(f"Could not parse completion_time for build {build_id}")
+    else:
+        # Build has no timestamp even though it's in COMPLETE state?
+        # This really shouldn't happen
+        raise ValueError(f"No completion_time for build {build_id}")
+    return completion_dt
+
+
+def _check_and_save_type(
+    component: dict[str, Any], softwarebuild: SoftwareBuild, save_product: bool, build_id: str
```
juspence, line 167: Is `component: dict[str, Any]` correct? That says the dict will always have string keys, but isn't it possible for the component to have a nested dict like the meta_attr?

If yes, this type hint should probably be just `dict` and then you don't need any import changes in this module.

EDIT: Oops, nevermind...I confused myself while thinking about this, what you have is fine.

corgi/tasks/brew.py
```diff
@@ -168,6 +130,70 @@ def slow_fetch_brew_build(
...
+        logger.info("Requesting fetch of nested build: (%s, %s)", b_id, build_type)
+        slow_fetch_brew_build.delay(
+            b_id, build_type, save_product=save_product, force_process=force_process
+        )
+
+
+def _check_and_load_errata(released_errata_tags, force_process) -> None:
+    """ " for builds with released errata tags set ProductComponentRelation"""
+    for e in released_errata_tags:
+        slow_load_errata.delay(e, force_process=force_process)
```
juspence: Without the check for missing errata_tags / the separate logger statement, this method is basically a one-liner. It could stay part of the main slow_fetch_brew_build task, unless we expect to reuse it somewhere else.
```suggestion
def _check_and_load_errata(released_errata_tags: list[str], force_process: bool) -> None:
    """For builds with released errata tags, create a ProductComponentRelation for each erratum"""
    for released_erratum in released_errata_tags:
        slow_load_errata.delay(released_erratum, force_process=force_process)
```

corgi/tasks/brew.py
```diff
@@ -110,47 +97,22 @@ def slow_fetch_brew_build(
...
         slow_save_taxonomy.delay(build_id, build_type)
 
-    # for builds with errata tags set ProductComponentRelation
-    # get_component_data always calls _extract_advisory_ids to set tags, but list may be empty
-    if not build_meta["errata_tags"]:
-        logger.info("no errata tags")
-    else:
-        for e in build_meta["errata_tags"]:
-            slow_load_errata.delay(e, force_process=force_process)
+    _check_and_load_errata(build_meta.get("released_errata_tags", []), force_process)
```
juspence:
```suggestion
    _check_and_load_errata(build_meta.get("released_errata_tags", ()), force_process)
```

corgi/tasks/brew.py
```diff
@@ -168,6 +130,70 @@ def slow_fetch_brew_build(
...
+
+def _check_and_load_errata(released_errata_tags, force_process) -> None:
+    """ " for builds with released errata tags set ProductComponentRelation"""
+    for e in released_errata_tags:
+        slow_load_errata.delay(e, force_process=force_process)
+
+
+def _check_and_load_brew_tags(
+    build_id: str, tags: list[str], build_type: str, softwarebuild: SoftwareBuild
+) -> int:
```
juspence: Is below missing `return 0` / `return new_relations_count` lines, or should the return type hint get fixed?
```suggestion
) -> None:
```

tests/data/errata/invalid.json
```diff
@@ -0,0 +1,16 @@
...
+      "status": "SHIPPED_LIVE"
+    },
+    "rhsa": {
+      "id": 120143,
+      "revision": 2,
+      "fulladvisory": "RHBA-2023:5018-02",
+      "status": "SHIPPED_LIVE"
+    }
+  }
+}
```
juspence, line 16: If this file is supposed to represent invalid data, should it be an empty JSON blob instead? Or if not, let's add a comment to the test that uses this file below, explaining what's invalid about it / why we expect it to fail.

juspence (approved):
This LGTM! I don't fully recall the context, but I think those comments about "refresh all tags on a build" needing changes were saying we can't rely on Brew builds to always have a correct list of (errata) tags. Stream tags are reliable and don't need changes.

So in some future MR, we should create ERRATA-type relations using SHIPPED_LIVE messages from ET instead of tag_added / tag_removed messages from Brew. That doesn't need to be part of this PR, and what you have here looks like a good start that can stand on its own. Thanks for working on this!

corgi/tasks/brew.py
jasinner: ok, inlined

corgi/tasks/errata_tool.py
jasinner: I think I'd added the necessary checks in calling code in this PR. These commits:

https://github.com/RedHatProductSecurity/component-registry/pull/610/commits/ddf01d185a176fb9effbbdf65328db374510529d
https://github.com/RedHatProductSecurity/component-registry/pull/610/commits/0a9eec672e356898f5c22ddb32465922947c0c3c

tests/data/errata/invalid.json
jasinner: There are 2 errata here, one of type rhba, and one of type rhsa. We only expect to get one type here, since we only looked up one erratum id. 
I added a comment to the test.

tests/test_tasks.py
jasinner: yeah, ok.

corgi/tasks/errata_tool.py
juspence: That handles all the `slow_load_errata.delay()` usages, but I thought there were one or two `app.send_task("slow_load_errata", *args, **kwargs)` usages that also needed to get fixed.

Since app.send_task is very indirect (string arg with a separate function name instead of a direct function call), I think PyCharm's "find usages" feature doesn't report these. Maybe I'm misremembering and those app.send_task calls to slow_load_errata don't actually exist.

jasinner: I had another look, but didn't see any send_task calls to slow_load_errata.

PR #611, opened 2023-09-14
CORGI-828: Add tagging for components, to support future work

juspence:
@jasinner FYI. I just got the existing tests to start passing, so you can look during your Friday if you'd like to and are curious.

This isn't ready for a full review yet, since I still need to confirm exact requirements and make sure I didn't do something silly here. I also need to add new tests and fix some more bugs.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -52,7 +52,7 @@ class ComponentFilter(FilterSet):
         model = Component
         # Fields that are matched to a filter using their Django model field type and default
         # __exact lookups.
-        fields = ("type", "namespace", "name", "version", "release", "arch", "nvr", "nevra")
+        fields = ("purl", "type", "namespace", "name", "version", "release", "arch", "nvr", "nevra")
 
     # Custom filters
     re_name = CharFilter(lookup_expr="iregex", field_name="name")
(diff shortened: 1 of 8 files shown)


juspence:
Closing due to comments from Rogue in email / Gchat. I'll add final bugfixes / tests and reopen this next week if we decide to continue with this approach.

juspence:
@RedHatProductSecurity/corgi-devs Please review, and sorry for the churn here. The final requirements are:

1. Tagging will not be used for CORGI-686 since an integration with SFM2 is needed, but will not be ready in time to meet a deadline.
2. Tagging will likely be used for other, future Compliance Engineering programs. Enabling this feature now avoids having to rush to meet a deadline.

juspence:
There are some changes to the API here, which I think are needed to make get_object() / get_queryset() work correctly in all cases. Mostly this just removes custom code in favor of Django filters, which is a small but breaking API change.

For example, before this change doing `/components?purl=value` for a nonexistent component would give a 404 status code and and a JSON dict like `{"detail": "Not found."}` as a response. After this change, the same request gives a 200 status code and a JSON dict like `{"count": 0, "results": []}` as a response.

This is consistent with other Django queryset filters / API parameters like `?re_purl=` and `?namespace=` - they already return an empty list when no results match.

juspence:
Due to the breaking API changes and Marek's feedback, we don't want to merge this now. We will wait until the absolute last possible minute given how much change is needed to enable this feature.

PR #612, opened 2023-09-15
CORGI-758: Handle Quarkus release errata

kgrant-rh:
Middleware errata don't have all their components attached, this change reads the purl of the root component of an already-processed SBOMer manifest to know which components were released.

### DIFF
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import typing
 from collections import defaultdict
@@ -261,3 +262,15 @@ class ErrataTool:
                 # map modular_rpms to build_ids. In slow_save_errata these are only saved as
                 # meta_attr so, let's just save the entire list for each build_id
                 variant_to_component_map[variant].append({build_id: modular_rpms})
+
+    def get_erratum_notes(self, erratum_id: int):
+        # Get the contents of the "Notes" (formerly "How to test") field from an erratum.
+        # Quarkus uses this field to associate an SBOM with an erratum.
+        erratum_json = self.get(f"api/v1/erratum/{erratum_id}?format=json")
+        try:
+            # All products that use SBOMer should have notes set
+            notes = json.loads(erratum_json["content"]["content"]["how_to_test"])
+        except json.JSONDecodeError:
+            logger.warning(f"Couldn't load Notes for erratum {erratum_id}")
+            return {}
+        return notes
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -263,14 +263,16 @@ class ErrataTool:
                 # meta_attr so, let's just save the entire list for each build_id
                 variant_to_component_map[variant].append({build_id: modular_rpms})
 
-    def get_erratum_notes(self, erratum_id: int):
+    def get_erratum_notes(self, erratum_id: int) -> dict:
         # Get the contents of the "Notes" (formerly "How to test") field from an erratum.
         # Quarkus uses this field to associate an SBOM with an erratum.
         erratum_json = self.get(f"api/v1/erratum/{erratum_id}?format=json")
         try:
             # All products that use SBOMer should have notes set
+            # NB: Though the field in the ET UI is called "Notes", the field name in the API
+            # is "how_to_test", which was the original purpose of the Notes field.
             notes = json.loads(erratum_json["content"]["content"]["how_to_test"])
-        except json.JSONDecodeError:
+        except json.JSONDecodeError as error:
             logger.warning(f"Couldn't load Notes for erratum {erratum_id}")
-            return {}
+            raise error
         return notes
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Any
 
+from corgi.core.constants import SBOMER_PRODUCT_MAP
 from corgi.core.models import Component
 
 logger = logging.getLogger(__name__)
@@ -99,9 +100,4 @@ class SbomerSbom:
 def is_sbomer_product(product: str, product_release: str) -> bool:
     """Identifies products for which SBOMer produces manifests and which need separate handling
     for release errata"""
-    products = {
-        "RHBQ": [
-            "Red Hat build of Quarkus Middleware",
-        ],
-    }
-    return product in products and product_release in products[product]
+    return product in SBOMER_PRODUCT_MAP and product_release in SBOMER_PRODUCT_MAP[product]
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -94,3 +94,14 @@ class SbomerSbom:
         # declared in the Components section above. At the moment,
         # only "dependsOn" type relationships are declared.
         self.dependencies = {d["ref"]: d["dependsOn"] for d in data["dependencies"]}
+
+
+def is_sbomer_product(product: str, product_release: str) -> bool:
+    """Identifies products for which SBOMer produces manifests and which need separate handling
+    for release errata"""
+    products = {
+        "RHBQ": [
+            "Red Hat build of Quarkus Middleware",
+        ],
+    }
+    return product in products and product_release in products[product]
--- a/corgi/tasks/pnc.py
+++ b/corgi/tasks/pnc.py
@@ -178,6 +178,6 @@ def slow_handle_pnc_errata_released(erratum_id: int, erratum_status: str) -> Non
             build_type=build.build_type,
             defaults={
                 "type": ProductComponentRelation.Type.ERRATA,
-                "meta_attr": {"components": component},
+                "meta_attr": {"component_purl": component.purl},
             },
         )
--- a/corgi/tasks/pnc.py
+++ b/corgi/tasks/pnc.py
@@ -4,6 +4,7 @@ import requests
 from celery_singleton import Singleton
 
 from config.celery import app
+from corgi.collectors.errata_tool import ErrataTool
 from corgi.collectors.models import CollectorErrataProductVariant
 from corgi.collectors.pnc import SbomerSbom
 from corgi.core.models import (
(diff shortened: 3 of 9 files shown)


kgrant-rh:

tests/data/pnc/erratum_notes.json
```diff
@@ -0,0 +1,15 @@
+{
+  "manifest" : {
+    "refs" : [ {
+      "type" : "purl",
+      "uri" : "pkg:maven/com.redhat.quarkus.platform/quarkus-bom@2.13.8.SP2-redhat-00001?type=pom"
```
juspence, line 5: This is mostly the same as the Maven purls we generate in Corgi. FYI the major differences I see are:

1. No `?repository_url=` qualifier in this purl. Corgi sets this to maven.repository.redhat.com - the exact URL is in corgi/core/constants.py I think.

2. I think Corgi doesn't set the `?type=` qualifier since we don't normally have that info. We could save the value here into the newly-created Component's meta_attr. Then the purl-building logic in corgi/core/models.py could add this value to the purl if it exists in the meta_attr.

corgi/collectors/errata_tool.py
```diff
@@ -261,3 +262,15 @@ def _parse_module(
...
+    def get_erratum_notes(self, erratum_id: int):
+        # Get the contents of the "Notes" (formerly "How to test") field from an erratum.
+        # Quarkus uses this field to associate an SBOM with an erratum.
+        erratum_json = self.get(f"api/v1/erratum/{erratum_id}?format=json")
+        try:
+            # All products that use SBOMer should have notes set
+            notes = json.loads(erratum_json["content"]["content"]["how_to_test"])
+        except json.JSONDecodeError:
+            logger.warning(f"Couldn't load Notes for erratum {erratum_id}")
+            return {}
```
juspence: Can we just raise the exception here if `json.loads()` fails? We've seen issues in other tasks get hidden because we only logged a warning and skipped loading that data. That caused a lot of missing data we had to go back and reprocess after fixing the bug.

corgi/collectors/pnc.py
```diff
@@ -94,3 +94,14 @@ def __init__(self, data: dict[str, Any]):
...
         self.dependencies = {d["ref"]: d["dependsOn"] for d in data["dependencies"]}
+
+
+def is_sbomer_product(product: str, product_release: str) -> bool:
+    """Identifies products for which SBOMer produces manifests and which need separate handling
+    for release errata"""
+    products = {
+        "RHBQ": [
+            "Red Hat build of Quarkus Middleware",
+        ],
```
juspence: The logic makes sense, but should the actual list live in corgi/core/constants.py? I'm thinking of a name like SBOMER_PRODUCT_MAPPING or something.

corgi/collectors/errata_tool.py
```diff
@@ -261,3 +262,15 @@ def _parse_module(
...
                 # meta_attr so, let's just save the entire list for each build_id
                 variant_to_component_map[variant].append({build_id: modular_rpms})
+
+    def get_erratum_notes(self, erratum_id: int):
+        # Get the contents of the "Notes" (formerly "How to test") field from an erratum.
+        # Quarkus uses this field to associate an SBOM with an erratum.
+        erratum_json = self.get(f"api/v1/erratum/{erratum_id}?format=json")
+        try:
+            # All products that use SBOMer should have notes set
+            notes = json.loads(erratum_json["content"]["content"]["how_to_test"])
```
juspence, line 274: Is the field name still "how_to_test", or should it be "notes" now?

corgi/collectors/errata_tool.py
```diff
@@ -261,3 +262,15 @@ def _parse_module(
                 # map modular_rpms to build_ids. In slow_save_errata these are only saved as
                 # meta_attr so, let's just save the entire list for each build_id
                 variant_to_component_map[variant].append({build_id: modular_rpms})
+
+    def get_erratum_notes(self, erratum_id: int):
```
```suggestion
    def get_erratum_notes(self, erratum_id: int) -> dict:
```

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+    # Check that the erratum is released
+    if erratum_status != "SHIPPED_LIVE":
+        raise ValueError(f"Invalid status {erratum_status} for erratum {erratum_id}")
+
+    # Get the purl(s) of the related component(s) from the erratum's Notes field
+    et = ErrataTool()
+    notes = et.get_erratum_notes(erratum_id)
+
+    related_purls = set()
+    for ref in notes.get("manifest", {}).get("refs", {}):
```
juspence: The last bit looks a bit off, but I might be confused. You have `.get("refs", {})` which implies refs is a dict. Then you iterate over each ref in the refs key, but iterating over a dict only gives you the key names.

So I don't think the check below for `if ref["type"] == "purl"` is going to work. Maybe refs is a dict with string keys and nested dicts as values? In that case, below should work:
```suggestion
    for ref in notes.get("manifest", {}).get("refs", {}).values():
```
Or maybe the default for `.get()` should be an empty tuple / list instead of an empty dict, to match the real data.

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+    if not related_purls:
+        raise ValueError(f"Erratum {erratum_id} had no associated purls")
+
+    # Check that there's a component matching the purls
+    root_components: set[Component] = set()
+    for purl in related_purls:
+        # There should only be one root component of SOURCE type
+        components = Component.objects.filter(meta_attr__purl_declared=purl).filter(
+            type=Component.Type.MAVEN
+        )
```
juspence: I'm not sure it matters, but putting the `type=MAVEN` bit first might help optimize the generated SQL by checking meta_attr, a very slow JSON field, on a smaller number of components.

EDIT: Actually I think it shouldn't matter, and you can combine the two .filter() calls into one, like `.filter(type=Component.Type.MAVEN, meta_attr__purl_declared=purl)`.

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+        if components.count() == 0:
+            logger.warning(
+                f"Erratum {erratum_id} refers to purl {purl} which matches no components"
+            )
+            continue
+        if components.count() > 1:
+            logger.warning(
+                f"Erratum {erratum_id} refers to purl {purl} which matches {components.count()}"
```
juspence: I think this will make three DB calls, so you could store the result of .count() to speed this up just a little bit:
```suggestion
        components_count = components.count()
        if components_count == 0:
            logger.warning(
                f"Erratum {erratum_id} refers to purl {purl} which matches no components"
            )
            continue
        elif components_count > 1:
            logger.warning(
                f"Erratum {erratum_id} refers to purl {purl} which matches {components_count}"
```
But I'm also not sure it's necessary. We have a unique constraint on the regular purl field. Even though you're using the meta_attr__purl_declared field, I'd be very surprised if we had two Maven components with different regular purls, but the same purl_declared.

One option might be to fix up our purl-generating logic for Maven components, so that we can raise an exception (during ingestion) if Corgi's purl isn't the same as SBOMer's purl_declared. Then the purl_declared will definitely be unique, and this check won't be needed.

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            )
+        root_components.add(components[0])
+
+    # Update relations with this erratum
+    # TODO: Is just the root component okay here, or do I need to recreate the entire tree?
+    for component in root_components:
+        if component.software_build is None:
+            logger.warning(
+                f"Component {component.purl} has no build, can't relate erratum {erratum_id}"
+            )
```
juspence: I think it's better to raise an error here. Alternatively, you can do:
```suggestion
            logger.warning(
                f"Component {component.purl} has no build, can't relate erratum {erratum_id}"
            )
            continue
```
And then the code below doesn't need to be wrapped in an else block or indented.

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            build = component.software_build
+            build_relation = ProductComponentRelation.objects.get(software_build=build)
+            ProductComponentRelation.objects.update_or_create(
+                external_system_id=erratum_id,
+                product_ref=build_relation.product_ref,
+                build_id=build.build_id,
+                build_type=build.build_type,
+                defaults={
+                    "type": ProductComponentRelation.Type.ERRATA,
+                    "meta_attr": {"components": component},
```
juspence: This won't append the component to the components field if it already exists - it will just get overwritten. So even if one build has many components, the relation's meta_attr will only list the last component we processed. That's fine if we expect only one (root) component to be processed for each build.

Maybe we should also store only the component.pk or purl values, instead of the whole component object.

juspence (approved):
I need to run out for some errands in just a bit, so didn't have time to look at tests. But overall this LGTM.

corgi/tasks/pnc.py
```diff
@@ -123,3 +124,62 @@ def slow_fetch_pnc_sbom(purl: str, product_data, build_data, sbom_data) -> None:
...
+            continue
+        if components.count() > 1:
+            logger.warning(
+                f"Erratum {erratum_id} refers to purl {purl} which matches {components.count()}"
+                " components; only handling the first"
+            )
+        root_components.add(components[0])
+
+    # Update relations with this erratum
+    # TODO: Is just the root component okay here, or do I need to recreate the entire tree?
```
juspence: We only need to create one relation for each build / erratum / variant combination. So only processing the root component is correct, assuming there's exactly one root component per manifest.

But I also see we're building a set of root_components in this manifest. That doesn't really fit the idea of a single "root" in the tree. Is that just for safety (handle duplicates reported in more than one place), and we ultimately have exactly one root component in each manifest?

tests/test_monitor.py
```diff
@@ -249,18 +249,33 @@ def test_umb_receiver_handles_shipped_errata():
...
+            "MOCK_ID", mock_id
+        )
     )
     # Messages with an invalid status should get filtered out by the topic selector
     mock_invalid_event.message.body = (
-        '{"errata_status": "DROPPED_NO_SHIP", "errata_id": MOCK_ID}'.replace("MOCK_ID", mock_id)
+        '{"errata_status": "DROPPED_NO_SHIP", "errata_id": MOCK_ID,'
+        ' "product": "Product", "release": "Red Hat release of Product"}'.replace(
+            "MOCK_ID", mock_id
+        )
```
juspence: If you make this into a multiline string (inside a tuple), all the different lines will get automatically concatenated together. But the call to `.replace()` is on the second line, while the value to be replaced is on the first line. Is that going to work OK?

If not, you could either make this a multi-line string (outside a tuple) using """ triple quotes, or you could just move the `.replace()` call up one line.

tests/test_monitor.py
```diff
@@ -249,18 +249,33 @@ def test_umb_receiver_handles_shipped_errata():
...
+    )
+    # Messages for Middleware/SBOMer products should be handled by separately
+    mock_sbomer_event.message.body = (
+        '{"errata_status": "SHIPPED_LIVE", "errata_id": MOCK_ID,'
+        ' "product": "RHBQ", "release": "Red Hat build of Quarkus Middleware"}'.replace(
+            "MOCK_ID", mock_id
+        )
```
juspence: As above.

tests/test_monitor.py
```diff
@@ -249,18 +249,33 @@ def test_umb_receiver_handles_shipped_errata():
...
+    mock_sbomer_event.message.address = address
     assert address.replace("topic://", VIRTUAL_TOPIC_ADDRESS_PREFIX) in addresses
 
-    mock_umb_event.message.body = '{"errata_status": "SHIPPED_LIVE", "errata_id": MOCK_ID}'.replace(
-        "MOCK_ID", mock_id
+    mock_umb_event.message.body = (
+        '{"errata_status": "SHIPPED_LIVE", "errata_id": MOCK_ID,'
+        ' "product": "Product", "release": "Red Hat release of Product"}'.replace(
+            "MOCK_ID", mock_id
+        )
```
juspence: As above.

tests/test_monitor.py
```diff
@@ -300,6 +315,31 @@ def test_umb_receiver_handles_shipped_errata():
...
+            handler.on_message(mock_sbomer_event)
+            mock_accept.assert_called_once_with(mock_sbomer_event.delivery)
+            mock_accept.reset_mock()
+
+            # Make sure invalid statuses work in the SBOMer erratum handler as well
+            mock_sbomer_event.message.body = mock_sbomer_event.message.body.replace(
+                "SHIPPED_LIVE", "DROPPED_NO_SHIP"
+            )
+            handler.on_message(mock_sbomer_event)
+            mock_accept.assert_called_once_with(mock_sbomer_event.delivery)
```
juspence, line 335: Should we also test that `handler.release()` gets called when `on_message()` raises an exception?

tests/data/pnc/erratum_notes.json
kgrant-rh: The PNC collector saves the purls listed in the SBOMer manifest in `purl_declared`, so we can get any arguments from there.

corgi/collectors/errata_tool.py
kgrant-rh: In the Errata Tool web UI, it's "Notes", but in the json data returned by the API, it's still "how_to_test". I'd talk to the ET team about changing the name of the field in the API, but I'd bet there are other tools out there that would break if they made such a change. Our approach here is (famous last words) temporary, for that reason also I'm not sure it's worth pursuing a change, but I'll add a comment to clarify.

corgi/tasks/pnc.py
kgrant-rh: Good catch, the default for refs should have been a list, not a dict.

corgi/tasks/pnc.py
kgrant-rh: I have CORGI-789 planned for this sprint, and I plan to address potential discrepancies between the SBOMer purl and Corgi's generated purls as part of that. When I make that change, I 'll also remove this check.

corgi/tasks/pnc.py
kgrant-rh: There should only be one root component per manifest, but there may be multiple root components in a single erratum which fixes multiple products.

corgi/tasks/pnc.py
kgrant-rh: Good call. I think this is okay, because there should only be one root component per build. I'll make sure to validate that in the next middleware meeting. Regarding the meta, I changed it to just store the purl. I was storing the component only because I was cribbing from the ET collector that creates these relations when it processes errata for Brew builds, but now that I look more closely, I realize that those attributes aren't necessary for manifest generation.

PR #613, opened 2023-09-18
CORGI-666 update released_components filter to use the relations table

jasinner:
Switch from using brew tags stored in SoftwareBuild meta_attr to using 'relations' reverse foreign key of SoftwareBuild, with a check to make sure they are of Errata type.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -6,13 +6,19 @@ from django.contrib.auth import get_user_model
 from rest_framework.authtoken.models import Token
 
 from corgi.collectors.appstream_lifecycle import AppStreamLifeCycleCollector
-from corgi.core.models import Component, ComponentNode, SoftwareBuild
+from corgi.core.models import (
+    Component,
+    ComponentNode,
+    ProductComponentRelation,
+    SoftwareBuild,
+)
 
 from .factories import (
     ChannelFactory,
     ComponentFactory,
     ComponentTagFactory,
     LifeCycleFactory,
+    ProductComponentRelationFactory,
     ProductFactory,
     ProductStreamFactory,
     ProductVariantFactory,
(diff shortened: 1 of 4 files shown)


jasinner:

tests/test_api.py
```diff
@@ -439,8 +444,11 @@ def test_latest_components_by_streams_filter(client, api_path):
 
 @pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
 def test_released_components_filter(client, api_path):
-    released_build = SoftwareBuildFactory(meta_attr={"released_errata_tags": ["RHBA-2023:1234"]})
-    unreleased_build = SoftwareBuildFactory(meta_attr={"released_errata_tags": []})
+    released_build = SoftwareBuildFactory()
+    unreleased_build = SoftwareBuildFactory()
+    ProductComponentRelation.objects.create(
+        type=ProductComponentRelation.Type.ERRATA, software_build=released_build
+    )
```
juspence: I'm not sure it matters, but I think there's a ProductComponentRelationFactory you can use here. That would avoid needing a new import and might make our test a little more robust (by randomly generating values for product_ref, external_system_id, and so on).

tests/test_manifests.py
```diff
@@ -503,13 +503,12 @@ def setup_products_and_components_upstreams():
...
+    build = SoftwareBuildFactory(build_id=1)
     if released:
-        meta_attr["released_errata_tags"] = ["RHBA-2023:1234"]
-    build = SoftwareBuildFactory(
-        build_id=1,
-        meta_attr=meta_attr,
-    )
+        ProductComponentRelation.objects.create(
+            type=ProductComponentRelation.Type.ERRATA, software_build=build
+        )
```
juspence: As before, it might be better to use the test factories here.

tests/test_model.py
```diff
@@ -1148,6 +1148,24 @@ def test_latest_filter():
     assert not ps.components.latest_components()
 
 
+@pytest.mark.django_db
+def test_released_filter():
+    sb = SoftwareBuildFactory()
+    c = ComponentFactory(software_build=sb)
+    # Make sure that a relation not of type Errata does not show in filter
+    ProductComponentRelation.objects.create(
```
juspence: This is fine as-is, but if you want to be really thorough, we could also check that a component with no relations doesn't end up in `released_components()`. We could also test the include=False case, and make sure the filter still works when there's more than one component:
```suggestion
    c = ComponentFactory(software_build=sb)
    c2 = ComponentFactory(software_build=sb)
    # Make sure that components with no relations don't show in filter
    assert c not in Component.objects.get_queryset().released_components()
    assert c in Component.objects.get_queryset().released_components(include=False)

    # Make sure that a relation not of type Errata does not show in filter
    ProductComponentRelation.objects.create(
```

juspence (approved):
LGTM - comments are minor so please feel free to ignore. And hope you're feeling better soonish.

tests/test_api.py
jasinner: Ok, I'll switch to the factory.

PR #616, opened 2023-09-22
add distinct to manifest_components where to prevent duplicates from …

jasinner:
…released_components filter

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1045,7 +1045,7 @@ class ComponentQuerySet(models.QuerySet):
         if quick:
             return roots
         else:
-            return roots.latest_components()
+            return roots.latest_components().distinct()
 
     def srpms(self, include: bool = True) -> models.QuerySet["Component"]:
         """Show only source RPMs by default, or only non-SRPMs if include=False"""
--- a/tests/test_manifests.py
+++ b/tests/test_manifests.py
@@ -249,6 +249,23 @@ def test_product_manifest_excludes_internal_components():
         assert "redhat.com/" not in package["name"]
 
 
+def test_manifest_no_duplicate_released_components():
+    component, stream, _, _ = setup_products_and_components_provides()
+    # Add another Errata relation type for the same build (one already created in
+    # setup_product_and_components_provides)
+    ProductComponentRelationFactory(
+        software_build=component.software_build, type=ProductComponentRelation.Type.ERRATA
+    )
+
+    unique_components = set()
+
+    for c in stream.components.manifest_components():
+        if c.purl not in unique_components:
+            unique_components.add(c.purl)
+        else:
+            assert False
+
+
 def test_manifest_cpes_from_variants():
     # setup_products_and_components_provides adds a variant with a cpe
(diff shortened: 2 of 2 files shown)


jasinner:
@RedHatProductSecurity/corgi-devs please review. I noticed duplicates with the new released_components filter. Caused by having multiple relations of type ERRATA for a single build.

tests/test_manifests.py
```diff
@@ -249,6 +249,23 @@ def test_product_manifest_excludes_internal_components():
...
+        software_build=component.software_build, type=ProductComponentRelation.Type.ERRATA
+    )
+
+    unique_components = set()
+
+    for c in stream.components.manifest_components():
+        if c.purl not in unique_components:
+            unique_components.add(c.purl)
+        else:
+            assert False
```
juspence, line 266: Minor nit, but it's a little simpler to understand if you just do:
```suggestion
    for purl in stream.components.manifest_components().values_list("purl", flat=True):
        assert purl not in unique_components
        unique_components.add(purl)
```

corgi/core/models.py
```diff
@@ -1045,7 +1045,7 @@ def manifest_components(self, quick=False) -> "ComponentQuerySet":
         if quick:
             return roots
         else:
-            return roots.latest_components()
+            return roots.latest_components().distinct()
```
juspence, line 1048: @jasinner This change is small enough, but all of a sudden the performance test started failing this morning. It didn't fail over the weekend or at all last week (as far as I can tell), so I think this MR was the cause.

I have no idea why, but now we're only seeing 1 component instead of 9,000 in `display_manifest_with_many_components()`. So it looks like we lost data in certain streams and the manifest is now incomplete.

juspence: Actually, maybe the performance test failure was also caused by this other PR that was merged earlier today, before the daily test run:

https://github.com/RedHatProductSecurity/component-registry/pull/604

I also see some test failures due to:
A requred env var that's missing in either CI / Tox settings or ops-repo / Django pod environment variables or both
A missing migration
SonarQube / linting failures
A manifest test failure due to incorrect components

Should we back out this change? It seems like the logic might be incorrect, if it caused certain manifests to report 0 components in both tests (sample data) and stage (real data).

corgi/core/models.py
```diff
@@ -1045,7 +1045,7 @@ def manifest_components(self, quick=False) -> "ComponentQuerySet":
         if quick:
             return roots
```
juspence, line 1046: Maybe we're missing a .distinct() on line 1046 as well, or you could combine both as part of line 1044. But I think this probably isn't the right place to add this logic. We should add `.distinct()` to the released_components queryset on line 1013 that has this bug, not to the later usages of that queryset.

Ultimately the manifest_components() logic here is actually something like:
```
roots = stream.components.root_components()  # Start with all components in the stream, find only the root components

released_roots = roots.released_components()  # Find all the released versions of those roots

return released_roots.latest_components()  # Report only the latest released version of a root - ignore newer versions that aren't released yet
```
But we need to fix this bug for all usages of the released_components filter, not just the manifest_components logic / queryset.

PR #617, opened 2023-09-25
CORGI-817 / CORGI-791 SBOM filter

jasinner:
This adds a no_manifest tag to product_streams which we want to exclude from manifests published on the customer portal. We'll have to update SDEngine to make use of this new product stream tag. 

```
api/v1/product_streams?tags=!no_manifest
```

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -35,13 +35,19 @@ class TagFilter(Filter):
             return queryset
         search_tags = value.split(",")
         for tag in search_tags:
+            exclude = False
+            if tag.startswith("!"):
+                tag = tag[1:]
+                exclude = True
             if ":" in tag:
                 tag_name, _, tag_value = tag.partition(":")
-                queryset = queryset.filter(
-                    tags__name__icontains=tag_name, tags__value__icontains=tag_value
-                )
+                queryset_kwargs = {"tags__name": tag_name, "tags__value": tag_value}
             else:
-                queryset = queryset.filter(tags__name__icontains=tag)
+                queryset_kwargs = {"tags__name": tag}
+            if exclude:
+                queryset = queryset.exclude(**queryset_kwargs)
+            else:
+                queryset = queryset.filter(**queryset_kwargs)
         return queryset
 
 
--- a/corgi/tasks/tagging.py
+++ b/corgi/tasks/tagging.py
@@ -1,0 +1,54 @@
+from celery.utils.log import get_task_logger
+from celery_singleton import Singleton
+
+from config.celery import app
+from corgi.core.models import Product, ProductStream, ProductStreamTag, ProductVersion
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+
+logger = get_task_logger(__name__)
+
+
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+)
+def apply_stream_no_manifest_tags():
+    tag_name = "no_manifest"
+    tag_value = ""
+    apply_middleware_stream_no_manifest_tags(tag_name, tag_value)
+    apply_rhel_8_9_z_stream_no_manifest_tags(tag_name, tag_value)
+    apply_managed_services_no_manifest_tags(tag_name, tag_value)
+
+
+def apply_middleware_stream_no_manifest_tags(tag_name: str, tag_value: str) -> None:
+    for stream_pk in Product.objects.filter(meta_attr__business_unit="Core Middleware").values_list(
+        "productstreams", flat=True
+    ):
+        stream = ProductStream.objects.get(pk=stream_pk)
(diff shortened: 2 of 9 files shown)


jasinner:

corgi/api/filters.py
```diff
@@ -37,11 +37,17 @@ def filter(self, queryset: QuerySet, value: str) -> QuerySet:
...
+                    queryset = queryset.exclude(tags__name=tag_name, tags__value=tag_value)
+                else:
+                    queryset = queryset.filter(tags__name=tag_name, tags__value=tag_value)
             else:
-                queryset = queryset.filter(tags__name__icontains=tag)
+                if tag.startswith("!"):
+                    tag = tag[1:]
+                    queryset = queryset.exclude(tags__name=tag)
+                else:
+                    queryset = queryset.filter(tags__name=tag)
```
juspence: We went from doing a case-insensitive substring match (using explicit __icontains) to doing a case-sensitive exact match (using implicit __exact / no explicit lookup given). LGTM if that's expected.

Minor suggestion: below can be simplified a bit so we're not repeating code four different ways:
```suggestion
            exclude = False
            if tag.startswith("!"):
                tag = tag[1:]
                exclude = True
            if ":" in tag:
                tag_name, _, tag_value = tag.partition(":")
                queryset_kwargs = {"tags__name": tag_name, "tags__value": tag_value}
            else:
                queryset_kwargs = {"tags__name": tag}
            if exclude:
                queryset = queryset.exclude(**queryset_kwargs)
            else:
                queryset = queryset.filter(**queryset_kwargs)
```

corgi/tasks/tagging.py
```diff
@@ -0,0 +1,58 @@
...
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+)
+def apply_tags():
+    apply_stream_no_manifest_tags()
+
+
+def apply_stream_no_manifest_tags():
```
juspence: I think we can just combine the two methods since the first is just a one-liner that calls the second:
```suggestion
def apply_stream_no_manifest_tags() -> None:
```
Should we also schedule this as a daily task, so that any new RHEL / middleware / managed services streams are automatically updated with this tag?

corgi/tasks/tagging.py
```diff
@@ -0,0 +1,58 @@
...
+    tag_name = "no_manifest"
+    tag_value = True
+    apply_middleware_stream_no_manifest_tags(tag_name, tag_value)
+    apply_rhel_8_9_z_stream_no_manifest_tags(tag_name, tag_value)
+    apply_managed_services_no_manifest_tags(tag_name, tag_value)
```
juspence: I guess we need to always pass a tag_value so that get_or_create uses an index and is atomic. But let's use an empty "" string default instead of a literal True / bool value.
```suggestion
    # Must pass tag_value to get_or_create so it's atomic
    tag_value = ""
    apply_middleware_stream_no_manifest_tags(tag_name, tag_value)
    apply_rhel_8_9_z_stream_no_manifest_tags(tag_name, tag_value)
    apply_managed_services_no_manifest_tags(tag_name, tag_value)
```

tests/test_api.py
```diff
@@ -682,6 +683,34 @@ def test_component_tags_duplicate(client, api_path):
...
+    assert ProductStream.objects.exclude(tags__name="manifest").count() == 1
+    response = client.get(f"{api_path}/product_streams?tags=!manifest")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
+
+    ProductStreamFactory(tag__name="manifest", tag__value="true")
+    assert ProductStream.objects.exclude(tags__name="manifest").count() == 1
+    response = client.get(f"{api_path}/product_streams?tags=!manifest")
+    assert response.status_code == 200
+    assert response.json()["count"] == 1
```
juspence: I think it's better to assert that `response[results][0]` is the stream we expect, instead of just checking the count:
```suggestion
    response = response.json()
    assert response["count"] == 1
    assert response["results"][0]["name"] == the_first_stream.name
```

juspence (approved):

corgi/tasks/tagging.py
jasinner: Ok, I set this up this way as I suspect we will have other tagging tasks to perform in future, but we can always refactor it back out when/if adding those. I also added  a daily task for this job, thank you!

corgi/tasks/tagging.py
jasinner: OK,  updated to empty string.

tests/test_api.py
jasinner: Good idea, thanks.

corgi/api/filters.py
jasinner: ok, updated. Yes, I meant to remove the case-insensitive sub string match. Since we will soon have a way to enumerating tags, we should be able to discover tags that way.

PR #618, opened 2023-09-25
Always run performance tests

juspence:
@RedHatProductSecurity/corgi-devs Please review. IMO it's much better to run performance tests on every MR instead of only once a day. Right now we have a failing pipeline on main, but it's hard to tell which PR introduced the regression, because we aren't running the tests often enough.


.gitlab-ci.yml
```diff
@@ -139,9 +139,6 @@ test-performance:
     - $DNF_WITH_OPTIONS install $RPM_REQUIREMENTS
     - python3.9 -m pip install tox
     - tox -e corgi -- -m performance --no-cov
-  only:
```
jasinner, line 142: I don't agree with this. I fixed the performance test in stage by regenerating the rhel-9.2.0 manifest. I guess the issue might have been caused by changes James was testing for the the latest_components stored procedure.

The reason we removed it was because it's testing changes which aren't from the current branch under test. It's annoying to have it fail when the changes are not related to what you are trying to merge.

jasinner (commented):
The other changes here apart from the re-enabling of the performance test on every merge are good, thank you.

.gitlab-ci.yml
juspence: Thanks for getting the test to pass again. I know it's annoying to see pipeline failures for stuff that's not related to your PR, but we'll still have failures either way.

The difference is whether we see failures the next day, and have to spend time debugging which PR / change caused issues, or whether we see the failure immediately and know "the previous PR before this one made a breaking change". Getting feedback about failures as soon as possible is much more useful IMO...or in other words, debugging this yesterday was a pain.

Are you sure you really can't tolerate having this be part of the pipeline? I don't think the performance tests fail that often, and never for no reason. All the failures i've seen have been rare issues with how data is linked together, or some breaking change we made to our API (and forgot to update tests).

jasinner: For me, I visit the pipelines page of Gitlab often enough that I think I would notice if the scheduled task starting failing, and I would go and investigate that. But if you think that isn't good enough, I guess we can re-enable the performance test in CI.

juspence: Yes, I'm watching it too, and I get emails when the scheduled job fails on main (which is why I filed CORGI-834 - thanks for picking that up).

But when there are multiple PRs merged in one day, we have to look at each to figure out which one caused the test failure. So having the performance test run on each PR will tell us "the PR before this one introduced a breaking change".

PR #620, opened 2023-09-26
CORGI-832 use existing umb certs for pyxis

jasinner:
Tested successfully on stage using Pyxis stage URL.


jasinner:

See also component-registry-ops/-/merge_requests/192

juspence (approved):
LGTM, thanks for fixing this!

corgi/collectors/pyxis.py
```diff
@@ -103,6 +104,13 @@
 """
 
 
+def check_client_cert(cert: tuple[Optional[str], Optional[str]]) -> TypeGuard[tuple[str, str]]:
+    """Convert the type Optional[str] returned by Django conf.settings to be a non-optional str"""
+    if not all(cert):
+        raise ValueError("Set CORGI_PYXIS_CERT and CORGI_PYXIS_KEY to get manifests from pyxis")
+    return True
+
+
```
juspence, line 113: This is a little complex...to fix mypy errors like this in the future, it should be enough to do below:
```
    if not settings.PYXIS_CERT or not settings.PYXIS_KEY:
        raise ValueError("Set CORGI_PYXIS_CERT and CORGI_PYXIS_KEY to get manifests from pyxis")
    # Both variables are now guaranteed to be non-null / non-empty strings
    # So type(cert) is guaranteed to be tuple[str, str] below
    cert = (settings.PYXIS_CERt, settings.PYXIS_KEY
```

juspence: Actually this seems to be causing issues on stage, I think becase typing_extensions is a test dependency for mypy type checking. So it exists in dev dependencies but not stage / prod (base.txt) dependencies.

PR #621, opened 2023-09-26
Allow SBOMer components with no license info

kgrant-rh:
Some components in SBOMer manifests may not have license information. This change prevents manifest processing from failing in those cases.

### DIFF
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -56,7 +56,7 @@ class SbomerSbom:
 
             # License info
             licenses = []
-            for _license in component.get("licenses", []):
+            for _license in component.get("licenses", ()):
                 if _license["license"].get("id"):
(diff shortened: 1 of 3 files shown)


corgi/collectors/pnc.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, data: dict[str, Any]):
 
             # License info
             licenses = []
-            for _license in component["licenses"]:
+            for _license in component.get("licenses", []):
```
juspence: LGTM! One super minor comment is that you can use a `()` tuple here to slightly improve speed and memory usage. An empty list is a little slower since it's still mutable, so the Python interpreter doesn't know if you'll append to it later. An empty tuple literal will never change, so the interpreter can optimize it a little more.

juspence (approved):

jasinner (approved):

PR #622, opened 2023-09-27
CORGI-834 pyxis fixups

jasinner:

### DIFF
--- a/corgi/tasks/pyxis.py
+++ b/corgi/tasks/pyxis.py
@@ -194,6 +194,11 @@ def save_component(component: dict, parent: ComponentNode) -> bool:
     mapping = {
         "rpm": Component.Type.RPM,
         "golang": Component.Type.GOLANG,
+        "npm": Component.Type.NPM,
+        "gem": Component.Type.GEM,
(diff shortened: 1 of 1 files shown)


corgi/tasks/common.py
```diff
@@ -23,6 +23,7 @@
 # But can also be due to database operation failures that should NOT be retried
 RETRYABLE_ERRORS = (
     DjangoInterfaceError,
+    HTTPError,
```
juspence: Sorry, I wasn't clear in the ticket. We want to retry the individual GET requests in the task, not the overall task itself. That's more efficient and also avoids issues like retrying POST / PUT / DELETE requests that aren't idempotent.

There are a few forms of retries you can use, but generally you'd add a for-loop or `@backoff` decorator around a particular request, and have it retry 5XX HTTP error codes only (server errors that a retry might be able to fix). We wouldn't want to retry e.g. 401 or 403 errors that won't go away on later requests.

jasinner: The Pyxis collector already has retry logic. I'll just remove this commit.

PR #623, opened 2023-09-28
CORGI-780: Handle private Quay images with no "latest" tag (image pull fails)

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have time (not urgent). I've confirmed this fixes the errors we're seeing when deployed to stage, and hopefully the pipeline won't fail overnight.


jasinner (approved):
LGTM

PR #624, opened 2023-09-29
CORGI-836 populate errata relation software_build fk

jasinner:
Since we changed the way the released_components filter works many streams are having their components filtered out by that filter. The cause of that seems to be be many missing ERRATA relations. 

I've got a script which looks for all streams which should have an SBOM, gets their root components and the release_errata_tags of those components software builds. It then calls slow_load_errata on those to create the ERRATA relations. However I noticed when testing this that the software_build foreign key of the ERRATA relation was not being populated. Most the time the SoftwareBuild object does exist, in that case we should populate the foreign key so that the new released_components filter will work correctly.

I filled a follow up task CORGI-837 to load missing SoftwareBuilds of all ERRATA relations as a scheduled task.

### DIFF
--- a/corgi/tasks/errata_tool.py
+++ b/corgi/tasks/errata_tool.py
@@ -1,3 +1,5 @@
+from typing import Any, DefaultDict
+
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
@@ -76,21 +78,7 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
         # Save PCR
         logger.info("Saving product components for errata %s", erratum_id)
         variant_to_component_map = et.get_erratum_components(str(erratum_id))
-        for variant_id, build_objects in variant_to_component_map.items():
-            for build_obj in build_objects:
-                for build_id, errata_components in build_obj.items():
-                    # Add to relations list as we go, so we can fetch them below
-                    build_ids.add(build_id)
-                    ProductComponentRelation.objects.update_or_create(
-                        external_system_id=erratum_id,
-                        product_ref=variant_id,
-                        build_id=build_id,
-                        build_type=build_type,
-                        defaults={
-                            "type": ProductComponentRelation.Type.ERRATA,
-                            "meta_attr": {"components": errata_components},
-                        },
-                    )
+        save_errata_relation(build_ids, build_type, erratum_id, variant_to_component_map)
 
     # If the number of relations was more than 0 check if we've processed all the builds
     # in the errata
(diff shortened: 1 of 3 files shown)


jasinner:
With the old version of scripts/kinit_in_container I was getting an error like:

```
Unspecified GSS failure: Server HTTP/errata.<snip>.com@REDHAT.COM not found in Kerberos database
```

tests/test_errata_data.py
```diff
@@ -226,10 +227,14 @@ def test_save_product_component_for_errata(
...
+    sb = SoftwareBuildFactory(build_id="1636922", build_type=BUILD_TYPE)
     slow_load_errata(erratum_id)
-    pcr = ProductComponentRelation.objects.filter(external_system_id=erratum_id)
-    assert len(pcr) == no_of_objs
+    pcrs = ProductComponentRelation.objects.filter(external_system_id=erratum_id)
+    assert len(pcrs) == no_of_objs
     assert mock_send.call_count == no_of_objs
+    for pcr in pcrs:
+        if pcr.build_id == "1636922":
+            assert pcr.software_build_id == sb.pk
```
juspence: Maybe below is a little cleaner:
```suggestion
        # If the relation uses this build's ID
        if pcr.build_id == sb.build_id:
            # assert it is linked to the build using the ForeignKey field
            assert pcr.software_build_id == sb.pk
        else:
            # else assert the ForeignKey is unset / other build IDs have not been fetched
            assert pcr.software_build_id is None
```

corgi/tasks/errata_tool.py
```diff
@@ -126,6 +114,36 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
+                    external_system_id=erratum_id,
+                    product_ref=variant_id,
+                    build_id=build_id,
+                    build_type=build_type,
+                    software_build=software_build,
+                    defaults={
+                        "type": ProductComponentRelation.Type.ERRATA,
+                        "meta_attr": {"components": errata_components},
+                    },
+                )
```
juspence: The other fields / keyword arguments here appear in a unique_together constraint, so finding a PCR matching them is OK (there's an index for this and update_or_create will be atomic).

I think having "software_build" as a keyword argument will cause issues though. If a relation exists where all the other fields match but software_build is unset, we'll fail to look it up / update it, and instead try to create a new PCR. Since all the other fields (which have a unique_together constraint) already exist in a different PCR, creating the new PCR with the same values will also fail.

I think the software_build field needs to be in the defaults dict instead:
```suggestion
                defaults = {
                    "type": ProductComponentRelation.Type.ERRATA,
                    "meta_attr": {"components": errata_components},
                }
                software_build = SoftwareBuild.objects.filter(
                    build_id=build_id, build_type=build_type
                ).first()
                if software_build:
                        # If software_build is None, this code will not add a "software_build": None
                        # key / value pair to the defaults dict, so we won't remove the linked
                        # software_build from any existing relations by mistake
                        # If software_build is None / doesn't exist, there shouldn't be any other
                        # relations with a linked build, right?
                        # Or if there are, maybe removing the linked build is what we want
                        defaults["software_build"] = software_build,
                ProductComponentRelation.objects.update_or_create(
                    external_system_id=erratum_id,
                    product_ref=variant_id,
                    build_id=build_id,
                    build_type=build_type,
                    defaults=defaults,
                )
```

juspence (approved):
I think there's a small bug with the way we look up the relations which have / don't have SoftwareBuilds, but otherwise LGTM.

We could also add one more test for this case: create an existing relation with no linked SoftwareBuild / null ForeignKey, but the same product_ref, external_system_id, build_type, build_id, and relation_type, before calling slow_load_errata. Then show that update_or_create correctly finds this relation and updates the ForeignKey field / links the SoftwareBuild.

I think right now, update_or_create will try and fail to look up a matching PCR to update / set the SoftwareBuild on, then try and fail to create a new relation with the same 5 fields and a linked / non-null SoftwareBuild.

corgi/tasks/errata_tool.py
jasinner: good idea, updated.

PR #625, opened 2023-10-03
CORGI-780: Fix DoesNotExist errors when manifesting services

juspence:
@RedHatProductSecurity/corgi-devs Please review. This definitely fixes the Component.DoesNotExist errors we currently see in our monitoring email (confirmed by deploying to stage), and should fix the SoftwareBuild.DoesNotExist errors as well (added this commit after I had already deployed).

We still see failing tasks / can't manifest services due to other errors, which these now-fixed errors were hiding. I'll continue working on the other errors in separate PRs, but this one should be complete enough to merge.

I also have a test written for some of the new logic to parse and save component metadata that Syft discovers. I still need to write tests for the overall mangen logic. Please let me know if we should discuss our use of purls / the information we save from a Syft scan, or if I've missed anything here.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1010,9 +1010,12 @@ class ComponentQuerySet(models.QuerySet):
         errata_relations = Q(software_build__relations__type=ProductComponentRelation.Type.ERRATA)
         if include:
             # Truthy values return only released components
-            return self.filter(errata_relations)
-        # Falsey values return only unreleased components
-        return self.exclude(errata_relations)
+            queryset = self.filter(errata_relations)
+        else:
+            # Falsey values return only unreleased components
+            queryset = self.exclude(errata_relations)
+        # Clear ordering and apply distinct() to avoid duplicates from above filter
+        return queryset.order_by().distinct()
 
     def root_components(self, include: bool = True) -> "ComponentQuerySet":
         """Show only root components by default, or only non-root components if include=False"""
@@ -1038,14 +1041,20 @@ class ComponentQuerySet(models.QuerySet):
         non_container_source_components = self.exclude(name__endswith="-container-source").using(
             "read_only"
         )
-        if settings.COMMUNITY_MODE_ENABLED:
-            roots = non_container_source_components.root_components()
-        else:
-            roots = non_container_source_components.root_components().released_components()
-        if quick:
-            return roots
-        else:
-            return roots.latest_components().order_by("pk").distinct("pk")
+        roots = non_container_source_components.root_components()
+        if not settings.COMMUNITY_MODE_ENABLED:
+            # Only filter in enterprise Corgi, where we have ERRATA-type relations
+            roots = roots.released_components()
+
+        if not quick:
+            # Only filter when we're actually generating a manifest
+            # not when checking if there are components to manifest, since it's slow
+            roots = roots.latest_components()
+
+        # Order by UUID to give stable results in manifests
+        # Other querysets should not define an ordering
+        # or should clear it with .order_by() if they use .distinct()
+        return roots.order_by("pk")
 
     def srpms(self, include: bool = True) -> models.QuerySet["Component"]:
         """Show only source RPMs by default, or only non-SRPMs if include=False"""
--- a/corgi/tasks/sca.py
+++ b/corgi/tasks/sca.py
@@ -12,13 +12,12 @@ import requests
 from celery.utils.log import get_task_logger
 from celery_singleton import Singleton
 from django.conf import settings
-from packageurl import PackageURL
+from django.db.models import Q
 
 from config.celery import app
 from corgi.collectors.brew import Brew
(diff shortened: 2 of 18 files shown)


corgi/core/models.py
```diff
@@ -1010,9 +1010,12 @@ def released_components(self, include: bool = True) -> "ComponentQuerySet":
...
             # Truthy values return only released components
-            return self.filter(errata_relations)
-        # Falsey values return only unreleased components
-        return self.exclude(errata_relations)
+            queryset = self.filter(errata_relations)
+        else:
+            # Falsey values return only unreleased components
+            queryset = self.exclude(errata_relations)
+        # Clear ordering and apply distinct() to avoid duplicates from above filter
+        return queryset.order_by().distinct()
```
juspence, line 1018: @jasinner You mentioned in another PR that we had a bug with some duplicate results in the released_components queryset. I left a comment on that PR that we had only fixed the manifest_components queryset, but the PR was already merged.

I think other usages of released_components could still have dupes, so I've tried to fix all the released_components usages here. Please let me know if this doesn't look right or if I misunderstood where the dupes were coming from.

jasinner: What happens when you combine release_components with manifest_components? I guess DJango is clever enough to only deduplicate once all the filter is complete? If so, think it's a good idea to add it here, otherwise I think we should add it only on the 'final' filter.

corgi/tasks/sca.py
```diff
@@ -86,58 +108,37 @@ def save_component(
...
-        syft_purl = PackageURL(
-            type=component_type,
-            name=meta.get("name", ""),
-            version=version,
-            qualifiers=qualifiers,
-        ).to_string()
-
     created = False
     name = meta.pop("name", "")
+    # TODO: Do we need to add epoch to the unique constraint?
```
jasinner: Probably yes. As long as it's always default to the same value for components without an epoch. Should be done in a separate PR / Task of course.

corgi/tasks/sca.py
```diff
@@ -86,58 +108,37 @@ def save_component(
...
+    epoch = meta.pop("epoch", 0)
     version = meta.pop("version", "")
     release = meta.pop("release", "")
     arch = meta.pop("arch", "noarch")
 
     namespace = Brew.check_red_hat_namespace(component["type"], version)
-    if namespace == Component.Namespace.REDHAT:
```
jasinner, line 120: This is now handled by Keith's changes in https://github.com/RedHatProductSecurity/component-registry/pull/626 ?

jasinner (approved):

corgi/tasks/sca.py
juspence: No, that change is unrelated to this one. Here, we were trying to get_or_create a new component. If that fails due to an IntegrityError, that means:

1. We couldn't get an existing component, because the name used as a keyword arg for the get() didn't match / was different somehow. This usually happens for UPPERcase vs lowercase names, or some-name with a dash vs some_name with an underscore.
2. We couldn't create a new component, because the purl we generated for the create() matched an existing purl. This happens when GitHub or Python components have a name like I described in step 1. In that case, the generated purl will be the same even though the names are slightly different.

To handle this IntegrityError, we tried to look up the existing duplicate component using the purl that Syft generated in meta["purl"]. Syft's purls don't perfectly match ours, and sometimes there is no Syft purl at all. So there was a lot of extra logic here to massage Syft's purl into the right shape, or generate one of our own, then try to find the duplicate component with the same purl.

But this logic will never work for RPMs without much more work that I don't think is a good idea. Syft's RPM purls look nothing like Corgi's. See the example in the test I added. There's no easy or sane way to build a purl here from Syft's data that always matches how Corgi builds the purl, for every type of component. So the diff you're commenting on is just me removing this old, bad logic - it doesn't work well and is already too complicated.

Instead, we'll just look up the duplicate components using their NEVRA, which is also unique. So we don't need a lot of complicated logic to try and build some purl. We'll just pass the component's type, name, version, release, and arch to the find_duplicate_component() function, without any extra processing.

corgi/core/models.py
juspence: We're already combining the released_components queryset with the manifest_components queryset. The problem is that adding the .distinct() only in the manifest_components queryset, means that other usages of released_components will still have duplicates.

Even if there are no other usages currently, we still need to fix the bug in the released_components logic. Otherwise the bug will pop up again when some other code starts calling released_components() without also calling manifest_components().

The test you added is still passing, so I assume that means there are still no duplicates in the manifest. And another test is still passing, so the ordering in the manifest is still based on each component's primary key / UUID.

I also updated one test to show that the API gives the wrong results before my change, and then the test starts passing again after the change I made here to fix this bug.

corgi/tasks/sca.py
juspence: I filed CORGI-849 for this.

PR #626, opened 2023-10-03
Store correct group ID and type for SBOMer purls

kgrant-rh:
A simple fix, more substantial changes to Maven purl handling to follow in CORGI-840.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1237,18 +1237,13 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         return PackageURL(type=str(self.type).lower(), **purl_data)
 
     def _build_maven_purl(self) -> dict[str, Union[str, dict[str, str]]]:
-        # Prefer some declared qualifiers to inferred qualifiers
-        declared_qualifiers = {}
-        if "purl_declared" in self.meta_attr:
-            declared_qualifiers = PackageURL.from_string(self.meta_attr["purl_declared"]).qualifiers
-
         qualifiers = {}
 
         classifier = self.meta_attr.get("classifier")
         if classifier:
             qualifiers["classifier"] = classifier
 
-        extension = declared_qualifiers.get("type") or self.meta_attr.get("type")
+        extension = self.meta_attr.get("type")
         if extension:
             qualifiers["type"] = extension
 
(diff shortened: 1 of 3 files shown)


corgi/core/models.py
```diff
@@ -1237,13 +1237,18 @@ def get_purl(self) -> PackageURL:
...
+            declared_qualifiers = PackageURL.from_string(self.meta_attr["purl_declared"]).qualifiers
+
         qualifiers = {}
 
         classifier = self.meta_attr.get("classifier")
         if classifier:
             qualifiers["classifier"] = classifier
 
-        extension = self.meta_attr.get("type")
+        extension = declared_qualifiers.get("type") or self.meta_attr.get("type")
```
juspence: Instead of storing this info twice, and having to look up / check two different fields, is it better to just store "all the SBOMer / PNC qualifiers" as individual keys in the meta_attr? E.g. set `meta_attr["classifier"] = component["classifier"]`, "type", and any others I'm forgetting in the PNC collector. We're already doing this for the group_id, so I'd prefer a consistent approach.

Setting everything in the collector means this code wouldn't need to change, but the lookup for `self.meta_attr.get("type")` would still give the right results. We'd have the type from PNC, if we were given it / if we ingested the component from an SBOMer manifest. Or we'd have the type discovered by Syft, if this component was discovered in an SCA scan instead of an SBOMer manifest.

I think this should be safe, since Syft metadata won't overwrite existing metadata if present. We do get_or_create in the SCA task instead of update_or_create to avoid wiping out meta_attr from Brew / PNC.

corgi/core/models.py
```diff
@@ -1237,13 +1237,18 @@ def get_purl(self) -> PackageURL:
...
 
     def _build_maven_purl(self) -> dict[str, Union[str, dict[str, str]]]:
+        # Prefer some declared qualifiers to inferred qualifiers
+        declared_qualifiers = {}
+        if "purl_declared" in self.meta_attr:
+            declared_qualifiers = PackageURL.from_string(self.meta_attr["purl_declared"]).qualifiers
+
         qualifiers = {}
 
         classifier = self.meta_attr.get("classifier")
```
juspence: Do we need to make changes here? If we use the approach below to check both Corgi data and PNC data, should a classifier from an SBOMer manifest be preferred over a classfier discovered by Syft?

juspence (approved):

corgi/core/models.py
kgrant-rh: Just to clarify, this isn't a concern if we just store the qualifiers in meta_attr in the PNC collector, right?

juspence: Yeah, either one or the other works as long as we're consistent. Whichever you'd prefer is fine.

PR #627, opened 2023-10-04
CORGI-836 add force_process option to loaderratadata

jasinner:
If an erratum was processed prior to it being shipped, and later the builds were updated and the errata shipped live the slow_load_errata task would not re-ingest the builds and add new relations. You have to use 'force_process' to load the erratum again.

### DIFF
--- a/corgi/tasks/management/commands/loaderratadata.py
+++ b/corgi/tasks/management/commands/loaderratadata.py
@@ -32,6 +32,12 @@ class Command(BaseCommand):
             help="Call tasks inline, not in a celery task",
         )
         parser.add_argument(
+            "-f",
+            "--force",
+            action="store_true",
+            help="Force process task",
+        )
+        parser.add_argument(
             "-p", "--product_variants", nargs="+", help="A list of variants to load errata for"
         )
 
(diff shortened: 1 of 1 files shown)


jasinner:
Using force_process is less than perfect because the old builds are still associated with the errata. However since we only processing shipped errata now, it shouldn't be a problem that will continue to occur in new data.

jasinner:

JimFuller-RedHat (approved):
LGTM

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -31,6 +31,12 @@ def add_arguments(self, parser: CommandParser) -> None:
             action="store_true",
             help="Call tasks inline, not in a celery task",
         )
+        parser.add_argument(
+            "-f",
+            "--force",
+            action="store_true",
+            help="Force process task",
+        )
```
juspence, line 39: Maybe I missed it - been a long time since I looked at argparse and Python management commands. But does this default to "false" if unset by the user?

To fix old builds still being associated with errata, should force_process delete the errata-type relations before recreating them? That seems likely to have side effects, so maybe it's not worth worrying about. We could also manually delete and force-reprocess all the existing errata-type relations in a migration (in a separate PR).

juspence (approved):

corgi/tasks/management/commands/loaderratadata.py
jasinner: Yes, this field will default to false.

I wanted to make this change as minimal as possible to fix the issue described in CORGI-836. It works as is, even though the older builds are still associated with the errata. If that starts to cause issues in the future we should manually clean them up. I created a task for that: CORGI-848

PR #629, opened 2023-10-04
CORGI-847: Release 1.3.10

juspence:
@RedHatProductSecurity/corgi-devs Please review. I've created this release-1.3.10 branch from the current 1.3.9 tag, so it doesn't include any of the latest changes in stage and only fixes this Gevent CVE.

If this all looks good, I'll immediately do the release when I get in first thing tomorrow morning. I don't want to do a full release since e.g. the Pyxis collector is still raising lots of errors in our monitoring emails, but please let me know if there's some other change(s) you think should be included.


jasinner (approved):
LGTM, thanks!

JimFuller-RedHat (approved):
LGTM - random question - did you consider cherry picking the original commit for this in main (e00e6bef3afb62d964231c679f5ef839c8f33dc7) ?

juspence:
> LGTM - random question - did you consider cherry picking the original commit for this in main ([e00e6be](https://github.com/RedHatProductSecurity/component-registry/commit/e00e6bef3afb62d964231c679f5ef839c8f33dc7)) ?

Merge conflicts :frowning: There were a bunch of conflicts I didn't want to deal with. It was simpler to just upgrade the one package I wanted (gevent) and let pip-compile automatically handle its dependencies, than try to resolve those conflicts correctly without accidentally updating something that didn't need it, or not updating some gevent dependency that did need to be updated.

juspence:
I tagged the previous `0867cfc499effafa2fd4cd5400e56a1ce8a1b113` commit as "1.3.10", which includes both the version number change and the Gevent dependency change, and released this to community and enterprise production. Then I dropped only the commit with the Gevent dependency change and force-pushed this branch, so that it now contains just the version number change.

I will now merge this version number change back into main / deploy it to stage so that the version numbers match between stage and production. The gevent dependency change is already included on the main branch / deployed to stage due to previous commits that were already merged. See the `e00e6bef3afb62d964231c679f5ef839c8f33dc7` commit that Jim noted above.

PR #630, opened 2023-10-05
CORGI-557: Fix bad migration and bug that creates RPMMOD nodes with no "/redhat/" in purls

juspence:
@RedHatProductSecurity/corgi-devs Please review again, sorry. The migration failed to run / apply in stage due to some bad code inside the for-loop. I've fixed that and run the migration successfully, but the for-loop shouldn't have been entered at all since I already cleaned up the dupes in stage manually (several months ago).

That means we've created (about 700) more ComponentNodes which are linked to RPMMODs, but which don't have `/redhat/` in their purls. All the RPMMOD components themselves definitely do have `/redhat/` in their purls, but there's a bug in save_node() where we create duplicate nodes when the purl on the component changes.

In detail, whenever we force-reprocess some build and update the object, its purl may change. Then save_node() does a get_or_create() using type, parent, and purl (the three unique fields for a ComponentNode) to find any existing node.

If the linked object's purl hasn't changed, this get() lookup will succeed, and the node will not be saved (nothing has changed, so the node doesn't need to be saved). If the linked object's purl has changed, ideally we'd do an update_or_create(), which would copy the new purl from the linked object to the node as part of the node.save() logic.

But using update_or_create() won't work. If the linked object's purl has changed, looking up the node using the new purl will fail. The node still has the old purl, so the get() / update() lookup will return no results and we will create() a new node with the new purl instead, while leaving behind the old (duplicate) node with the old purl.

So I added extra logic in the Component model, to copy any updated purl to all the existing nodes that are already linked to that Component, whenever we call .save() and the Component's purl has changed. I also added a test which failed before this change, but passes afterwards.


jasinner (approved):
LGTM thanks

juspence:
@RedHatProductSecurity/corgi-devs As we discussed, I've added another migration to handle duplicate container data. I'm still working on fixing the duplicate binary RPM and duplicate Maven data, but I think this first bit is good enough to merge on its own.

That will fix all the root component issues I know of, so update_or_create() / .save() will only be blocked for a small subset of components (hopefully) and not the roots of any trees. The bug creating this duplicate data will also be fixed, so we shouldn't create any more of it / need to clean up anything else in the future.

PR #632, opened 2023-10-09
CORGI-818: add active product stream filter to components

JimFuller-RedHat:
Filtering components to just those components in active product streams makes the search domain smaller and the first step in addressing CORGI-818. 

A few open questions:

* Is **active** the right url param name (alternately we could make it more explicit eg. **active_product_streams**)
* Do we want to match semantics of product_streams ex. _product_streams?active=all_  though it seems odd to union boolean with scalar value 'all'
* Do we want to emulate latest components semantics when set to False (eg. exclude active components) ? 
* What sort do we impose ?

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -120,6 +120,11 @@ class ComponentFilter(FilterSet):
         method="filter_gomod_components",
     )
 
+    active_streams = BooleanFilter(
+        label="Show components from active streams",
+        method="filter_active_streams",
+    )
+
     @staticmethod
     def filter_gomod_components(
         qs: QuerySet[Component], _name: str, value: bool
@@ -194,6 +199,16 @@ class ComponentFilter(FilterSet):
         # Truthy values return the filtered queryset (only latest components)
         # Falsey values return the excluded queryset (only older components)
         return queryset.latest_components_by_streams(include=value)
+
+    @staticmethod
+    def filter_active_streams(
+        queryset: ComponentQuerySet, _name: str, value: bool
+    ) -> QuerySet["Component"]:
+        """Show only components from active streams when True"""
+        if value in EMPTY_VALUES:
+            # User gave an empty ?param= so return the unfiltered queryset
+            return queryset
+        return queryset.active_streams(include=value)
 
 
 class ProductDataFilter(FilterSet):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1796,3 +1796,62 @@ def test_product_streams_exclude_components(client, api_path):
     assert response.status_code == 200
     response = response.json()
     assert response["exclude_components"] == ["starfish", "seahorse"]
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_latest_components_by_active_filter(client, api_path):
+    active_stream = ProductStreamFactory(active=True)
+    other_active_stream = ProductStreamFactory(active=True)
+    inactive_stream = ProductStreamFactory(active=False)
+
+    for active_name in "red", "orange":
+        new_srpm = SrpmComponentFactory(name=active_name)
+        new_srpm.productstreams.add(active_stream)
+        for arch in "aarch64", "x86_64":
+            new_binary_rpm = BinaryRpmComponentFactory(
+                name=active_name,
+                arch=arch,
+            )
+            new_binary_rpm.sources.add(new_srpm)
+            new_binary_rpm.productstreams.add(active_stream)
+
+    for active_name in "blue", "green":
+        new_srpm = SrpmComponentFactory(name=active_name)
+        new_srpm.productstreams.add(active_stream)
+        new_srpm.productstreams.add(other_active_stream)
+        for arch in "aarch64", "x86_64":
+            new_binary_rpm = BinaryRpmComponentFactory(
+                name=active_name,
+                arch=arch,
+            )
+            new_binary_rpm.sources.add(new_srpm)
+            new_binary_rpm.productstreams.add(active_stream)
+            new_binary_rpm.productstreams.add(other_active_stream)
+
+    for inactive_name in "purple", "pink":
+        inactive_component = BinaryRpmComponentFactory(
+            name=inactive_name,
+        )
+        inactive_component.productstreams.add(inactive_stream)
+
+    for both_name in "brown", "yellow":
+        both_component = BinaryRpmComponentFactory(
+            name=both_name,
+        )
+        both_component.productstreams.add(active_stream)
+        both_component.productstreams.add(inactive_stream)
+
+    response = client.get(f"{api_path}/components")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 16
+
+    response = client.get(f"{api_path}/components?active_streams=True")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 14
+
+    response = client.get(f"{api_path}/components?active_streams=False")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-* refactored latest component filter into a pgsql stored proc
 
 ### Changed
-* added exclude_components to product_stream
-* 
+* refactored latest component filter into a pgsql stored proc
+
+### Added
+* added exclude_components to /api/v1/product_streams
+* include_inactive_streams to /api/v1/components
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1111,6 +1111,14 @@ class ComponentQuerySet(models.QuerySet):
         # Falsey values return the excluded queryset (only non-SRPM components)
         return self.exclude(SRPM_CONDITION)
 
+    def active_streams(self, include: bool = True) -> "ComponentQuerySet":
+        """Show only components in active product streams"""
+        if include:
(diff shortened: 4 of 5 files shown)


JimFuller-RedHat:
no clue why test-performance is failing - nothing I did seems related ... though maybe it is, will investigate.

jasinner:
I agree the active=all 'all' value on ProductStreams is odd. We could replace the value 'all' with 'false'. The default search should default to only active, and if we want to include inactive ProductStreams as well we should set the value to 'false'. I don't think there is any value in only returning components from inactive streams.

For components the filter might be better as include_inactive and make the non default value 'true' to make it a bit more explicit what to expect in the results. As you suggest, it's probably even better to call the filter include_inactive_streams, with 'streams' in the name.

Perhaps ordering by created_at?

jasinner (approved):

JimFuller-RedHat:
> I agree the active=all 'all' value on ProductStreams is odd. We could replace the value 'all' with 'false'. The default search should default to only active, and if we want to include inactive ProductStreams as well we should set the value to 'false'. I don't think there is any value in only returning components from inactive streams.
> 

If it is ok with team I am not going to revisit existing semantics on **/api/v1/product_streams** ... in this PR.

> For components the filter might be better as include_inactive and make the non default value 'true' to make it a bit more explicit what to expect in the results. As you suggest, it's probably even better to call the filter include_inactive_streams, with 'streams' in the name.
>

agreed 
 
> Perhaps ordering by created_at?

this would change sort order on the default invoke of **/v1/api/components** ... to be clear are we ok with that (I am as I think it prob might even be more useful ) ?

JimFuller-RedHat:
lets wait to get @juspence thoughts on this .

corgi/api/views.py
```diff
@@ -627,6 +629,9 @@ def list(self, request: Request, *args: tuple, **kwargs: dict) -> Response:
                 return Response({"count": productstreams.count(), "results": serializer.data})
             if view == "summary":
                 self.serializer_class = ComponentListSerializer
+            # default is to emit components only from active streams sorted by 'created_at'
+            if not include_inactive_streams:
+                self.queryset = self.queryset.filter(productstreams__active=True).using("read_only")
```
juspence: Adding custom logic / code is a little bug-prone, and has to be annotated with the OpenAPI parameter above (new line 599) to appear in the schema.

The logic here is aso rather complex, e.g. do we need this parameter on any of the other branches? Refactoring in the future will be tricky, and that refactoring is needed if we want to add tags (due to some subtle lurking bugs which are already present here).

It's much easier to add a new django-filter to the ComponentFilter class (around line 63 of the corgi/api/filters.py module). That filter can default to filtering "only active" if no value is given, or if include_inactive_streams=False is given. Otherwise if include_inactive_streams=True is given, it can show "all components" / not do any filtering.

If we're keeping the current approach, I would suggest a small change:
```suggestion
            # default is to emit components only from active streams sorted by 'created_at'
            if not include_inactive_streams:
                # Maybe causes duplicates if one component is linked to multiple active streams, so add .distinct() for safety:
                self.queryset = self.get_queryset().filter(productstreams__active=True).distinct()
```

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
...
+        for arch in "aarch64", "x86_64", "src":
+            new_component = ComponentFactory(
+                name=active_name,
+                arch=arch,
+                type=Component.Type.RPM,
+            )
+            new_component.productstreams.add(active_stream)
+
+    for inactive_name in "blue", "green":
+        inactive_component = ComponentFactory(
```
juspence: Let's reuse the BinaryRPM or SourceRPM factories here.

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
...
+    inactive_stream = ProductStreamFactory(active=False)
+
+    for active_name in "red", "orange":
+        for arch in "aarch64", "x86_64", "src":
+            new_component = ComponentFactory(
+                name=active_name,
+                arch=arch,
+                type=Component.Type.RPM,
+            )
+            new_component.productstreams.add(active_stream)
```
juspence:
```suggestion
        new_srpm = SourceRpmComponentFactory(
            name=active_name
        )
        new_srpm.productstreams.add(active_stream)
        for arch in "aarch64", "x86_64":
            new_binary_rpm = BinaryRpmComponentFactory(
                name=active_name,
                arch=arch,
            )
            new_binary_rpm.sources.add(new_srpm)
            new_binary_rpm.productstreams.add(active_stream)
```

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
     assert response.status_code == 200
     response = response.json()
     assert response["exclude_components"] == ["starfish", "seahorse"]
+
+
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_latest_components_by_active_filter(client, api_path):
+    active_stream = ProductStreamFactory(active=True)
+    inactive_stream = ProductStreamFactory(active=False)
```
juspence: Let's test a component that is included in both an active and an inactive stream, to make sure it shows up both when the filter is true (showing only components in active streams), and when the filter is false (showing only components in inactive streams).

Let's also test a component that is in multiple active streams. I think the current queryset will give duplicates in that case, but I'm not 100% sure and adding a test for this will prevent future regressions in any refactoring.

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
...
+
+    response = client.get(f"{api_path}/components?include_inactive_streams=True")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 8
+
+    response = client.get(f"{api_path}/components?include_inactive_streams=False")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 8
```
juspence: `include_inactive_streams=False` means "show me only components in active streams", right? Shouldn't the result count be 6 here? Two names "red" and "orange" times three arches gives 6 results?

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
...
+
+    response = client.get(f"{api_path}/components")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 6
+
+    response = client.get(f"{api_path}/components?include_inactive_streams=True")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 8
```
juspence: It looks like the result is the same whether or not you set the filter to True or False, so I think there's a bug in the logic somewhere.

I guess `include_inactive_streams=True` means "show me components in both active and inactive streams", since the default is "only active" and we're including inactive as well (True).

So having 8 results here seems correct for this case. But maybe I've misunderstood the logic, and we only want to see the 2 inactive components (named "blue" and "green") here.

corgi/api/views.py
```diff
@@ -627,6 +629,9 @@ def list(self, request: Request, *args: tuple, **kwargs: dict) -> Response:
                 return Response({"count": productstreams.count(), "results": serializer.data})
             if view == "summary":
                 self.serializer_class = ComponentListSerializer
+            # default is to emit components only from active streams sorted by 'created_at'
+            if not include_inactive_streams:
```
juspence: There's a bug here. We're getting a string parameter, but treating it like a bool. So not giving any value defaults to an empty "" string or None / NULL parameter, which in Python is boolean False. The test correctly shows only 6 results in this case.

If we pass a string "true", "false", or "garbage", Python treats that as a boolean True (there is a string value / the string is not empty). So we see results for components in both active and inactive streams, even if the user passed "false" / wants to see only components in active streams.

Django-filters handle parsing the string parameter values for you, so truthy values like "1", "true", "True", and "TRUE" all become a boolean True, while other values like "0", "false", "False", and "FALSE" are correctly converted to a boolean False.

tests/test_api.py
```diff
@@ -1796,3 +1796,40 @@ def test_product_streams_exclude_components(client, api_path):
...
+        inactive_component = ComponentFactory(
+            name=f"inactive_component_{inactive_name}",
+            type=Component.Type.RPM,
+        )
+        inactive_component.productstreams.add(inactive_stream)
+
+    response = client.get(f"{api_path}/components")
+    assert response.status_code == 200
+    response = response.json()
+    assert response["count"] == 6
```
juspence: This case looks right. We default to showing "only components in active streams" when no parameter is given, and there are two active names ("red" and "orange") times three architectures, so we get 6 results like I expect.

juspence (commented):
There's at least one bug here, so I think the simpler approach is to use Django-filters instead of custom code. But the current approach is also fine if you fix the definite bug and add a test for the possible bug.

Re: comments above, sort order needs to stay the same to avoid Django-REST-Framework pagination bugs that Adrian and Ondrej can describe more fully. All I remember is that sort order must be on either a field with unique=True, or all fields appearing together in a UniqueConstraint, which should be the case today. Since created_at isn't unique and could have collisions, we shouldn't use it for sorting.

I agree that active=all is a wart we could fix in a future PR, but it doesn't need to be part of this one.

tests/test_api.py
JimFuller-RedHat: was copying status quo elsewhere in the tests, but can use

corgi/api/filters.py
```diff
@@ -195,6 +200,16 @@ def filter_latest_components_by_streams(
         # Falsey values return the excluded queryset (only older components)
         return queryset.latest_components_by_streams(include=value)
 
+    @staticmethod
+    def filter_active_streams(
+        queryset: ComponentQuerySet, _name: str, value: bool
+    ) -> QuerySet["Component"]:
+        """Show only"""
```
juspence: Looks like you forgot to fill in the comment here, otherwise LGTM! Thanks for switching to the django-filter approach.

juspence (approved):

PR #633, opened 2023-10-10
CORGI-824 exclude all test directory from syft scan

jasinner:
Syft is detecting components incorrectly from test data. I adjusted our existing exclude filter to filter out all directories with 'test' in the path. That covered 2 existing excludes as well as this new one with the rpmlint source code.

I will write a script to delete the existing bad components from rpmlint RPMs and attach it to CORGI-824


jasinner:

corgi/core/migrations/0095_clean_rpmlint_invalid_syft_deps.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def clean_rpmlint_invalid_syft_deps(apps, schema_editor):
+    Component = apps.get_model("core", "Component")
+
+    for rpmlint_src in Component.objects.filter(name="rpmlint", arch="src"):
+        # Delete any provide which wasn't detected by koji.listRPMs (the noarch rpm only)
+        rpmlint_src.provides.exclude(meta_attr__source__contains="koji.listRPMs").delete()
```
juspence, line 9: Should we schedule / rerun the SCA task for these SRPMs after we delete the non-Koji / SCA provides?

corgi/core/migrations/0095_clean_rpmlint_invalid_syft_deps.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+
+
+def clean_rpmlint_invalid_syft_deps(apps, schema_editor):
+    Component = apps.get_model("core", "Component")
+
+    for rpmlint_src in Component.objects.filter(name="rpmlint", arch="src"):
```
juspence:
```suggestion
    for rpmlint_src in Component.objects.filter(name="rpmlint", type="RPM", arch="src"):
```

juspence (approved):
LGTM, comments are minor and sorry for the delay in getting to reviews lately.

corgi/core/migrations/0095_clean_rpmlint_invalid_syft_deps.py
jasinner: No need. It won't pick up anything new.

PR #634, opened 2023-10-10
fix mypy error with builds?ofuri endpoint and released components filter for rhel ga streams

jasinner:


JimFuller-RedHat (approved):
LGTM

PR #635, opened 2023-10-10
CORGI-834: Fix more Pyxis collector errors

juspence:
@RedHatProductSecurity/corgi-devs I'm still working on tests for this, but opening for review now since I lost part of Monday.

### DIFF
--- a/corgi/web/templates/product_manifest.json
+++ b/corgi/web/templates/product_manifest.json
@@ -83,7 +83,7 @@
       "packageFileName": "NOASSERTION",
       "SPDXID": "SPDXRef-{{obj.uuid}}",
       "supplier": "Organization: Red Hat",
-      "versionInfo": "{% if obj.epoch %}{{obj.epoch|escapejs}}-{% endif %}{{obj.version|escapejs}}{% if obj.release %}-{{obj.release|escapejs}}{% endif %}"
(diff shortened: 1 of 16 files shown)


corgi/web/templates/product_manifest.json
jasinner, line 1: What is the motivation for this? If the component is an RPM we won't include the epoch, or release in the versionInfo.

jasinner (commented):
Overall looks good to me thanks. Only concern is with the final commit which changes versionInfo in product manifest template.

corgi/web/templates/product_manifest.json
juspence: This is a template for a product, and "obj" here is a ProductStream (today - could be ProductVariant, ProductVersion, etc. in the future). AKA we're providing data for the last package in the list / the "packages" key. You can tell based on the licenseComments field, which is talking about "no licenses available for products, only components".

There is no obj.epoch or obj.release here, but Django templates are very forgiving and just convert obj.unknown_props to empty strings. The if statements end up always being False, so we still report just "obj.version", I think. But this was a bug I should have caught in a previous review (oops), so fixing it now.

jasinner: Ok, that makes sense to me to update it to obj.version for a ProductStream then.

PR #636, opened 2023-10-11
fix components missing from inactive (rhel-9.2.0) stream manifests

jasinner:
This should fix the broken performance test in stage and populate RHEL GA manifests (which are now inactive) with components.


jasinner:
@JimFuller-RedHat can you check this? I've verified it fixes the number of packages in the (inactive) rhel-9.2.0 manifest

jasinner:
performance-test is fixed by this so merge to fix problem CI pipeline.

PR #637, opened 2023-10-12
CORGI-621: Fix NEVRA, NVR formatting when version, release both empty "" strings

juspence:
@RedHatProductSecurity/corgi-devs Please review this nice, simple fix for a formatting error. There's also a migration that should fix the existing data, but I'm not sure about the logic in one query.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1630,15 +1630,21 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         return self.type == Component.Type.RPM and self.arch == "src"
 
     def get_nvr(self) -> str:
+        # Many GOLANG components don't have a version or release set
+        # so don't return an oddly formatted NVR, like f"{name}-"
+        version = f"-{self.version}" if self.version else ""
         release = f"-{self.release}" if self.release else ""
-        return f"{self.name}-{self.version}{release}"
+        return f"{self.name}{version}{release}"
 
     def get_nevra(self) -> str:
         epoch = f":{self.epoch}" if self.epoch else ""
+        # Many GOLANG components don't have a version or release set
+        # so don't return an oddly formatted NEVRA, like f"{name}-.{arch}"
+        version = f"-{self.version}" if self.version else ""
         release = f"-{self.release}" if self.release else ""
         arch = f".{self.arch}" if self.arch else ""
 
-        return f"{self.name}{epoch}-{self.version}{release}{arch}"
+        return f"{self.name}{epoch}{version}{release}{arch}"
 
     def _build_repo_url_for_type(self) -> str:
         """Get an upstream repo URL based on a purl"""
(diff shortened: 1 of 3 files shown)


corgi/core/migrations/0092_fix_nevra_formatting.py
```diff
@@ -0,0 +1,33 @@
...
+    # arch is not present in the NEVRA for this component type (whatever it may be)
+    Component.objects.filter(
+        nevra__endswith="-", epoch=0, version="", release="", arch="noarch"
+    ).update(nvr=F("name"), nevra=F("name"))
+
+    # arch is present in the NEVRA for this component type (whatever it may be)
+    Component.objects.filter(
+        nevra__endswith="-.noarch", epoch=0, version="", release="", arch="noarch"
+    ).update(nvr=F("name"), nevra=functions.Concat(F("name"), Value(".noarch")))
+    # TODO: Why does this query return results in stage?
```
jasinner: I think nevra only really useful for REDHAT RPMs or OCI types. For those Components we should append the .noarch suffix, but otherwise leave it off the nevra.

corgi/core/models.py
```diff
@@ -1575,15 +1575,21 @@ def is_srpm(self):
...
-        return f"{self.name}-{self.version}{release}"
+        return f"{self.name}{version}{release}"
 
     def get_nevra(self) -> str:
         epoch = f":{self.epoch}" if self.epoch else ""
+        # Many GOLANG components don't have a version or release set
+        # so don't return an oddly formatted NEVRA, like f"{name}-.{arch}"
+        version = f"-{self.version}" if self.version else ""
         release = f"-{self.release}" if self.release else ""
         arch = f".{self.arch}" if self.arch else ""
```
jasinner, line 1645: Perhaps we could add an additional check here for namespace = REDHAT, and type in ["RPM","CONTAINER_IMAGE"] and only append the arch in that case?

jasinner (approved):
I think the changes suggested are fine as they are. However if you want to invest some more time, and agree with the suggested improvements to NEVRA please go ahead. I don't think we make use of nevra, or nvr in manifests so our time is probably invested in other tasks contributing to the GA release of SBOM.

corgi/core/migrations/0092_fix_nevra_formatting.py
juspence: I agree it's not useful otherwise, but having a consistent format is a little simpler for our users (and us) I think. Always having ".noarch" is redundant e.g. for PyPI and Golang packages, but made the migration in this PR very easy to write.

In other words, the consistency means we can just find any components which have arch=noarch, but NEVRAs missing ".noarch", and automatically know they need to be updated, without having to check the namespace or type as well. I'm not strongly opposed if you really want to make the change though - feel free to file a follow-up ticket.

corgi/core/models.py
juspence: See above - I'm going to work on other tasks as you say, and I think the current code is simpler and easier to maintain. But I'm not strongly opposed if you really would like to get this change in.

PR #639, opened 2023-10-15
update django-celery-results==2.5.1, psycopg2==2.9.9

JimFuller-RedHat:
some perfunctory updates to dependencies:

* **django-celery-results**: addresses generation of idle (unclosed) db TX with updates to django celery results table
* **psycopg2**: minor perf improvements communicating with db


jasinner (approved):
LGTM

PR #641, opened 2023-10-16
CORGI-818: Search provides upstreams latest components

JimFuller-RedHat:
Enable searching of root components by their provides and upstreams combined with latest_components_by_streams.

The following searches root components by iregex matching on 'pywebkitgtk' against component names provided by the root component:
```
api/v1/components?re_provides_name=pywebkitgtk&latest_components_by_stream=True&include_fields=purl,namespace,name&root_components=True
```
similar for upstreams
```
api/v1/components?re_upstreams_name=pywebkitgtk&latest_components_by_stream=True&include_fields=purl,namespace,name&root_components=True
```

and will be adding more tests.

**Note** - we could use existing re_provides and re_upstreams but those are based on purls ... somewhat uncontained in terms of searching. Griffon is free to use both

While I was in the area I optimised a number of things
* refactor ofuri into django filter
* ensure we get distinct results with common sorting from api/v1/components
* minor fix on latest_components_by_streams where we were returning dupe product streams
* added associated gin indexes


JimFuller-RedHat:
gah, need to use GinTrgmIndex (and update dep to use it) ... will fix

corgi/api/filters.py
```diff
@@ -210,6 +222,30 @@ def filter_active_streams(
...
+        if isinstance(model, Product):
+            components_for_model = queryset.filter(products__ofuri=value)
+        elif isinstance(model, ProductVersion):
+            components_for_model = queryset.filter(productversions__ofuri=value)
+        elif isinstance(model, ProductStream):
+            components_for_model = queryset.filter(productstreams__ofuri=value)
+        elif isinstance(model, ProductVariant):
+            components_for_model = queryset.filter(productvariants__ofuri=value)
+        else:
+            components_for_model = queryset
```
jasinner: Should this return a 404 response instead indicating that the Product with the specified ofuri couldn't be found?

corgi/api/views.py
```diff
@@ -566,8 +566,8 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
     filterset_class = ComponentFilter
 
     def get_queryset(self) -> QuerySet[Component]:
-        # we always want a distinct result from REST API
-        return self.queryset.distinct()
+        # ensure we always return distinct set of components from REST API and enforce a common sort
+        return self.queryset.order_by("name").distinct()
```
jasinner: Justin said on [pull/632](https://github.com/RedHatProductSecurity/component-registry/pull/632#pullrequestreview-1671346151) that "sort order must be on either a field with unique=True, or all fields appearing together in a UniqueConstraint, which should be the case today. " Therefore candidates are "uuid", "purl" or the combination of ("name", "type", "arch", "version", "release"). If not we will likely run into bugs paginating results as OSIDB did.

jasinner (commented):
Looking good, thank you! I think we can't use the suggested order_by("name") for ComponentQuerySet however.

corgi/api/filters.py
JimFuller-RedHat: this is now just setting queryset - was unsure of embedding http behaviour here but yes I can otherwise the options are 1) throw error (as before) 2) return original queryset 3) return empty queryset ...

corgi/api/views.py
JimFuller-RedHat: ah right ... let me investigate - will probably remove and keep status quo

jasinner (approved):

corgi/core/models.py
```diff
@@ -925,7 +925,9 @@ def latest_components(
                 .using("read_only")
             )
         if ofuri:
-            components = components.filter(productstreams__ofuri=ofuri).using("read_only")
+            components = (
+                components.filter(productstreams__ofuri=ofuri).distinct().using("read_only")
+            )
```
juspence: You don't need .using("read_only") here, since you're just appending additional filters to a queryset that already defines which DB to use. See line 925 above where .using("read_only") is already defined for this "components" / self queryset.

I also think we shouldn't define .using("read_only") in any models.py code, since we don't know whether it's being called by the API (which should use read-only views most of the time), or a Celery worker / task (which should use read/write views when updating e.g. the latest component's data).

Instead, let's add .using("read_only") in the API code, and add nothing in the Celery task code, just after we call this latest_components() queryset. Adding nothing means "use the read/write DB by default", so the only place we need extra / special logic should be in the read-only parts of the API.

corgi/api/views.py
```diff
@@ -567,33 +566,11 @@ class ComponentViewSet(ReadOnlyModelViewSet):  # TODO: TagViewMixin disabled unt
...
-            components_for_model = self.queryset.filter(productvariants__ofuri=ofuri)
-        else:
-            # No matching model instance found, or invalid ofuri
-            raise Http404
-        return components_for_model.root_components().latest_components(
-            model_type=model_type,
-            ofuri=ofuri,
-        )
+        # ensure we always return distinct set of components from REST API
+        return self.queryset.distinct()
```
juspence: Instead of overriding standard Django methods, can we remove this and just call .distinct() directly wherever it's needed?

Or you can add this to the queryset class property on line 556 above, so it applies to all usages of self.get_queryset(). Then you don't have to remember to add it manually in each place .distinct() is needed, if you're fine with the performance penalty of always calling SELECT DISTINCT even when it's not needed.

Or at the very least, let's make sure if we keep our custom get_queryset(), that we call super().get_queryset().distinct() instead of accessing the raw self.queryset property directly.

corgi/core/migrations/0096_auto_20231016_1919.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        # install gin/trgrm postgres extensions
+        BtreeGinExtension(),
+        TrigramExtension(),
+        # setting gin indexes with gin_trgm_ops does not work when define on
+        # django Component model indexes - so we set them manually
+        migrations.RunSQL(
+            "CREATE INDEX IF NOT EXISTS compons_search_name ON core_component USING gin (name gin_trgm_ops);"  # noqa: E501
```
juspence: Even if we can't create the index on the model using standard Django features, can we at least add a comment reminding us that this index exists? Otherwise I think we'll forget about it / it's likely to break in the future if we refactor something important.

It may also be helpful to add a test for all these custom indexes we keep adding. You can write a standard Django query or RawSQL statement that should use the index, then assert that "INDEX SCAN" appears in the query's .explain() output.

juspence (commented):
Not doing a full review, just leaving a comment here so I get updates about this PR. When you're ready for review you can tag at-corgi-devs so the whole team gets an email alert that review is needed. Thanks again for moving more logic into Django filters!

corgi/core/migrations/0096_auto_20231016_1919.py
JimFuller-RedHat: tests are coming which includes this

corgi/core/models.py
juspence: We don't want to default to read_only, since Celery tasks need to read and write data. If they read from read_only, then write back to the default DB, we risk heisenbugs.

For example, changes in the default DB which have not been replicated to the read_only replica, will be overwritten when an updated object is saved. Normally AWS replication lag is very low, but we have a SignalFX dashboard that monitors this and sometimes there are spikes.

That's also why we're not using a DB router (either a library or our own) like OSIDB does. Those routers send all reads to the replica DB, even if the object will later be used in a write call / saved back to the default DB.

I know it's complex, but it's only needed in one place (API code) and you can set a default distinctness / ordering / DB replica usage for each ViewSet's default queryset. In other words, there should be no need to add methods like this in many places as long as you're using self.get_queryset(), and not creating a new queryset like MyModel.objects.filter().

JimFuller-RedHat: thx for additional info - not questioning the reason why we are doing routing, more questioning why it is in the app layer - usually the solution for this kind of thing is one layer lower then the app itself ... there are several ways to route SELECT or INSERT to the right db/replica ... not a now thing

PR #643, opened 2023-10-18
CORGI-843 container errata builds

jasinner:
This might be easier to review commit-wise. It's a fairly small change to the ErrataTool collector to use a different url for non-rpm Errata. It also adds a migration which finds all streams with brew_tags and tries reloads their errata. I limited it to streams with brew_tags as they are the ones with released container errata.

I wanted to reuse some existing code from the prod_defs task and loaderratadata management command so I refactored some functions there so I could call them from the migration.

### DIFF
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -262,13 +262,6 @@ class ErrataTool:
         shipped_live = errata_type_details[0]["status"] == "SHIPPED_LIVE"
         return erratum_id, shipped_live
 
-    def get_builds_for_errata(self, errata_id: int) -> list[int]:
-        build_ids = set()
-        for variant in self.get_erratum_components(str(errata_id)).values():
-            for component in variant:
-                build_ids.update(component.keys())
-        return list(build_ids)
-
     def _parse_module(
         self,
         module_name: str,
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -26,6 +26,7 @@ class ErrataTool:
     """Interface to the Errata Tool APIs."""
 
     GSSAPI_AUTH = HTTPSPNEGOAuth()
+    SHIPPED_LIVE_SEARCH = "show_state_SHIPPED_LIVE=1"
 
     def __init__(self):
         self.session = requests.Session()
@@ -45,7 +46,7 @@ class ErrataTool:
     def get_paged(
         self, path: str, page_data_attr: str = "data", pager: str = "page[number]"
     ) -> list[dict[str, Any]]:
-        """Generator to iterate over data from paged Errata Tool API endpoint."""
+        """Iterate over data from paged Errata Tool API endpoint."""
         params = {pager: 1}
         data: list[dict[str, Any]] = []
         while True:
@@ -321,13 +322,41 @@ class ErrataTool:
             product["product_versions"] = product_versions
         return products
 
+    def _do_errata_search(
+        self, search_values: str, product_variants: list[str], container_only: bool = False
+    ) -> list[int]:
+        if not search_values:
+            raise ValueError("search_values cannot be an empty string")
+        if container_only:
+            search_values += "&content_type[]=docker"
+        product_errata = self.get_paged(f"api/v1/erratum/search?{search_values}")
+        all_errata: set[int] = set()
+        for erratum in product_errata:
+            logger.debug(f"Found shipped erratum {erratum['id']}")
+            # If we are passed a list of variants, check that there are builds matching those
+            # variants on the erratum before returning that erratum_id
+            if not product_variants:
+                all_errata.add(erratum["id"])
+                continue
+            is_rpm = "rpm" in erratum["content_types"]
+            builds_url = self._get_url_by_erratum_type(erratum["id"], is_rpm)
+            builds = self.get(builds_url)
+            for product_version in builds.values():
+                for build in product_version["builds"]:
+                    for build_data in build.values():
+                        for variant in build_data["variant_arch"].keys():
+                            if variant in product_variants:
+                                logger.info(f"Found variant {variant} in {erratum['id']}")
+                                all_errata.add(erratum["id"])
+        return list(all_errata)
+
     def get_errata_matching_variants(
         self, product_variants: list[str], container_only: bool = False
-    ) -> set[str]:
-        """Given a list of variants in a single ET product, find all errata shipping builds to those
+    ) -> list[int]:
+        """Given a list of variants in a single ET product, load all errata shipping builds to those
         variants"""
         if not product_variants:
-            return set()
+            return []
         product_ids = set()
         for et_variant in CollectorErrataProductVariant.objects.filter(name__in=product_variants):
             variant_version = et_variant.product_version
@@ -336,28 +365,22 @@ class ErrataTool:
         # make sure there is only 1 product id for the given variants
         if len(product_ids) > 1:
             raise ValueError(f"Variants had more than 1 product id. Found: {product_ids}")
-
+        elif len(product_ids) == 0:
+            raise ValueError(f"Variants passed do not exist in DB: {product_variants}")
         product_id = product_ids.pop()
         logger.info(f"Searching shipped errata for product with id {product_id}")
         # Unfortunately we can't use the variant id directly here
-        search_values = f"show_state_SHIPPED_LIVE=1&product[]={product_id}"
-        if container_only:
-            search_values += "&content_type[]=docker"
-        product_errata = self.get_paged(
-            f"api/v1/erratum/search?{search_values}",
-            page_data_attr="data",
-        )
+        search_values = f"{self.SHIPPED_LIVE_SEARCH}&product[]={product_id}"
+        return self._do_errata_search(search_values, product_variants, container_only)
 
-        all_errata: set[str] = set()
-        for erratum in product_errata:
-            logger.info(f"Found shipped erratum {erratum['id']}")
-            is_rpm = "rpm" in erratum["content_types"]
-            builds = self.get(self._get_url_by_erratum_type(erratum["id"], is_rpm))
-            for product_version in builds.values():
-                for build in product_version["builds"]:
-                    for _, build_data in build.items():
-                        for variant in build_data["variant_arch"].keys():
-                            if variant in product_variants:
-                                logger.info(f"Found variant {variant} in {erratum['id']}")
-                                all_errata.add(str(erratum["id"]))
-        return all_errata
+    def get_errata_for_releases(
+        self, releases: list[int], container_only: bool = False
+    ) -> set[int]:
+        all_errata: set[int] = set()
+        if not releases:
+            return all_errata
+        search = self.SHIPPED_LIVE_SEARCH
+        for release in releases:
+            search += f"&release[]={release}"
+        # When doing a search for errata by releases, don't filter by product_variants
+        return set(self._do_errata_search(search, [], container_only))
--- a/corgi/collectors/errata_tool.py
+++ b/corgi/collectors/errata_tool.py
@@ -215,6 +215,11 @@ class ErrataTool:
 
         return dict(variant_to_component_map)
 
+    @staticmethod
+    def _get_url_by_erratum_type(erratum_id: str, is_rpm: bool) -> str:
+        endpoint = "builds_list.json" if is_rpm else "builds"
+        return f"api/v1/erratum/{erratum_id}/{endpoint}"
+
     def save_variant_cdn_repo_mapping(self):
         """Fetches all existing Errata Tool Variants and the CDN repos they are configured with.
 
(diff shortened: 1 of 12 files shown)


corgi/collectors/errata_tool.py
```diff
@@ -220,13 +220,6 @@ def get_errata_key_details(self, name: str) -> tuple[int, bool]:
         shipped_live = errata_type_details[0]["status"] == "SHIPPED_LIVE"
         return erratum_id, shipped_live
 
-    def get_builds_for_errata(self, errata_id: int) -> list[int]:
```
JimFuller-RedHat, line 223: I love the smell of 'deleted code' in the morning ;)

corgi/collectors/errata_tool.py
```diff
@@ -209,7 +212,7 @@ def get_variant_cdn_repo_mapping() -> dict[str, list[str]]:
         }
         return variants_with_repos
 
-    def get_errata_key_details(self, name: str) -> tuple[int, bool]:
+    def get_errata_key_details(self, name: str) -> tuple[int, bool, bool]:
```
JimFuller-RedHat, line 217: observation - always difficult to know if a function returning tuples is better then just atomic getters ... both approaches can complexify other code ... this is fine, just opining.

corgi/tasks/prod_defs.py
```diff
@@ -346,6 +340,16 @@ def _parse_cpes_from_brew_tags(
         return []
 
 
+def brew_tag_to_et_prefixes(brew_tag: str) -> tuple[str, str]:
+    """Match brew tags in prod_defs with those from ET by reducing them to common prefixes"""
+    trimmed_brew_tag = brew_tag.removesuffix("-candidate")
```
JimFuller-RedHat, line 345: I wonder if we should parse these tags into their composite parts eg. we may find ourselves having to munge them in other places.

corgi/core/migrations/0096_load_missing_container_errata_builds.py
```diff
@@ -0,0 +1,60 @@
...
+    ProductStream = apps.get_model("core", "ProductStream")
+    CollectorErrataProductVersion = apps.get_model("collectors", "CollectorErrataProductVersion")
+
+    variants_by_stream = defaultdict(set)
+    # Allow this migration to be re-run in case of timeout or other failure
+    # Only get the streams we haven't processed yet
+    for stream in (
+        ProductStream.objects.annotate(num_components=Count("components"))
+        .filter(num_components__gt=0)
+        .exclude(tags__name="no_manifest")
```
JimFuller-RedHat, line 21: random observation: if this tag is some kind official 'minted' tag that we use in the system (for example collectors set this) ... then maybe we want to consider having a constant defined for this.

JimFuller-RedHat (approved):
apart from failing CI - LGTM though my grasp of the finer details are not very up to date

corgi/core/migrations/0096_load_missing_container_errata_builds.py
jasinner: Good idea, I created a constant in the tagging task.

corgi/tasks/prod_defs.py
jasinner: That would be overkill in my opinion.

corgi/tasks/tagging.py
```diff
@@ -5,6 +5,8 @@
 from corgi.core.models import Product, ProductStream, ProductStreamTag, ProductVersion
 from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
+NO_MANIFEST_TAG = "no_manifest"
```
juspence, line 8: I thought in a previous PR, you used "!manifest" with an exclamation point. And then you added logic to use Django's filter() method for regular tags, but Django's exclude() method for tags with ! at the beginning. As part of the Django-filters custom code for the ?tags= API parameter, somewhere in corgi/api/filters.py I think?

If that's right, can we get rid of that custom code now that we just use a specific "no_manifest" tag with no special ! exclamation pojnt syntax? Or did I dream that? :grin:

corgi/collectors/errata_tool.py
```diff
@@ -218,14 +223,8 @@ def get_errata_key_details(self, name: str) -> tuple[int, bool]:
             )
         erratum_id = errata_type_details[0]["id"]
         shipped_live = errata_type_details[0]["status"] == "SHIPPED_LIVE"
-        return erratum_id, shipped_live
-
-    def get_builds_for_errata(self, errata_id: int) -> list[int]:
```
juspence, line 265: Are we just removing this because it was unused? It's fine either way, I just wanted to make sure.

tests/test_errata_data.py
```diff
@@ -482,3 +490,16 @@ def test_invalid_errtaum_details(requests_mock):
         )
     with pytest.raises(ValueError):
         ErrataTool().get_errata_key_details("123456")
+
+
+def test_parse_container_errata_components(requests_mock):
+    erratum_id = 97738
+    build_id = 2177946
+    with open("tests/data/errata/errata_container_builds.json", "r") as remote_source_data:
+        requests_mock.get(
```
juspence, line 706: You might assert `requests_mock.call_count == 1` at the end, or whatever the attribute name is. This mock is special and doesn't have the same assert_called_once() method that other mocks do.

corgi/collectors/errata_tool.py
```diff
@@ -152,7 +152,7 @@ def get_erratum_components(self, erratum_id: str):
         }
         """
         variant_to_component_map: typing.DefaultDict[str, list] = defaultdict(list)
-        builds = self.get(f"api/v1/erratum/{erratum_id}/builds_list.json")
+        builds = self.get(self._get_url_by_erratum_type(erratum_id, is_rpm))
```
juspence: This is minor but it's a little nicer to have code like this split across two lines, so that if we get a traceback it's easier to see which part failed.
```suggestion
        builds_url = self._get_url_by_erratum_type(erratum_id, is_rpm)
        builds = self.get(builds_url)
```

juspence (commented):
Not finished looking yet, but need to submit this before I forget.

corgi/collectors/errata_tool.py
```diff
@@ -268,3 +267,42 @@ def get_erratum_notes(self, erratum_id: int) -> dict:
...
+        product_errata = self.get_paged(
+            f"api/v1/erratum/search?{search_values}",
+            page_data_attr="data",
+        )
+
+        all_errata: set[str] = set()
+        for erratum in product_errata:
+            logger.debug(f"Found shipped erratum {erratum['id']}")
+            is_rpm = "rpm" in erratum["content_types"]
+            builds = self.get(self._get_url_by_erratum_type(erratum["id"], is_rpm))
```
```suggestion
            url = self._get_url_by_erratum_type(erratum["id"], is_rpm)
            builds = self.get(url)
```

corgi/collectors/errata_tool.py
```diff
@@ -268,3 +267,42 @@ def get_erratum_notes(self, erratum_id: int) -> dict:
...
+            for product_version in builds.values():
+                for build in product_version["builds"]:
+                    for _, build_data in build.items():
```
```suggestion
                    for build_data in build.values():
```

corgi/tasks/management/commands/loaderratadata.py
```diff
@@ -58,44 +57,8 @@ def handle(self, *args, **options) -> None:
...
-                self.stderr.write(
-                    self.style.ERROR(f"Variants had more than 1 product id. Found: {product_ids}")
-                )
-                sys.exit(1)
-
-            product_id = product_ids.pop()
-            self.stdout.write(
-                self.style.SUCCESS(f"Searching shipped errata for product with id {product_id}")
-            )
-            # TODO switch this to use releases instead of Product
```
juspence, line 79: I see we didn't keep this comment in the new code. It doesn't have to be done in this PR, but just noting in case it was missed (or maybe you already tried it and it didn't work?)

corgi/core/migrations/0098_load_missing_container_errata_builds.py
```diff
@@ -0,0 +1,63 @@
...
+def load_missing_container_errata_build(apps, schema_editor):
+    ProductStream = apps.get_model("core", "ProductStream")
+    CollectorErrataProductVersion = apps.get_model("collectors", "CollectorErrataProductVersion")
+
+    variants_by_stream = defaultdict(set)
+    # Allow this migration to be re-run in case of timeout or other failure
+    # Only get the streams we haven't processed yet
+    for stream in (
+        ProductStream.objects.annotate(num_components=Count("components"))
+        .filter(num_components__gt=0)
```
juspence: Instead of using .annotate(), is it possible to just do .exclude(components__isnull=True) instead?

If not, and if counting all the components on all the streams causes issues, it might be easier to exclude the other streams first and do the .annotate() last. That should minimize the number of streams we have to check and annotate.

I remember .annotate() not working properly if it's not in the right place, so I'd like to avoid that (by not using it at all, if possible), but I also don't remember where "the right place" is. There's no need to change anything here if you've already tested this and found it works.

corgi/core/migrations/0098_load_missing_container_errata_builds.py
```diff
@@ -0,0 +1,63 @@
...
+        .exclude(tags__name=NO_MANIFEST_TAG)
+        .exclude(meta_attr__has_key="container_errata_checked")
+    ):
+        if stream.brew_tags:
+            for brew_tag in stream.brew_tags.keys():
```
juspence: Minor, but I think you can just do this in one line, since brew_tags defaults to an empty list:
```suggestion
        for brew_tag in stream.brew_tags:
```

corgi/core/migrations/0098_load_missing_container_errata_builds.py
```diff
@@ -0,0 +1,63 @@
...
+                trimmed_brew_tag, sans_container_released = brew_tag_to_et_prefixes(brew_tag)
+                variants = CollectorErrataProductVersion.objects.filter(
+                    brew_tags__overlap=[trimmed_brew_tag, sans_container_released]
+                ).values_list("variants__name", flat=True)
+                variants_by_stream[stream.name].update(variants)
+
+    et = ErrataTool()
+    for stream, variants in variants_by_stream.items():
```
juspence: You build a mapping of parent streams to their child variants in the for-loop above, then process each stream one at a time in the for-loop here.

Since we're already looping over the stream above, can't we combine the two loops and just:
1. Get the stream's list of variants
2. Get the list of errata_matching_variants using the stream's list of variants, as found in step 1 above

corgi/core/migrations/0098_load_missing_container_errata_builds.py
```diff
@@ -0,0 +1,63 @@
...
+        for errata in stream_container_errata:
+            current_slow_load_errata.delay(errata, force_process=True)
+        # Set these streams so they are not re-processed next run
+        checked_stream = ProductStream.objects.get(name=stream)
+        checked_stream.meta_attr["container_errata_checked"] = True
+        checked_stream.save()
+
+    # remove all the temporary container_errata_checked meta_attr keys
+    checked_streams = []
+    for stream in ProductStream.objects.filter(meta_attr__ship_checked=True):
```
juspence: Is there a typo here?
```suggestion
    for stream in ProductStream.objects.filter(meta_attr__container_errata_checked=True):
```

juspence (approved):
LGTM, most comments are minor and can be ignored I think.

corgi/tasks/tagging.py
jasinner: The '!' is excluding streams with the tag. So the search I proposed to use in SDEngine to get stream which should have a manfest is 'api/v1/product_streams?tags=!no_manifest'. 
The change you are commenting on here just pulls the string value "no_manifest" out to a constant.

corgi/collectors/errata_tool.py
jasinner: Yes, we are removing because it's unused. There is a clue in the commit message for this change.

corgi/tasks/management/commands/loaderratadata.py
jasinner: I pulled the data in from ET as part of pull/649 but didn't find it very useful for finding stream errata.

corgi/collectors/errata_tool.py
jasinner: ok

corgi/collectors/errata_tool.py
jasinner: Done

corgi/collectors/errata_tool.py
jasinner: fixed

tests/test_errata_data.py
jasinner: Done.

corgi/core/migrations/0098_load_missing_container_errata_builds.py
jasinner: Well, I have tested it and I know it works. But I also tested your suggestion and it works well, so I'll adopt it, thanks you. 
 I chose this query because it's the same one we use in the update_manifests task. Probably good to update it htere as well.

corgi/core/migrations/0098_load_missing_container_errata_builds.py
jasinner: Oh yeah, thanks. Anyway, I moved this to the exclude part of the query instead.

corgi/core/migrations/0098_load_missing_container_errata_builds.py
jasinner: I simplified the migration quite a bit since you left this comment, there is now only 1 loop.

corgi/core/migrations/0098_load_missing_container_errata_builds.py
jasinner: I have now removed this.

corgi/tasks/management/commands/loaderratadata.py
jasinner: Actually I'm now using the releases data for stream which don't match variants via brew_tags, eg rhes-3.5

jasinner:
I added a bunch of tests for the new slow_load_stream_errata task.

jasinner:
I ended up dropping the actual code change here to switch the endpoints based on type as the issue of missing builds described in CORGI-843 was fixed in Errata Tool. I kept the rest of the refactoring so that we could utilize the new slow_load_stream_errata task. I also kept the migration so that we could reload the data fixed in ET.

PR #644, opened 2023-10-18
Accept SPDX license strings in SBOMer components

kgrant-rh:
CycloneDX components can include either a list of individual licenses, or an SPDX license string. This change updates the SBOMer manifest parser to accept either.

### DIFF
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -65,10 +65,17 @@ class SbomerSbom:
             # License info
             licenses = []
             for _license in component.get("licenses", ()):
-                if _license["license"].get("id"):
-                    licenses.append(_license["license"].get("id"))
-                if _license["license"].get("name"):
-                    licenses.append(_license["license"].get("name"))
+                # Licenses may be either listed individually or
+                # as an SPDX license string
+                if "expression" in _license:  # SPDX string
+                    licenses.append(_license["expression"])
+                elif "license" in _license:  # Individual license
+                    if _license["license"].get("id"):
+                        licenses.append(_license["license"].get("id"))
+                    if _license["license"].get("name"):
+                        licenses.append(_license["license"].get("name"))
+                else:  # bomref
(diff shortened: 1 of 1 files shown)


corgi/collectors/pnc.py
```diff
@@ -65,10 +65,17 @@ def __init__(self, data: dict[str, Any]):
...
-                    licenses.append(_license["license"].get("name"))
+                # Licenses may be either listed individually or
+                # as an SPDX license string
+                if "expression" in _license:  # SPDX string
+                    licenses.append(_license["expression"])
+                elif "license" in _license:  # Individual license
+                    if _license["license"].get("id"):
+                        licenses.append(_license["license"].get("id"))
+                    if _license["license"].get("name"):
+                        licenses.append(_license["license"].get("name"))
```
juspence, line 76: How do we actually process the list of licenses we build here? Do we just join them all together with " OR " to make a longer SPDX expression out of each name (plus the existing expression, if any)? If yes, LGTM I think.

EDIT: Actually, oops, this is an existing bug so doesn't need to be fixed here. But aren't we collecting the same data twice? From the original list of licenses (license dicts), we get both of the "license" key's "name" subkey and the license key's "id" subkey for each license dict, on lines 73 and lines 75 I mean.

So I think we're storing the same license twice...like "GPLv2 and GNU-General-Public-License-V2". In the future we could use "elif" to prefer the ID first, then fall back to the name only if the ID isn't present. But again we don't need to worry about this now.

juspence (approved):

corgi/collectors/pnc.py
kgrant-rh: The list of licenses and expressions is joined with semicolons and stored via set_license_declared_safely. If " OR " is a better choice, I can use that instead.

juspence: I think let's leave it for now - we can always use " OR " instead of semicolons and store only one ID for each license (instead of two / name and ID) in the future. This is good enough on its own to go ahead and merge.

PR #645, opened 2023-10-19
Add ET Releases to Collector models

jasinner:
The main changes here are the 2nd and 3rd commits. The first of those commits adds the ET Release data to the Collector models. The other one persists most of the additional ET data not stored into the primary ET Collector model fields into a new meta_attr JSONField attribute.

This will be useful for the data migration in CORGI-843 and CORGI-737 for finding CPEs for some streams.

I also deleted one environment variant converted it to a test constant to make the developer setup experience slightly simpler.

This feature is blocked by CORGI-867 so I added a commit to fix that also.


jasinner:

jasinner:
I ran the load_variants() function in stage and it successfully fixed the issue described in CORGI-867

tox.ini
```diff
@@ -12,7 +12,6 @@ passenv =
     CORGI_DB_PORT
     # Internal hostnames or URLs that appear in build metadata; used in tests
     CORGI_APP_INTERFACE_URL
-    CORGI_TEST_DOWNLOAD_URL
```
juspence, line 15: I haven't looked in a while but I think this var might also be defined in Gitlab CI variables, and maybe other places too. Let's make sure we fix those as well.

corgi/collectors/models.py
```diff
@@ -26,6 +27,15 @@ class CollectorErrataProductVersion(CollectorErrataModel):
...
 
+class CollectorErrataRelease(CollectorErrataModel):
+    brew_tags = fields.ArrayField(models.CharField(max_length=1024), default=list)
+    product_versions = models.ManyToManyField(
+        CollectorErrataProductVersion, related_name="releases"
+    )
+    is_active = models.BooleanField(default=False)
+    enabled = models.BooleanField(default=False)
+
+
```
juspence, line 38: Sorry for the dumb questions, I don't know much about these. But could you briefly explain what a Release in ET actually is and how it gets used?

I also see both are linked to ET Product Versions - is it possible for one release to include many variants, and for one variant to be included in multiple different releases? Or are they completely unrelated?

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
+            # Allow variants to update their et_id, see CORGI-867
             et_product_variant, created = CollectorErrataProductVariant.objects.update_or_create(
-                et_id=variant["id"],
+                name=attributes.pop("name"),
                 defaults={
                     "cpe": cpe,
                     "product_version": et_product_version,
-                    "name": variant["attributes"]["name"],
+                    "et_id": variant["id"],
+                    "meta_attr": attributes,
```
juspence, line 158: LGTM, thanks for fixing this! Let's switch all the other update_or_create usages to use the unique name field instead of the ET ID.

I also wonder if we should delete and recreate all the Collector models in Corgi each day. Or maybe we need to handle models which are removed / renamed but keep the same ID / deleted and recreated in ET with the same name but different IDs / etc. in some other way.

Let's also add more tests for some of this logic.

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
+            attributes = variant["attributes"]
+            product_version = attributes["relationships"].pop("product_version")
+            product_version_id = product_version["id"]
             try:
                 et_product_version = CollectorErrataProductVersion.objects.get(
                     et_id=product_version_id
                 )
             except CollectorErrataProductVersion.DoesNotExist:
-                logger.warning("Did not find product version with id %s", product_version_id)
+                logger.warning(f"Did not find product version with id {product_version_id}")
```
juspence, line 144: Should this raise an error instead of just logging a warning?

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
 
+    def load_releases(self) -> None:
+        releases = self.get_paged("api/v1/releases", page_data_attr="data")
+        for release in releases:
+            attributes = release["attributes"]
+            name = attributes.pop("name")
+            relationships = release.pop("relationships")
+            release, created = CollectorErrataRelease.objects.update_or_create(
+                et_id=release.pop("id"),
+                name=name,
```
juspence, line 114: We should only use one field which is unique=True in update_or_create keyword arguments, or all the fields that appear together in a UniqueConstraint.

So I think name alone is enough and et_id can be part of the defaults like other methods. I know both et_id and name are technically unique, but they don't appear in a constraint together so I'm not sure if using both will cause some subtle issue.

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
                     defaults={"name": product_version["name"], "product": et_product},
```
juspence, line 84: Did you forget to add `"meta_attr": product_version` here, or was it left off on purpose? For example, if it only has two keys "id" and "name" that we already store elsewhere?

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
+            update = False
             if pv.brew_tags != brew_tags:
+                update = True
                 pv.brew_tags = brew_tags
+                logger.info(f"Updated Product Version {pv.name} Brew Tags to {pv.brew_tags}")
+            if pv.meta_attr != pv_detail_data:
+                update = True
+                pv.meta_attr = pv_detail_data
+                print(f"Updated Product Version {pv.name} meta_attr")
+            if update:
```
juspence, line 103: This will work, and I think this is safe if we only call this collector logic from a singleton Celery task once a day. But if any other code updates or manipulates the Collector models, let's put this inside a transaction.

Maybe it's even better to delete all the Collector models each day before we recreate them, so that we can clean up stale data that was removed from ET, and also handle updates in a slightly simpler way. Or maybe that will cause problems elsewhere, I'm not sure what links to these models or how they're used.

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
                 pv.save()
-                logger.info("Updated Product Version %s Brew Tags to %s", pv.name, pv.brew_tags)
 
+    def load_releases(self) -> None:
+        releases = self.get_paged("api/v1/releases", page_data_attr="data")
+        for release in releases:
+            attributes = release["attributes"]
+            name = attributes.pop("name")
+            relationships = release.pop("relationships")
```
juspence, line 111: Let's keep this relationships key a part of the attributes dict instead of popping it, but still pop off relationships["brew_tags"]. That way, other non-Brew tag relationships (if any) will still get saved into the meta_attr instead of being lost.

corgi/collectors/errata_tool.py
```diff
@@ -75,40 +84,82 @@ def load_et_products(self):
...
                 if created:
-                    logger.info("Created ET ProductVersion: %s", et_product_version.name)
+                    logger.info(f"Created ET ProductVersion: {et_product_version.name}")
+
+    def load_product_versions(self) -> None:
         for pv in CollectorErrataProductVersion.objects.get_queryset():
             pv_details = self.get(f"api/v1/products/{pv.product.et_id}/product_versions/{pv.et_id}")
-            brew_tags = [t.removesuffix("-candidate") for t in pv_details["data"]["brew_tags"]]
+            pv_detail_data = pv_details["data"]
+            brew_tags = [t.removesuffix("-candidate") for t in pv_detail_data.pop("brew_tags")]
```
juspence, line 93: Not related to this PR, but we fiddle with Brew tag suffixes in another part of the code too, I think...maybe the prod-defs collector?

Do we need to combine all that logic in one place, or should we strip even more suffixes here, like "-container-released" or whatever?

juspence (approved):
I think this looks good, but let's add a few more tests to make sure we handle e.g. models being updated (or recreated with different names or IDs) in ET.

corgi/collectors/errata_tool.py
jasinner: No, there are no additional details here, as these are just the product_versions from the product relationshipts. In the load_product_version function we set the meta_attr. In this function we save the product_versions and their relationships to products, then in load_product_versions we set the meta_attr and brew_tags which are not present in the ET products API endpoint.

corgi/collectors/errata_tool.py
jasinner: Good idea, I will update it to recreate the collector models each time.

corgi/collectors/errata_tool.py
jasinner: Perhaps just creating a constant for "-candidate" will give a maintainer a clue that the 2 functions both adjust the raw values? I think it makes sense to adjust the raw values in both function as we currently do. One is brew_tag data coming from ET, another is brew_tag data coming from prod_defs. 

You did make me realize that we also should probably remove the '-candidate' suffix from the new load_et_releases function as well, so that brew tag matching can behave the same between product models and collector models. I'll adjust that.

Brew tags in ET have '-candidate' tags because Engineers never attached released builds to an Errata. ET switches the brew tags to '-released' once they are attached to an Erratum and released. That's why the ET data and prod_defs data differ. Some products don't use the brew tag candidate and released suffixes at all.

corgi/collectors/models.py
jasinner: To be honest I don't know much about Releases. I don't know what their relationship to variants are, if any. Probably they are only related via product versions. I do know that:

- They can be used in a errata search endpoint as a way of grouping Errata
- They declare relationships on both product versions and brew_tags
- As described in CORGI-737 they can be used to override brew_tags defined in product versions when attached builds to an erratum.

tox.ini
jasinner: Yup, good idea. I'll remove the variable from Gitlab CI after we merge this.

corgi/collectors/errata_tool.py
jasinner: I'll try deleting and recreating the models.

corgi/collectors/errata_tool.py
jasinner: There are many of these warning logs for each run of the load_et_productions task today. Hopefully there will be less once we delete/recreate the models. If that's the case raising an error would make more sense.

corgi/collectors/errata_tool.py
jasinner: ok

corgi/collectors/errata_tool.py
jasinner: I updated the whole function to recreate the objects, so we no longer use update_or_create here.

jasinner:
I got some integrity errors when recreating the collector models, so I pushed to a different branch 'et-models-recreate'. I might abandon that approach.

juspence:
> I got some integrity errors when recreating the collector models, so I pushed to a different branch 'et-models-recreate'. I might abandon that approach.

I think it's fine to use the current approach if deleting and recreating doesn't work. We can look into it further in a future PR, it doesn't have to be done here if it's difficult or causing problems.

corgi/collectors/models.py
juspence: Ack, current code is fine as-is I think. I just wanted to check if we should add a link from Releass to Variants, so I think the current code is fine. We can always add those links in the future if needed later.

tox.ini
juspence: I just checked again and I see the .gitlab-ci.yml pipeline definition also uses this value. Once you remove that and merge this, I think there should be no other usages and deleting the variable in the remote Gitlab CI variables should be safe. Thanks!

jasinner:
Closed in favour of pull/649

jasinner:
> I think this looks good, but let's add a few more tests to make sure we handle e.g. models being updated (or recreated with different names or IDs) in ET.

I added a bunch of test for the new slow_load_stream_errata task.

PR #646, opened 2023-10-19
Handle external build system URLs

kgrant-rh:
Some external build system references don't include all the parameters that internal build system references include.

### DIFF
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -40,10 +40,15 @@ class SbomerSbom:
             # build info
             if "supplier" in component and component["supplier"].get("name") == "Red Hat":
                 component["namespace"] = Component.Namespace.REDHAT
+                # Build URL info for Red Hat internal build systems is in the format
+                # { "comment": "pnc-build-id",
+                #   "url": "https://pnc-url.redhat.com/pnc-rest/v2/builds/XXXXXXXX",
+                #   "type": "build-system" }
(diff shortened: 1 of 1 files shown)


corgi/collectors/pnc.py
```diff
@@ -40,10 +40,15 @@ def __init__(self, data: dict[str, Any]):
...
+                # Build URL info for Red Hat internal build systems is in the format
+                # { "comment": "pnc-build-id",
+                #   "url": "https://pnc-url.redhat.com/pnc-rest/v2/builds/XXXXXXXX",
+                #   "type": "build-system" }
+                # External build systems may be missing one or more of these parameters
                 build_urls = {
                     ref["comment"]: ref["url"]
                     for ref in component.get("externalReferences", {})
-                    if ref["type"] == "build-system"
+                    if ("comment" in ref and "url" in ref and ref["type"] == "build-system")
```
juspence, line 51: LGTM, only question is if we ever have a build-system ref we want to include, but which points to some non-RH system. For example, Fedora Koji. In that case, maybe we wouldn't have comment or url keys in the ref, but still want to include it somehow? 

I don't know what data an upstream build system ref would have / what it actually looks like, so skipping any upstream refs like that is fine. Maybe it's something we want to include in the future, or maybe that never happens and it's a moot point.

EDIT: Whoops, this is PNC-only code so disregard :smile: Only other nit might be updating an existing test to handle this, but definitely not a priority given we need to get this manifest out.

PR #647, opened 2023-10-20
Minor: set provides_name, sources_name, and upstreams_name filters to distinct

JimFuller-RedHat:
otherwise we get duplicate results which was an oversight on my part for CORGI-818.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -90,11 +90,11 @@ class ComponentFilter(FilterSet):
 
     # Normally we are interested in retrieving provides,sources or upstreams of a specific component
     sources = CharFilter(lookup_expr="purl")
-    sources_name = CharFilter(field_name="sources", lookup_expr="name")
+    sources_name = CharFilter(field_name="sources", lookup_expr="name", distinct=True)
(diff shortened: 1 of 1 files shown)


corgi/api/filters.py
```diff
@@ -90,11 +90,11 @@ class Meta:
 
     # Normally we are interested in retrieving provides,sources or upstreams of a specific component
     sources = CharFilter(lookup_expr="purl")
-    sources_name = CharFilter(field_name="sources", lookup_expr="name")
+    sources_name = CharFilter(field_name="sources", lookup_expr="name", distinct=True)
```
juspence, line 93: If we always want distinct results for the /components endpoint, can't we add .distinct() directly to the ComponentViewSet's queryset, in one place, instead of applying it to every filter in this file?

juspence (approved):

juspence (approved):
Let's add a test that confirms this fixes the issue, otherwise LGTM.

corgi/api/filters.py
JimFuller-RedHat: in a previous PR your comment not do this motivated this approach ;)

PR #648, opened 2023-10-20
Latest filter regressions

juspence:
@RedHatProductSecurity/corgi-devs Please review (probably commit-by-commit is easiest). There are several changes here:

1. Make sure that the ?ofuri= parameter in the API continues to work, even when the stream the user gave is inactive

2. Make the ofuri= parameter in the latest_components() queryset and manifest_components() queryset required, since we no longer support calling these without a stream

3. Reuse code between the latest_components() and latest_components_by_streams() querysets

4. Return the correct number of results in the latest_components() queryset when the include= parameter is set to True but there are no latest components

5. Fix usages of .using(), .order_by(), and .distinct() that could give incorrect results. DB routing and ordering should generally be applied in the API at the ViewSet level.

These parameters should not be defined as part of individual QuerySets, since they are reused in multiple places with different ordering / DB requirements. For example, read-only API code vs. writable API code and Celery task code require different DBs.

There are also some notes I've added to tests:
1. Unshipped components not part of any stream no longer appear in the "latest_components" queryset, even when the ofuri given is an empty "" string and the include_inactive_streams= parameter is set to True.

This may or may not impact OpenLCS and certain other users. If this is expected and not a bug, please let me know and I'll add a note to the CHANGELOG.md

2. We hide results or choose the wrong result (wrong 1 of N choices) in the API when two components have the same namespace and name. This can happen e.g. for source RPMs and their related RPM modules, which have two different type=RPM&arch=src / type=RPMMOD&arch=noarch type / arch pairs, respectively, for the same name.

In the previous latest filter code, we grouped by type, namespace, name, and arch first to avoid showing incorrect results, then picked the latest epoch + version + release from each group.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1179,27 +1179,31 @@ def test_latest_components_queryset(client, api_path, stored_proc):
         # A NEVRA like "PyYAML version 1.2.3 with no release or architecture"
         # could be a PyPI package, or an upstream RPM
         # We want to see the latest version of both components
-        # Create PyPI components for src architecture only
+        # Create PyPI components once per loop, like src architecture components
         # with the same namespace, name and version as UPSTREAM noarch components
         # and no release, arch, or software_build
+        # TODO: We should use RPMMOD instead of PyPI here, but then the test fails
+        #  because we reintroduced a bug we fixed previously
+        #  Whe hide results if two components have the same name and namespace
+        #  but different types or arches, for example a source RPM and its related RPM module
         older_unrelated_component = ComponentFactory(
-            type=Component.Type.PYPI,
-            namespace=older_upstream_component.namespace,
+            type=Component.Type.RPMMOD,
+            namespace=older_component.namespace,
             name=older_component.name,
             version=older_component.version,
-            release=older_upstream_component.release,
+            release=older_component.release,
             arch=older_upstream_component.arch,
-            software_build=older_upstream_component.software_build,
+            software_build=older_component.software_build,
         )
         older_unrelated_component.productstreams.add(stream)
         newer_unrelated_component = ComponentFactory(
             type=older_unrelated_component.type,
-            namespace=older_upstream_component.namespace,
+            namespace=newer_component.namespace,
             name=newer_component.name,
             version=newer_component.version,
-            release=older_upstream_component.release,
+            release=newer_component.release,
             arch=older_upstream_component.arch,
-            software_build=older_upstream_component.software_build,
+            software_build=newer_component.software_build,
         )
         newer_unrelated_component.productstreams.add(stream)
         components[(older_unrelated_component.type, name, older_unrelated_component.arch)] = (
@@ -1214,6 +1218,11 @@ def test_latest_components_queryset(client, api_path, stored_proc):
         ofuri=stream.ofuri,
         include=True,
     )
+    # Latest_components is now really "latest_root_components" for a particular stream
+    # We no longer show latest versions of non-root components
+    # or latest versions of unshipped components, which are not linked to any stream
+    # This may have implications for OpenLCS
+    # As well as other users of the ?latest_components_by_streams= filter / API parameter
     assert len(latest_components) == 2
     for component in latest_components:
         assert (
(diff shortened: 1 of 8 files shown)


tests/test_model.py
```diff
@@ -1179,27 +1179,31 @@ def test_latest_components_queryset(client, api_path, stored_proc):
         # A NEVRA like "PyYAML version 1.2.3 with no release or architecture"
         # could be a PyPI package, or an upstream RPM
         # We want to see the latest version of both components
-        # Create PyPI components for src architecture only
+        # Create PyPI components once per loop, like src architecture components
         # with the same namespace, name and version as UPSTREAM noarch components
         # and no release, arch, or software_build
+        # TODO: We should use RPMMOD instead of PyPI here, but then the test fails
```
jasinner, line 1185: Perhaps the simplest solution here to is fix CORGI-859 so that the latest filter no longer includes both the RPMMOD and it's SRPM.

jasinner:
I'll let @JimFuller-RedHat comment on the other suggestions here. I do think prefetch_related should be parameterized based on the ofuri model type as suggested by Justin.

JimFuller-RedHat (approved):
thx @juspence for getting to this - apart from the failing CI LGTM though will test first and make suggestions

JimFuller-RedHat (approved):
apart from CI failure, nothing I can add - LGTM

tests/test_model.py
```diff
@@ -1179,27 +1179,31 @@ def test_latest_components_queryset(client, api_path, stored_proc):
...
+        #  because we reintroduced a bug we fixed previously
+        #  Whe hide results if two components have the same name and namespace
+        #  but different types or arches, for example a source RPM and its related RPM module
```
juspence, line 1188: > apart from CI failure, nothing I can add - LGTM

@JimFuller-RedHat @jasinner The CI failure is expected, see this TODO comment in test_model.py which shows the issue. We previously tested for this bug, but after the changes to / new logic in the latest filter, the test is not correct anymore and always passes no matter what.

This PR only fixes the test to demonstrate the issue, I do not have a fix for the actual bug.

Fixing the test (to use a root component / RPMMOD type instead of PyPI type) shows we reintroduced this bug as part of the latest filter refactoring.

As noted in the PR description, to fix this bug, we must rewrite the latest filter to group by type, namespace, name, and arch. Grouping only by namespace and name is not enough.

See the query below for a real-world example. In stage,  we incorrectly get only 1 result (we group all component types together, then pick 1 latest firefox version from this list). In production, we correctly get 1 latest RPM version and 1 latest RPMMOD version for the "firefox" name (we group each type separately, then pick 1 latest firefox version from the separate list of versions for each type).

/api/v1/components?name=firefox&root_components=True&latest_components_by_streams=True&product_streams=o:redhat:rhel:8.8.0.z

I also found another bug. In stage, we get multiple results / versions for the same RPM name, even with the latest filter turned on. We still don't see any result for the module. We should have only 1 RPM result and 1 RPMMOD result, as seen in production:

/api/v1/components?name=libreoffice&root_components=True&latest_components_by_streams=True&product_streams=o:redhat:rhel:8.8.0.z

I could fix the first issue (RPMMOD result hides RPM result, or vice-versa) but I'm not sure what's causing the second issue. I think we should just revert this latest filter refactoring for now and continue working on this in a branch, until we fix all the bugs and correctly test for the same.

JimFuller-RedHat: ah right - 

today we group by name, namespace - I think just adding back **type** fixes - though I question why we need arch ?  I can do that work based on this PR (and raise a new PR).

As for multiple results - this is probably another form of cross join fun (which happens when django generates superfluous LEFT INNER JOINS which I discovered last week) - I thought I addressed all forms of it - will investigate.

JimFuller-RedHat: will add back arch as well ;)

tests/test_model.py
juspence: There is still a bug even if we don't treat RPMMOD components as roots. For example, a container, source RPM, and managed service Github repo, all with the same name, would be considered the same component. We would only show 1 "latest" version of a random type, and hide all the other types.

juspence:
Closing since Jim added this to his PR for CORGI-818. Thanks for fixing!

PR #649, opened 2023-10-23
Add ET Releases to Collector models and recreate models

jasinner:
Continuation of /pull/645, this time recreating the models on each run of the load_et_products task.


jasinner:
These changes where approved by @juspence in pull/645.

PR #651, opened 2023-10-24
CORGI-818: Add back type arch latest filters

JimFuller-RedHat:
have taken the good work of @juspence ( in https://github.com/RedHatProductSecurity/component-registry/pull/648) and enhanced stored proc with **type** and **arch** to avoid collision - we can generalise this filter more as we can now narrow down with _active_streams=True_

**Note**: I could just as easily add [0f85669](https://github.com/RedHatProductSecurity/component-registry/pull/651/commits/0f85669222e01bba64a11fd72d8e2d2711eac3e5) to https://github.com/RedHatProductSecurity/component-registry/pull/648

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -907,15 +907,17 @@ class ComponentQuerySet(models.QuerySet):
         components: "ComponentQuerySet", model_type: str, ofuri: str, include_inactive_streams: bool
     ) -> Iterable[str]:
         return (
-            components.values("name", "namespace")
+            components.values("type", "namespace", "name", "arch")
             .order_by()  # required to avoid cross-join fun
             .distinct()
             .annotate(
                 latest_version=Func(
                     Value(model_type),
                     Value(ofuri),
+                    F("type"),
                     F("namespace"),
                     F("name"),
+                    F("arch"),
                     Value(include_inactive_streams),
                     function="get_latest_component",
                     output_field=models.UUIDField(),
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -2,7 +2,7 @@ import logging
 import re
 import uuid as uuid
 from abc import abstractmethod
-from typing import Any, Iterator, Union
+from typing import Any, Iterable, Iterator, Union
 
 from django.apps import apps
 from django.conf import settings
@@ -902,35 +902,14 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
 class ComponentQuerySet(models.QuerySet):
     """Helper methods to filter QuerySets of Components"""
 
-    def latest_components(
-        self,
-        model_type: str = "ProductStream",
-        ofuri: str = "",
-        include: bool = True,
-        include_inactive_streams=False,
-        include_non_root_components=False,
-    ) -> "ComponentQuerySet":
-        """Return an ofuri's components from latest builds across all product streams."""
-
-        # the concept of 'latest component' is only relevent within product boundaries
-        # we want to constrain by product type/ofuri which is why we pass in model_type and ofuri
-        # into the get_latest_component stored proc annotation
-
-        components = self
-        if not (include_non_root_components):
-            components = (
-                components.select_related("software_build")
-                .prefetch_related("productstreams")
-                .root_components()
-                .using("read_only")
-            )
-        if ofuri:
-            components = components.filter(productstreams__ofuri=ofuri).distinct()
-
-        latest_components_uuids = set(
+    @staticmethod
+    def _latest_components_func(
+        components: "ComponentQuerySet", model_type: str, ofuri: str, include_inactive_streams: bool
+    ) -> Iterable[str]:
+        return (
             components.values("name", "namespace")
-            .order_by("name")  # required to avoid cross-join fun
-            .distinct("name", "namespace")
+            .order_by()  # required to avoid cross-join fun
+            .distinct()
             .annotate(
                 latest_version=Func(
                     Value(model_type),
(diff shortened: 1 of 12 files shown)


corgi/core/models.py
```diff
@@ -943,102 +922,97 @@ def latest_components(
...
+        """Return components from latest builds in a single stream."""
+
+        # the concept of 'latest component' is only relevant within product boundaries
+        # we want to constrain by product type/ofuri which is why we pass in model_type and ofuri
+        # into the get_latest_component stored proc annotation
+
+        components = (
+            self.root_components()
+            .select_related("software_build")
+            # TODO: What if model_type isn't ProductStream?
```
jasinner: Should we remove this TODO statement is prefetching productstreams is the correct and only model to prefetch here?

corgi/core/models.py
```diff
@@ -1089,16 +1063,9 @@ def manifest_components(self, ofuri: str = "", quick=False) -> "ComponentQuerySe
         if not quick:
```
jasinner, line 1071: I guess we should remove the 'quick' function parameter as it no longer has any affect.

tests/test_model.py
```diff
@@ -1179,27 +1179,31 @@ def test_latest_components_queryset(client, api_path, stored_proc):
...
         # could be a PyPI package, or an upstream RPM
         # We want to see the latest version of both components
-        # Create PyPI components for src architecture only
+        # Create PyPI components once per loop, like src architecture components
         # with the same namespace, name and version as UPSTREAM noarch components
         # and no release, arch, or software_build
+        # TODO: We should use RPMMOD instead of PyPI here, but then the test fails
+        #  because we reintroduced a bug we fixed previously
+        #  Whe hide results if two components have the same name and namespace
+        #  but different types or arches, for example a source RPM and its related RPM module
```
jasinner: These comments need fixing.

```suggestion
        # Create RPMMOD components once per loop, like src architecture components
        # with the same namespace, name and version as UPSTREAM noarch components
        # and no release, arch, or software_build. Otherwise we hide results if two 
        # components have the same name and namespace but different types or arches, 
        # for example a source RPM and its related RPM module
```

tests/test_model.py
```diff
@@ -1214,7 +1218,12 @@ def test_latest_components_queryset(client, api_path, stored_proc):
         ofuri=stream.ofuri,
         include=True,
     )
-    assert len(latest_components) == 2
+    # Latest_components is now really "latest_root_components" for a particular stream
+    # We no longer show latest versions of non-root components
+    # or latest versions of unshipped components, which are not linked to any stream
+    # This may have implications for OpenLCS
+    # As well as other users of the ?latest_components_by_streams= filter / API parameter
```
jasinner: I think we can remove this comment.

jasinner (approved):

jasinner:
My feedback here is minor. Happy to merge this.

corgi/core/models.py
JimFuller-RedHat: dynamically set prefetch based on model_type now

corgi/core/models.py
JimFuller-RedHat: Want to avoid adding any more potential blast radius to this PR ... something for the (near) future.

JimFuller-RedHat:
as per conversation with @jasinner have merged and we are always free to consider more changes in future PRs

tests/test_database.py
```diff
@@ -105,7 +105,7 @@ def test_get_latest_component_stored_proc(stored_proc):
 
     with connection.cursor() as cursor:
         cursor.execute(
-            "select * from get_latest_component('ProductStream','o:redhat:openshift-enterprise:3.11.z','REDHAT','ansible-runner',True);"  # noqa: E501
+            "select * from get_latest_component('ProductStream','o:redhat:openshift-enterprise:3.11.z','RPM','REDHAT','ansible-runner','src',True);"  # noqa: E501
```
juspence, line 108: Can we get rid of the `# noqa: E501` here, and just fix whatever error that represents?

corgi/core/models.py
```diff
@@ -902,143 +902,125 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
             .values_list("latest_version", flat=True)
-            .using("read_only")
+        )
+
+    def latest_components(
+        self,
+        ofuri: str,
+        model_type: str = "ProductStream",
+        include: bool = True,
+        include_inactive_streams=False,
```
juspence, line 934:
```suggestion
        include_inactive_streams: bool = False,
```

corgi/core/models.py
```diff
@@ -902,143 +902,125 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+        # we want to constrain by product type/ofuri which is why we pass in model_type and ofuri
+        # into the get_latest_component stored proc annotation
+
+        product_prefetch = "productstreams"
+        if model_type == "Product":
+            product_prefetch = "products"
+        if model_type == "ProductVersion":
+            product_prefetch = "productversions"
+        if model_type == "ProductVariant":
+            product_prefetch = "productvariants"
```
juspence, line 948:
```suggestion
product_prefetch = f"{model_type.lower()}s"

File: corgi/core/models.py [hunk 8]
```diff
@@ -902,143 +902,125 @@ def get_product_details(variant_names: tuple[str], stream_names: list[str]) -> d
...
+        components = (
+            self.root_components()
+            .select_related("software_build")
+            .prefetch_related(product_prefetch)
+            .filter(productstreams__ofuri=ofuri)
+        )
```
juspence, line 954: We always filter on productstreams__ofuri here, even if the user is looking at a different model.
```suggestion
        lookup = {f"{product_prefetch}__ofuri": ofuri}
        components = (
            self.root_components()
            .select_related("software_build")
            .prefetch_related(product_prefetch)
            .filter(**lookup)
        )

juspence (commented):
These fixes LGTM, sorry I didn't have time to get to this before.

I still see a few issues which can be handled in a follow-up PR. But the bigger issue is we still have multiple "latest" results even for a single SRPM in a single stream:

/api/v1/components?name=libreoffice&type=RPM&arch=src&latest_components_by_streams=True&product_streams=o:redhat:rhel:8.8.0.z

We get only 1 result in prod like we expect, but there are 5 results in stage.

JimFuller-RedHat:
thx ! will test & raise PR in my morning

PR #652, opened 2023-10-24
CORGI-557: Fix duplicate RPMs which may have bad namespace, arch, NEVRA, or purl

juspence:
@RedHatProductSecurity/corgi-devs Please review. The modules and containers have finished being cleaned up in stage, so I've moved onto RPMs in the first migration here:

1. There are 40,000-ish "noarch" binary RPMs which have their namespace set to "REDHAT" correctly, but are missing "/redhat/" in their purls.
2. There are 40,000-ish binary RPMs with an empty "" string arch. This was an old bug in our ingestion code which is now fixed. They need to have their arch changed to "noarch", then the NEVRA and purl must be updated to match.
3. Almost all of the components in point 2 also have their namespace set to "REDHAT" correctly, but are missing "/redhat/" in their purls.

The components in point 2 are likely duplicates of the components in the point 1. If this is correct (the migration checks), we just delete them instead of trying to merge their descendant nodes. Merging the data for RPMs is more complicated for reasons noted in the migration, and I've spent enough time on this already.

The second migration here fixes duplicate nodes for remote-source components, for example MAVEN, NPM, and PYPI packages. Every single remote-source component type has nodes where the purl doesn't match the linked component. The migration looks to see if there is a duplicate node (same type and parent, purl matches linked component like we expect). If a duplicate is present, we just delete the bad node. Otherwise, we update the bad node so that the purl matches the component.

Again, we don't check the descendant counts or do any reprocessing here if the bad node has a duplicate. The logic to do this correctly would be even more complicated than in the previous PRs. Any descendants we lose are either legacy data from a bug we've already fixed, so we don't want to keep them around anyway (e.g. test dependencies from Syft). Or the missing descendants will get recreated the next time we run SCA on a root that uses this remote-source component.

I still need to recheck the exact numbers of components that will be affected, and try to identify why the purls don't match, but in the meantime this is ready for review. And I hope this is the last PR for this issue.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,12 +1,9 @@
 import logging
 
 import pytest
-from rest_framework.request import Request
-from rest_framework.test import APIRequestFactory
 from rest_framework.viewsets import GenericViewSet
 
 from corgi.api import views
-from tests.factories import ProductFactory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


tests/test_views.py
jasinner, line 1: Thanks for fixing this!

jasinner (approved):
Seems too to me. Thanks for writing these migrations.

juspence:
@jasinner The performance tests are still failing due to recent latest filter changes, but I'm going to merge this anyway so I can start running the migrations here. I hope it will be finished by the end of my day on Friday. If not and this causes issues for you Monday, please feel free to back out the migrations if needed. I can always run them manually as a script.

I also added the same changes you already tested to the 0098 migration. That already completed in stage, but failed for you locally due to a last-minute change that had a small glitch. It should run fine in prod now (it will take a long time but will eventually complete).

PR #655, opened 2023-10-26
CORGI-872: Report CPEs for root components in manifests

juspence:
@RedHatProductSecurity/corgi-devs Please review. This is a farily simple change to start reporting CPEs for root components in product-based manifests, and the component being manifested in component-based manifests.

For the latter case, "the component being manifested" may not actually be a root component, e.g. we support manifesting arch-specific containers and binary RPMs. As long as they "provide" some other component, we can report useful data about them in a manifest, so I made sure we support reporting their CPEs as well.

Given above, we can start reporting CPEs for all components in a manifest (provides and upstreams) if we want to, although for now I only added this logic to root components.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1815,8 +1815,26 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
         return ComponentNode.objects.filter(pk__in=root_node_pks).using(using)
 
     @property
-    def build_meta(self):
-        return self.software_build.meta_attr
+    def cpes(self) -> QuerySet:
+        """Build and return a list of CPEs from all Variants this Component relates to"""
+        # For each Variant-type relation, get the linked Variant's CPE directly
+        # Remove any duplicates, and return the CPEs in sorted order so manifests are stable
+        return (
+            self.productvariants.exclude(cpe="")
+            .values_list("cpe", flat=True)
+            .distinct()
+            .order_by("cpe")
+        )
+
+        # For Stream-type relations, we no longer link any Variants
+        # since this caused data quality issues, where Components were linked to Variants
+        # that didn't actually ship those Components
+        # We could look at all the linked streams to get CPEs from all their child Variants
+        # But this would cause the same problem - we'd report too many Variants / incorrect CPEs
+
+        # TODO: We return an empty QuerySet if no Variants are linked, or none have CPEs
+        #  But do we want to raise an error / make manifesting fail,
+        #  whenever we can't find CPEs for some root component?
 
     def _build_download_url_for_type(self) -> str:
         """Get a source code or binary download URL based on a purl"""
--- a/corgi/web/templates/component_manifest.json
+++ b/corgi/web/templates/component_manifest.json
@@ -45,7 +45,12 @@
     {
       "copyrightText": {% if obj.copyright_text %}"{{obj.copyright_text|escapejs}}"{% else %}"NOASSERTION"{% endif %},
       "downloadLocation": {% if obj.download_url %}"{{obj.download_url|escapejs}}"{% else %}"NOASSERTION"{% endif %},
-      "externalRefs": [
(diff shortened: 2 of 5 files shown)


corgi/core/models.py
```diff
@@ -1815,8 +1815,26 @@ def get_roots(self, using: str = "read_only") -> QuerySet[ComponentNode]:
         return ComponentNode.objects.filter(pk__in=root_node_pks).using(using)
 
     @property
-    def build_meta(self):
-        return self.software_build.meta_attr
```
juspence, line 1819: FYI this property is unused today.

corgi/core/models.py
```diff
@@ -1815,8 +1815,26 @@ def get_roots(self, using: str = "read_only") -> QuerySet[ComponentNode]:
...
+
+        # For Stream-type relations, we no longer link any Variants
+        # since this caused data quality issues, where Components were linked to Variants
+        # that didn't actually ship those Components
+        # We could look at all the linked streams to get CPEs from all their child Variants
+        # But this would cause the same problem - we'd report too many Variants / incorrect CPEs
+
+        # TODO: We return an empty QuerySet if no Variants are linked, or none have CPEs
+        #  But do we want to raise an error / make manifesting fail,
+        #  whenever we can't find CPEs for some root component?
```
juspence, line 1837: Looking for opinions here.

jasinner: We can use the relations table to look up errata type relations, get the product_ref and then CPE from the collector models for stream-type relations. 

I think we should sanity check that the CPE matches one identified for the stream though.

jasinner (approved):

corgi/core/models.py
juspence: Looks like you edited your comment - in email I see it says "just skip adding the CPE".

The problem is not finding the CPEs for streams. I actually had code originally that looked up the productstreams linked to this component, and fetched CPEs from each of their child variants. But after I wrote that code, I realized it was very fiddly and made the test much more complicated.

Using all the CPEs from all the child variants of the linked stream is also a little incorrect I think. For variant-type relations, we link exactly one variant to this component, so we know exactly which Variants ship the component and what the right CPEs are.

For stream-type relations, currently we only link the stream itself to the component. Previsouly, we used to link the stream and all the child variants to the component. That caused issues where we reported components being included in variants that didn't actually ship them.

Reporting "all the child variant CPEs of a linked stream" would cause the same issue right? We'd report incorrect CPEs / variants which don't actually ship the component I think. It seemed error-prone, so I just left that out for now. The comment just above this diff window describes that decision / the same stuff I wrote here.

Replying to your updated comment, you suggest using the relations table and then the Collector model CPEs. But using the relations table means this only works for root components, and you recommended keeping our options open for CORGI-857. There are also some cases where we manifest a non-root component (see the PR description), and I'd like to report CPEs for "the component we're manifesting" even in those cases as well.

I actually wrote code to query the relations table at first, but it required more queries than I was expecting and using component.productvariants (and optionally component.productstreams too) keeps the implementation much simpler.

I'm also not sure about using the collector models instead of the ProductVariant models directly. Is there a reason to prefer one over the other? Right now I think we handle changes / model deletions in both prod-defs and ET models equally well, so neither should have issues with reporting stale data I hope. Both models have a "cpe" property we can refer to.

I think we're only creating ProductVariants from errata_info now (not from brew_tags as well) so shouldn't the list of CPEs from Variants be the same as the list of CPEs from collector models?

Finally, you say "sanity-check the CPE (from the collector model) matches one identified for the stream". Only Variants have CPEs right? Or do you mean using the product_stream.cpes property? I also had code that used this, but it actually requires lots of DB queries and I think the current code is simpler. And you mention errata-type relations, but those are only for variants.

juspence:
@jasinner I see you approved so I'm going to merge this as-is, but please let me know if I should make some changes here.

PR #658, opened 2023-10-27
CORGI-871, CORGI-879: Fix sbomer code and historical data

juspence:
@RedHatProductSecurity/corgi-devs Please review. This should:

1. Ensure we always create valid SPDX license expressions (CORGI-871)
2. Fix historical data for quarkus-bom root components
3. Make sure new quarkus-bom root components are Maven-type
4. Stop adding a ?repository_url= qualifier to Red Hat Maven component purls
5. Make certain errors a little more visible and clean up some code

I hope that 2, 3, and 4 are enough to keep Corgi purls identical to SBOMer purls. That should allow SDEngine to find quarkus-bom root components / new Quarkus versions in Corgi, using just the SBOMer purl as given in the Notes field of some SHIPPED_LIVE erratum. Then SDEngine doesn't have to write any complex code to find which quarkus-bom component / Quarkus version needs to get manifested.

For number 4, I didn't add a migration to fix all the existing Maven-type purls. Thanks to the bugfix in COGI-557, these should get automatically updated the next time we call save() on the component, AND the purls on the nodes will be updated as well / data will not get out of sync. Updating both the component and node purls in a migration is very fiddly, but I'm happy to write one if people prefer.

I'm definitely not as familiar with this part of the code, so please let me know if I've missed something or a change here doesn't look quite right. Thank you!

### DIFF
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -75,12 +75,19 @@ class SbomerSbom:
                 if "expression" in _license:  # SPDX string
                     licenses.append(_license["expression"])
                 elif "license" in _license:  # Individual license
+                    # Prefer license IDs, only fallback to names if no ID is available
                     if _license["license"].get("id"):
                         licenses.append(_license["license"].get("id"))
-                    if _license["license"].get("name"):
+                    elif _license["license"].get("name"):
                         licenses.append(_license["license"].get("name"))
+                    else:
+                        raise NotImplementedError(
+                            "Unsupported license type (no ID or name) in SBOMer manifest"
+                        )
                 else:  # bomref
-                    raise NotImplementedError("Unsupported license type in SBOMer manifest")
+                    raise NotImplementedError(
+                        "Unsupported license type (bomref) in SBOMer manifest"
+                    )
 
             component["licenses"] = licenses
 
--- a/corgi/core/constants.py
+++ b/corgi/core/constants.py
@@ -17,7 +17,7 @@ CORGI_COMPONENT_TAXONOMY_VERSION = "v1"
 # There's also an Early Access repo: https://maven.repository.redhat.com/earlyaccess/all
 # But anything in there should become GA eventually
 # and we don't want the purls to change over time / we don't know when GA will happen
-RED_HAT_MAVEN_REPOSITORY = "https://maven.repository.redhat.com/ga"
+RED_HAT_MAVEN_REPOSITORY = "https://maven.repository.redhat.com/ga/"
 
 # Map MPTT node levels in our product taxonomy to model names as defined in models.py
 NODE_LEVEL_MODEL_MAPPING = {
--- a/corgi/core/migrations/0100_fix_duplicate_sbomer_components.py
+++ b/corgi/core/migrations/0100_fix_duplicate_sbomer_components.py
@@ -4,6 +4,11 @@ from django.db.models import F, Value, functions
 from corgi.core.constants import RED_HAT_MAVEN_REPOSITORY
 
 LICENSE_DECLARED_RAW_FIELD = F("license_declared_raw")
+PURL_FIELD = F("purl")
+
+# Identical except for a trailing / slash we just added
+RED_HAT_MAVEN_REPOSITORY_OLD = "https://maven.repository.redhat.com/ga"
+
 SEMICOLON_VALUE = Value(";")
 SPDX_OR_VALUE = Value(" OR ")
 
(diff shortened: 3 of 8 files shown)


corgi/core/migrations/0102_fix_duplicate_sbomer_components.py
```diff
@@ -0,0 +1,67 @@
...
+        if "type" not in component.meta_attr:
+            component.meta_attr["type"] = "pom"
+        if "group_id" not in component.meta_attr:
+            component.meta_attr["group_id"] = "com.redhat.quarkus.platform"
+        # TODO: Other qualifiers or meta_attr values?
+
+        bad_purl = component.purl
+        good_purl = bad_purl.replace("pkg:generic/", "pkg:maven/com.redhat.quarkus.platform/", 1)
+
+        if "type=" not in good_purl:
```
kgrant-rh: Maybe it would be worth using PackageURL here to normalize the qualifiers here.

corgi/core/migrations/0102_fix_duplicate_sbomer_components.py
```diff
@@ -0,0 +1,67 @@
...
+            component.license_declared_raw = component.license_declared_raw.replace(";", " OR ")
+
+        if "type" not in component.meta_attr:
+            component.meta_attr["type"] = "pom"
+        if "group_id" not in component.meta_attr:
+            component.meta_attr["group_id"] = "com.redhat.quarkus.platform"
+        # TODO: Other qualifiers or meta_attr values?
+
+        bad_purl = component.purl
+        good_purl = bad_purl.replace("pkg:generic/", "pkg:maven/com.redhat.quarkus.platform/", 1)
```
kgrant-rh: All of these components should have purl_declared in meta_attr, if you want to sanity check against that.

kgrant-rh:
Looks good to me. Thanks for fixing the components that were already processed.

kgrant-rh (approved):

jasinner (commented):
I'm against taking repository_url out of the Maven purls. If we are trying to following the specification it should be included. Without that, users have to assume the components can be found in Maven central when they can not. Perhaps we could add the repository_url back when we render these components in an SBOM? Alternatively I think it's better to store them with repostitory_url and have SDEngine remove that for matching to ET data.

juspence:
> I'm against taking repository_url out of the Maven purls. If we are trying to following the specification it should be included. Without that, users have to assume the components can be found in Maven central when they can not. Perhaps we could add the repository_url back when we render these components in an SBOM? Alternatively I think it's better to store them with repostitory_url and have SDEngine remove that for matching to ET data.

@jasinner Ack, I don't necessarily want to have different purls between Corgi's API and Corgi's manifests, particularly for only one type of component. But @mprpic is the one who will have to do the SDEngine-side work if we don't handle it in Corgi somehow.

As mentioned in Gchat, SDEngine will have to track the purl as well as "the purl without qualifiers", then use the ?re_purl= API parameter / filter to look up a matching component in Corgi. I guess that "purl without qualifiers" should be regex-escaped, so that `?re_purl=pkg:rpm/name@version.el9` looks up a literal prefix, instead of trying to do regex matches with the . dot in the .el9 release part.

Please let me know once we've decided on an approach, and I'll either merge this as-is (purls will match exactly, no SDEngine work needed), or I'll fix the migration here to keep the repository_url (and SDEngine will need some work).

juspence:
Based on discussions with middleware (in Gchat), they've updated the purls the cve-mapping tool generates. So Corgi can keep the ?repository_url= qualifier in the purls we generate, purls will still match between Corgi and Errata Tool, purls will be stable and not change over time, and SDEngine shouldn't need to make any changes I think.

I'm going to fix the migration in this PR to match what I wrote above, then I'll merge. Please let me know if I missed something and I'll open follow-up PRs as needed.

corgi/core/migrations/0102_fix_duplicate_sbomer_components.py
juspence: Ideally I'd reuse the Component model's _build_maven_purl() function, but I think custom model code isn't available in migrations so I'd be reimplementing that logic here, then calling the PackageURL library to turn the dict we'd build into a purl. I probably should have done that anyway if I'd thought about it, but I did check the data and the values we have to transform are very limited in this case.

For quarkus-bom components, we only have to worry about the type= qualifier from SBOMer and the repository_url= qualifier from Corgi. There's no classifier= qualifier for any of these components, so we know exactly what to add where (in alphabetical order / type= last) and the surgery is simple enough to write here.

corgi/core/migrations/0102_fix_duplicate_sbomer_components.py
juspence: Good idea, I'll manually check the data after the migration runs (and do follow-up PRs to fix stuff if needed). I guess they should be identical except for the ?repository_url= qualifier we're adding. I'd like to merge this before the sprint ends, so checking as part of the migration was a missed opportunity I guess. Sorry :frowning:

jasinner (approved):
Looks good to me now, thanks.

PR #664, opened 2023-11-02
CORGI-678 pyxis image data

jasinner:

Lookup container build repository names from Pyxis and use them as the component name and repository url. If an existing container build with the same name label is looked up a second time, use the repository name from the previous build of the package so save time processing the build. Only block the slow_fetch_brew_build task when we encounter a new name label.

Please review commit wise.


jasinner:
@RedHatProductSecurity/corgi-devs I added 2 commits here to handle the case where a container build we first fetch from Brew is for a previously unseen name label and therefore repository. Later when that build is pushed by ET we would not have updated it's name to the repository name. Now we handle that situation.

corgi/collectors/brew.py
```diff
@@ -451,10 +451,12 @@ def _create_image_component(
...
         }
         if name_label:
-            image_component["meta"]["repository_url"] = f"{CONTAINER_REPOSITORY}/{name_label}"
+            image_component["meta"]["name_from_label"] = name_label
             name_label_parts = name_label.rsplit("/", 1)
             if len(name_label_parts) == 2:
-                image_component["meta"]["name_from_label"] = name_label_parts[1]
+                image_component["meta"]["repository_url"] = f"{CONTAINER_REPOSITORY}/{name_label}"
+                image_component["meta"]["name"] = name_label_parts[1]
+
```
juspence: I'm not sure about this change to the `name_from_label` field. Unless we go back and reprocess all containers, this is going to create inconsistent results in our data. Newer containers will have the `name` field with the last part of the combined name and label, but older containers will not have any `name` field.

Newer containers will always have the `name_from_label` field, and store the original value before any processing. Older containers will always have the `name_from_label` field, but store only the last part of the combined name and label. The original value is lost and no longer present anywhere (I think). So we'd have to handle two different data formats for the same field.

Maybe it's better to keep the `name_from_label` field logic unchanged (store just the last part), and add the combined name and label to a new `name_and_label` field. Then your migration can check for the new `name_and_label` field, and reprocess only the containers which don't have this set.

corgi/collectors/brew.py
```diff
@@ -588,8 +590,8 @@ def get_container_build_data(self, build_id: int, build_info: dict[str, Any]) ->
 
         # During collection we are only able to inspect docker config labels on
         # attached arch specific archives. We do this loop here to save the description, license,
-        # name label, and repository url also on the index container object at the root of the tree.
-        for attr in ("description", "license", "name_from_label", "repository_url"):
+        # also on the index container object at the root of the tree.
+        for attr in ("name", "description", "license", "name_from_label", "repository_url"):
```
juspence: Looks like a typo here, where you accidentally removed part of the comment? Or maybe that was on purpose, but you forgot to change the code to match?
```suggestion
        # name, label, and repository_url also on the index container object at the root of the tree.
        for attr in ("name", "description", "license", "name_from_label", "repository_url"):
```
If you make my suggested change above, I guess `name` should become `name_and_label`.

corgi/core/models.py
```diff
@@ -1339,7 +1334,7 @@ def _build_container_purl(self) -> dict[str, Any]:
         # Note that if no container digest format matched, the digest below is ""
         # and the constructed purl has no .version attribute
         return dict(
-            name=purl_name,
+            name=self.name,
```
juspence, line 1337: Maybe this will cause issues, I'm leaving this here as a TODO reference comment for myself during review.

tests/test_model.py
```diff
@@ -262,7 +262,9 @@ def test_component_model():
 
 def test_container_purl():
     # TODO: Failed due to "assert arch not in purl"
-    container = ContainerImageComponentFactory()
+    # With the changes introduced in CORGI-678 the container name should be equal to the last part
+    # of the repository_url
+    container = ContainerImageComponentFactory(name="node-exporter-rhel8")
```
juspence: To make this a little more clear, I would do below:
```suggestion
    repo_name = "node-exporter-rhel8"
    # With the changes introduced in CORGI-678 the container name should be equal to the last part
    # of the repository_url
    container = ContainerImageComponentFactory(name=repo_name)
```
And then remove the repo_name redefinition from the old line 275 / new line 277 below.

tests/test_model.py
```diff
@@ -277,7 +279,6 @@ def test_container_purl():
     container.meta_attr = {
         "digests": {CONTAINER_DIGEST_FORMATS[0]: example_digest},
         "repository_url": repository_url,
-        "name_from_label": repo_name,
```
juspence, line 280: Now we're setting `name_from_label` to the entire `name_label_parts`, right? And we have a new `name` field that's just the last `name_label_parts[1]` value.

I don't think we're using the meta_attr anywhere in this test directly, but the old code would do different things depending on these values. I think we should still set them to some "unused" string, and confirm they no longer affect the purls we generate. In other words, let's confirm the old code really is gone, and the new code only uses the container name directly.

corgi/tasks/pyxis.py
```diff
@@ -58,7 +96,6 @@ def _slow_fetch_pyxis_manifest_for_repository(
 ) -> bool:
 
     # According to the PURL spec, the name part is the last fragment of the repository url
```
juspence: Ah, I think I forgot to make a small doc update. When I saw "fragment" written here, I thought it was referring to URL fragments like `/path/to/page#fragment_link_to_some_heading. Maybe below is more clear:
```suggestion
    # According to the PURL spec, the name part is the last slash-separated path part of the repository url
```

corgi/tasks/errata_tool.py
```diff
@@ -108,6 +109,17 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
+            # container builds can get a new name. Make sure we check for that here and update the
+            # container name if so.
+            if is_container:
+                build = SoftwareBuild.objects.filter(
+                    build_id=build_id, build_type=SoftwareBuild.Type.BREW
+                ).first()
+                if build:
+                    # Let this propagate a error
+                    nvr = build.meta_attr.get("nvr")
+                    slow_update_name_for_container_from_pyxis.delay(nvr)
```
juspence: This won't fail the slow_load_errata task if the NVR key is missing. Below will:
```suggestion
                    # Let this propagate a error
                    nvr = build.meta_attr["nvr"]
                    slow_update_name_for_container_from_pyxis.delay(nvr)
```
Or if you want the slow_load_errata task to succeed, and only the scheduled task should fail, maybe we should says that: "Let this task succeed, but fail the scheduled task if the build has no NVR"

corgi/core/migrations/0103_get_latest_repo_names_from_pyxis.py
```diff
@@ -0,0 +1,29 @@
...
+def get_latest_repo_names_from_pyxis(apps, schema_editor) -> None:
+    brew_container_nvrs = (
+        Component.objects.filter(
+            type=Component.Type.CONTAINER_IMAGE, software_build__build_type=SoftwareBuild.Type.BREW
+        )
+        .values_list("nvr", flat=True)
+        .iterator()
+    )
+    for nvr in brew_container_nvrs:
+        slow_update_name_for_container_from_pyxis.delay(nvr)
```
juspence: I don't think this will completely update all containers. Only the root component / noarch container is linked to the build, so only that will get updated. Maybe below is what you wanted:
```suggestion
    Component = apps.get_model("core", "Component")
    brew_index_containers = Component.objects.filter(
        type="OCI", arch="noarch", meta_attr__name_and_label__isnull=True
    ).select_related("software_build").filter(
        software_buid__build_type="BREW"
    ).iterator()
    for index_container in brew_index_containers:
        slow_update_name_for_container_from_pyxis.delay(index_container.nvr)
        for child_container_nvr in brew_index_containers.provides.filter(type="OCI").values_list("nvr", flat=True).iterator():
            slow_update_name_for_container_from_pyxis.delay(child_container_nvr)
```
We could avoid scheduling a lot of duplicate tasks, if the lookup task handles this last piece of logic for us. In other words, look up each index container only once, and don't look up all the child containers as well. Then once we find the index container's public name, update all the names on all the child containers as well.

Or maybe the task is already doing that, so your original code is almost correct? We still need the extra filtering for containers with no name_and_label / `meta_attr__name_and_label__isnull=True)`. And I think the select_related() should help speed things up. It will make one query in total and join the tables in the DB before returning query results. Without that, we will end up making one query per each index container, to acess its linked SoftwareBuild object.

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
         return f"{self.name}"
+
+
+class CollectorPyxisModel(models.Model):
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    meta_attr = models.JSONField(default=dict)
+
+    class Meta:
+        abstract = True
+        ordering = ["uuid"]
```
juspence, line 99: Do we need to define a default ordering here? I guess by default, each query will either give as a list of models ordered by PK (maybe), or an unordered list of models (more likely). In the latter case, we can specify `.order_by("pk")` only on the queries that need it. That avoids some overhead for sorting queries that don't need it.

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                name="unique_pyxis_repo",
+                fields=["registry", "name"],
+            )
+        ]
+
+    def __str__(self):
+        return f"{self.name}"
```
juspence: If we expect the combo of registry and name to be unique together, maybe it's better to return that in our string representation instead of just the name.
```suggestion
    def __str__(self):
        return f"{self.registry}/{self.name}"
```

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    tags = fields.ArrayField(models.CharField(max_length=200), default=list)
+    creation_date = models.DateTimeField(null=True)
+    pyxis_id = models.CharField(unique=True, max_length=50)
+
+
+class CollectorPyxisImageRepository(CollectorPyxisModel):
+    name = models.CharField(max_length=200)
+    registry = models.CharField(max_length=200)
+    images = models.ManyToManyField(CollectorPyxisImage, related_name="repos")
+    image_advisory_id = models.CharField(max_length=50, default="")
```
juspence, line 127: Is image_advisory_id an erratum ID, like RHSA-2023:1234? Does it belong on the CollectorPyxisImage model instead?

It might also be nice to see some header comments explaining where / how each of these models are used. The names are clear enough, but of course I don't know what I don't know.

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    name_label = models.CharField(max_length=200, default="")
+    tags = fields.ArrayField(models.CharField(max_length=200), default=list)
+    creation_date = models.DateTimeField(null=True)
+    pyxis_id = models.CharField(unique=True, max_length=50)
+
+
+class CollectorPyxisImageRepository(CollectorPyxisModel):
+    name = models.CharField(max_length=200)
+    registry = models.CharField(max_length=200)
+    images = models.ManyToManyField(CollectorPyxisImage, related_name="repos")
```
juspence: At first I thought this model was representing "one registry / repo", like quay.io or quay.io/some-org, which could have many different images inside of it.

But maybe this model actually represents "one image in one registry", like quay.io/some-org/some-image instead? If that's correct, let's use a ForeignKey to link to the CollectorPyxisImage model and not a ManyToManyField:
```suggestion
class CollectorPyxisImageRepository(CollectorPyxisModel):
    name = models.CharField(max_length=200)
    registry = models.CharField(max_length=200)
    images = models.ForeignKey(CollectorPyxisImage, related_name="repos")
```

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    # Pyxis doesn't have a distinct Image for the "noarch" container like Corgi has, it only has
+    # image objects for the arch specific containers, so we mirror that here. This value represents
+    # the sha256 hash from the arch specific container purl version eg:
+    # pkg:oci/ose-ovirt-csi-driver-operator@
+    # --> sha256:596d18e9a4be00f7c5f1ef536b0d34b3a3a5b826807650cc866bbb7976b237b4 <--
+    # ?arch=aarch64
+    # &repository_url=registry.redhat.io/openshift4/ose-ovirt-csi-driver-operator
+    # &tag=v4.12.0-202301042354.p0.gfeb14fb.assembly.stream
+    image_id = models.CharField(max_length=200)
+    arch = models.CharField(max_length=20, default="")
```
juspence: Should this default to "noarch", or not define any default and explicitly require an arch to be provided?

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    nvr = models.TextField()
+    name_label = models.CharField(max_length=200, default="")
```
juspence, line 114: If we're talking about a container image that's shipped in / available from a public repository, it has to have a name and label by definition, right? Let's get rid of the default here if so. Or maybe there are edge cases where we don't know this value?

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+    tags = fields.ArrayField(models.CharField(max_length=200), default=list)
+    creation_date = models.DateTimeField(null=True)
```
juspence: Again, if the image exists it must have a creation date. Let's enforce that our code sets one, instead of allowing it to sometimes be NULL / None.

corgi/collectors/models.py
```diff
@@ -88,3 +88,54 @@ class CollectorRPMRepository(models.Model):
...
+        ordering = ["uuid"]
+
+
+class CollectorPyxisImage(CollectorPyxisModel):
+    # This corresponds to the sha256 hash we use to lookup the image index or "noarch" container eg:
+    # pkg:oci/ose-ovirt-csi-driver-operator@
+    # --> sha256:2b55f17547ddcbdbefddde7372bc3ddfba9b66dfe76d453e479f9efd51514656 <--
+    # ?repository_url=registry.redhat.io/openshift4/ose-ovirt-csi-driver-operator
+    # &tag=v4.12.0-202301042354.p0.gfeb14fb.assembly.stream
+    manifest_list_digest = models.CharField(max_length=200)
```
juspence: I think this is basically the same as image_id below, except for the noarch container instead of the arch-specific container. If that's right, can we call it "noarch_image_id" instead?

Maybe it should also be a ForeignKey link to another CollectorPyxisImage, which we would mock up for the noarch container even though it doesn't exist in Pyxis. That might support bulk lookups / updates / deletes for one "container" and all of its various architectures.

Or maybe we should add a separate CollectorPyxisNoarchImage model, and store most of the common information there (e.g. NVR, name_label). That would avoid duplicating this info on each architecture-specific image. Then each child container / PyxisImage can have a ForeignKey link to its parent NoarchImage.

juspence (commented):
Nowhere near done, but this review is long overdue and I want to submit at least some feedback while I continue looking. Thanks for the great work work on this - using the public container names will be a huge improvement.

tests/test_pyxis_collector.py
```diff
@@ -17,3 +19,35 @@ def test_get_manifest_data():
...
+        repo = image.repos.first()
+        assert repo.name == "openshift4/ovirt-csi-driver-rhel8-operator"
+        assert repo.registry == "registry.access.redhat.com"
+
+    image_index_id = CollectorPyxisImage.objects.values_list(
+        "manifest_list_digest", flat=True
+    ).distinct()
+    assert image_index_id.count() == 1
+    assert (
+        image_index_id.first()
```
juspence:
```suggestion
    assert (
        image_index_id.get() =
```

tests/test_pyxis_collector.py
```diff
@@ -17,3 +19,35 @@ def test_get_manifest_data():
...
+    assert CollectorPyxisImageRepository.objects.count() == 1
+
+    for image in CollectorPyxisImage.objects.all():
+        assert image.repos.count() == 1
+        assert image.nvr == nvr
+        assert image.name_label == "openshift/ose-ovirt-csi-driver-operator"
+        assert image.tags == ["v4.12.0-202301042354.p0.gfeb14fb.assembly.stream"]
+        repo = image.repos.first()
+        assert repo.name == "openshift4/ovirt-csi-driver-rhel8-operator"
+        assert repo.registry == "registry.access.redhat.com"
```
juspence: Let's add a test for the case where one image ships to multiple different repos.

tests/test_pyxis_collector.py
```diff
@@ -17,3 +19,35 @@ def test_get_manifest_data():
...
     assert manifest == example
+
+
+@pytest.mark.django_db
+def test_get_image_by_nvr(requests_mock):
+    nvr = (
+        "ose-cluster-ovirt-csi-operator-container-v4.12.0-202301042354.p0.gfeb14fb.assembly.stream"
+    )
+    with open("tests/data/pyxis/images_by_nvr.json") as data:
+        requests_mock.get(f"{PYXIS_REST_API_URL}/v1/images/nvr/{nvr}", text=data.read())
```
juspence, line 36: At the end of the test, you may want to assert that requests_mock.call_count == some_expected_number

tests/test_pyxis_collector.py
```diff
@@ -17,3 +19,35 @@ def test_get_manifest_data():
...
+
+    get_image_by_nvr(nvr)
+
+    assert CollectorPyxisImage.objects.count() == 4
+    assert CollectorPyxisImageRepository.objects.count() == 1
+
+    for image in CollectorPyxisImage.objects.all():
```
juspence, line 52:
```suggestion
    for image in CollectorPyxisImage.objects.get_queryset()
```

tests/test_pyxis_collector.py
```diff
@@ -17,3 +19,35 @@ def test_get_manifest_data():
...
+        assert image.repos.count() == 1
+        assert image.nvr == nvr
+        assert image.name_label == "openshift/ose-ovirt-csi-driver-operator"
+        assert image.tags == ["v4.12.0-202301042354.p0.gfeb14fb.assembly.stream"]
+        repo = image.repos.first()
```
juspence: You can just do .get() and then there's no need to check the .count():
```suggestion
        repo = image.repos.get()
```

corgi/collectors/brew.py
jasinner: It's a good idea. Although the current migration doesn't actually ingest the brew builds, so processed images would not set the new `name_from_label` field. So we can't use that to filter.

corgi/collectors/models.py
jasinner: I think RHTAP doesn't require the use of this field. At least with RHTAP we don't need to use the field, because we already know the external name.

corgi/collectors/brew.py
jasinner: I added a new meta field called `name_from_label_raw` and set it before trying to split the value. I left the old `name_from_label` field unchanged even though it duplicates the `name` field now.

tests/test_model.py
jasinner: good idea

corgi/tasks/pyxis.py
jasinner: fixed

corgi/core/migrations/0103_get_latest_repo_names_from_pyxis.py
jasinner: The slow_update_name_for_container_from_pyxis handles setting the name on child containers because it does a look up by NVR, which is the same for the image index and it's arch specific containers. I did add an exclude filter for the new Container Component meta_attr field `name_from_label_raw` to avoid this migration taking any action on any container builds ingested after the associated change in the Brew collector.

corgi/collectors/models.py
jasinner: The point of these collector models is to cache the same fields from Pyxis. So one can check the documentation there for an explanation of the fields. Also why image_advisory_id is added to this ImageRepository model instead of PyxisImage.

corgi/collectors/models.py
jasinner: You are right, I'll do that. Thanks.

corgi/collectors/models.py
jasinner: No, I don't think anything is enforcing this, and I heard it will not be added by container builds in RHTAP.

corgi/collectors/models.py
jasinner: ok

corgi/collectors/models.py
jasinner: Again, these models are just to Cache data from Pyxis. I want to keep the names and structure similar to what's in Pyxis.

corgi/collectors/models.py
jasinner: OK.

corgi/collectors/brew.py
```diff
@@ -588,8 +591,15 @@ def get_container_build_data(self, build_id: int, build_info: dict[str, Any]) ->
 
         # During collection we are only able to inspect docker config labels on
         # attached arch specific archives. We do this loop here to save the description, license,
-        # name label, and repository url also on the index container object at the root of the tree.
-        for attr in ("description", "license", "name_from_label", "repository_url"):
+        # also on the index container object at the root of the tree.
+        for attr in (
+            "name",
```
juspence, line 596: I suggested "name" as the new field, but you're using "name_from_label_raw" (which is a better choice, good idea). So I think this "name" here can go away, but maybe I'm confused.

tests/test_errata_data.py
```diff
@@ -267,6 +271,9 @@ def test_save_product_component_for_errata(
     build_list_url = f"{settings.ERRATA_TOOL_URL}/api/v1/erratum/{erratum_id}/builds_list.json"
     requests_mock.get(build_list_url, text=build_list)
     sb = SoftwareBuildFactory(build_id="1636922", build_type=BUILD_TYPE)
+    container_build = SoftwareBuildFactory(build_id="1628358", build_type=SoftwareBuild.Type.BREW)
+    container_build.meta_attr["nvr"] = "rh-dotnet31-container-3.1-18"
+    container_build.save()
```
juspence, line 276: I hope you can set all these at creation time, and then you don't need .save():
```suggestion
    container_build = SoftwareBuildFactory(build_id="1628358", build_type=SoftwareBuild.Type.BREW, meta_attr={"nvr": "rh-dotnet31-container-3.1-18"})
```

corgi/collectors/models.py
```diff
@@ -88,3 +89,56 @@ class CollectorRPMRepository(models.Model):
...
+    creation_date = models.DateTimeField()
+    pyxis_id = models.CharField(unique=True, max_length=50)
+
+
+class CollectorPyxisImageRepository(CollectorPyxisModel):
+    name = models.CharField(max_length=200)
+    registry = models.CharField(max_length=200)
+    images = models.ForeignKey(
+        CollectorPyxisImage, related_name="repos", on_delete=django.db.models.deletion.CASCADE
+    )
```
juspence: Thanks for switching to a ForeignKey. I forgot to note in my earlier suggestion we should use the singular "image" in this case since only one object will be linked:
```suggestion
    image = models.ForeignKey(
        CollectorPyxisImage, related_name="repos", on_delete=django.db.models.deletion.CASCADE
    )
```

juspence (approved):
I haven't looked at the task code at all, but I'll be on PTO tomorrow and all next week, we have lots of new tests here, and we can always do follow-up PRs as needed. So I guess this is fine to merge.

Or if you'd rather wait and work on other tickets, I will finish looking as soon as I get back on Monday the 27th.

corgi/tasks/errata_tool.py
```diff
@@ -108,6 +109,17 @@ def slow_load_errata(erratum_name: str, force_process: bool = False) -> None:
...
+            # container builds can get a new name. Make sure we check for that here and update the
+            # container name if so.
+            if is_container:
+                build = SoftwareBuild.objects.filter(
+                    build_id=build_id, build_type=SoftwareBuild.Type.BREW
+                ).first()
+                if build:
+                    # Let this propagate an error if the nvr key is missing
+                    nvr = build.meta_attr["nvr"]
+                    slow_update_name_for_container_from_pyxis.delay(nvr)
```
juspence, line 122: Marek thought of a bug related to Brew container builds, I filed the ticket but don't have the number handy. We look for a build with type="BREW" here, but we actually receive UMB messages from Pyxis in two cases:

1. For containers which are built in RHTAP and are not in Brew at all. The build is created with type"PYXIS".
2. For containers which were originally built in Brew, but which also end up in Pyxis. The build is created with type="BREW".

For the second case, they really are Brew builds and not Pyxis builds. But since we still get a UMB message from Pyxis for these builds, we create a duplicate SoftwareBuild with type="PYXIS". So we're storing twice as many builds as we need - not great, but not a huge problem.

Then we find the root component for this "Pyxis" build, which is an index container that already exists and is linked to a Brew build. We unlink the index container from the Brew build we already had, then relink the index container to the Pyxis build we just created.

The reverse may happen as well (not sure), if the Pyxis UMB message is received first, and then the Brew "brew.build.complete" UMB message is received after.

In this case, would your code here to filter for builds with type="BREW" give the wrong results? I think we'd fail to update the NVR for some containers, if they were originally linked to some Brew build but later got relinked to a Pyxis build. The NVR might still use the internal Brew value and not the public value from Pyxis. Or maybe I'm confused - it's been a long week, sorry.

corgi/collectors/models.py
juspence: I think the risk here is that we store lots of the same values across many model instances, one for each architecture that some container image is built for. We might update one arch-specific image / model instance, but forget to update the others.

I think it's better to deduplicate that data into a separate model, then have the child models (for each architecture-specific container) all refer to the common model instances (for the overall container NVR) using a ForeignKey. But I'm not strongly opposed to the current approach if you'd prefer to leave it as-is.

corgi/tasks/common.py
```diff
@@ -118,3 +124,61 @@ def run_external(
...
+        # we might erase the license that OpenLCS provided
+        # They cannot erase any licenses we set ourselves (when the field is not empty)
+        # API endpoint blocks this (400 Bad Request)
+        # Use .update() to avoid possible race conditions
+        # We only update if the current license in the DB doesn't match the new value
+        Component.objects.filter(pk=obj.pk).exclude(
+            license_declared_raw=license_declared_raw
+        ).update(license_declared_raw=license_declared_raw)
+    # else the value from Brew is an empty string
+    # so don't overwrite any value OpenLCS may have submitted
```
juspence, line 145: This might be dangerous to move to common...I wrote the logic here with the assumption that only the Brew collector / tasks called this function. I think that's changed anyway due to PNC / middleware using this as well, so I guess it's fine to keep here for now. But maybe we need to revisit how we avoid overwriting RHTAP / Pyxis and PNC / Brew license data, plus whatever OpenLCS submits to license_declared_raw (if anything).

corgi/collectors/models.py
jasinner: done

corgi/collectors/models.py
jasinner: That would involve creating another model type for image index containers. Which would be even further from what we have in Core models. I'm going to leave it the way it is.

corgi/collectors/models.py
jasinner: Actually the main reason we have multiple images is that we have one for each arch. So a single advisory is correct, as that advisory released all the binary artifacts for the build.

corgi/collectors/models.py
jasinner: Actually I think the many-to-many was correct. While it's not "one registry / repo", like quay.io or quay.io/some-org it is more like quay.io/some-org/some-image/noarch-container. But under that `noarch-container`, we have several binary (arch-specific) containers, each with a unique image_id.

corgi/collectors/models.py
jasinner: Actually I moved this to the ImageRepo model and added it to the unique together constraint. I also moved the tags as well.

corgi/collectors/brew.py
jasinner: No, we need to copy name down too.

corgi/tasks/errata_tool.py
jasinner: Nothing to change in this part of the code. The root component created by processing a Pyxis UMB message already uses the last part of the repository path as the component name.

tests/test_pyxis_collector.py
jasinner: ok

PR #666, opened 2023-11-02
CORGI-901: Handle Red Hat Maven roots correctly, e.g. quarkus-bom

juspence:
@RedHatProductSecurity/corgi-devs Please review. This change supports reporting root components like quarkus-bom inside our stream-based manifests (and the ?root_components= filter / parameter in the API). These components have Maven type and are in the Red Hat namespace, but we can't add `Q(type="MAVEN", namespace="REDHAT")` to the existing root_components_condition.

That would cause us to show Red Hat Maven children of the quarkus-bom roots, like agroal-api, as well as completely unrelated Red Hat Maven components, as roots in all cases (both the API and manifests). So the API and other non-middleware manifests for components like UBI and streams like RHEL, which just happen to provide Red Hat Maven components, would both start treating these as roots and reporting incorrect relationships. 

If we don't do anything, the existing stream-based manifests for middleware (Quarkus) will continue to be missing lots of data (all the real Maven root components, their children, and relationships between them).

Since we fixed the code (and data) to only link root components to software builds, we can now rely on this link being present to indicate "root component status". We should not change this behavior again in the future (e.g. don't start linking binary RPMs or arch-specific containers to builds) or else the root_components_condition and filter will break.


jasinner (commented):
I think we'll also have to adjust the latest filter stored procedure to stop filter based on type, arch etc, and use the foreign key to SoftwareBuild as in constants.py.

While would rather clean up the root components as described in CORGI-882 and use the graph to help us determine graph, especially once we start to rely on the graph more heavility. However I don't think there is anything wrong with using SoftwareBuild FK as you have done here. Especially if it up

juspence:
@jasinner Thanks again for reminding me about the latest filter using the same root_components_condition. I moved the SQL to constants.py and tried to pull out common parts as snippets that are easier to change, instead of having to edit the same logic in many different places.

I think it should work the same as before, but now when you change ROOT_COMPONENTS_CONDITION, there's a reminder right below it to also fix ROOT_COMPONENTS_SQL and deploy the updated storedproc in a new migration.

corgi/core/constants.py
```diff
@@ -52,12 +52,15 @@
...
-)
+# Only root components should be linked to software builds
+# This lets us distinguish between Red Hat Maven components which are / are not roots
+# Using some combination of type, namespace, and arch will not work
+# since some Red Hat Maven components are roots (e.g. quarkus-bom)
+# but others are provided by / children of these roots (e.g. agroal-api)
+ROOT_COMPONENTS_CONDITION = Q(software_build_id__isnull=False)
+# If you change above, fix below to match
+# then deploy the updated GET_LATEST_COMPONENT_STOREDPROC_SQL in a new migration
+ROOT_COMPONENTS_SQL = "core_component.software_build_id IS NOT NULL"
```
jasinner: software_build_id is not a column on core_component

```suggestion
ROOT_COMPONENTS_SQL = "core_component.software_build_uuid IS NOT NULL"
```

jasinner (approved):

jasinner:
I like that you refactored the Stored Procedure into reusable chunks, thanks.

jasinner:
I just realized you also need to adjust the stored_proc function in test/conftest.py to use this new function.

corgi/core/constants.py
```diff
@@ -69,3 +72,117 @@
...
+    "latest_epoch := component_epoch; "
+    "latest_version := component_version; "
+    "latest_release := component_release; "
+    "END IF;"
+)
+
+# We are using Postgres functions instead of pure stored procedures
+# These constants should escape \d sequences to ensure migration is applied
+# correctly. Manual insertion of these functions will need to unescape.
+GET_LATEST_COMPONENT_STOREDPROC_SQL = f"""
```
JimFuller-RedHat, line 112: This modularity seems superfluous to me eg:

1) what does modularity in python/django give us ? for me it obfuscates

2) **get_latest_components** will 'go away' the moment we get into graphdb ... we will continue to use rpmver funcs

alternately we could modularise in pg itself ...  but as the lifetime of get_latest_component is short ... not really worth the effort.

juspence: The advantage here is that I don't have to make the same changes to the root components logic (or the where clause select condtions, or ... etc.) in mutiple places. Some of these changes were needed ayway since I'm updating which entries are considered roots.

Having the same logic in multiple places just increases the risk of me making a typo. Actually I already did typo a few things in an earlier version, which made tests start failing and it took me a rather long time to spot the error. Hence the decision to just pull out the common bits and change them once.

JimFuller-RedHat: sorry for late reply - I am fine either way - just highlighting that sacrificing comprehension for maintenance shows up as a constant 'cost' for any person reading the code ... my personal bias is towards comprehension eg. I will always choose a little bit of duplication and a little less reuse for instant comprehension most days.

corgi/core/constants.py
juspence: software_build_id is an implicit field added by the software_build ForeignKey.
Using this field lets you refer to the ID of the related object without making another query to a separate table:
```
>>> root.software_build
<SoftwareBuild: SoftwareBuild object (f836be3a-4376-4827-b93a-0333a5318880)>
>>> root.software_build_id
UUID('f836be3a-4376-4827-b93a-0333a5318880')
```

I don't see any software_build_uuid column:
```
>>> root.software_build_uuid
  File "<console>", line 1, in <module>
AttributeError: 'Component' object has no attribute 'software_build_uuid'
```

juspence: Actually we're both right...software_build_id is what you'd use in Python, but it has to be software_build_uuid in the SQL.

PR #668, opened 2023-11-03
Handle percent encoded repository_url

kgrant-rh:
Now that the purls in erratum notes fields include repository_url, the purl has to be percent-decoded to match purls in Corgi.


kgrant-rh:

mprpic (approved):

PR #671, opened 2023-11-07
Record PNC components as direct children of a root node

kgrant-rh:
A simple change to record all components in an SBOMer manifest as direct children of a root component, rather than replicating the relationships section of the manifest.


kgrant-rh:
@RedHatProductSecurity/corgi-devs Please review. I decided to keep this scoped only to CORGI-880, rather than incorporating other management commands or other refactoring.

jasinner (approved):

PR #672, opened 2023-11-08
avoid runtime warning and failing tests/test_sca.py because of changes in Python3.11 to avoid CVE-2007-4559

jasinner:
This change should be compatible with future Python versions.


JimFuller-RedHat (approved):
apart from mypy failure LGTM

PR #675, opened 2023-11-09
CORGI-899: Fully manifest services and report any permissions issues at end, instead of stopping on first error

juspence:
@RedHatProductSecurity/corgi-devs Please review, I'm going to add tests after lunch. This has to go out ASAP so I'll be doing a patch release tomorrow if this looks good and works in stage (after merging).


jasinner (approved):
LGTM, thanks

gunicorn_config.py
```diff
@@ -24,12 +20,8 @@
 if not running_dev():
     # Saves memory in the worker process, but breaks --reload
     preload_app = True
-    # avoid restarting gunicorn and leave it to pod restarts to handle memory leaks
-    max_requests = 0
```
JimFuller-RedHat, line 28: this is set here for a reason ... we do not want gunicorn to restart ... as it results in Bad Gateway errors for in flight http responses being served ... if there are memory leaks (which we should not have) let pod deal with it

gunicorn_config.py
```diff
@@ -24,12 +20,8 @@
...
     graceful_timeout = 800  # if a restart must happen then let it be graceful
     keepalive = 60  # specifically this should be a value larger then nginx setting
-    # ref: https://github.com/benoitc/gunicorn/issues/1978
-    max_requests_jitter = 0
```
JimFuller-RedHat, line 32: as per the ref ... we set this as a guard for max_requests_jitter logic to never engage

gunicorn_config.py
```diff
@@ -1,9 +1,5 @@
-# avoid MonkeyPatchWarning error/warnings
-from gevent import monkey
-
 from config.utils import running_dev
 
-monkey.patch_all(thread=False, select=False)
```
JimFuller-RedHat, line 6: doing this monkey patching here resolved some variants of bad gateway ... I would leave this asis

gunicorn_config.py
juspence: This is the default setting, adding the config option here does nothing.

gunicorn_config.py
juspence: This is the default setting, adding the option here does nothing.

JimFuller-RedHat: while tests revealed that this setting indeed had no behaviour I thought it best to explicitly set to 0 with code comment referring to additional uncontained behaviour regardless of max_requests=0 setting

gunicorn_config.py
juspence: When running locally I saw the Celery-beat container would sometimes restart and there was a warning about monkeypatching multiple times. The restarts were due to a different issue in my local environment, but removing this while debugging did get rid of the warning.

In general, we should only need to monkeypatch once as part of preload = True, which gunicorn already handles for us. Monkeypatching multiple times, too soon, or too late can cause other issues. Finally, monkeypatching in a local dev environment isn't correct since we aren't preloading Gunicorn. This supports automatic reload whenever config / files change. I'll back out this change if I see monkeypatch warnings in stage.

gunicorn_config.py
```diff
@@ -1,9 +1,5 @@
-# avoid MonkeyPatchWarning error/warnings
```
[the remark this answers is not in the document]
juspence: See my replies below. These changes were made while debugging local environment issues after an OS reinstall and fix warnings.

gunicorn_config.py
JimFuller-RedHat: +1 ... main concern is avoiding Bad Gateways ... testing revealed this change mitigated some of the issues in this area. It might be other mitigations elsewhere now make this not applicable.

gunicorn_config.py
juspence: I did look at the ref but the user says "I expect that --max-requests-jitter has no effect when --max-requests is set to 0." Setting both to 0 is the default and this bug was fixed years ago, so I don't think we need the config options.

JimFuller-RedHat: checking, yes I think at the time ... we had older gunicorn then its been updated since (I think I or someone else did that) ... thanks lets remove it

PR #678, opened 2023-11-10
Temporary workaround for Quarkus manifests

kgrant-rh:
A temporary workaround to deal with conflicting purls in Quarkus manifests.


juspence (approved):
I really thought the existing code should handle this, but LGTM since we need to get this working ASAP.
FYI you can run ansible-playbook commands manually to build an image from a particular commit and deploy to stage. You don't need to actually merge this and wait for the CI pipeline. That also makes it easier to make changes, redeploy, and test again without doing lots of MRs.

PR #682, opened 2023-11-13
CORGI-857 cpe per component

jasinner:
I did this hoping that it would help with CORGI-602 but unfortunately I didn't help. However this does extend the work we did in CORGI-872 to work with streams which don't use errata_info to define which builds make them up.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1778,7 +1778,7 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
 
     @property
     def cpes(self) -> QuerySet:
-        """Build and return a list of CPEs this Component relates to"""
+        """Build and return a list of CPEs from all Variants this Component relates to"""
         # For each Variant-type relation, get the linked Variant's CPE directly
         # Remove any duplicates, and return the CPEs in sorted order so manifests are stable
         if self.productvariants.exists():
@@ -1790,7 +1790,9 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
             )
         elif self.software_build:
             build_variants = (
-                ProductComponentRelation.objects.filter(software_build=self.software_build)
+                ProductComponentRelation.objects.filter(
+                    software_build=self.software_build, type=ProductComponentRelation.Type.ERRATA
+                )
                 .values_list("product_ref", flat=True)
                 .distinct()
             )
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -17,6 +17,7 @@ from mptt.models import MPTTModel, TreeForeignKey
 from packageurl import PackageURL
 from packageurl.contrib import purl2url
 
+from corgi.collectors.models import CollectorErrataProductVariant
 from corgi.core.constants import (
     CONTAINER_DIGEST_FORMATS,
     EL_MATCH_RE,
@@ -1777,25 +1778,30 @@ class Component(TimeStampedModel, ProductTaxonomyMixin):
 
     @property
     def cpes(self) -> QuerySet:
-        """Build and return a list of CPEs from all Variants this Component relates to"""
+        """Build and return a list of CPEs this Component relates to"""
         # For each Variant-type relation, get the linked Variant's CPE directly
         # Remove any duplicates, and return the CPEs in sorted order so manifests are stable
-        return (
-            self.productvariants.exclude(cpe="")
-            .values_list("cpe", flat=True)
-            .distinct()
-            .order_by("cpe")
-        )
+        if self.productvariants.exists():
+            return (
+                self.productvariants.exclude(cpe="")
+                .values_list("cpe", flat=True)
+                .distinct()
+                .order_by("cpe")
+            )
+        elif self.software_build:
+            build_variants = (
+                ProductComponentRelation.objects.filter(software_build=self.software_build)
+                .values_list("product_ref", flat=True)
+                .distinct()
+            )
+            if build_variants:
+                return (
+                    CollectorErrataProductVariant.objects.filter(name__in=build_variants)
+                    .values_list("cpe", flat=True)
+                    .order_by("cpe")
+                )
 
-        # For Stream-type relations, we no longer link any Variants
-        # since this caused data quality issues, where Components were linked to Variants
-        # that didn't actually ship those Components
-        # We could look at all the linked streams to get CPEs from all their child Variants
-        # But this would cause the same problem - we'd report too many Variants / incorrect CPEs
-
-        # TODO: We return an empty QuerySet if no Variants are linked, or none have CPEs
-        #  But do we want to raise an error / make manifesting fail,
-        #  whenever we can't find CPEs for some root component?
+        return ProductVariant.objects.none()
 
     def _build_download_url_for_type(self) -> str:
         """Get a source code or binary download URL based on a purl"""
--- a/corgi/web/templates/product_manifest.json
+++ b/corgi/web/templates/product_manifest.json
@@ -4,15 +4,12 @@
       "copyrightText": {% if component.copyright_text %}"{{component.copyright_text|escapejs}}"{% else %}"NOASSERTION"{% endif %},
       "downloadLocation": {% if component.download_url %}"{{component.download_url|escapejs}}"{% else %}"NOASSERTION"{% endif %},
 
-      "externalRefs": [{% for cpe in component.cpes %}
-        {# Make sure any component cpe is found in the product cpe list #}
-        {% if cpe in cpes %}
+      "externalRefs": [{% for cpe in component.cpes %}{# Make sure any component cpe is found in the product cpe list #}{% if cpe in cpes %}
         {
           "referenceCategory": "SECURITY",
           "referenceLocator": "{{cpe}}",
           "referenceType": "cpe22Type"
-        },{% endif %}
-        {% endfor %}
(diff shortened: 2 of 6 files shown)


jasinner:

corgi/core/models.py
```diff
@@ -1774,25 +1775,30 @@ def get_roots(self, using: str = "read_only") -> QuerySet[ComponentNode]:
...
+                self.productvariants.exclude(cpe="")
+                .values_list("cpe", flat=True)
+                .distinct()
+                .order_by("cpe")
+            )
+        elif self.software_build:
+            build_variants = (
+                ProductComponentRelation.objects.filter(software_build=self.software_build)
+                .values_list("product_ref", flat=True)
+                .distinct()
```
juspence: This is finding all relations, but I think we should filter only for relations that refer to variants:
```suggestion
                ProductComponentRelation.objects.filter(software_build=self.software_build, type__in=ProductComponentRelation.VARIANT_TYPES)
                .values_list("product_ref", flat=True)
                .distinct()
```
This also requires that the relation be processed. That's fine if it's what you want. But we could also report CPEs even for Variant product_refs from unprocessed relations. We'd just need to filter the relations on `build_id=self.software_build.build_id, build_type=self.software_build.build_type` instead of using `software_build=self.software_build` directly.

corgi/core/models.py
```diff
@@ -1774,25 +1775,30 @@ def get_roots(self, using: str = "read_only") -> QuerySet[ComponentNode]:
...
+            )
+            if build_variants:
+                return (
+                    CollectorErrataProductVariant.objects.filter(name__in=build_variants)
+                    .values_list("cpe", flat=True)
+                    .order_by("cpe")
```
juspence, line 1803: I thought we didn't want to access Collector models from Core model code. Is there a reason we don't just filter using the regular ProductVariant model? That should still return more data than self.productvariants. If we haven't saved the taxonomy for this component yet, self.productvariants would either be empty or incomplete / there will be no links to the models, but the ProductVariant models themselves will still exist and have all the right CPEs.

Whatever we do, we should also exclude empty "" string CPEs from being reported here.

corgi/core/models.py
```diff
@@ -1774,25 +1775,30 @@ def get_roots(self, using: str = "read_only") -> QuerySet[ComponentNode]:
 
     @property
     def cpes(self) -> QuerySet:
-        """Build and return a list of CPEs from all Variants this Component relates to"""
```
juspence, line 1780: You deleted "from all Variants" in the new code. But aren't we still getting CPE values only from Variants? Or did you plan to eventually do something like below, in addition to variants?
1. From the relations table, look up all product_refs for this component
2. Filter only the stream-type relations and look up the related streams
3. Report CPEs for all the stream's child Variants, using each stream's "cpes" property

corgi/web/templates/product_manifest.json
```diff
@@ -3,12 +3,16 @@
     {
       "copyrightText": {% if component.copyright_text %}"{{component.copyright_text|escapejs}}"{% else %}"NOASSERTION"{% endif %},
       "downloadLocation": {% if component.download_url %}"{{component.download_url|escapejs}}"{% else %}"NOASSERTION"{% endif %},
+
       "externalRefs": [{% for cpe in component.cpes %}
+        {# Make sure any component cpe is found in the product cpe list #}
+        {% if cpe in cpes %}
```
juspence: Where is this CPEs variable defined? Also, I think you need to update the component_manifest.json template as well, right?

Finally, the for loop, comment, and if statement need to all be on the same line. This also applies to the endif / endfor bits below. Otherwise whitespace will end up in the template and it may break the JSON formatting.

juspence (commented):
I'm not sure I understand the change, it seems incomplete. Right now we only report CPEs for root components, but I thought in the ticket we wanted to report CPEs for all components.

If that's correct, you can just add the `{% for cpe in component.cpes %}` logic to the externalRefs key for provided and upstream components as well. Right now this for-loop only exists for root components. This should be enough to start reporting CPEs for all components, changes to the Component model's "cpes" property shouldn't be needed.

Your changes to the "cpes" property will mostly work fine I think, but I'm not sure what edge-case you're trying to handle. I say "mostly" because self.software_build is only set for root components, and again I thought we were trying to report CPEs for all components.

tests/test_manifests.py
```diff
@@ -132,8 +134,12 @@ def test_component_manifest_backslash():
 
 
 @pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
-def a_test_slim_rpm_in_containers_manifest(stored_proc):
-    containers, stream, rpm_in_container = setup_products_and_rpm_in_containers()
```
juspence, line 136: Do you know when we changed this test to be prefixed with `a_`? That disables the test so it hasn't been running at all.

tests/test_manifests.py
```diff
@@ -163,6 +169,20 @@ def a_test_slim_rpm_in_containers_manifest(stored_proc):
...
+        if pk in container_pks:
+            found_cpe = False
+            for ref in package["externalRefs"]:
+                if ref["referenceType"] == "cpe22Type":
+                    # The CPE matches the one returned by setup_products_and_rpm_in_containers()
+                    assert ref["referenceLocator"] == cpe
+                    found_cpe = True
+            # There was at least one CPE
+            assert found_cpe
+
```
juspence: We're already testing this below, but I guess it doesn't hurt to add it here as well. You can also break early after you set `found_cpe = True`.

tests/test_model.py
```diff
@@ -174,6 +175,32 @@ def test_component_cpes():
         assert unused_variant.cpe not in cpes
 
 
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
+def test_component_without_variant_cpes():
```
juspence, line 179: Let's add a docstring comment:
```suggestion
@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
def test_component_without_variant_cpes():
    """Test the CPEs property on the Component model finds CPEs
        even when no Variants are linked directly to that Component"""
```

corgi/core/models.py
jasinner: Good idea to also filter by type. Although I think filtering by ERRATA type is probably best since that is the best indicator that the build has been released to that Variant. Errata type relations shouldn't ever not have a linked SoftwareBuild foreign key, so leaving that part of the filter unchanged.

corgi/core/models.py
jasinner: ProductVariant models don't exist for brew_tag streams (like openshift-4.*) which we are targeting here. We would have make a bunch of other changes to the code to allow that linking. For example, setting up a many-to-many relationship between stream and variant to allow variants to be linked to multiple streams. Also changes to the save_product_taxonomy functions to prevent builds linked to ProductVariants matched via brew_tags from being also linked to ProductStreams. 

Using the Collector models here seems like a good compromise to me. Although if we really wanted to be strict about that we should probably allow ProductVariant models to have their productstreams, etc be null so that we can create ProductVariant and use those parentless ProductVariants in situations like this.

corgi/core/models.py
jasinner: I put the old comment back, it's still Variants.

corgi/web/templates/product_manifest.json
jasinner: The cpes variable is define only for ProductManifestFile in corgi/core/files.py. The point of this is to make sure that we're only specifying the component's CPEs where they the product's CPEs list. Components can ship to multiple products. We couldn't potentially report all CPEs for a components in the component_manifest.json, but that's beyond the scope of this PR in my opinion.

I moved the for, comment and if statement only a single line.

tests/test_manifests.py
jasinner: I don't know.

tests/test_manifests.py
jasinner: ok

tests/test_model.py
jasinner: ok

jasinner:
> I'm not sure I understand the change, it seems incomplete. Right now we only report CPEs for root components, but I thought in the ticket we wanted to report CPEs for all components.
> 
> If that's correct, you can just add the `{% for cpe in component.cpes %}` logic to the externalRefs key for provided and upstream components as well. Right now this for-loop only exists for root components. This should be enough to start reporting CPEs for all components, changes to the Component model's "cpes" property shouldn't be needed.
> 
> Your changes to the "cpes" property will mostly work fine I think, but I'm not sure what edge-case you're trying to handle. I say "mostly" because self.software_build is only set for root components, and again I thought we were trying to report CPEs for all components.

You misunderstood the modivation for this change. The change to assist in determining which CPE a root component relates to when there are no linked ProductVariants for the component, for example, for openshift-4.x streams which use brew_tag relations to determine which builds make up the product.

corgi/core/models.py
juspence: Ack, using the Collector Variants is fine in this case. Thanks for explaining the edge-case we're handling with the brew_tag streams.

juspence (approved):
LGTM, thanks for explaining the problem with the brew_tag streams.

jasinner:
CORGI-857 is fixed by the changes merged in PR#723, verified in: PR#732

PR #683, opened 2023-11-14
CORGI-915: Fix "quarkus-bom" component providing itself / duplicate data in manifests

juspence:
@kgrant-rh Quick review please. I still need to regenerate Quarkus data in prod after migration number 100 ran, and I think this fix will solve the issue Jens Reimann reported.

I think the existing tests would have covered this issue, except that the sample data we load only has the root component once. It doesn't duplicate the root component into the first entry in the list of components. I changed it to match what I think SBOMer really sends us, but please let me know if that isn't right.

### DIFF
--- a/corgi/collectors/pnc.py
+++ b/corgi/collectors/pnc.py
@@ -31,8 +31,14 @@ class SbomerSbom:
         # separately for each dependency they have or fulfill.
         # FIXME ------
         # Temporary workaround for component purl conflicts
+        root_component = data["metadata"]["component"]
         self.components = {}
-        created_nvrs = set()
+        # A set of tuples, pre-populated with the root component's name and version
+        created_nvrs = {(root_component["name"], root_component["version"])}
+        # The first entry should be a duplicate of the root component
+        # Don't create an extra node for it, otherwise quarkus-bom will provide quarkus-bom
+        # so it will appear twice in our manifeats, and IDs will not be unique
+        # this breaks parsing of SPDX manifests by clients
         for component in data["components"]:
             if (component["name"], component["version"]) in created_nvrs:
                 continue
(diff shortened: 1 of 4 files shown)


corgi/collectors/pnc.py
```diff
@@ -33,7 +33,11 @@ def __init__(self, data: dict[str, Any]):
         # Temporary workaround for component purl conflicts
         self.components = {}
         created_nvrs = set()
-        for component in data["components"]:
+        # Skip the first entry, which is a duplicate of the root component
+        # Otherwise we will create an extra node, quarkus-bom will provide quarkus-bom
+        # so it will appear twice in our manifeats, and IDs will not be unique
+        # this breaks parsing of SPDX manifests by clients
+        for component in data["components"][1:]:
```
mprpic: Are we 100% sure that this will always be the first entry in the SBOMer manifest?

juspence: Nope, see new commit I just pushed :slightly_smiling_face: 

This should be correct based on the data SBOMer sends today, but obviously that doesn't guarantee they won't change things in the future. So I changed the code to pre-populate the set of unique NVRs with the root component's NVR.

Then if the first component is also the root component, we won't create a duplicate, but if the first component is something else, we'll still process it correctly instead of skipping it and missing data.

kgrant-rh: The first component should, afaik, be the same as the "root" (viz. the metadata) component because it has to have a bomref in the components section to be referenced in the relationships section. Obviously, that _could_ change, but I think that would be an exceptional circumstance.

kgrant-rh (approved):
Looks good. Nice catch on listing the root component first in the list of NVRs.

PR #687, opened 2023-11-20
Minor: refactor out unneeded functions

JimFuller-RedHat:
* removed unnec functions
* added **epoch** to default component field filters
* added **re_downstreams_name** to component filters


jasinner (commented):
The API changes look fine. I wonder why we are adding test files and mpp-openshift-corgi tags in .common_ci_setup though.

PR #688, opened 2023-11-23
save taxonomy for containers with SOURCE descendants but not upstreams

jasinner:
This adds a data migration to core models to persist the 'upstreams' (and other product and component taxonomy) field of Container Components where that Container has more than one SOURCE type node in it's dependents.


jasinner:
@juspence can you please review this. I think this might have been missing from your previous PRs related to CORGI-798

corgi/core/migrations/0106_save_container_taxonomy.py
```diff
@@ -0,0 +1,30 @@
...
+
+from django.db import migrations
+from corgi.core.models import Component
+from corgi.tasks.common import slow_save_taxonomy as current_slow_save_taxonony
+
+
+def save_container_taxonomy(apps, schema_editor) -> None:
+    for root_container_no_upstreams in (
+        Component.objects.exclude(software_build__isnull=True)
+        .filter(type="OCI", upstreams=None)
```
juspence: I think we should filter (narrow the search results) before we try to make the DB iterate over and exclude from a (very large) queryset:
```suggestion
        Component.objects.filter(type="OCI", arch="noarch")
```
I'm also not sure filtering based on a many-to-many upstreams field will work efficiently. Maybe it's better to just check `root_container_no_upstreams.upstreams.exists()` in the for-loop below instead.

juspence (approved):
I don't see a problem with running this just to be safe, but I thought in my previous MR I checked our data in stage. I hope I noted the checks I ran in a ticket or PR somewhere, as well as the assumptions / justifications for the exact changes I made.

As far as I remember, I looked for but don't recall ever seeing an index container / root node with SOURCE-type child container nodes. And I couldn't find anywhere in our code that these were created. I think that's why I skipped checking for and dealing with any SOURCE-type descendants.

Maybe our data has changed since then, though, or maybe my check wasn't right, or maybe data in prod didn't match stage. Did you see some particular issue that made CORGI-798 look incomplete?

I guess the upstream_go_modules for an index container might be created as SOURCE-type descendants, and wouldn't get reported as "upstreams" of the container until we save the taxonomy. I thought that had been handled or wasn't relevant due to...something else I did. Maybe the upstream Go modules always have PROVIDES type as well?

Or maybe I was only concerned about and checking for SOURCE-type child containers, not SOURCE-type descendants of other types. If that's the issue you're having, I guess this will fix the problem.

jasinner:
> I don't see a problem with running this just to be safe, but I thought in my previous MR I checked our data in stage. I hope I noted the checks I ran in a ticket or PR somewhere, as well as the assumptions / justifications for the exact changes I made.
> 
> As far as I remember, I looked for but don't recall ever seeing an index container / root node with SOURCE-type child container nodes. And I couldn't find anywhere in our code that these were created. I think that's why I skipped checking for and dealing with any SOURCE-type descendants.
> 
> Maybe our data has changed since then, though, or maybe my check wasn't right, or maybe data in prod didn't match stage. Did you see some particular issue that made CORGI-798 look incomplete?
> 
> I guess the upstream_go_modules for an index container might be created as SOURCE-type descendants, and wouldn't get reported as "upstreams" of the container until we save the taxonomy. I thought that had been handled or wasn't relevant due to...something else I did. Maybe the upstream Go modules always have PROVIDES type as well?
> 
> Or maybe I was only concerned about and checking for SOURCE-type child containers, not SOURCE-type descendants of other types. If that's the issue you're having, I guess this will fix the problem.

Not SOURCE-type child container nodes but SOURCE-type child github nodes. The value I checked was the container from the description of CORGI-798 see:

```
In [1]: import corgi.core.models as m
In [2]: c = m.Component.objects.get(purl="pkg:oci/quay-rhel8@sha256:585573ba879bb94317ca19729c067f83067c201b5398fb4f971ef431cf6e4f32?repository_url=registry.redhat.
   ...: io/quay/quay-rhel8&tag=v3.10.0-1")
In [3]: c.upstreams.values("purl")
Out[3]: <ComponentQuerySet []>

In [4]: from corgi.tasks.common import slow_save_taxonomy
In [5]: sb = c.software_build
In [6]: slow_save_taxonomy(sb.build_id,sb.build_type)
...

In [7]: c.refresh_from_db()
In [8]: c.upstreams.values("purl")
Out[8]: <ComponentQuerySet [{'purl': 'pkg:github/prometheus/pushgateway@007ba874bead1b9ad2253d89e3adeb16a73fd012'}, {'purl': 'pkg:github/quay/quay@ba29a40b80f637aecabb264062488b63c5157bc4'}, {'purl': 'pkg:github/quay/config-tool@5f9333ce08793c186bc1adeee34e8ea3806538fb'}]>
```

corgi/core/migrations/0108_save_container_taxonomy.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.2.20 on 2023-11-23 04:48
+
+from django.db import migrations
+
+from corgi.core.models import Component
+from corgi.tasks.common import slow_save_taxonomy as current_slow_save_taxonony
+
+
+def save_container_taxonomy(apps, schema_editor) -> None:
+    for root_container in Component.objects.filter(type="OCI", arch="noarch").iterator():
```
juspence, line 11: You're importing the current Component model, but in this case I think we should use the historical `apps.get_model("core", "Component")` version for safety / to avoid changes that could break this migration.

That does mean below will have to do `ComponentNode.objects.filter(type="SOURCE", parent=None, purl=root_container.purl) instead of accessing the `root_container.cnodes` property.

corgi/core/migrations/0108_save_container_taxonomy.py
```diff
@@ -0,0 +1,24 @@
...
+        if root_container.upstreams.exists():
+            continue
```
juspence: I guess this is fine, but there's a chance the upstreams may exist (for upstream Go modules or something else) while still being incomplete / missing other values. You could skip this check, save 1 call to the DB each loop, and just aggressively save the taxonomy for all root containers, whether or not they have upstreams.

corgi/core/migrations/0108_save_container_taxonomy.py
```diff
@@ -0,0 +1,24 @@
...
+        if root_container.cnodes.first().get_descendants().filter(type="SOURCE").exists():
```
juspence: root_container.cnodes.first() shouldn't be none, but there might be gaps in our data. It might be more safe to check the node exists before calling get_descendants().

Alternatively, we could be more aggressive and try to enforce that one root container has exactly one node (for the one build it appears in / is linked to), by using .get() instead of .first().

juspence (approved):
My comments below are minor, it's probably fine to merge as-is.

Re: your comments above, yep. My notes say that we should have SOURCE-type child nodes under each index container root, which are just the Brew sources plus the upstream Go modules. And I think the comment I was remembering was "we never have SOURCE-type child nodes that are linked to OCI components". I believe the SOURCE-type child nodes are only linked to e.g. GitHub components like you showed above.

I hope that means the code I wrote before is correct, and I just didn't bother to save the taxonomies after all those changes. Maybe that was an oops or maybe I thought there were too many / we would just handle all this as part of some larger reprocessing effort.

But maybe my earlier PR wasn't correct and some of those container upstreams / Brew "sources" got converted to PROVIDES-type nodes instead of SOURCE-type nodes. I don't remember if this actually happened or not - it's been too long since I worked on this code. Please feel free to file a ticket or ping me if you notice any incorrect results.

corgi/core/migrations/0108_save_container_taxonomy.py
jasinner: Ok, removed this check.

corgi/core/migrations/0108_save_container_taxonomy.py
jasinner: OK, now using apps.get_model

corgi/core/migrations/0108_save_container_taxonomy.py
jasinner: checking to make sure cnodes.first() is not None.

PR #691, opened 2023-11-25
Minor: tweak to re_provides

JimFuller-RedHat:
Minor refactor of re_provides and re_provides_name. Added perf test while I was in the area.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -1075,9 +1075,17 @@ class ComponentQuerySet(models.QuerySet):
         """Show only components in active product streams"""
         if include:
             # Truthy values return the filtered queryset (only components in active streams)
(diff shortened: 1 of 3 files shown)


corgi/core/models.py
```diff
@@ -1075,9 +1075,17 @@ def active_streams(self, include: bool = True) -> "ComponentQuerySet":
         """Show only components in active product streams"""
         if include:
             # Truthy values return the filtered queryset (only components in active streams)
-            return self.filter(productstreams__active=True).distinct()
+            return (
+                self.filter(productstreams__active=True)
```
JimFuller-RedHat, line 1079: we do this to avoid multiples from cross join hits against productstream

PR #693, opened 2023-11-27
CORGI-767 add migration to copy license scan results from srpm to binary rpms

jasinner:
I didn't use openlcs_scan_url in the filter since that was cleared from production in CORGI-916. Using license_concluded_raw should give the same results.


jasinner:

corgi/core/migrations/0107_copy_srpm_license_data_to_binary_rpms.py
```diff
@@ -0,0 +1,32 @@
...
+
+def copy_srpm_license_data_to_binary_rpms(apps, schema_editor) -> None:
+    for scanned_srpm in (
+        Component.objects.filter(type=Component.Type.RPM, arch="src")
+        .exclude(license_concluded_raw="")
+        .iterator()
+    ):
+        for binary_rpm in scanned_srpm.provides.filter(type=Component.Type.RPM).exclude(
+            license_concluded_raw=""
+        ):
```
juspence: You might also want to use .iterator() here, and I'm not sure the logic is right. With the current .exclude() code, we'll skip binary RPMs which are missing a license, and only update binary RPMs which already had scan results reported separately by OpenLCS.

I think instead, we want to use .filter() to skip any binary RPMs where OpenLCS already uploaded a scan result (they do this for newer RPMs). Then we'd only copy the license from the SRPM to the binary RPM if OpenLCS didn't upload any scan result (older RPMs, before they made a change to their scanning process).
```suggestion
        for binary_rpm in scanned_srpm.provides.filter(
            type=Component.Type.RPM,
            license_concluded_raw="",
        ).iterator():
```

corgi/core/migrations/0107_copy_srpm_license_data_to_binary_rpms.py
```diff
@@ -0,0 +1,32 @@
...
+            binary_rpm.openlcs_scan_version = scanned_srpm.openlcs_scan_version
+            binary_rpm.license_concluded_raw = scanned_srpm.license_concluded_raw
+            binary_rpm.license_declared_raw = scanned_srpm.license_declared_raw
+            binary_rpm.copyright_text = scanned_srpm.copyright_text
+            binary_rpm.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
```
juspence: If you use .update() like I suggested above, you can add below here safely:
```suggestion
    atomic = False
    dependencies = [
```
Then the migration will run outside of a transaction (using .update() makes this safe and avoids race conditions). That means the migration can pick up where it left off after any timeouts. And I think timeouts are likely since the volume of data we need to iterate over and update is very large.

corgi/core/migrations/0107_copy_srpm_license_data_to_binary_rpms.py
```diff
@@ -0,0 +1,32 @@
...
+    ):
+        for binary_rpm in scanned_srpm.provides.filter(type=Component.Type.RPM).exclude(
+            license_concluded_raw=""
+        ):
+            binary_rpm.openlcs_scan_url = scanned_srpm.openlcs_scan_url
+            binary_rpm.openlcs_scan_version = scanned_srpm.openlcs_scan_version
+            binary_rpm.license_concluded_raw = scanned_srpm.license_concluded_raw
+            binary_rpm.license_declared_raw = scanned_srpm.license_declared_raw
+            binary_rpm.copyright_text = scanned_srpm.copyright_text
+            binary_rpm.save()
```
juspence: We could also use .update() on the queryset instead of setting properties on each binary RPM individually, which should be faster (fewer DB queries). Typing from memory, but I think something like below will work:
```suggestion
        scanned_srpm.provides.filter(
            type=Component.Type.RPM,
            license_concluded_raw="",
        ).update(
            copyright_text=Value(scanned_srpm.copyright_text)
            license_concluded_raw=Value(scanned_srpm.license_concluded_raw)
            license_declared_raw=Value(scanned_srpm.license_declared_raw)
            openlcs_scan_url=Value(scanned_srpm.openlcs_scan_url)
            openlcs_scan_version=Value(scanned_srpm.openlcs_scan_version)
            )
```

juspence (approved):
LGTM aside from logic issue with filter() vs. exclude() noted below. Other comments are just minor possible enhancements.

PR #695, opened 2023-11-28
CORGI-854 / CORGI-923: Fix mangen-related bugs

juspence:
@RedHatProductSecurity/corgi-devs Please review this fix for bad data in stage. Each day, we delete old managed service data, then rerun Syft on all Github repos and Quay images for each managed service. This change addresses the removed TODO comment, where we weren't fully cleaning up old data from previous runs of our managed services tasks.

There's (I think) another bug as well. Syft creates components like "@rails/actiontext" with a combined version string like "version-release" and an empty release string like "". I tried fixing this (a very long time ago) just for managed services, so that we'd get a version like "version" and a release like "release". This ended up causing odd failures I didn't have time to go investigate, so I reverted this change.

However, some components with the correct "version" and "release" were left behind. Now Syft fails to create the same e.g. "@rails/actiontext" component with a combined "version-release" string, due to unique constraints on the purl. So the second commit here fixes Syft again, so that the version and release are reported correctly in separate fields.

I hope this should avoid the errors we saw before, since now we only split Syft's version-release into separate fields if absolutely necessary (when we already had a separate version and release from another Syft scan, or from Brew). Thoughts welcome. I'd like to deploy this to stage (without merging) to confirm it really fixes the issue. If so I'll add tests for this.

Finally, I just added a commit that should fix manifest generation for managed services. Previously, we only generated a manifest locally if a stream had "released components" that are publicly available for customers. This required ProductComponentRelations with type "ERRATA" (for most products, including RHEL) or "COMPOSE" (a special case for RHEL only).

Now we consider "APP_INTERFACE" relations / components as always being released. I think this is correct, since we only manifest the production version of the services, so the components we discover should already be running in the service's environment.

We still won't publish manifests for services to the Customer Portal, since all the managed service streams always have the "no_manifest" tag. But generating the manifest locally means Griffon searches will start working, which unblocks the VM team from switching to Corgi and abandoning the legacy mangen.

### DIFF
--- a/corgi/tasks/managed_services.py
+++ b/corgi/tasks/managed_services.py
@@ -16,8 +16,7 @@ from corgi.core.models import (
     ProductStream,
     SoftwareBuild,
 )
-from corgi.tasks.brew import slow_save_taxonomy
-from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
+from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS, slow_save_taxonomy
 from corgi.tasks.sca import save_component
 
 logger = get_task_logger(__name__)
@@ -280,7 +279,8 @@ def cpu_manifest_service(stream_name: str, service_components: list) -> None:
                 type=ProductComponentRelation.Type.APP_INTERFACE,
             )
 
-        slow_save_taxonomy(build.build_id, build.build_type)
+        # Give the transaction time to commit, before looking up the build we just created
+        slow_save_taxonomy.apply_async(args=(build.build_id, build.build_type), countdown=10)
 
         logger.info(
             f"Finished processing service component {service_component['name']} "
--- a/corgi/tasks/managed_services.py
+++ b/corgi/tasks/managed_services.py
@@ -35,7 +35,12 @@ class MultiplePermissionExceptions(Exception):
         super().__init__(combined_message)
 
 
-@app.task(base=Singleton, autoretry_for=RETRYABLE_ERRORS, retry_kwargs=RETRY_KWARGS)
+@app.task(
+    base=Singleton,
+    autoretry_for=RETRYABLE_ERRORS,
+    retry_kwargs=RETRY_KWARGS,
+    soft_time_limit=settings.CELERY_LONGEST_SOFT_TIME_LIMIT,
+)
 def refresh_service_manifests() -> None:
     """Collect data for all Red Hat managed services from product-definitions
     and App Interface, then remove and recreate all services"""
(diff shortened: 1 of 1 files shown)


corgi/core/models.py
```diff
@@ -1012,6 +1012,7 @@ def released_components(self, include: bool = True) -> "ComponentQuerySet":
             software_build__relations__type__in=(
                 ProductComponentRelation.Type.ERRATA,
                 ProductComponentRelation.Type.COMPOSE,
+                ProductComponentRelation.Type.APP_INTERFACE,
```
jasinner: I don't think this is a good idea. We don't want these to accidentally make their way to the customer portal, as it's RH internal information. Griffon should work without them and meet the 'manifest' requirement that way.

In other words instead of having a flat 'SBOM' for each managed service, the requirement should be that the api/v1/components?ofuri=< managed service ofuri > returns the correct components for the service.

jasinner:
Apart from the the released components relation type changes, this looks good to me. Thanks.

corgi/core/models.py
juspence: To meet the manifest requirement, Florencio is using trying to search using the `griffon service product-manifest` command to see all the components in one combined list, preformatted nicely with all the information our manifest documents include.

Even if we make this change, the managed service streams still should not publish manifests to the portal, right? We have the "no_manifest" tag on the stream which prevents this. SDEngine will not publish any manifests for any streams with this tag, I thought.

But we still need to fix the manifest generation locally, even if we don't publish them to the Customer Portal for external customers. Then Griffon searches like above will work and the manifests will exist for internal stakeholders to use.

juspence: Discussed in standup - we really don't want to risk problems here so I removed this commit. But it's actually not possible for VM to generate product manifests on demand like we thought.

The released_components() filter still applies whether we pregenerate manifests ourselves or ask users to generate them on demand. So the managed services streams still have no released components and there's nothing to manifest. Using the component-based manifests for each root component in the stream (there are only a few) is still a possible workaround.

juspence:
@jasinner Thanks for reviewing. I removed the commit to treat managed services as released, like we discussed in standup. I've also added another commit to handle CORGI-923, but haven't merged it yet due to some issues.

Right now, we have lots of tasks in our slow queue, so the daily slow_save_taxonomy() tasks we need for mangen can't run. My fix to save the taxonomies inline caused timeouts, so I switched to using Celery's custom task / queue priorities instead.

But after testing locally, tasks without an explicit priority default to 0 / the highest priority. That means all the current Pyxis tasks in the slow queue, which were already sent without a priority, will go to the front of the queue and still block any new taxonomy-saving tasks. This is true even if I give the taxonomy-saving tasks an explicit priority 0 / the highest priority.

Is it safe to merge and deploy this, then clear out those Pyxis tasks? We can reschedule them immediately after / rerun the migration, and then they'll go to the back of the queue instead. See my change to the migration, which sets them to priority 9 instead.

EDIT: Actually to get everything working, I had to make much more invasive changes. Going to open a separate PR for these shortly, and going to just merge this one once pipeline works.

JimFuller-RedHat:
sounds like a good plan to me

PR #690, opened 2023-11-24
CORGI-859 exclude modular srpms from latest

jasinner:
Adjust the root_components filter and latest_components stored procedure to exclude modular SRPM components. I didn't adjust the indexes for the stored procedure as it was a little complex to do so, but a bit of testing showing that we are still doing an index scan on the existing index eg:

```
In [1]: import corgi.core.models as m

In [2]: stream = m.ProductStream.objects.get(name="rhel-9.1.0.z")

In [3]: print(stream.components.root_components().filter(name="apache-commons-io").values("nevra").explain)
<bound method QuerySet.explain of <ComponentQuerySet [{'nevra': 'apache-commons-io:1-2.8.0-8.el9.src'}, {'nevra': 'apache-commons-io:1-2.8.0-7.el9.src'}]>>

In [4]: print(stream.components.root_components().filter(name="apache-commons-io").values("nevra").explain())
Nested Loop  (cost=1.97..22.29 rows=1 width=69)
  ->  Index Scan using compon_latest_type_name_idx on core_component  (cost=0.85..8.81 rows=6 width=85)
        Index Cond: (name = 'apache-commons-io'::text)
        Filter: (((arch)::text <> 'src'::text) OR ((release)::text !~~ '%.module%'::text) OR ((type)::text <> 'RPM'::text))
  ->  Index Only Scan using core_component_productst_component_id_productstre_10bd2bae_uniq on core_component_productstreams  (cost=1.12..304.78 rows=270 width=16)
        Index Cond: ((component_id = core_component.uuid) AND (productstream_id = '1ba5a9d3-865a-4ee5-aa16-f05809e80e6d'::uuid))
```

and

```
print(stream.components.root_components().latest_components(ofuri=stream.ofuri,include_inactive_streams=True).explain())
Nested Loop  (cost=1.97..4382.88 rows=1 width=1117)
  ->  Index Scan using compon_latest_idx on core_component  (cost=0.85..4334.04 rows=22 width=1117)
        Index Cond: (uuid = ANY ('{fc989eb8-78e5-402d-a3df-b9f5ee9a3e56,a76efb0e-0c69-43e6-844c-2b54436c80a6,...,af17c072-16e4-4759-923c-af01d5094501}'::uuid[]))
        Filter: (((arch)::text <> 'src'::text) OR ((release)::text !~~ '%.module%'::text) OR ((type)::text <> 'RPM'::text))
[...]

PR #700, opened 2023-12-01
Minor: fix performance test

JimFuller-RedHat:
The new count for display_product_stream_with_many_roots() is correct as per @jasinner ... adjusting performance test to reflect recent change in modular_srpm exclusion (#690 ).


juspence (approved):
LGTM, thanks for fixing!

PR #701, opened 2023-12-01
CORGI-923: Don't save taxonomy inline to avoid timeouts, use task priorities instead

juspence:
@RedHatProductSecurity/corgi-devs Please review again, sorry. My other attempt to just save the taxonomies inline, as part of the service manifesting task, caused lots of timeouts. So this very hacked-together and invasive change is needed - or another approach if you can think of a better one. This change means we end up with 4 lists in Redis for each queue in Celery - one each for priorities 0 through 2, 3 through 5, 6 through 8, and 9 (the maximum).

I only added priorities to tasks in the slow queue, since only this queue is getting clogged currently. We can prioritize tasks in the CPU queue as well, or change my guesses about priorities, if needed in the future.

Tasks scheduled manually, or in code, without any priority argument get the priority= defined in the task definition, or else the highest (0) priority by default. This is probably what we want.

Daily tasks, as well as e.g. taxonomy-saving tasks, have the highest priority. Once we create data, I assume we want to link it and make it searchable in the API as soon as possible.

Less common tasks, e.g. errata or PNC tasks, don't seem to fill up our queues too often. So I've given these priority 3 to be in the next queue.

Brew and Pyxis tasks are very common and seem to clog the queues fairly often. I've given all these tasks priority 6 by default so they're in their own queue to start with, but we could adjust in the future depending on the context. For example, slow_fetch_brew_build() triggered by our UMB listener could have priority 3, but slow_fetch_brew_build() triggered by itself (for nested builds) could have priority 6 instead.

Bulk-data loading tasks are the least important and should always have priority 9 to be in the last queue.

If this looks good, we'll need to merge and deploy this, clear out all the tasks from the recent migration, then reschedule them to be in the correct (last) queue with the lowest / 9 priority. The current tasks were sent without any priority argument, so will get the highest / 0 priority by default.


jasinner (approved):
This looks good to me, thank you. I tried it out in my development environment briefly and didn't find any issues with the functionality.

PR #702, opened 2023-12-04
dont recheck container names from pyxis on failed migration

jasinner:
If a migration takes longer than the timeout period for an init-container in OpenShift we have to run it again manually. That means that if a migration takes longer than 10 min it fail. This change prevents the 0105_get_latest_repo_names_from_pyxis migration from rechecking the same containers if it's run more than once.


jasinner:
@juspence  I'm going to merge this now to prevent the 106, and 107 migrations from being run in stage, and avoid needing to undo those later. Once [pull/701](https://github.com/RedHatProductSecurity/component-registry/pull/701) is merged we should purge the celery slow queue and run the 105 migration manually using `python3 manage.py migrate core` to add message priorities to the `slow_update_name_for_container_from_pyxis` tasks started by this migration.

PR #703, opened 2023-12-04
Minor: tweak re_provides_name so it is more stable during ingestion

JimFuller-RedHat:
django generates sql that is more or less stable in terms of query planner ... this seems to be ok during times of high ingestion

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.core.validators import EMPTY_VALUES
-from django.db.models import QuerySet, Subquery
+from django.db.models import QuerySet
 from django.http import Http404
 from django_filters.rest_framework import BooleanFilter, CharFilter, Filter, FilterSet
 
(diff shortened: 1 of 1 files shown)


corgi/api/filters.py
```diff
@@ -187,10 +188,14 @@ def filter_provides_name(
         if value in EMPTY_VALUES:
             # User gave an empty ?param= so return the unfiltered queryset
             return queryset
-        source_uuids = (
-            queryset.prefetch_related("sources").filter(name=value).values("sources").distinct()
+        sources = list(
+            queryset.prefetch_related("sources")
+            .filter(name=value)
```
juspence, line 193: @JimFuller-RedHat I think this is going to prefetch the "sources" links for every row in the queryset, then filter the queryset to only the row(s) with the matching name. Doing below instead should prefetch only the sources for the row(s) with the matching name, if I understand correctly:
```suggestion
            queryset.filter(name=value)
            .prefetch_related("sources")
```

But is the logic in these filter\_\*provides\_\*() methods correct? Based on the name, I'd expect them to:
1. Find all parent components, regardless of name
2. Filter above to show only parent components, where the child / provided component has a matching name (or purl)

Instead, this method is actually:
1. Finding only parent components with a matching name (or purl)
2. Returning all the provided components of the above, whether or not the child / provided component has a matching name (or purl)

JimFuller-RedHat: that prefetch will go away ... needed to test in situ on stage ... also this function is correct in matching name eg. filter_re_provides_name filter function does name__iregex test.

Note: this testing is about what exact sql is being generated by django ... we now emitting a nice minimal:

```
SELECT DISTINCT "core_component_sources"."to_component_id"
  FROM "core_component"
  LEFT OUTER JOIN "core_component_sources"
    ON ("core_component"."uuid" = "core_component_sources"."from_component_id")
 WHERE "core_component"."name"::text ~* 'webkitgtk'
```

PR #705, opened 2023-12-05
CORGI-602 create inferred variants types from brew tag matching

jasinner:
This is the first part of resolving CORGI-602 and is required in order to group by variant for latest_components stored procedure. The key difference here is that instead of also relating all components to these variants, we skip that linking for inferred types. I also took advantage of this new linking to remove the 'brew_tags_from_variants' attribute of ProductStream, and replace it with a property which uses the new inferred variants. Next is to complete CORGI-811 which is why I left a few TODOs in the prod_defs update_products task


jasinner:
@RedHatProductSecurity/corgi-devs please review. There are some failing tests, I'll fix those with this next PR. I don't plan on merging this as is because I think it might cause Variants to move back and forth between streams on runs of the update_products task. I'll include these commits in a follow up PR which will add many-to-many relationships between the product models.

tests/test_products.py
```diff
@@ -505,6 +505,60 @@ def test_brew_tag_matching(variant_name, cpe, brew_tag, stream_name, requests_mo
...
+    variant_name = "8Base-Quay-3"
+    cpe = "cpe:/a:redhat:quay:3::el8"
+    CollectorErrataProductVariant.objects.create(
+        et_id=3042, name=variant_name, product_version=et_product_version, cpe=cpe
+    )
+
+    result = _parse_variants_from_brew_tags(brew_tags)
+    assert result == {variant_name: cpe}
+
+    # Now add another brew_tag to the ProductVersion, and verify a single variant is returned
```
juspence, line 527: The test makes sense, but I'm not sure how useful it is in terms of correctness. You only create one CollectorErrataProductVariant above, so there's only ever one variant in the DB that can be returned.

It might be better to create multiple variants for the same version, as well as different versions with one or two variants each. In the latter case, you could also test both versions with non-overlapping Brew tags (version 1: "tag A", version 2: "tag B") and overlapping Brew tags (version 1: "tag A", version 2: "tag A").

tests/test_products.py
```diff
@@ -505,6 +505,60 @@ def test_brew_tag_matching(variant_name, cpe, brew_tag, stream_name, requests_mo
...
+@pytest.mark.django_db
+def test_create_inferred_variants():
+    variant_name = "8Base-Quay-3"
+    cpe = "cpe:/a:redhat:quay:3::el8"
+    variants_and_cpes = {variant_name: cpe}
+    product_stream = ProductStreamFactory()
+    product = product_stream.products
+    product_version = product_stream.productversions
+    stream_node = ProductNode.objects.create(parent=None, object_id=product_stream.pk, obj=product_stream)
+    _create_inferred_variants(variants_and_cpes, product, product_version, product_stream, stream_node)
```
juspence: This doesn't quite match real code...I would recommend creating the full taxonomy structure using the Product as the root node, the ProductVersion as the child node, and the ProductStream as the grandchild node, just to be thorough / make sure the test and code work the same way.

corgi/tasks/prod_defs.py
```diff
@@ -215,19 +215,46 @@ def parse_product_version(
...
         trimmed_brew_tag, sans_container_released = brew_tag_to_et_prefixes(brew_tag)
-        variants = CollectorErrataProductVersion.objects.filter(
+        variant_data = CollectorErrataProductVersion.objects.filter(
             brew_tags__overlap=[trimmed_brew_tag, sans_container_released]
-        ).values_list("variants__name", flat=True)
-        variant_names.update(variants)
-    return list(variant_names)
+        ).values("variants__name", "variants__cpe")
+        for variant in variant_data:
+            distinct_variants[variant["variants__name"]] = variant["variants__cpe"]
```
juspence: Should we exclude variants without any CPE?

You can also use .values_list() here, then just unpack the tuple. We only care about the name / CPE values and not the keys themselves, so there's no need to ask for a dict:
```suggestion
        ).values_list("variants__name", "variants__cpe")
        for variant_name, variant_cpe in variant_data:
            if not variant_cpe:
                continue
            distinct_variants[variant_name] = variant_cpe
```

corgi/tasks/prod_defs.py
```diff
@@ -215,19 +215,46 @@ def parse_product_version(
...
+    for variant, cpe in variants_and_cpes.items():
+        variant, created = ProductVariant.objects.update_or_create(
+            name=variant,
+            defaults={
+                "cpe": cpe,
+                "products": product,
+                "productversions": product_version,
+                "productstreams": product_stream
+            }
+        )
```
juspence: Should this be a get_or_create() here? Otherwise we'll find the existing Variant, which might be linked to many streams, and update it so that it's only linked to one stream (the one given).
```suggestion
        variant, created = ProductVariant.objects.get_or_create(
            name=variant,
            defaults={
                "cpe": cpe,
                "products": product,
                "productversions": product_version,
                "productstreams": product_stream,
            }
        )
        if not created:
            # Atomically link this existing Variant to some new Products / Versions / Streams
            # Harmless and will not add duplicates if any models given above were already linked
            variant.products.add(product)
            variant.productversions.add(product_version)
            variant.productstreams.add(product_stream)
```

corgi/core/migrations/0107_add_product_node_type.py
```diff
@@ -0,0 +1,46 @@
...
+            model_name="productnode",
+            name="unique_pnode_get_or_create",
+        ),
+        migrations.RemoveConstraint(
+            model_name="productnode",
+            name="unique_pnode_get_or_create_for_null_parent",
+        ),
+        migrations.AddField(
+            model_name="productnode",
+            name="type",
```
juspence: "type" is a built-in Python name and reusing it as a field name will cause problems. Can we call the field node_type instead?

corgi/core/migrations/0107_add_product_node_type.py
```diff
@@ -0,0 +1,46 @@
...
+                choices=[("DIRECT", "Direct"), ("INFERRED", "Inferred")],
+                default="DIRECT",
+                max_length=20,
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="productnode",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("parent__isnull", False)),
+                fields=("object_id", "parent", "type"),
```
juspence: The old condition said that the same ProductModel object couldn't appear more than once underneath the same parent. Do we expect a Variant to be linked to a ProductStream twice, as both an inferred Variant and a direct Variant?

If not, let's leave the constraint unchanged.

corgi/core/migrations/0107_add_product_node_type.py
```diff
@@ -0,0 +1,46 @@
...
+                name="unique_pnode_get_or_create",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="productnode",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("parent__isnull", True)),
+                fields=("object_id", "type"),
```
juspence: Same as above - let's leave the constraint unchanged if we aren't going to link the same Variant to the same Stream twice.

corgi/core/models.py
```diff
@@ -110,7 +126,7 @@ class Meta(NodeModel.Meta):
 
     @staticmethod
     def get_node_pks_for_type(
-        qs: QuerySet["ProductNode"], mapping_model: type, lookup: str = "__pk"
+        qs: QuerySet["ProductNode"], mapping_model: Type[object], lookup: str = "__pk"
```
juspence, line 127: Using "type" as a variable name is causing a collision here, and while looking for alternatives I found out typing.Type is deprecated as well. I'd recommend using "node_type" for the field name instead.

tests/test_manifests.py
```diff
@@ -313,20 +318,20 @@ def test_manifest_cpes_from_cpe_lookup(stored_proc):
...
+    assert stream.cpes == (
+        "cpe:/a:redhat:test:1",
+        "cpe:/o:redhat:enterprise_linux:8",
+    )
 
     manifest = json.loads(stream.manifest)
     product_data = manifest["packages"][-1]
     cpes_in_manifest = set(ref["referenceLocator"] for ref in product_data["externalRefs"])
-    assert {test_cpe, another_test_cpe} == cpes_in_manifest
+    assert set(sorted([test_cpe, variant.cpe])) == cpes_in_manifest
```
juspence: There's no reason to call set(sorted()) because the outer set() will remove any ordering applied by the inner sorted(). Python sets have no defined ordering.

tests/test_manifests.py
```diff
@@ -313,20 +318,20 @@ def test_manifest_cpes_from_cpe_lookup(stored_proc):
 
 @pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
 def test_manifest_cpes_from_patterns_and_brew_tags(stored_proc):
-    stream, variant = setup_product()
+    stream, variant = setup_product(variant_node_type=ProductNode.ProductNodeType.INFERRED)
```
juspence, line 321: This is changing the assumption of the test - before, we were checking that a stream with directly-linked variants reported the correct CPEs, then deleting the Variant and checking the stream's direct .cpes, cpes_from_brew_tags, and cpes_matching_patterns all had the correct results.

Can we keep this test as-is, using directly-linked variants, and add a new test for the new inferred variants?

corgi/core/models.py
```diff
@@ -422,25 +424,67 @@ def builds(self) -> QuerySet:
 
     @property
     def cpes(self) -> tuple[str, ...]:
-        """Return CPEs for all descendant variants."""
-        variant_cpes = (
+        """Return CPEs for direct descendant variants if they exist. Otherwise, return a union
+        of indirect descendant variants and descendant streams cpes_matching_patterns"""
```
juspence, line 425: Can we keep this property and its usages unchanged, then add a new indirect_cpes property for only the places where we need CPEs from inferred variants or cpes_matching_patterns?

We could add a wrapper property all_cpes that returns this property (direct CPEs) plus the indirect_cpes, if we need both together in more than a few places. And in that case it might make sense to rename this property direct_cpes instead.

I don't want to force the logic to be combined here since there might be times we only want the direct CPEs, only the indirect ones, etc. Combining both types here also makes the logic hard to follow and means a major change to many of our existing tests.

corgi/tasks/prod_defs.py
```diff
@@ -433,16 +471,16 @@ def parse_errata_info(
...
+                #  possible. If not possible create both
                 node, node_created = ProductNode.objects.get_or_create(
+                    parent=product_stream_node,
                     object_id=product_variant.pk,
-                    defaults={
-                        "parent": product_stream_node,
-                        "obj": product_variant,
-                    },
+                    type=ProductNode.ProductNodeType.DIRECT,
+                    defaults={"obj": product_variant},
```
juspence: I thought we would only create inferred variants where a direct variant didn't already exist. I don't understand the comment about the Quarkus stream having both.

If we really do have both Variant types under the same parent, I guess this is correct. But if only one Variant type exists for a given stream, let's just use parent= and object_id= as the keyword arguments.

corgi/tasks/errata_tool.py
```diff
@@ -290,15 +290,13 @@ def _get_errata_search_criteria(stream_name) -> tuple[list[str], list[int]]:
...
-            f"Tried to load errata for a stream without brew_tags or variants: {stream_name}"
-        )
-    return variant_names, release_ids
+    release_ids = stream.meta_attr.get("releases_from_brew_tags", [])
+    if release_ids:
+        return [], release_ids
+    # Could be a stream with no build info, composes or yum_repositories
+    logger.warning(
+        f"Tried to load errata for a stream without brew_tags or variants: {stream_name}"
+    )
```
juspence, line 310: I think before, we preferred to return a stream's variants_from_brew_tags, and only returned releases_from_brew_tags if the variants didn't exist.

Now we only report release_ids, but the if condition isn't needed. You can just return `[]. release_ids` in all cases, and that will return the same `[], []` tuple of empty lists.

juspence (commented):
I need to look over this much more thoroughly...the motivation makes sense but too many of the implementation details are escaping me.

For now here's some feedback to chew on, and I'll try to finish the rest ASAP.

juspence:
> I also took advantage of this new linking to remove the 'brew_tags_from_variants' attribute of ProductStream, and replace it with a property which uses the new inferred variants.

This sounds fine but can we do it (and maybe the CPE changes too, if possible) in a separate PR? I want to keep changes here as minimal as possible so I don't miss something when reviewing.

corgi/tasks/prod_defs.py
jasinner: In the next PR which adds a many-to-many relationship between ProductVariant and ProductStream I've allowed the productstreams field of ProductVariant to be null so that we can avoid specifying those stream relationships up front and instead create the Product graph first, then populate the productstreams relationship of variant. Using update_or_create will allow the cpes to be updated, although I think that should never happen.

corgi/core/migrations/0107_add_product_node_type.py
jasinner: Good point, I'll update this and the migration.

corgi/core/migrations/0107_add_product_node_type.py
jasinner: Good point, I'll update this and the migration.

corgi/core/models.py
jasinner: The field name is mapping_model, 'type' here was the type of the object. I started getting mypy errors the way it was, which is why I updated it.

tests/test_manifests.py
jasinner: ok

tests/test_manifests.py
jasinner: The assumptions of the tests are the same. The field cpes_from_brew_tags is not longer a list, but instead a property calculated from the inferred variants.

corgi/core/models.py
jasinner: If you want the inferred cpes you can still use cpes_from_brew_tags. I moved this functionality from files.py class which was used to calculate the cpes for that case, leaving behind the hardcoded CPES for RHEL GA etc. Keeping it here means that the API and manifest will report the same CPE list.

corgi/tasks/prod_defs.py
jasinner: I'll update this to change the type to DIRECT for any matching stream parent and variant object_id.

jasinner:
> > I also took advantage of this new linking to remove the 'brew_tags_from_variants' attribute of ProductStream, and replace it with a property which uses the new inferred variants.
> 
> This sounds fine but can we do it (and maybe the CPE changes too, if possible) in a separate PR? I want to keep changes here as minimal as possible so I don't miss something when reviewing.

I probably shouldn't have said "took advantage of" because brew_tags_from_variants attributes is the main reason we were inspecting brew_tag on variants previously. It's really a key piece of the existing functionality which should be changed as part of this PR.

corgi/tasks/errata_tool.py
jasinner: The check is there to see if we need to print the warning or not.

corgi/core/migrations/0107_add_product_node_type.py
jasinner: ok.

tests/test_products.py
jasinner: I've fixed this in the next PR.

corgi/tasks/prod_defs.py

tests/test_products.py
jasinner: Sure, that would be nice, but it's all a matter of time. This is better than what was in the code before, and I don't think this area of the code is buggy. It's been working in production for quite some time, and I haven't changed how it works.

jasinner:
Closed in favor of https://github.com/RedHatProductSecurity/component-registry/pull/719

PR #706, opened 2023-12-05
Minor: remove unneeded prefetch

JimFuller-RedHat:


juspence (commented):
@JimFuller-RedHat Removing the prefetch is fine. But I still don't understand the way we do filtering, and you reported it's not showing the right results anyway.

We use a method= keyword argument / custom code we wrote to filter provides_name, re_provides, and re_provides_name. But we use a built-in Django filter lookup (for example, name__iregex on line 111) to filter provides, sources, sources_name, re_sources, and re_sources_name. Which approach is right?

I'm expecting the \*provides\* and \*sources\* filters to show "all components which have the name (or purl) given as a provide / source". So if I enter "bash" into provides_name, I should see all the root components (generally SRPMs index containers and modules, but sometimes others) which have "bash" listed in their "provides" field.

And If I enter "bash" into sources_name, I should see all the child components which are bundled into the bash binary RPMs, as well as the bash binary RPMs themselves (as children of the bash source RPM). I should also see child components (if any) that are included in some "bash" container or RPM module. These all have "bash" listed in their "sources" field.

I think right now these filters are reversed. First the \*provides\* filters find components with the matching name / purl, then they list all the sources of the matched components. So we don't filter on the contents of the "provides" field at all.

I think this is causing the bug you reported with missing container / OCI-type results. Is my understanding correct, or is this a docs bug? Let's also confirm we have enough tests, the tests are correct, and the assertions in them haven't been changed recently.

PR #711, opened 2023-12-06
CORGI-854: Ignore permissions errors in Quay

juspence:
@RedHatProductSecurity/corgi-devs Please review. I tested the first commit in stage and it wasn't enough to fix all the issues. So I added the third commit to deal with specific known errors, and not hide every single kind of subprocess error like we were doing before.

The second commit should finish off the migration using memory-safe, race-condition free .update() functions. So the migration shouldn't be killed halfway through due to running out of memory, can still make progress if my rsh session times out (atomic = False), and won't corrupt or lose data due to running at the same time as other Celery tasks / DB queries.


jasinner (approved):
LGTM

So if the last part of the migration fails to remove the named_check keys we remove that part, re-run the migration, so it completes, and then either leave the keys or remove them manually?

PR #717, opened 2023-12-08
Fix miscellaneous bugs

juspence:
@RedHatProductSecurity/corgi-devs Please review. The commit messages have more details, but these changes are needed to fix incorrect results in the API, errors in our monitoring email, and so on.

### DIFF
--- a/corgi/tasks/sca.py
+++ b/corgi/tasks/sca.py
@@ -18,6 +18,7 @@ from config.celery import app
 from corgi.collectors.brew import Brew
 from corgi.collectors.go_list import GoList
 from corgi.collectors.syft import Syft
+from corgi.core.constants import ROOT_COMPONENTS_CONDITION
 from corgi.core.models import Component, ComponentNode, SoftwareBuild
 from corgi.tasks.common import RETRY_KWARGS, RETRYABLE_ERRORS
 
@@ -206,8 +207,12 @@ def cpu_software_composition_analysis(build_uuid, force_process: bool = False):
 
     component_qs = Component.objects.filter(software_build=software_build)
     try:
-        # Get root component for this build.
-        root_component = component_qs.root_components().exclude(type=Component.Type.RPMMOD).get()
+        # Get root component for this build, using the condition instead of the queryset
+        # because the predefined root_components() Django queryset also excludes modular SRPMs
+        # but we must still run SCA on these
+        root_component = (
+            component_qs.filter(ROOT_COMPONENTS_CONDITION).exclude(type=Component.Type.RPMMOD).get()
+        )
     except Component.DoesNotExist as exc:
         # None of the components were root components
         module_qs = component_qs.filter(type=Component.Type.RPMMOD)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * added **epoch** to component filters, ex. /api/v1/components?epoch=0
 * added **re_downstreams_name** to component filters, ex. /api/v1/components?re_downstreams_name=foo
 
+### Changed
+* Exclude modular source RPMs (type="RPM", arch="src", release__contains=".module") from manifests,
+and from the API when using the root_components=True filter
+
 ## [1.4.1] - 2023-11-15
 
 ### Changed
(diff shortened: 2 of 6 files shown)


corgi/api/filters.py
```diff
@@ -102,16 +102,16 @@ class Meta:
...
     upstreams_name = CharFilter(field_name="upstreams", lookup_expr="name", distinct=True)
     downstreams = CharFilter(lookup_expr="purl")
 
     # otherwise use regex to match provides,sources or upstreams purls
     re_sources = CharFilter(field_name="sources", lookup_expr="purl__iregex", distinct=True)
     re_sources_name = CharFilter(field_name="sources", lookup_expr="name__iregex", distinct=True)
-    re_provides = CharFilter(method="filter_re_provides_purl", distinct=True)
-    re_provides_name = CharFilter(method="filter_re_provides_name", distinct=True)
+    re_provides = CharFilter(field_name="provides", lookup_expr="purl__iregex", distinct=True)
+    re_provides_name = CharFilter(field_name="provides", lookup_expr="name__iregex", distinct=True)
```
JimFuller-RedHat, line 114: do not change this

JimFuller-RedHat:
I would hold off on any changes for now - things are in a lot of motion ... should calm down soon.

corgi/api/filters.py
juspence: Sure, per discussion on Slack I've removed this for now. Please let me know when you've finished testing and I'll do a separate PR for this fix if still needed

corgi/core/models.py
```diff
@@ -1046,7 +1046,7 @@ def manifest_components(self, ofuri: str, quick=False) -> "ComponentQuerySet":
         non_container_source_components = self.exclude(name__endswith="-container-source").using(
             "read_only"
         )
-        roots = non_container_source_components.root_components()
+        roots = non_container_source_components.root_components().exclude(MODULAR_SRPM_CONDITION)
```
jasinner: This has the effect of hiding both the non-modular and modular SRPM from the manifest. I think it's because of the interaction between root_components and latest_components. When the latest_components function runs it again calls root_components (on line 943 of models.py) and includes the modular SRPM, so it get it's component UUID as the latest, but doesn't find that in the original query so filters both out. I found this by re-generating the rhel-9.1.0.z manifest on this branch and greping the manifest for 'apache-commons-io'. We only find the binary rpm with nevra 'apache-commons-io-2.11.0-2.module+el9.2.0+18047+4ad4bba4.noarch'. But neither of the apache-commons-io SRPMs.

@JimFuller-RedHat  We probably need to adjust the latest_components function so that it doesn't include a call to root_components again. I'm thinking perhaps we should incorporate the grouping done in _latest_components_func directly into the stored procedure?

Anyway, I think we should leave this commit out of this PR, and revisit the requirements when we do CORGI-823

jasinner:
I'm fine with the other changes in this PR, apart from the changes related to the root_component condition.

corgi/core/models.py
juspence: Bah, thanks for letting me know. The goal here is just to fix the silly SCA errors about "couldn't find a root component to run on".I thought we could achieve that by only excluding the modular SRPMs from the manifest and not from the API / queryset filters that the SCA task uses.

I guess I'll just fix the task instead, so that it can find either an SRPM (including a modular SRPM) or index container, but not a module.

corgi/tasks/sca.py
```diff
@@ -206,8 +207,12 @@ def cpu_software_composition_analysis(build_uuid, force_process: bool = False):
 
     component_qs = Component.objects.filter(software_build=software_build)
     try:
-        # Get root component for this build.
-        root_component = component_qs.root_components().exclude(type=Component.Type.RPMMOD).get()
+        # Get root component for this build, using the condition instead of the queryset
+        # because the predefined root_components() Django queryset also excludes modular SRPMs
+        # but we must still run SCA on these
+        root_component = (
+            component_qs.filter(ROOT_COMPONENTS_CONDITION).exclude(type=Component.Type.RPMMOD).get()
```
JimFuller-RedHat, line 214: ah interesting, nice catch

corgi/core/models.py
JimFuller-RedHat: @jasinner re comment about root_components and latest components ... sounds reasonable to me

JimFuller-RedHat (approved):
**LGTM**

corgi/core/models.py
juspence: Fixed in latest commit - I've updated only the task logic for finding the root components to run SCA on, and left the actual ROOT_COMPONENTS_CONDITION / root_components() queryset unchanged.

This does mean modular SRPMs will now be excluded from the API as well as from manifests, so I've added a CHANGELOG.md note describing this.

EDIT: Oops, I didn't refresh this page so missed some of the comments posted here. I merged based on Jim's review below...please let me know if I should revert this one part.

PR #718, opened 2023-12-08
CORGI-854: Fix timeouts for large services and git errors for buggy repos

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have time. This should fix (I hope) the last two errors in our monitoring email for managed services tasks (other than known and expected permissions issues).

I'm out of time for this week, but will go back and look over tests next week, then fill in any gaps given all the recent additions and changes.


corgi/core/models.py
```diff
@@ -324,16 +324,19 @@ def save_product_taxonomy(self):
 
         return None
 
-    def disassociate_with_product(self, product_model_name: str, product_pk: str) -> None:
-        """Remove the product references from all components associated with this build."""
+    def disassociate_with_product(
```
jasinner, line 327: This looks fine. More an FYI. I have changed the way this works in PR712 to clear and reset the relations for each component related to the software build based on what's in the relations table. It will make this change redundant, but if you need it now please merge as is.

jasinner (approved):

corgi/core/models.py
juspence: I think that will still work...the final version of this code deletes the old relation in the relations table, then calls this method, so the function should still remove all the old service names. Then we create new data, add a new relation, and save the new taxonomies.

There are probably other things we can do to speed things up so maybe we can fix the redundant deletion logic in the future. I've run in stage for a few days and everything seems to work, plus I think I'm out of time on this one, so will merge as-is. Thanks for reviewing!

PR #719, opened 2023-12-12
CORGI-811 many to many models

jasinner:
After PR#704 created a INFERRED ProductNode type, this PR changes the relationship between streams and variants to be many-to-many.


jasinner:
@RedHatProductSecurity/corgi-devs please review. I plan to merge this PR before completing CORGI-602.

I know 'mypy' is failing but I'm not sure how to fix it. Seems the ProductModel as the wrong productstreams type, but I'm not sure what I should set it to.

jasinner:
See the CORGI-811 task for a comparison on the product hierarchy from stage compared with the one produced with these code changes in dev.

corgi/tasks/prod_defs.py
```diff
@@ -214,19 +214,46 @@ def parse_product_version(
...
+        )
+        if created:
+            logger.info(f"Created ProductVariant {variant} as inferred variant of {product_stream.name} ProductStream")
+
+        # TODO set type to INFERRED
+        ProductNode.objects.get_or_create(
+            object_id=variant.pk,
+            parent=parent,
+            defaults={"obj": variant}
+        )
```
juspence: Weren't these TODOs supposed to be fixed in this PR? Or is there a third PR that will address these? Either way I think you can just add a `"type": ProductNode.ProductNodeType.INFERRED` key to the defaults dict here.

Then an INFERRED node will be created when either a DIRECT or INFERRED node doesn't already exist for this variant and parent. Or am I imissing something about how we need to create these?

corgi/tasks/prod_defs.py
```diff
@@ -214,19 +214,46 @@ def parse_product_version(
...
+
+def _create_inferred_variants(variants_and_cpes: dict[str, str], product: Product, product_version: ProductVersion, product_stream: ProductStream, parent: ProductNode) -> None:
+    for variant, cpe in variants_and_cpes.items():
+        variant, created = ProductVariant.objects.update_or_create(
+            name=variant,
+            defaults={
+                "cpe": cpe,
+                "products": product,
+                "productversions": product_version,
+                "productstreams": product_stream
```
juspence: I still don't understand how update_or_create() is supposed to work here. If variants and streams now have a many-to-many relationship, I think we can't set these properties on creation (or update them) and have to use `variant.productstreams.add(product_stream)` instead.

This is because Django doesn't allow you to directly assign values to the reverse (many) side of a related set, even if you're only setting a single value. You have to use the special methods like .add(), .set(), or .clear() instead.

Even if I'm mistaken and this code works, won't it give the wrong results? If the Variant already exists and will be updated, we want to keep it linked to any existing streams, and just .add() a new link to the product_stream given, right? The code we have now will unlink it from all the other streams and relink to only the one given.

corgi/tasks/prod_defs.py
```diff
@@ -214,19 +214,46 @@ def parse_product_version(
...
+            }
+        )
+        if created:
+            logger.info(f"Created ProductVariant {variant} as inferred variant of {product_stream.name} ProductStream")
+
+        # TODO set type to INFERRED
+        ProductNode.objects.get_or_create(
+            object_id=variant.pk,
+            parent=parent,
```
juspence, line 259: I think the code we have here is correct - we do get_or_create() on all the fields that appear together in a unique constraint. And we want to say "both the variant and the parent are unique together", so that the same variant can be linked to multiple parents - it's not enough to say only the variant by itself is unique.

But in other places, we still do get_or_create() using only an object_id= keyword argument, which may give the wrong results now that the unique constraint has been changed. I think we now need to update all of the get_or_create() and update_or_create() usages for the ProductNode model (not just the ones in this module) to use both object_id= and parent= as keyword arguments.

One example is line 328 below, but I can't comment on it directly because it's outside the diff window.

corgi/tasks/prod_defs.py
```diff
@@ -214,19 +214,46 @@ def parse_product_version(
...
-        variant_names.update(variants)
-    return list(variant_names)
+        ).values("variants__name", "variants__cpe")
+        for variant in variant_data:
+            distinct_variants[variant["variants__name"]] = variant["variants__cpe"]
+
+    return distinct_variants
+
+
+def _create_inferred_variants(variants_and_cpes: dict[str, str], product: Product, product_version: ProductVersion, product_stream: ProductStream, parent: ProductNode) -> None:
```
juspence: Will the parent argument ever be different from product_stream.pnodes.get()? In other words, will a stream ever have more than 1 node, or 0 nodes? Or will a Variant ever be linked to something other than a stream?

If not, maybe we can get rid of the parent arg and just rely on looking up the given product_stream's node directly.

And I guess the same applies to the other arguments besides product_stream. This method could reuse product_stream.products and product_stream.productversions instead of having to pass in a separate value (and risking bugs if we pass the wrong one).

corgi/tasks/prod_defs.py
```diff
@@ -302,6 +330,9 @@ def parse_product_stream(
             "obj": product_stream,
         },
     )
+
+    _create_inferred_variants(variants_and_cpes, product, product_version, product_stream, product_stream_node)
+
     parse_errata_info(errata_info, product, product_stream, product_stream_node, product_version)
```
juspence: Should we create the INFERRED variants last, after we create the DIRECT variants from errata_info? That way we always default to having DIRECT variant nodes, and only create INFERRED variant nodes if some matching DIRECT variant node doesn't already exist.

tests/test_products.py
```diff
@@ -14,12 +14,12 @@
...
     ProductStream,
     ProductVariant,
-    ProductVersion,
+    ProductVersion, ProductNode,
 )
 from corgi.tasks.prod_defs import (
     _find_by_cpe,
     _match_and_save_stream_cpes,
-    update_products,
+    update_products, _parse_variants_from_brew_tags, _create_inferred_variants,
```
juspence: I think you'll need to fix the formatting here before the pipeline will pass.

corgi/core/migrations/0107_add_product_node_type.py
```diff
@@ -0,0 +1,46 @@
...
+                name="unique_pnode_get_or_create",
+            ),
+        ),
+        migrations.AddConstraint(
+            model_name="productnode",
+            constraint=models.UniqueConstraint(
+                condition=models.Q(("parent__isnull", True)),
+                fields=("object_id", "type"),
+                name="unique_pnode_get_or_create_for_null_parent",
+            ),
```
juspence: I guess based on this code here, we still expect to have both DIRECT and INFERRED nodes for the same Variant linked to the same parent? LGTM if so. Or if not, let's enforce uniqueness on only the object_id and parent for the first constraint.

For the second constraint, do we expect to ever create both DIRECT and INFERRED Variant nodes as the root of some tree, without any parent stream nodes at all? I think the answer is no, so technically the econd condition can just stay unchanged.

corgi/core/models.py
```diff
@@ -125,20 +141,6 @@ def get_node_pks_for_type(
...
-        # "product_version" -> "ProductVersion"
-        mapping_model = target_model.title().replace("_", "")
-        # No .distinct() since __name on all ProductModel subclasses + Channel is always unique
-        return (
-            product_model.pnodes.get()
-            .get_family()
-            .filter(level=MODEL_NODE_LEVEL_MAPPING[mapping_model])
-            .values_list(f"{target_model}__name", flat=True)
-        )
-
```
juspence, line 141: Was this renamed to get_related_names_of_type() below, or is this being removed for some other reason?

corgi/core/models.py
```diff
@@ -421,17 +423,49 @@ def builds(self) -> QuerySet:
 
     @property
     def cpes(self) -> tuple[str, ...]:
-        """Return CPEs for all descendant variants."""
-        variant_cpes = (
+        """Return CPEs for direct descendant variants if they exist. Otherwise, return a union
+        of indirect descendant variants and descendant streams cpes_matching_patterns"""
+        if self is ProductVariant:
+            return (self.cpe,)  # type: ignore[attr-defined]
```
juspence, line 423: I think the old code handled the case where we access the some_variant.cpes property instead of some_variant.cpe directly. Did you see a bug with that code?

If not, we can remove this and rely on the old code - `if direct_variant_cpes` should be True so we should return only this Variant's CPE.

Or we can keep this bit, and then we can remove the include_self=True in the call to get_descendants(). There may also be other assumptions that need to be cleaned up.

corgi/core/models.py
```diff
@@ -421,17 +423,49 @@ def builds(self) -> QuerySet:
...
+            .get_descendants(include_self=True)
+            .using("read_only")
+            .filter(
+                level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"],
+                type=ProductNode.ProductNodeType.INFERRED,
+            )
+            .values_list("productvariant__cpe", flat=True)
+            .distinct()
+        )
+        distinct_cpes_matching_patterns = set([cpe for cpe_patterns in cpes_matching_patterns for cpe in cpe_patterns])
```
juspence: I think we should remove all the .distinct() calls. They're wasted effort and a little misleading - even if the DB returns distinct values, the same CPE could technically appear in both indirect_variant_cpes and cpes_matching_patterns. So your logic at the end here (line 462) is the correct solution, and the extra .distinct() calls aren't needed.

You can also save a few CPU cycles by getting rid of the outer [] square brackets on line 462. The outer set() constructor will take a "value for value in container" generator expression directly, so there should be no need to build the inner list first and then convert it to a set - you can just build the set directly.

It would also be nice if we could move this logic for indirect CPEs to its own method, for easier reading and testing. We can just call this new indirect_cpes() method from the end of the cpes property and return whatever values it finds, if there are no direct variant cpes.

corgi/core/models.py
```diff
@@ -421,17 +423,49 @@ def builds(self) -> QuerySet:
...
+        distinct_cpes_matching_patterns.update(self.qs_to_tuple_no_empty_strings(indirect_variant_cpes))
+        return tuple(distinct_cpes_matching_patterns)
```
juspence: I think it's easier and less risky to just call .update() first, then convert the whole set to a tuple once at the end of the method. Then if we ever touch this code in the future, we don't risk adding back an empty CPE to our results if we add a new way of looking up CPEs (and forget to strip the empty CPEs from it).
```suggestion
        distinct_cpes_matching_patterns.ipdate(indirect_variant_cpes)
        return self.qs_to_tuple_no_empty_strings(distinct_cpes_matching_patterns)
```

corgi/core/models.py
```diff
@@ -421,17 +423,49 @@ def builds(self) -> QuerySet:
...
+
+    def qs_to_tuple_no_empty_strings(self, direct_variant_cpes: QuerySet) -> tuple[str, ...]:
```
juspence, line 466: If you change above, you can use Iterable for the type hint instead:
```suggestion
    def qs_to_tuple_no_empty_strings(self, direct_variant_cpes: Iterable[str]) -> tuple[str, ...]:
```

corgi/core/models.py
```diff
@@ -488,6 +522,23 @@ def save_product_taxonomy(self):
         # All ProductModels have a set of descendant channels
         self.channels.set(channels)
 
+    def get_related_names_of_type(self, mapping_model: type) -> list[str]:
+        """For a given ProductModel instance find all directly related nodes with the given model type
+        and return their names"""
+        mapping_key = mapping_model.__name__
+        # No .distinct() since __name on all ProductModel subclasses + Channel is always unique
+        attribute_name = mapping_key.lower()
+        pnode = self.pnodes.exclude(type=ProductNode.ProductNodeType.INFERRED).first()
```
juspence: In the old code, I used .get() to enforce that models had exactly one node. There are probably other places with this assumption that will need to be fixed.

For this code here, I get that we sometimes expect a Variant to have multiple nodes now. But can we raise an error if no node is found at all?

It's also helpful if you can edit the existing code instead of deleting the old code and adding the new code, so that the diff is easier to parse. You can do the editing in one commit, then move the method in another commit, so commit-by-commit reviewing still works without having to jump around to compare stuff.

corgi/core/models.py
```diff
@@ -488,6 +522,23 @@ def save_product_taxonomy(self):
         # All ProductModels have a set of descendant channels
         self.channels.set(channels)
 
+    def get_related_names_of_type(self, mapping_model: type) -> list[str]:
+        """For a given ProductModel instance find all directly related nodes with the given model type
+        and return their names"""
+        mapping_key = mapping_model.__name__
+        # No .distinct() since __name on all ProductModel subclasses + Channel is always unique
```
juspence: We do use .distinct() now, since multiple nodes can point to the same model, so I think this comment can go away.

corgi/core/models.py
```diff
@@ -488,6 +522,23 @@ def save_product_taxonomy(self):
...
+        attribute_name = mapping_key.lower()
+        pnode = self.pnodes.exclude(type=ProductNode.ProductNodeType.INFERRED).first()
+        if pnode:
+            return list(
+                pnode.get_family()
+                .filter(level=MODEL_NODE_LEVEL_MAPPING[mapping_key])
+                .values_list(f"{attribute_name}__name", flat=True)
+                .distinct()
```
juspence: Based on line 531 above, I assume we only want to report names for related models / nodes with DIRECT type, not INFERRED type. Should we filter on type=DIRECT instead of excluding type=INFERRED?

I think we need to put that logic here (filter DIRECT or exclude INFERRED) as well. Currently, we might find e.g. a stream with a DIRECT-type node, then look for the productvariant__name values of all its children. Then we'd report both DIRECT-type and INFERRED-type variant children of a DIRECT-type stream. Or maybe that's correct?

Having the filtering logic above means we won't report any names (e.g. stream parents) for an INFERRED variant node, even though in this case (starting from a Variant) all the ancestor names will always have DIRECT type. Is that correct, or should we move the filtering logic to here and then get rid of above?

corgi/core/models.py
```diff
@@ -848,18 +899,18 @@ class Meta:
...
     In other words, builds relate to variants and streams through the PCR table
     Components relate to builds, and products / versions relate to variants / streams
     We want to link all build components to their related products, versions, streams, and variants
     """
-    if variant_names:
-        variant_streams = ProductVariant.objects.filter(name__in=variant_names).values_list(
-            "productstreams__name", flat=True
-        )
+    for variant in ProductVariant.objects.filter(name__in=variant_names):
+        variant_streams = variant.get_related_names_of_type(ProductStream)
```
juspence, line 915: I think we're trying to avoid including stream parents of INFERRED variants, right? Can we just keep the old code, but add a little more filtering logic like below, instead of needing to write a custom method and make many DB calls for each Variant?
```suggestion
    if variant_names:
        variant_streams = ProductVariant.objects.filter(name__in=variant_names, pnodes__type=ProductNode.ProductNodeType.DIRECT).values_list("productstreams__name", flat=True)
```
I think if I've correctly understood the code changes I've seen so far, we'll create nodes for INFERRED-type variants. And we'll link those inferred Variant objects to the parent stream, just like the DIRECT-type Variants.

But even if a Variant has both types of nodes, won'tl they both point to the same parent stream? Is there ever a case where a Variant is a direct child of one stream and an inferred child of a different stream? If not, then I think we can just tweak the old code slightly and still get the correct stream names (only when a direct relationship exists).

If Variants can point to two different streams using direct and indirect relationships, then I need to go back and redo this whole review...

Do we ever expect a Variant to have more than one INFERRED-type node, whether or not it has a DIRECT-type node? In other words, can one Variant really get linked to lots of streams? Or will Variants always be linked to exactly 1 stream (either DIRECT or INFERRED) or 2 streams (both DIRECT and INFERRED)?

corgi/core/models.py
```diff
@@ -1780,6 +1831,7 @@ def cpes(self) -> QuerySet:
         """Build and return a list of CPEs from all Variants this Component relates to"""
         # For each Variant-type relation, get the linked Variant's CPE directly
         # Remove any duplicates, and return the CPEs in sorted order so manifests are stable
+        # TODO review this now that we have INFERRED variants
         return (
             self.productvariants.exclude(cpe="")
```
juspence: I think you can just filter a little more and still get the correct results:
```suggestion
            self.productvariants.filter(pnodes__type=ProductNode.ProductNodeType.DIRECT).exclude(cpe="")
```

juspence (commented):
Still not done but there's a lot here...sorry for the dumb questions. I'm also looking commit-by-commit so maybe you already fixed some of these in a later commit, and I just haven't gotten to it yet.

Feel free to ignore if so, I'll delete comments that don't apply anymore once I finish the whole review.

corgi/core/models.py
jasinner: I wasn't being used, so I renamed and moved it also I believe.

corgi/core/models.py
jasinner: I guess this check is superfluous because we override the cpes property for ProductVariant model. Although I think it doesn't hurt to keep it here. 
I'll remove the  include_self=True in the call to get_descendants() as it is redundant.

corgi/core/models.py
jasinner: Ok, I'll clean this up. Thanks.

corgi/core/models.py
jasinner: We don't want to return an error, we just don't want to return anything.

corgi/core/models.py
jasinner: We do expect that a Variant will have more than one INFERRED-type node. One Variant can get linked to lots of streams. What I changed based on your feedback from the previous PR was removing 'type' from the uniqueness constraint of ProductNode so that we can only ever have one relationship between the same obj and parent (variant and stream), so in that case (which occurs today for quarkus-2.13) we set the type to DIRECT.

We could inline this call to get_related_names_of_type. But I use it in more than one place, with a differnt mapping model in either call.

corgi/core/models.py
juspence: Why not though? Don't we expect a Variant (whether direct or inferred) to always have at least one node?

corgi/core/migrations/0111_save_variant_components.py
```diff
@@ -0,0 +1,44 @@
...
+
+from corgi.core.models import ProductVariant, ProductNode, ProductComponentRelation
+from corgi.tasks.common import slow_save_taxonomy
+
+
+def save_product_taxonomy_for_inferred_variants(apps, schema_editor) -> None:
+    meta_attr_key = "inferred_variants_saved"
+    inferred_variants = set()
+    for pnode in ProductNode.objects.filter(
+        level=3, node_type=ProductNode.ProductNodeType.INFERRED
```
juspence, line 13: Instead of hardcoding level=3 here, can you import and use the constant PRODUCT_NODE_LEVEL_MAPPING or whatever it's called?

corgi/core/migrations/0111_save_variant_components.py
```diff
@@ -0,0 +1,44 @@
...
+            )
+        checked_variant = ProductVariant.objects.get(name=inferred_variant)
+        checked_variant.meta_attr[meta_attr_key] = True
+        checked_variant.save()
+
+
+class Migration(migrations.Migration):
+    # rely on slow_save_taxonomy being a task singleton to eliminate duplicate tasks if this
+    # migration is run in quick succession (due to init-container timeout)
+    atomic = False
```
juspence, line 39: This migration isn't safe to run outside of a transaction. You need to use .filter() and .update() instead of .save(), remove `atomic = False` here, or wrap the "fetch and update" logic above in a `with transaction.atomic()` block.

corgi/core/migrations/0111_save_variant_components.py
```diff
@@ -0,0 +1,44 @@
...
+            .values("build_id", "build_type")
+            .distinct()
+        ):
+            kwargs = software_build_details
+            kwargs["save_components"] = False
+            slow_save_taxonomy.apply_async(
+                kwargs=kwargs,
+                priority=6,
+            )
+        checked_variant = ProductVariant.objects.get(name=inferred_variant)
```
juspence, line 31: You add pnode.obj.name to the set, and then try to look up the matching variant object here. I think it's simpler to just add pnode.obj to the set, and then set `inferred_variant.meta_attr[meta_attr_key] = True` directly.

corgi/core/models.py
```diff
@@ -424,28 +424,27 @@ def cpes(self) -> tuple[str, ...]:
...
         direct_variant_cpes = (
             self.pnodes.get_queryset()
-            .get_descendants(include_self=True)
+            .get_descendants()
             .using("read_only")
             .filter(level=MODEL_NODE_LEVEL_MAPPING["ProductVariant"])
             .exclude(node_type=ProductNode.ProductNodeType.INFERRED)
             .values_list("productvariant__cpe", flat=True)
             .order_by("productvariant__cpe")
-            .distinct()
```
juspence: Do we still need the .distinct() here, in case we return direct_variant_cpes directly?

corgi/core/models.py
```diff
@@ -424,28 +424,27 @@ def cpes(self) -> tuple[str, ...]:
...
             .filter(level=MODEL_NODE_LEVEL_MAPPING["ProductStream"])
             .values_list("productstream__cpes_matching_patterns", flat=True)
-            .distinct()
-        )
-        distinct_cpes = set(
-            [cpe for cpe_patterns in cpes_matching_patterns for cpe in cpe_patterns]
         )
+        distinct_cpes = set(cpe for cpe_patterns in cpes_matching_patterns for cpe in cpe_patterns)
         distinct_cpes.update(self.distinct_inferred_variant_cpes())
         return tuple(sorted(distinct_cpes))
```
juspence, line 450: If nothing else will call distinct_inferred_variant_cpes() directly, we can simplify this code (below) and stop doing extra work by removing the conversion to a tuple, and by removing .distinct() on line 461. That method can just return a queryset of cpes, without evaluating the queryset.

If other code will call it directly, we should continue to use .distinct() and strip empty values, then return a tuple.

corgi/core/models.py
```diff
@@ -424,28 +424,27 @@ def cpes(self) -> tuple[str, ...]:
...
 
+        return self._get_indirect_cpes()
+
+    def _get_indirect_cpes(self):
         cpes_matching_patterns = (
             self.pnodes.get_queryset()
             .get_descendants(include_self=True)
             .using("read_only")
             .filter(level=MODEL_NODE_LEVEL_MAPPING["ProductStream"])
             .values_list("productstream__cpes_matching_patterns", flat=True)
```
juspence: The logic here doesn't seem correct. `_get_indirect_cpes()` may return empty strings because we don't strip these values from cpes_matching_patterns. We only strip empty "" string CPEs from inferred_variant_cpes (later / below).

On line 447, we can update the set of distinct_cpes with the queryset of inferred_variant_cpes. There's no need to convert to a tuple first - set.update() should work with any iterable argument.

After we combine all the values we want into one set, we should convert that to a tuple with no empty values - let's do this conversion only once at the very end, so we don't forget. Then we'd return `self.qs_to_tuple_no_empty_strings(distinct_cpes)` / strip the empty CPEs on line 449.

corgi/core/models.py
```diff
@@ -424,28 +424,27 @@ def cpes(self) -> tuple[str, ...]:
...
-            .distinct()
-        )
-        distinct_cpes = set(
-            [cpe for cpe_patterns in cpes_matching_patterns for cpe in cpe_patterns]
         )
+        distinct_cpes = set(cpe for cpe_patterns in cpes_matching_patterns for cpe in cpe_patterns)
         distinct_cpes.update(self.distinct_inferred_variant_cpes())
         return tuple(sorted(distinct_cpes))
```
juspence, line 449: Is it important to sort the tuple of CPEs we return? We do this sorting here but not in two other places. If yes, let's add the sorting logic as part of the `qs_to_tuple_no_empty_strings()` method. Otherwise, let's remove sorted() from line 449 below.

corgi/core/models.py
```diff
@@ -536,7 +534,7 @@ def _get_direct_pnode_over_inferred(self):
         """Always return a DIRECT pnode if it exists, otherwise return INFERRED"""
         result = None
         for pnode in self.pnodes.get_queryset():
-            if pnode.type == ProductNode.ProductNodeType.DIRECT:
```
juspence: Maybe I missed this before but can't we just filter the pnodes queryset, so that we only return exactly the nodes we want from the DB, without having to iterate?
```
node = self.pnodes.filter(type=ProductNode.ProductNodeType.DIRECT).first()
if node:
    # Prefer the first direct node, if any
    return node

# If the Variant exists, it must be linked to something / have at least one node
# but we have no DIRECT type nodes, so return the INFERRED type node that must exist
node = self.pnodes.exclude(type=ProductNode.ProductNodeType.DIRECT).first()
if node:
    return node

# else raise an error, this shouldn't happen
```

corgi/tasks/prod_defs.py
```diff
@@ -470,14 +472,11 @@ def parse_errata_info(
...
                 )
-                # TODO linking this variant to multiple streams. quarkus-2.13 stream for example now
-                #  has both DIRECT and INFERRED relationships. Prefer DIRECT over INFERRED if
-                #  possible. If not possible create both
                 node, node_created = ProductNode.objects.get_or_create(
                     parent=product_stream_node,
                     object_id=product_variant.pk,
-                    type=ProductNode.ProductNodeType.DIRECT,
-                    defaults={"obj": product_variant},
+                    content_type=ContentType.objects.get_for_model(ProductVariant),
```
juspence: Sorry if I'm confused but why did this change? When reviewing commit-by-commit and you fix my comments in a new commit, I can't see the older discussion / history. I think you need to include keyword arguments that match only the fields in a unique constraint - above, that's just parent and object_id. That will do a get() lookup using an index to find a matching node.

If the get() fails, we want to create a new node with DIRECT type and link the node to its related object. In that case, you can just add both "node_type" and an "obj" key to the defaults dict below. Looking up the content_type as a keyword argument above will give the wrong results.

tests/conftest.py
```diff
@@ -69,6 +70,22 @@ def setup_product(
...
+    pvariant_node = ProductVariantNodeFactory(obj=pvariant, parent=psnode, node_type=variant_node_type)
+
+=======
+    pvariant_node = ProductVariantNodeFactory(
+        obj=pvariant, parent=psnode, node_type=variant_node_type
+    )
+>>>>>>> Stashed changes
+    variant = pvariant_node.obj
+
+>>>>>>> Stashed changes
```
juspence: It looks like something didn't unstash properly here - was this committed by mistake?

tests/factories.py
```diff
@@ -140,6 +140,63 @@ class Meta:
     # link model using reverse relationship to child models
     tag = factory.RelatedFactory(ProductVariantTagFactory, factory_related_name="tagged_model")
 
+<<<<<<< Updated upstream
+<<<<<<< Updated upstream
+=======
+=======
+>>>>>>> Stashed changes
```
juspence: More stuff that didn't unstash properly.

tests/test_model.py
```diff
@@ -89,6 +90,12 @@ def test_cpes():
         products=p1,
         productversions=pv2,
         productstreams=ps2,
+=======
+    ProductVariantNodeFactory(obj=pvariant2, parent=ps3_node)
+    pvariant3 = ProductVariantFactory(cpe="cpe:/o:redhat:inferred_variant:8")
+    ProductVariantNodeFactory(
+        obj=pvariant3, parent=ps3_node, node_type=ProductNode.ProductNodeType.INFERRED
+>>>>>>> Stashed changes
```
juspence: More stash issues in this file as well.

tests/test_products.py
```diff
@@ -239,7 +239,57 @@ def test_stream_variants_with_old_builds(mock_remove, requests_mock):
 
 
 @pytest.mark.django_db
+<<<<<<< Updated upstream
```
juspence: Last stash issues I see in this commit.

corgi/tasks/prod_defs.py
juspence: This is still relevant, you've added a "type" key but removed this "obj" key. We need both "type" and "obj" in the defaults dict, then only object_id= and parent= keyword arguments above.

corgi/core/models.py
```diff
@@ -487,9 +551,29 @@ def save_product_taxonomy(self):
 
         # All ProductModels have a set of descendant channels
         self.channels.set(channels)
+        self.ofuri = self.get_ofuri()
```
juspence, line 554: Why are we now setting ofuri only when the taxonomy is saved, and not each time we call .save()? Won't this cause problems for new variants that have just been created and don't have their taxonomies saved yet?

I guess maybe the issue is having a stale ofuri, if the taxonomy / parent stream for a variant changes? I think it's better for the ofuri to change both during the model .save() and whenever the taxonomy changes / is saved. That gives us a value when the variant is first created, and updates the value if the variant gains a new direct parent.

jasinner:
Closed in favour of 

https://github.com/RedHatProductSecurity/component-registry/pull/722
and
https://github.com/RedHatProductSecurity/component-registry/pull/723

PR #720, opened 2023-12-14
Select minimum set of components for latest filters

JimFuller-RedHat:
* order is significant in django **api/filters.py**
* ensured we narrow down active streams in **latest_components_by_streams** when desired
* ensured we select minimal set of components in **_latest_components_func**

**Note**: this PR will be applied to stage and then I can run final statistic run to confirm all is well with concomitant ingestion activity - will revert if that test is negative.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -128,6 +128,12 @@ class ComponentFilter(FilterSet):
         method="filter_root_components",
         label="Show only root components (source RPMs, index container images)",
     )
+
+    active_streams = BooleanFilter(
+        label="Show components from active streams",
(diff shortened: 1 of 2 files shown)


corgi/api/filters.py
```diff
@@ -128,6 +128,12 @@ class Meta:
         method="filter_root_components",
         label="Show only root components (source RPMs, index container images)",
     )
+
```
JimFuller-RedHat, line 131: order is significant in django filters - this ensures **active_streams** is applied before **latest_components_by_streams** to avoid superfluous processing

PR #721, opened 2023-12-14
Remove custom code that's identical to old Django-filter, fix queryset ordering

juspence:
@JimFuller-RedHat Quick review please. Like we discussed in my last PR, this removes the custom code for the \*provides\* filter methods. I see the latest version is now the same as the old Django-filter code we had before, and I'd like to avoid maintaining custom logic whenever possible.

I also tweaked the latest_components() function slightly - there was one queryset that used .distinct() without clearing the ordering inherited from the self / parent queryset. This may give the wrong results in some cases, if the parent queryset defines a non-default ordering.

### DIFF
--- a/corgi/api/filters.py
+++ b/corgi/api/filters.py
@@ -102,7 +102,7 @@ class ComponentFilter(FilterSet):
     sources = CharFilter(lookup_expr="purl")
     sources_name = CharFilter(field_name="sources", lookup_expr="name", distinct=True)
     provides = CharFilter(lookup_expr="purl")
-    provides_name = CharFilter(method="filter_provides_name", distinct=True)
+    provides_name = CharFilter(field_name="provides", lookup_expr="name", distinct=True)
     upstreams = CharFilter(lookup_expr="purl")
     upstreams_name = CharFilter(field_name="upstreams", lookup_expr="name", distinct=True)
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat (approved):
apart from one comment about ordering of active_streams, LGTM!

corgi/api/filters.py
```diff
@@ -164,35 +158,10 @@ class Meta:
...
-    @staticmethod
-    def filter_re_provides_name(
-        queryset: ComponentQuerySet, _name: str, value: str
-    ) -> QuerySet["Component"]:
-        """Filter components by name regex search on component provides."""
-        if value in EMPTY_VALUES:
-            # User gave an empty ?param= so return the unfiltered queryset
-            return queryset
-        return queryset.filter(provides__name__iregex=value).distinct()
+    active_streams = BooleanFilter(
```
JimFuller-RedHat, line 161: please put this back - ordering was intentional eg. we want active streams to filter before latest_components

juspence: Thanks a bunch for catching this! I put the ordering back and added a comment to remind us about needing to filter active before latest. Feel free to move latest to the bottom as well (for safety) if you'd like to.

PR #722, opened 2023-12-19
CORGI-811 create inferred variants

jasinner:
This commit is about replacing the cpes_from_brew_tags attributes with a property. The new property of ProductStream looks up Variants linked with INFERRED ProductNode types. The ProductStream class does included INFERRED variant types directly in it's productvariants foreign key relationship. 

Instead of combining cpes_from_patterns with cpes_from_brew_tags when generating a manifest, we now do it in the ProductModels `cpe` property.

I replaced the unused existing get_node_names_for_type with a new method, get_related_names_of_type looks for related ProductNodes of a particular type, but excludes INFERRED variants. It's used from the save_product_taxonomy function to avoid associating INFERRED variants with Components.

In the next PR I will address the issue of multiple streams sharing a common INFERRED variant by making that relationship many-to-many. Also the create_inferred_variants function doesn't populate the foreign keys of the ProductVariant it creates as in the next PR I populate those using the ProductNode tree. 

Ignore any TODOs are they are addressed in the next PR. Ignore all tests are they are updated in the next PR.


jasinner:

jasinner:
It's not possible to make all the test pass here, and this can't be merged stand-alone because it will cause INFERRED variants to move back and forth between streams. We really need the changes here to be combined with the changes in PR#723

PR #723, opened 2023-12-19
CORGI-811 many to many models 2

jasinner:
This is an updated version of PR#719 which fixes some of the issues identified in that PR. It also moves all the changes into 2 commits, with detailed commit messages explaining the assumptions made in each. I've incorporated the changes from earlier reviews into their respective commit.


corgi/core/models.py
```diff
@@ -685,15 +744,13 @@ class ProductVariant(ProductModel):
...
     )
     productversions = models.ForeignKey(
-        "ProductVersion", on_delete=models.CASCADE, related_name="productvariants"
+        "ProductVersion", on_delete=models.CASCADE, related_name="productvariants", null=True
     )
     # Below creates implicit "productvariants" field on ProductStream
-    productstreams = models.ForeignKey(
-        ProductStream, on_delete=models.CASCADE, related_name="productvariants"
-    )
+    productstreams = models.ManyToManyField(ProductStream, related_name="productvariants")
```
jasinner, line 753: Need some help to fix mypy failure here.

jasinner:

corgi/core/models.py
```diff
@@ -522,27 +548,29 @@ def save_product_taxonomy(self):
...
+                    .filter(level=MODEL_NODE_LEVEL_MAPPING[mapping_key])
+                    .exclude(node_type=ProductNode.ProductNodeType.INFERRED)
+                    .values_list(f"{attribute_name}__name", flat=True)
+                    .distinct()
+                )
+                results.extend(list(query))
+        return results
 
     def save(self, *args, **kwargs):
-        self.ofuri = self.get_ofuri()
```
jasinner, line 495: We can't create the ofuri until the ProductNode tree is created.

corgi/tasks/prod_defs.py
```diff
@@ -453,35 +435,28 @@ def parse_errata_info(
...
                         "meta_attr": {
                             "et_product": et_product_name,
                             "et_product_version": et_pv_name,
                         },
                     },
                 )
-                node, node_created = ProductNode.objects.get_or_create(
+                node, node_created = ProductNode.objects.update_or_create(
                     parent=product_stream_node,
                     object_id=product_variant.pk,
```
jasinner, line 449: Can't use obj here because it causes a failure related to reverse lookup of relationship. That's why I've split obj to object_id, and content_type. Although the content_type here should never change.

juspence: These kinds of errors make me suspect there are more bugs in this change. Can you share exactly what error you saw? I also think in general we should:

Create errata_info variants and nodes first
Use `update_or_create(parent=stream_node, object_id=variant.pk, defaults={"type": "DIRECT", "obj": variant})` so that an existing node with INFERRED type is updated to have DIRECT type. This handles the case when an existing stream is updated to use errata_info instead of brew_tags in prod-defs.

Create brew_tag variants and nodes last
Use `get_or_create(parent=stream_node, object_id=variant.pk, defaults={"type": "INFERRED", "obj": variant})` so that an existing node with DIRECT type is preferred / reused, and we only create an INFERRED node if a DIRECT node for the same variant / parent combo doesn't already exist. This handles the case when a stream defines both brew_tags and errata_info.

jasinner: Ok, I made this change, and also added a test for the scenario where an inferred variant is changed to an direct one.

PR #725, opened 2023-12-19
CORGI-956: Release 1.4.2

juspence:
@RedHatProductSecurity/corgi-devs FYI. I'll keep an eye on our failed tasks email tomorrow and try to fix any blockers before shutdown. But I hope this is a relatively safe change.


juspence:
The cherry-picked commit here is tagged as "1.4.2" and deployed to production. This patch release includes only this change, not everything else in main / stage currently.

PR #726, opened 2023-12-20
CORGI-602 manifest latest by variant

jasinner:
Includes commits from #723 

This fixes a bug with the latest_components stored_procedure which allows it to work with ProductVariants. It also adjusts the latest_components filter to calculate the latest package for each ProductVariant in a ProductStream where multiple ProductVariants are related.


corgi/core/migrations/0113_remove_active_streams_filter_from_latest_variants_cursor.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 3.2.22 on 2023-12-15 05:41
```
juspence: I think we already merged all the migrations before this one, right? We just need this new one to update the latest filter after your changes in constants.py?

corgi/core/models.py
```diff
@@ -1002,15 +1032,23 @@ def latest_components(
...
-                return Component.objects.none()
-            else:
-                # Show only the older / non-latest components
-                # But there are no latest components to hide??
-                # So show everything / return unfiltered queryset
-                return self
+        elif include:
+            # no latest components, don't do any further filtering
+            return Component.objects.none()
+        return self
```
juspence, line 1034: It might be useful to keep the comment explaining what case the final "return" statement is handling, after all the "if" / "elif" statements above.

tests/test_model.py
```diff
@@ -1571,6 +1571,80 @@ def test_latest_filter_with_inactive(stored_proc):
     assert len(latest_components) == 0
 
 
+@pytest.mark.django_db(databases=("default", "read_only"), transaction=True)
```
juspence, line 1579: Do these tests need access to the read-only DB? Shouldn't all the queryset methods / all the logic in models.py be using the regular writable DB instance? "read_only" should only be used for read-only API functions I think.

corgi/core/models.py
```diff
@@ -984,16 +985,45 @@ def latest_components(
...
-        components = self.root_components().prefetch_related(product_prefetch)
-
-        latest_components_uuids = set(
-            self._latest_components_func(components, model_type, ofuri, include_inactive_streams)
-        )
+        latest_components_uuids: set[str] = set()
+        if model_type == "ProductStream":
+            stream_with_multiple_variants = (
+                ProductStream.objects.annotate(variants_count=Count("productvariants"))
+                .filter(ofuri=ofuri, variants_count__gte=2)
```
juspence: I don't know if this will work or not, but it might be better to filter on ofuri first, then annotate the variants_count for only the one ofuri we're interested in. Annotating an entire model like we do here (all the streams in the ProductStreams table) killed the DB when I tried this in another place.

I say I don't know if it will work, since trying the other way (filter first, then annotate) seemed to give incorrect results for me. In that case, I was filtering and annotating the same model (find all X components, then annotate Y components). I think here it should be OK, since we want to filter the streams, then look up and annotate their variants. So the result you want to annotate shouldn't be filtered out.

corgi/core/models.py
```diff
@@ -984,16 +985,45 @@ def latest_components(
...
+                .first()
```
juspence: I see above we don't enforce that ofuri is unique across all models of a given type. Since there could be two streams with the same ofuri, I think we should add a check that this queryset had exactly 1 result (if we found a matching ofuri), or 0 results (if there was no matching ofuri, or if that stream had only 1 variant). Then we can raise an error if more than one matching stream was found.

corgi/core/models.py
```diff
@@ -984,16 +985,45 @@ def latest_components(
...
+            if stream_with_multiple_variants:
+                # calculate the latest uuids for each variant
+                for variant_ofuri in stream_with_multiple_variants.productvariants.values_list(
+                    "ofuri", flat=True
+                ):
+                    variant_latest_uuids = self._latest_pks_by_ofuri(
+                        components,
+                        "ProductVariant",
+                        variant_ofuri,
+                        include_inactive_streams=True,
```
juspence: Do we still need this part here after your change to constants.py above? Should we pass through the original value of include_inactive_streams? Or should we not pass it through at all, since it shouldn't apply when finding latest pks for variant ofuris?

corgi/core/models.py
```diff
@@ -1002,15 +1032,23 @@ def latest_components(
...
+        components: "ComponentQuerySet",
+        model_type: str,
+        ofuri: str,
+        include_inactive_streams: bool = False,
+    ) -> set[str]:
+        product_prefetch = f"{model_type.lower()}s"
+        components.prefetch_related(product_prefetch)
+        return set(
+            self._latest_components_func(components, model_type, ofuri, include_inactive_streams)
+        )
```
juspence: I think you can return an Iterable[str] just like _latest_components_func does. I believe you can call set() or set.update() with just an iterable (e.g. a Django queryset, doesn't have to be a set already), and then we avoid doing lots of unneeded / redundant set conversions.

The final result after set.update() will still be a set like you're expecting, and I think that means we won't report the same component version / NEVRA twice, even if it appears in multiple variants for this stream. Maybe that's worth adding a test for.

juspence (approved):
I'm not sure I fully understand all the implications here, but what I do see LGTM.

Before we merge this, though, I think we need to fix some existing bugs. As I mentioned in a previous review, the current code assumes that each ProductModel instance (e.g. one stream or one variant) always has exactly one node.

We should have, but didn't. update all the usages of ProductNode get(), get_or_create(), and update_or_create() to handle the case when there are multiple nodes for a single related object. This is causing errors in our daily monitoring email and blocking some data-loading tasks. Can you please check over all those usages and make sure they're updated, before we make any further changes here?

Maybe the new code assumes that streams / variants will never have more than one DIRECT-type node. Then we only have many INFERRED-type nodes for one linked object, or exactly one DIRECT-type node and one INFERRED-type node. If that's correct, we can just update all e.g. `ProductNode.objects.get(pk=related_obj.pk)` usages to look something like `ProductNode.objects.get(pk=related_obj.pk, type="DIRECT")` instead.

corgi/core/migrations/0113_remove_active_streams_filter_from_latest_variants_cursor.py
jasinner: yeah, that was a problem with my rebase sorry. Should be fixed now.

jasinner: yes, I fixed that problem with the rebase now, thanks.

tests/test_model.py
jasinner: Sorry, fixed this now.

jasinner:
> I'm not sure I fully understand all the implications here, but what I do see LGTM.
> 
> Before we merge this, though, I think we need to fix some existing bugs. As I mentioned in a previous review, the current code assumes that each ProductModel instance (e.g. one stream or one variant) always has exactly one node.
> 
> We should have, but didn't. update all the usages of ProductNode get(), get_or_create(), and update_or_create() to handle the case when there are multiple nodes for a single related object. This is causing errors in our daily monitoring email and blocking some data-loading tasks. Can you please check over all those usages and make sure they're updated, before we make any further changes here?
> 
> Maybe the new code assumes that streams / variants will never have more than one DIRECT-type node. Then we only have many INFERRED-type nodes for one linked object, or exactly one DIRECT-type node and one INFERRED-type node. If that's correct, we can just update all e.g. `ProductNode.objects.get(pk=related_obj.pk)` usages to look something like `ProductNode.objects.get(pk=related_obj.pk, type="DIRECT")` instead.

The problem is not with get_or_create but rather `pnodes.get()` for a ProductVariant object. I've added another commit which adds ProductNode linking Channels to each ProductVariant pnode instead now, and a corresponding test.

corgi/core/models.py
jasinner: I see what you mean. I switched to using ProductStream.objects.get(ofuri=ofuri) then getting a values_list of productvariant "ofuri" values instead.

corgi/core/models.py
jasinner: I switched to using ProductStream.objects.get(ofuri=ofuri) which will raise a DoesNotExist or MultipleObjectsReturned error.

corgi/core/models.py
jasinner: You are right, not passing it through at all seems best.

corgi/core/models.py
jasinner: I added it back.

PR #729, opened 2023-12-20
Fix bad query that killed DB, and start deleting Brew builds again

juspence:
No rush to review, just trying to clean out my TODO list before shutdown. This code has many tests already, and was running in stage for a few days, but was disabled because a bad query used up all the temporary space in the DB.

The new approach is copied from the rewritten mangen collector, which has been running in stage for about a week now and doesn't have the same issues. I didn't reuse that code exactly because deleting services builds vs. deleting Brew builds requires slightly different logic for what we clean up.


jasinner (approved):
LGTM, thanks.

JimFuller-RedHat:
@juspence / @jasinner is this still applicable ? If not should we close and revisit in the future ?

juspence:
Yes, this is still relevant. I didn't merge it when I opened it because it was right before shutdown, then I was busy with mangen and I wanted to release just those changes to prod before I made this one in stage.

It's probably fine to merge now since you guys did the release, but I see the pipeline failed so you can continue to ignore it if you're busy - it's not urgently needed, it just slows how quickly the DB size / number of Brew builds grows.

PR #730, opened 2023-12-20
Deduplicate failed tasks in monitoring email

juspence:
Credit to Martin for this one. I'm trying to clear out my TODO list before shutdown and finally got around to copying this very useful code from SDEngine.

It may not help us as much, since those SDEngine tasks usually have the same args / kwargs and errors can be deduplicated easily. In Corgi the failure-spam we see is usually for "fetch Brew build 1, 2, 3..." and so on, so there's very little to deduplicate. But it will help in some cases.


jasinner (approved):
LGTM, thanks

JimFuller-RedHat:
@juspence / @jasinner is this still applicable ? If not should we close and revisit in the future ?

JimFuller-RedHat:
hoping @juspence is ok with us merging this nice work

PR #733, opened 2024-01-09
CORGI-955: Add filtering to Pyxis UMB listener and reenable collection

juspence:
@RedHatProductSecurity/corgi-devs Please review when you have some time. This should solve some of the performance issues we were previously seeing with the Pyxis collector.

Previously we would collect all builds from Pyxis, 95% of which were built in Brew (by Metaxor) and were already available / were duplicated in Corgi through the Brew collector. We only need the extra 5% of Pyxis containers which are now being built in the new RHTAP build system (and are not available at all in Brew).

The queue which processes these tasks was constantly filling up, and many tasks to fetch the manifest from Pyxis were timing out. We therefore disabled the collector to avoid blocking other tasks. This PR will help with the first problem (we will have a lot less Pyxis tasks to process), so reenabling should be relatively safe.

The issue with timeouts may be due to large Brew container builds, so will be solved by this PR to only collect RHTAP container builds. But the timeouts may also be due to general Pyxis API slowness, in which case I'll need to do more research and then open a follow-up PR. I'll keep an eye on things over the next few days to see how full the queue gets, and if the tasks to fetch RHTAP container builds are slow or time out.

The ticket also has a link to a Gchat thread where this implementation is discussed / the Pyxis team confirmed the values to filter on if you're curious.

### DIFF
--- a/corgi/monitor/consumer.py
+++ b/corgi/monitor/consumer.py
@@ -11,8 +11,7 @@ from corgi.collectors.pnc import is_sbomer_product
 from corgi.tasks.brew import slow_fetch_brew_build, slow_update_brew_tags
 from corgi.tasks.errata_tool import slow_handle_shipped_errata
 from corgi.tasks.pnc import slow_fetch_pnc_sbom, slow_handle_pnc_errata_released
-
-# from corgi.tasks.pyxis import slow_fetch_pyxis_manifest
+from corgi.tasks.pyxis import slow_fetch_pyxis_manifest
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


corgi/monitor/consumer.py
```diff
@@ -206,16 +205,19 @@ def sbomer_complete(event: Event) -> bool:
...
-        json.loads(event.message.body)
+        message = json.loads(event.message.body)
+
         try:
-            pass
-            # Temporarily disabled due to slowness and lots of duplicate data
-            # slow_fetch_pyxis_manifest.delay(
-            #     message["entityData"]["_id"]["$oid"],
-            # )
+            created_by = message["entityData"]["created_by"]
```
jasinner, line 211: Should we also skip processing if the entityData/created_by field is missing, or you design it to fail in that case? I guess either is OK.

Perhaps we should add a test for that case also?

jasinner (approved):
Fine to merge as is. Just added a suggestion inline.

corgi/monitor/consumer.py
juspence: Yes, the code we had originally would also fail if the entityData was missing, so that behavior hasn't changed. I assume based on all the messages I've seen, and Ralph Bean's original tests, that this part of the message is required and always present.

It is possible for the "last_updated_by" key (similar to but different from the "created_by" key) and some other fields to be present with a null value. If this also happens for the "created_by" key, the string comparison will return false in that case, like we expect.

I assume based on the presence of keys with null values that the keys themselves will always be present, or if not we'll raise a KeyError exception which is handled on the new line 217 below. The old test covers the exception-handling logic.

PR #734, opened 2024-01-09
Revert "CORGI-811 many to many models 2" because it makes managed services tasks time out

juspence:
Reverts RedHatProductSecurity/component-registry#723 because it makes managed services tasks time out or fail due to deadlocks / transactions that now take way too long. I'm not merging this but I need the commit to exist in Github so I can pull it locally and start testing.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -2,7 +2,7 @@ import logging
 import re
 from abc import abstractmethod
 from typing import Any, Iterable, Iterator, Type, Union
-from uuid import uuid4
+from uuid import UUID, uuid4
 
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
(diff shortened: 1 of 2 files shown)


juspence:
@RedHatProductSecurity/corgi-devs Quick review please. Based on discussion in standup I'm not actually reverting a large part of PR #723's changes. Instead, I've only added back the old taxonomy-unlinking method. This should be safe since:

1) We're only using it to process managed service streams
2) These ProductStreams always have a many-to-one relationship with their ancestors, e.g. one ProductStream has exactly one parent ProductVersion and exactly one grandparent Product
3) These ProductStreams never have any child ProductVariants, so there are no many-to-many relationships we'd need to add special logic for

corgi/core/models.py
```diff
@@ -2152,6 +2152,70 @@ def get_upstreams_purls(self, using: str = "read_only") -> QuerySet:
         """Return only the purls from the set of all upstream nodes"""
         return self.get_upstreams_nodes(using=using).values_list("purl", flat=True).distinct()
 
+    def disassociate_with_product(self, product_refs: Iterable[ProductModel]) -> None:
+        """Disassociate this component with the passed in ProductModel and any child ProductModels
+        in that product's hierarchy. This is the reverse of what happens in save_product_taxonomy.
+        """
+        # DANGER: This code is only needed, and only safe, for managed service streams
```
JimFuller-RedHat, line 2159: u might consider embeding 'managed_services' directly in the function name

JimFuller-RedHat (approved):
a quick review LGTM ... unsure how we would test this (or if we even want to test this).

corgi/core/models.py
juspence: Yeah, probably better to make this code check for and work only on ProductStreams, not any other model. I was being lazy and just copied what we had before. Fixed now.

juspence:
Everything's timing out currently anyway, so this should improve things. I'll deploy without merging tomorrow morning and rerun the managed services data-loading tasks to see if it helps, then merge it if so.

juspence:
This seems to help with timeouts, doesn't seem to cause any other issues, and is really just putting back the code we had before, so going to merge.

PR #735, opened 2024-01-15
only calculate latest for all variants when generating a manifest

jasinner:
I noticed this change had a major impact on latest filter performance, so switched to only using it when generating manifests.

### DIFF
--- a/corgi/core/models.py
+++ b/corgi/core/models.py
@@ -992,6 +992,7 @@ class ComponentQuerySet(models.QuerySet):
         model_type: str = "ProductStream",
         include: bool = True,
         include_inactive_streams: bool = False,
+        include_all_variants: bool = False,
     ) -> "ComponentQuerySet":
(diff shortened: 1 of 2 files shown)


JimFuller-RedHat:
did we fail perf tests when this was deployed - or maybe nothing has deployed after this change as our perf test currently is run only against stage ... thinking of automating smoke, reconciliation and perf tests on a regular basis independent of deployments.

jasinner:
> did we fail perf tests when this was deployed - or maybe nothing has deployed after this change as our perf test currently is run only against stage ... thinking of automating smoke, reconciliation and perf tests on a regular basis independent of deployments.

The 'test_display_product_stream_with_many_roots' performance test against stage started to fail after deploying PR#726, see gitlab pipelines run against main since then.

JimFuller-RedHat (approved):
LGTM

PR #736, opened 2024-01-16
Fix GoList test that fails locally

dmonzonis:
In a local environment, the test test_sca.py::test_golist_scan_files fails, in particular for the podman-4.3.1-814b7b0.tar.gz package. It complains about a dependency which is imported from github.com/onsi/ginkgo/ginkgo, which appears to not be a go package.

Changing the dependency to github.com/onsi/ginkgo fixes the issue and the test passes locally.

For some reason, this test does not fail on the CI.


jasinner (approved):
The original test works for me in dev, however I'm happy to update the test archive as it also works for me.

PR #737, opened 2024-01-16
Add tox as a dev dependency

dmonzonis:
tox is the tool used to run the tests both locally and in the CI pipeline, but it is not added as a dependency in any of the requirement files. Therefore, it has to be installed manually and every time the environment is synced using `pip-sync requirements/dev.txt`, it is removed from the environment.

This commit adds tox in the test.txt requirement file (and therefore in the dev.txt file which depends on it). A couple of packages have also been updated (namely packaging and platformdirs) to avoid conflicts with the dependencies of tox.


requirements/lint.txt
```diff
@@ -75,9 +75,9 @@ pathspec==0.11.1 \
     --hash=sha256:2798de800fa92780e33acca925945e9a19a133b715067cf165b8866c15a31687 \
     --hash=sha256:d8af70af76652554bd134c22b3e8a1cc46ed7d91edcdd721ef1a0c51a84a5293
     # via black
-platformdirs==3.2.0 \
-    --hash=sha256:d5b638ca397f25f979350ff789db335903d7ea010ab28903f57b27e1b16c2b08 \
-    --hash=sha256:ebe11c0d7a805086e99506aa331612429a72ca7cd52a1f0d277dc4adc20cb10e
+platformdirs==4.1.0 \
```
juspence, line 78: This is a major version upgrade. Is there any lower version that still satisifies Tox's requirements, like 3.2.1 or 3.3.0? If not, have you checked the release notes for any potential problems?

juspence (commented):
Maybe there was some reason we excluded Tox to begin with? I don't really remember.

In our CI pipeline, we use a base image that already has Tox installed, and we do something similar in a few other places I believe. I'm worried this change will cause some weird issues with version conflicts in CI, if there's a different Tox version in the base container and in the venv.

For local development, you can install Tox systemwide with `sudo dnf install tox`, which gives me 4.11.1 on Fedora 39. I guess we should add a note about this to our docs - I'd be happy to merge that instead.

You can also tag (at)corgi-devs to request review from anyone on the team.

requirements/lint.txt
dmonzonis: Before tox, platformdirs was only being used by black
```
platformdirs==3.2.0
└── black==22.1.0 [requires: platformdirs>=2]
```

The only major change from 3.2.0 to 4.1.0 is a change in the [site_cache_dir](https://github.com/platformdirs/platformdirs/blob/main/CHANGES.rst#platformdirs-400-2023-11-10), but black only uses [user_cache_dir](https://github.com/psf/black/blob/b7c3a9fedd4cfcc6a6a88aacc7b0f599b63d4716/src/black/cache.py#L12) to store a cache for linting.

And sadly it looks like tox does require the latest version:
```
platformdirs==4.1.0
├── black==22.1.0 [requires: platformdirs>=2]
├── tox==4.12.0 [requires: platformdirs>=4.1]
└── virtualenv==20.25.0 [requires: platformdirs>=3.9.1,<5]
    └── tox==4.12.0 [requires: virtualenv>=20.25]
```

dmonzonis:
Wouldn't it be a good idea to make sure that everyone running the environment has the same version of tox? This would ensure reproducibility. If I were to install tox in an OS other than Fedora 39, the version might be different and things may not work as expected. The same logic applies to all packages that are run in the environment.

As for the CI pipeline using an image with tox pre-installed, I don't really see the point of doing it that way. Why don't we use the same environment as the one used for developing, specifying the version in the requirements, to make it consistent and reproducible? In fact, it's possible that this could be the cause of what happens in #736, where a test fails locally but not in the CI.

juspence:
@dmonzonis We can do it that way but it would be a bigger change. Please update the .gitlab-ci.yml file and the other stuff which I can't discuss on a public site - I've sent you the details.

dmonzonis:
@juspence I've modified the dependency to be inside dev.txt instead of test.txt. The issue of making it consistent with the tox version run in the CI will be tackled in a different PR in the future.

juspence (approved):
LGTM!

PR #738, opened 2024-01-17
Fix links in user guide

dmonzonis:
The links to the project's OpenAPI specification were broken, as they were pointing to the corgi repo (now component-registry) at the master branch (now main). This commit fixes the link to the correct one.

The anchor to the REST API product definitions was also not working, and now it correctly links to the section.

Since the same information about checking out the OpenAPI specification to learn about endpoint usage was said twice and very close together, one of the sentences has been removed to avoid redundancy.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -6,15 +6,12 @@ Component Registry exposes a REST API that any number of clients can connect to,
 front-end application to serve as a web client.
 
 For more in-depth information about the resources served by the API, see section
-(REST API Resource Definitions)[#rest-api-resource-definitions].
+[REST API Resource Definitions](#rest-api-resource-definitions).
 
-The [OpenAPI specification](https:///github.com/RedHatProductSecurity/corgi/-/blob/master/openapi.yml) provides
+The [OpenAPI specification](https:///github.com/RedHatProductSecurity/component-registry/blob/main/openapi.yml) provides
 developer level documentation for endpoint usage.
 
 ### Fetching data
-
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat (approved):
LGTM

docs/user_guide.md
```diff
@@ -6,16 +6,13 @@ Component Registry exposes a REST API that any number of clients can connect to,
...
+[REST API Resource Definitions](#rest-api-resource-definitions).
 
-The [OpenAPI specification](https:///github.com/RedHatProductSecurity/corgi/-/blob/master/openapi.yml) provides
+The [OpenAPI specification](https:///github.com/RedHatProductSecurity/component-registry/blob/main/openapi.yml) provides
 developer level documentation for endpoint usage.
 
 ### Fetching data
 
-[REST API docs](https:///github.com/RedHatProductSecurity/corgi/-/blob/main/openapi.yml) provide detailed
-usage on all endpoints.
```
juspence, line 17: It's fine to remove this but please remove the section header as well.

juspence (approved):
LGTM, thanks for fixing this!

docs/user_guide.md
dmonzonis: I feel like the section header still makes sense, as parent of the more specific subsections "Retrieving components" and "Retrieving component detail", the same way the following section "Searching for components" also has a couple of specific subsections.

juspence: Oops, you're right. I thought both this section header and the one below it were at the same level, so we would have a header above an empty section. But actually "Retrieving Components" is one level below "Fetching data", so it's fine to keep.

PR #739, opened 2024-01-18
CORGI-833 et product version names

jasinner:
Match streams to Errata Tool products via CPEs and their variants. Calculate the 'external_name' property on demand for each ProductStream, and use it in all references to the stream in SBOM files. 

In some cases more than one ProductStream can share an external name. This occurs when 2 ProductStreams have the same brew_tags defined. However whenever the 'external_name' is shared, but the brew_tags are different, I've hardcoded the external_name to be an uppercase version of the existing stream name from product-definitions. This prevents 2 streams from sharing an SBOM where they shouldn't, yet allows 2 ProductStreams with this same contents to share an SBOM.

There is a full listing of ProductStream and external_name attached to CORGI-833 ticket.


jasinner:

corgi/api/serializers.py
```diff
@@ -565,7 +565,7 @@ def get_build_count(instance: ProductModel) -> int:
 
     @staticmethod
     def get_manifest(instance: ProductStream) -> str:
-        manifest_path = f"{instance.name}-{instance.pk}.json"
+        manifest_path = f"{instance.external_name}.json"
```
juspence, line 568: Please keep the `instance.pk` here. According to my memory of the spec, the filename or something should be a UUID which is updated whenever "a new version is released" or something changes, and also should / must match the documentNamespace key in the manifest itself. I don't remember the exact details which are a little fiddly.

We don't really implement this today, but we can at least match (or get close to) the right format by including the UUID. And if we ever recreate data in prod, or migrate data from stage to prod, the PK values / UUIDs will be different as an indicator that "significant changes have occurred".

corgi/core/files.py
```diff
@@ -71,10 +70,12 @@ def render_content(self) -> str:
         released_components = components.manifest_components(ofuri=self.obj.ofuri)  # type: ignore[attr-defined] # noqa: E501
         distinct_provides = self.obj.provides_queryset  # type: ignore[attr-defined]
         distinct_upstreams = self.obj.upstreams_queryset  # type: ignore[attr-defined]
-        cpes = self._get_stream_cpes(self.obj)
+        cpes = self.obj.cpes  # type: ignore[attr-defined]
+        external_name = self.obj.external_name  # type: ignore[attr-defined]
```
juspence, line 94: If these properties are defined on the ProductModel class and not on ProductStream, you don't need the type: ignore here. If the properties are invoked on a Variant, they can just return the name or CPE of the Variant directly since those come from Errata Tool and are already public.

Otherwise, the properties can just return what they do for streams today - find the child Variants and return their CPEs, or calculate the external name of the stream like you've implemented here.

corgi/web/templates/product_manifest.json
```diff
@@ -1,4 +1,11 @@
 {# ProductModel manifest, obj is a ProductModel subclass here #}{% extends "base_manifest.json" %}{% load base_extras %}
+{% block document %}
+  "documentDescribes": [
+    "SPDXRef-{{external_name}}"
```
juspence: In the component-based manifest, we're still using obj.uuid as the SPDXRef for the whole document. Let's also do that here in the product-based manifests to be consistent. I also think that SPDXRefs either require UUIDs, or are very limited in the characters they allow, so using the UUID is simpler. But my memory of the spec is very fuzzy.

corgi/tasks/manifest.py
```diff
@@ -16,31 +19,69 @@
...
 def update_manifests():
+    external_names = set()
     for ps in ProductStream.objects.annotate(num_components=Count("components")).filter(
         num_components__gt=0
     ):
-        cpu_update_ps_manifest.delay(ps.name)
+        # Don't regenerate a manifests for streams with matching external names, the content will
+        # be the same. This happens for the following streams, which share the same brew_tags, and
+        # variants
+        # CERTSYS-10.4-RHEL-8: ['certificate_system_10.4', 'certificate_system_10.4.z']
```
juspence, line 35: This makes sense, but what will happen to the existing manifests we've already published? I guess right now we have two mostly-identical manifests for both streams. The only differences should be the filename on the customer portal, and the stream name we report inside the manifest document.

After we release this change to production, do we need to manually clean up all the manifests we've already published?

Is it possible, even though the streams have the same brew_tags and variants, that they have different e.g. yum_repos, errata_info, or some other relation type? If we do have manifests for both streams on the portal, have you compared the two and confirmed there really aren't any differences in which components are part of each stream?

corgi/tasks/manifest.py
```diff
@@ -16,31 +19,69 @@
...
+        # RHEL-7-FAST-DATAPATH: ['fdp-el7', 'fdp-el7-ovs']
+        # GITOPS-1.2-RHEL-8: ['gitops-1.2', 'gitops-1.2.z']
+        # JAEGER-1.20-RHEL-8: ['jaeger-1.20.0', 'jaeger-1.20.3', 'jaeger-1.20.4']
+        # OPENJDK TEXT-ONLY: ['openjdk-11', 'openjdk-17', 'openjdk-1.8']
+        # OPENSHIFT-PIPELINES-1.7-RHEL-8: ['pipelines-1.7', 'pipelines-1.7.1']
+        # OPENSHIFT-PIPELINES-1.8-RHEL-8: ['pipelines-1.8', 'pipelines-1.8.1']
+        # RHEL-8-RHACM-2.7: ['rhacm-2.7', 'rhacm-2.7.z']
+        # RHEL-8-RHACM-2.8: ['rhacm-2.8', 'rhacm-2.8.z']
+        #
+        # Since we use the external name as the filename, all streams will share the same manifest
```
juspence, line 46: I think we need to keep the UUID as part of the filename in order to comply with the SPDX spec, but maybe I'm misremembering and that was only a SHOULD suggestion, not a MUST requirement.

corgi/web/templates/component_manifest.json
```diff
@@ -1,4 +1,11 @@
 {# Component manifest, obj is a Component here #}{% extends "base_manifest.json" %}{% load base_extras %}
+{% block document %}
+  "documentDescribes": [
+    "SPDXRef-{{obj.uuid}}"
+  ],
+  "documentNamespace": "https://access.redhat.com/security/data/sbom/beta/spdx/{{obj.name|escapejs}}-{{obj.uuid}}",
+  "name": "{{obj.name|escapejs}}",
```
juspence, line 7: If this manifest is for e.g. a UBI8 container, do we end up using the public name from Pyxis here, or is this still the internal name from the Brew NVR?

corgi/web/templates/component_manifest.json
```diff
@@ -79,4 +86,8 @@
       "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",
       "relationshipType": "GENERATES",
       "spdxElementId": "SPDXRef-{{node_id}}"
-    },{% endfor %}{% endif %}{% endblock relationships %}
\ No newline at end of file
+    },{% endfor %}{% endif %}    {
+      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# Document describes stream or component being manifested #}
```
juspence:
```suggestion
      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",{# Document describes component being manifested #}
```

corgi/web/templates/component_manifest.json
```diff
@@ -79,4 +86,8 @@
       "relatedSpdxElement": "SPDXRef-{{obj.uuid}}",
       "relationshipType": "GENERATES",
       "spdxElementId": "SPDXRef-{{node_id}}"
-    },{% endfor %}{% endif %}{% endblock relationships %}
\ No newline at end of file
+    },{% endfor %}{% endif %}    {
```
juspence: Please take care not to add or remove whitespace anywhere in the template - those changes will directly end up in the final manifest. You may want to generate a few stream and component manifests using the new code here, then compare them with manifests we've already generated in stage, and confirm there are no whitespace changes.
```suggestion
    },{% endfor %}{% endif %}
    {
```

corgi/web/templates/product_manifest.json
```diff
@@ -1,4 +1,11 @@
...
+  ],
+  "documentNamespace": "https://access.redhat.com/security/data/sbom/beta/spdx/{{external_name|escapejs}}",
```
juspence, line 6: The last part of the documentNamespace (the filename part) either SHOULD or MUST match either the actual filename, or the documentDescribes SPDXRef, or both, I believe. I think it's safer to just replace obj.name with external_name and not make any other changes.
```suggestion
  "documentNamespace": "https://access.redhat.com/security/data/sbom/beta/spdx/{{external_name|escapejs}}-{{obj.uuid}}",
```

If you really want to make this change as part of SBOM GA, please just double-check the spec quickly and maybe also run the online validator against a few sample manifests you generate:
https://tools.spdx.org/app/validate/

corgi/web/templates/product_manifest.json
```diff
@@ -102,7 +109,12 @@
...
     {
-      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}", 
+      "relatedSpdxElement": "SPDXRef-{{external_name}}",
       "relationshipType": "PACKAGE_OF",
       "spdxElementId": "SPDXRef-{{component.uuid}}"
-    },{% endfor %}{% endblock relationships %}
\ No newline at end of file
+    },{% endfor %}
+    {
+      "relatedSpdxElement": "SPDXRef-{{external_name}}",{# Document describes stream or component being manifested #}
```
juspence:
```suggestion
      "relatedSpdxElement": "SPDXRef-{{external_name}}",{# Document describes stream being manifested #}
```

corgi/core/models.py
```diff
@@ -427,7 +428,11 @@ def builds(self) -> QuerySet:
     def cpes(self) -> tuple[str, ...]:
         """Return CPEs for direct descendant variants if they exist. Otherwise, return a union
         of indirect descendant variants and descendant streams cpes_matching_patterns"""
-        if self is ProductVariant:
+        if isinstance(self, ProductStream):
+            hardcoded_cpes: set[str] = cpe_lookup(self.name)
+            if hardcoded_cpes:
+                return tuple(hardcoded_cpes)
```
juspence: If the cpe_lookup method has a return type annotation like set[str], you shouldn't need an explicit type annotation on this variable here I hope.

What happens if we're looking at a ProductStream, but there are no hardcoded cpes? I guess we skip the elif and fall through to the code below?

Maybe you can add a comment below the new line 436 describing the else case: either we're a product / version and not a stream / variant, or we were a stream but had no hardcoded CPEs. Github won't let me suggest / comment there since it's outside the diff window.

corgi/tasks/manifest.py
```diff
@@ -16,31 +19,69 @@
...
+        if ps.name not in external_names:
+            cpu_update_ps_manifest.delay(ps.name, ps.external_name)
+            external_names.add(ps.external_name)
+
+
+def extract_lines(files) -> Generator[tuple[int, str], None, None]:
+    for line in enumerate(files):
+        if '"created":' not in line[1]:
```
juspence: I like the idea of doing this, but I also feel like it would be better in a separate PR. I'm happy to review multiple PRs, but keeping each one small makes reviewing easier and also reduces risk. We can pick and choose which to deploy and which to revert, if needed.

The code you have currently is going to iterate over every single line in the file, and then iterate again over every single character in that line, which will be very, very slow especially for large manifests. It may also behave incorrectly if we ever have another "created:" key anywhere else in the manifest (unlikely but possible).

I think the more performant way might be to do something like below.
1. At the start of the task, record datetime.now() as the new_start_time and use this as the "created" timestamp
2. Look up the last successful run for the cpu_update_ps_manifest task with this stream name, and grab its TaskResult (a tuple of old_start_time and old_hash)
3. Just generate the manifest like we did in the old code
4. Use `new_manifest = ps.manifest.replace(f"created: {new_start_time}", f"created: {old_start_time}", 1)` to replace the new manifest's created date with the old one
5. Take the new_hash of the new_manifest
6. Compare it to the old_hash of the old_manifest you got in step 2
7. If they are different, publish ps.manifest and return a tuple of (new_start_time, new_hash)
8. If they are the same, don't publish ps.manifest. Just return a tuple of (old_start_time, old_hash)

For step 2, if no successful task run is found, just assume you need to regenerate the manifest.

Step 4 avoids needing to iterate over the whole file - you find the value you care about and replace only that. After you replace 1 value (the third argument - how many items to replace), the replace() function can stop looking / ignore the rest of the data. You can do the same with re.sub() I believe. This should be a lot faster since (I think) the created key is near the top of the file.

juspence (commented):
I'm not finished looking but I need to go finish the service accounts stuff before my day ends. Here's some feedback in the meantime, we can discuss if you have questions.

corgi/api/serializers.py
jasinner: The problem with using `instance.pk` is that now multiple streams share an SBOM file where the `external_name` shares a value between 2 streams. Perhaps if we want to make the file name distinct between updates we should append the created date to the external name.

corgi/core/files.py
jasinner: That's not true. The cpes property is defined on ProductModel but removing type:ignore from there results in a Mypy error:

`corgi/core/files.py:73: error: "TimeStampedModel" has no attribute "cpes"  [attr-defined]`

corgi/web/templates/product_manifest.json
jasinner: As mentioned above, using the `obj.uuid` is really an option any more as multiple streams can now share an SBOM.

corgi/web/templates/component_manifest.json
jasinner: `obj.name` is the public name from Pyxis for container images.

corgi/web/templates/component_manifest.json
jasinner: updated.

corgi/web/templates/product_manifest.json
jasinner: updated.

corgi/tasks/manifest.py
jasinner: For certificate_system 10.4, and 10.4.z the SBOM contents match as expected in stage. For some reason in prod they are different, however the 2 root components are the same, so perhaps some ordering is different or similar.

```
>>> import corgi.core.models as m
>>> cert104 = m.ProductStream.objects.get(name="certificate_system_10.4")
>>> cert104.components.root_components().released_components().values("purl")
<ComponentQuerySet [{'purl': 'pkg:rpmmod/redhat/redhat-pki@10:8060020220420152504:07fb4edf'}, {'purl': 'pkg:rpmmod/redhat/redhat-pki@10:8060020220824122647:07fb4edf'}]>
>>> cert104z = m.ProductStream.objects.get(name="certificate_system_10.4.z")
>>> cert104z.components.root_components().released_components().values("purl")
<ComponentQuerySet [{'purl': 'pkg:rpmmod/redhat/redhat-pki@10:8060020220420152504:07fb4edf'}, {'purl': 'pkg:rpmmod/redhat/redhat-pki@10:8060020220824122647:07fb4edf'}]>
```

I guess it is possible for 2 streams to have matching external_names but have different yum_repos, or composes. However in practice that does not occur in the data we have in product_definitions today.

Once we release this change to production we will need to remove all the current files, regenerate and publish new ones.

corgi/web/templates/product_manifest.json
jasinner: I did take a look at the spec but didnt' find anything indicating we need a uuid in the filename. Good idea, I will check the validator.

corgi/core/files.py
juspence: Yep, that's because we said self.obj was a TimeStampedModel and not a ProductModel for some reason I don't remember without looking a lot more at the code. Maybe we wanted to support self.obj being a Component as well?

Let's just leave it for now.

corgi/web/templates/product_manifest.json
juspence: We need to use only self.external_name for the filename and documentNamespace. We need to stop including the UUID to make sure we have one unique file for each external name, even when two different streams share the same external name.

But here inside the file, we don't need to make changes right? Let's try to minimize the amount of stuff we're changing. Or at the very least please look over the SPDX spec's requirements for filenames, documentNamespace, and SPDXRef formats / allowed characters before making these changes.

EDIT: I double-checked the spec and the documentNamespace / filename changes are fine. For the SPDXRef changes, https://spdx.github.io/spdx-spec/v2.3/package-information/#721-description says:

> Format: "SPDXRef-"[idstring], where [idstring] is a unique string containing letters, numbers, ., and/or -.

I don't think we can guarantee the format of the external names, which may contain _ underscores, : colons, or / slashes. So I'd prefer to keep using the UUID for every SPDXRef. That also means we're consistent between component-based and product-based manifests.

corgi/tasks/manifest.py
juspence: Can you paste a diff here? It makes me worry we either have an undiscovered bug in our code, or some kind of data quality issue. Obviously the diff in createdTimestamp is expected, but the ordering should be consistent and fixed regardless of which stream we're manifesting.

corgi/web/templates/product_manifest.json
juspence: https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#65-spdx-document-namespace-field says:

> The URI shall be unique for the SPDX document including the specific version of the SPDX document. If the SPDX document is updated, thereby creating a new version, a new URI for the updated document shall be used. There may only be one URI for an SPDX document and only one SPDX document for a given URI.

We comply with this by saying "we create new versions only when data is migrated from stage to prod". Only major updates create "new versions" of our documents with new UUIDs. This isn't really what the spec intends but is the most reasonable solution for us today.

If you do open that separate PR for only updating the SBOMs when the content has changed, we could start adding a random UUIDv4 to the filename and documentNamespace instead. SDEngine would then need to unpublish the old manifest before publishing the new one.

https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#652-intent says:

> Although it is not required, the URI can be constructed in a way which provides information on how the SPDX document can be found. For example, the URI can be a URL referencing the SPDX document itself, if it is available on the internet.

> A best practice for creating the URI for SPDX documents available on the public internet is `https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]`

Having the UUID in the documentNamespace, and making that match the filename, is only a "best practice". It's not even a SHOULD suggestion, nevermind a MUST requirement, so I think it's fine to remove the UUID from both places.

However, we need to be consistent with the component-based manifests. We can remove UUIDs from the filename and documentNamespace for both types of manifests.

Or we can generate a manifest for every ProductStream, if its external name has not already been processed (not part of the set). But we can still keep the UUID (for the first external name / stream we analyzed) as part of the documentNamespace and the filename, even if we don't consider UUIDs when figuring out "which streams / external names already generated a manifest?"

corgi/api/serializers.py
juspence: See comment below, I think it's fine after rereading the spec.

corgi/tasks/manifest.py
juspence: Only recommended and not required, see other comments.

corgi/web/templates/product_manifest.json
jasinner: The published [ubi component manifest](https://access.redhat.com/security/data/sbom/beta/spdx/) don't include the UUID in the file name. SDengine then builds the filename from the repository_url and version). Therefore we should do the same for the DocumentNamespace for component manifests, perhaps even expose an external_name attribute of the component on the API.

corgi/tasks/manifest.py
jasinner: I updated the comparison code to read the created_at date and document_uuid from the existing file (using a streaming parser to reduce memory usage), I then create a new document with those old values, compare the hash of this new value with the old document contents, and only render a new manifest if the hashes don't match.

corgi/web/templates/product_manifest.json
jasinner: I changed this to generate a `document uuid` for each generation of the manifest file and using it in the documentDescribes and package references. Because of the same_content check in the cpu_update_ps_manifest task this will now be updated whenever the stream manifest content is updated.

corgi/web/templates/product_manifest.json
juspence: Yep, if we don't have this in the UBI / other component manifests, then removing it from product manifests as well sounds good to me. As long as we're consistent :)

corgi/tasks/manifest.py
juspence: The concern isn't really memory but time, since we have to read the whole file which could be very large / slow. Using the TaskResult model's ".result" attribute should be much simpler. There's also a get_last_success_for_task() method you can reuse to find the previous TaskResult where we generated a manifest for this stream.

I guess what you have is fine for now, but please consider above if we start seeing even more timeouts in these manifest-generating tasks.

corgi/core/files.py
```diff
@@ -65,16 +78,23 @@ class ProductManifestFile(ManifestFile):
...
+    def render_content(self, created_at: str = "", document_uuid: str = "") -> str:
         components = self.obj.components  # type: ignore[attr-defined]
         # As this is a ProductManifestFile so we can assume self.obj has an ofuri
         released_components = components.manifest_components(ofuri=self.obj.ofuri)  # type: ignore[attr-defined] # noqa: E501
         distinct_provides = self.obj.provides_queryset  # type: ignore[attr-defined]
         distinct_upstreams = self.obj.upstreams_queryset  # type: ignore[attr-defined]
-        cpes = self._get_stream_cpes(self.obj)
-
+        cpes = self.obj.cpes  # type: ignore[attr-defined]
+        external_name = self.obj.external_name  # type: ignore[attr-defined]
```
juspence, line 94: When self.obj is a Component model instance, will there be an external_name attribute defined?

corgi/core/files.py
```diff
@@ -65,16 +78,23 @@ class ProductManifestFile(ManifestFile):
...
+        if not created_at:
+            created_at = self.get_created_at()
+        if not document_uuid:
+            document_uuid = f"SPDXRef-{uuid.uuid4()}"
         kwargs_for_template = {
             "obj": self.obj,
+            "external_name": external_name,
+            "document_uuid": document_uuid,
```
juspence, line 102: You're checking the document's hash to see if the new version has changed vs. the old version. You'll have to replace all the occurrences of this document_uuid for that check to work properly. Noting here in case I forget later and that's not handled below, but I still think this improvement would be better in its own PR.

corgi/core/fixups.py
```diff
@@ -94,3 +94,22 @@
 def cpe_lookup(product_stream_name: str) -> set[str]:
     """Manual CPE overrides for streams which cannot be matched with variants"""
     return set(supported_stream_cpes.get(product_stream_name, []))
+
+
+external_names = {
+    # Can't parse the stream version properly
```
juspence: Can you clarify this comment explaining why we can't parse the names? e.g. do we expect streams to have no more than one - dash?

corgi/core/fixups.py
```diff
@@ -94,3 +94,22 @@
...
+
+
+def external_name_lookup(product_stream_name: str) -> str:
+    stream_prefixes = ("dts-", "rhel-br-", "mtr-", "quay-")
+    # 'dts' streams share variants with 'rhscl', however the component sets are different
+    # 'rhel-br' streams share variants with 'rhel' however the components sets are different
+    # 'mtr-' streams share the 'RHEL-8-MTR-1' variant, but have distinct brew_tags
+    # 'quay-' streams share the 'QUAY-3-RHEL-8' variant, but have distinct brew_tags
+    for stream_prefix in stream_prefixes:
+        if product_stream_name.startswith(stream_prefix):
```
juspence: This assumes the product_stream_name is always lowercase. I hope that's always the case in prod-defs, but maybe for safety we could do below:
```suggestion
        if product_stream_name.lower().startswith(stream_prefix):
```

corgi/core/models.py
```diff
@@ -740,6 +745,53 @@ def upstreams_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
...
+        # search all variants here, not just this stream's productvariants
+        et_versions_matching_cpes = (
+            ProductVariant.objects.filter(cpe__in=self.cpes)
+            .exclude(et_product_version="")
+            .values_list("et_product_version", flat=True)
+            .distinct()
+        )
+        variants_matching_cpe_length = len(et_versions_matching_cpes)
+        # There is a single matching et_product_version for this streams cpes
+        if variants_matching_cpe_length == 1:
```
juspence: Do we need this variable later? If not I think it's fine to just do `if len(...) == 1`. Or this is fine too if you like it better.

corgi/core/models.py
```diff
@@ -740,6 +745,53 @@ def upstreams_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
...
+            return et_versions_matching_cpes[0].upper()
+        else:
```
juspence: Since you use `return` above, you don't need to write `else` here and can avoid indenting the code again. You can just add a comment like `# else we have more than one matching et_product_version` if you want to be clear about the logic. Or what you have is fine, just FYI.

corgi/core/models.py
```diff
@@ -740,6 +745,53 @@ def upstreams_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
...
+            et_versions_with_match = (
+                ProductVariant.objects.filter(
+                    cpe__in=self.cpes, et_product_version__contains=self.version.removesuffix(".z")
+                )
+                .values_list("et_product_version", flat=True)
+                .distinct()
```
juspence: Do we need to exclude empty "" string et_product_versions here as well?

corgi/core/models.py
```diff
@@ -740,6 +745,53 @@ def upstreams_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
...
+            et_products_matching_cpes = (
+                ProductVariant.objects.filter(cpe__in=self.cpes)
+                .exclude(et_product="")
+                .values_list("et_product", flat=True)
+                .distinct()
+            )
+            # There is a single matching et_product with this stream's cpes
+            if len(et_products_matching_cpes) == 1:
+                return f"{et_products_matching_cpes[0]}-{self.version}".upper()
+            elif len(et_products_matching_cpes) > 1:
```
juspence: Else we had no matches, I think...what happens then? We just default to returning the internal self.name from the prod-defs stream data?

I think that behavior is fine, but maybe we can add a comment above the final return statement explaining all the else-cases that might lead there.

corgi/tasks/manifest.py
```diff
@@ -1,48 +1,108 @@
...
-        logger.info(f"Generating manifest for {product_stream}")
-        with open(f"{settings.STATIC_ROOT}/{product_stream}-{ps.pk}.json", "w") as fh:
-            fh.write(ps.manifest)
+        if not same_contents(output_file, ps):
+            logger.info(f"Generating manifest for {product_stream}")
+            with open(output_file, "w") as fh:
+                fh.write(ps.manifest)
+                return True
+        else:
+            logger.info(f"Not updating {output_file} with same contents")
```
juspence: FYI it's a little more readable to just log a message and return False at the top if the two files have the same_contents(). After that, we can just write out the file and return True without an else block. That avoids heavily nesting / indenting code where it isn't needed.

corgi/tasks/manifest.py
```diff
@@ -1,48 +1,108 @@
...
+        # GITOPS-1.2-RHEL-8: ['gitops-1.2', 'gitops-1.2.z']
+        # JAEGER-1.20-RHEL-8: ['jaeger-1.20.0', 'jaeger-1.20.3', 'jaeger-1.20.4']
+        # OPENJDK TEXT-ONLY: ['openjdk-11', 'openjdk-17', 'openjdk-1.8']
+        # OPENSHIFT-PIPELINES-1.7-RHEL-8: ['pipelines-1.7', 'pipelines-1.7.1']
+        # OPENSHIFT-PIPELINES-1.8-RHEL-8: ['pipelines-1.8', 'pipelines-1.8.1']
+        # RHEL-8-RHACM-2.7: ['rhacm-2.7', 'rhacm-2.7.z']
+        # RHEL-8-RHACM-2.8: ['rhacm-2.8', 'rhacm-2.8.z']
+        #
+        # Since we use the external name as the filename, all streams will share the same manifest
+        if ps.name not in external_names:
```
juspence: Should this be `if ps.external_name not in external_names`?

corgi/core/models.py
```diff
@@ -740,6 +745,53 @@ def upstreams_queryset(self, using: str = "read_only") -> QuerySet["Component"]:
...
+                for et_product in et_products_matching_cpes:
+                    # There is a matching et_product where the et_product name is found in the
+                    # stream name. For example rhn_satellite_6.8 has multiple products, "SAT-TOOLS",
+                    # and "SATELLITE" use the stream's name to favour "SATELLITE" found in that case
+                    if self.name.upper().find(et_product) > 0:
+                        return f"{et_product}-{self.version}".upper()
+        return self.name.upper()
```
juspence: Can you add a comment explaining why we return the .upper() case names? Is this to be consistent with external_names_lookup(), which also returns uppercase names, so we don't generate a manifest for the same stream twice? I think we're using uppercase names to mark "some condition occurred", but I don't remember what that condition was and I can't find the comment describing it.

If we're using the external name as the filename, won't this give us UPPERCASE vs lowercase files on the Customer Portal? Could that be an issue if we sometimes have differently-cased names for the same stream?

corgi/web/templates/product_manifest.json
```diff
@@ -102,7 +109,12 @@
...
     {
-      "relatedSpdxElement": "SPDXRef-{{obj.uuid}}", 
+      "relatedSpdxElement": "{{document_uuid}}",
       "relationshipType": "PACKAGE_OF",
       "spdxElementId": "SPDXRef-{{component.uuid}}"
-    },{% endfor %}{% endblock relationships %}
\ No newline at end of file
+    },{% endfor %}
+    {
+      "relatedSpdxElement": "{{document_uuid}}",{# Document describes stream or component being manifested #}
```
juspence:
```suggestion
      "relatedSpdxElement": "{{document_uuid}}",{# Document describes stream being manifested #}
```

corgi/web/templates/component_manifest.json
juspence: I hope I'm looking at the latest diff but it looks like it still says "stream or component" here.

corgi/core/files.py
```diff
@@ -28,8 +29,20 @@ def file_name(self) -> str:
...
-        kwargs_for_template = {"obj": self.obj}
+    @staticmethod
+    def get_created_at():
+        dt = datetime.now()
+        return dt.strftime("%Y-%m-%dT%H:%M:00Z")
+
+    def render_content(self, created_at: str = "", document_uuid: str = "") -> str:
+        if not created_at:
+            created_at = self.get_created_at()
+        document_namespace = f"{self.obj.name.replace('/', '_')}-{self.obj.version}"  # type: ignore[attr-defined]
```
juspence: I've lost track but do we need to remove / slashes in the documentNamespace? The requirement for "only letters, numbers, . dots, and - dashes" just applies to the SPDXRef format I thought.

Do we also need to use the external_name here instead? How will that work for Component model instances?

corgi/web/templates/product_manifest.json
juspence: Sorry, I'm still confused. We only need the documentNamespace to have a UUID in it per the spec. We're going to break the spec and not have a UUID in the documentNamespace. Instead we'll just have the external name as both the filename and documentNamespace. This is fine, we don't really have a better solution.

We can't use the external name for the SPDXRefs, and we don't need them to be random or regenerated each time the manifest content changes. Why not just keep the old UUID?

It's not an issue that multiple streams may have the same external name. For each one of those external names, we'll only generate one manifest for one object (whether stream A or stream B), So we'll end up with only one SPDXRef - either stream A's UUID or stream B's UUID.

juspence (approved):
I've reviewed as much as I can, but didn't look at tests and am having a hard time understanding some of the changes here. It would be very helpful in the future if you could split these into separate PRs whenever possible, for example:

1. One PR to add the external names property, model changes and migrations, logic to calculate the names, etc.
2. One PR for manifest changes to start using the new properties, update the documentNamespace and SPDXRef formats, etc.
3. One PR to only regenerate manifests when the content changes

corgi/tasks/manifest.py
```diff
@@ -1,48 +1,108 @@
...
+            external_names.add(ps.external_name)
+
+
+def same_contents(existing_file: str, stream: ProductStream) -> bool:
+    logger.info(f"Checking if manifest is updated for {stream.name}")
+    existing_file_obj = Path(existing_file)
+    if existing_file_obj.is_file():
+        # use json_stream to reduce memory usage when reading the old file as a dict
+        with existing_file_obj.open("r") as existing_fh:
+            existing_data = json_stream.load(existing_fh)
```
juspence: I don't think this is actually going to reduce memory. You're iterating over one line of the file at one time, instead of reading the whole document at once, but the dict of existing_data still has to contain all the data at the end. My previous comments about iterating over the whole file were related to speed, not memory use.

corgi/web/templates/component_manifest.json
```diff
@@ -1,4 +1,11 @@
 {# Component manifest, obj is a Component here #}{% extends "base_manifest.json" %}{% load base_extras %}
+{% block document %}
+  "documentDescribes": [
+    "SPDXRef-{{obj.uuid}}"
+  ],
+  "documentNamespace": "https://access.redhat.com/security/data/sbom/beta/spdx/{{namespace|escapejs}}",
```
juspence: Should the last part of this line say `document_namespace` instead of just `namespace`? I may also have missed where we defined this variable - in files.py?

corgi/core/files.py
jasinner: No, but this function is defined on ProductManifestFile, not ManifestFile, and external_name is not used in the product_manifest.json template.

corgi/core/fixups.py
jasinner: The version is not after the last slash for these streams. I'll update the comment.

corgi/web/templates/product_manifest.json
jasinner: I think it's more elegant to use a document-uuid instead of the stream uuid as the document can describe multiple streams, and might switch between them. This also means the document-uuid will be updated each time is updated, which is a good idea since we are really publishing the manifest of the latest builds in that stream.

corgi/tasks/manifest.py
jasinner: The certificate-system-10.4 stream doesn't have any latest_components in prod. It's strange because both have the same set of root components. However I don't think it's worth spending a great deal of time investigating it as the streams SBOM's are correct in stage. We should check the 2 streams again after we do our next prod deploy.

```
>>> stream = m.ProductStream.objects.get(name="certificate_system_10.4.z")
>>> stream.components.root_components().released_components().values_list("purl",flat=True)
<ComponentQuerySet ['pkg:rpmmod/redhat/redhat-pki@10:8060020220420152504:07fb4edf', 'pkg:rpmmod/redhat/redhat-pki@10:8060020220824122647:07fb4edf']>
>>> stream.components.root_components().released_components().latest_components(ofuri=stream.ofuri,include_inactive_streams=True).values_list("purl",flat=True)
<ComponentQuerySet ['pkg:rpmmod/redhat/redhat-pki@10:8060020220824122647:07fb4edf']>
```

```
>>> stream = m.ProductStream.objects.get(name="certificate_system_10.4")
>>> stream.components.root_components().released_components().values_list("purl",flat=True)
<ComponentQuerySet ['pkg:rpmmod/redhat/redhat-pki@10:8060020220420152504:07fb4edf', 'pkg:rpmmod/redhat/redhat-pki@10:8060020220824122647:07fb4edf']>
>>> stream.components.root_components().released_components().latest_components(ofuri=stream.ofuri,include_inactive_streams=True).values_list("purl",flat=True)
<ComponentQuerySet []>
```

I think the probably the `include_inactive_streams` setting is not being passed through to the stored procedure in prod. Although that part of the code in production looks OK to me.

corgi/core/models.py
jasinner: Ok, removed the set[str] type annotation and added a comment about the else case.

corgi/tasks/manifest.py
jasinner: The problem is the new document_uuid SPDXRef  now. I guess we could also store that in the task results. OK, I'll see if I can rework this so we don't have the json_stream dependency, or parse the old file looking for these values.

corgi/core/files.py
jasinner: Yes, I do replace the document_uuid.

corgi/core/models.py
jasinner: Done.

corgi/core/files.py
jasinner: This is something which SDEngine does today when it publishes component manifests. I'm just making sure that the documentNamespace value matches the filename as it will be published in SDEngine.

corgi/tasks/manifest.py
jasinner: I'll see if I can get these values from the TaskResults.

corgi/web/templates/component_manifest.json
jasinner: The value `namespace` was correct. But I updated it to `document_namespace` now as I think it's better.

corgi/core/models.py
jasinner: Added a comment. It's just to match the ET naming style.

corgi/core/fixups.py
jasinner: Updated.

corgi/web/templates/product_manifest.json
jasinner: Sorry, neglected to push these changes. Pushed now.

corgi/core/models.py
jasinner: removed, thanks.

corgi/core/models.py
jasinner: No because we're filtering on et_product_version in the main `include` filter in the previous line.

corgi/core/models.py
jasinner: ok.

corgi/tasks/manifest.py
jasinner: yes, thank you!

corgi/tasks/manifest.py
jasinner: ok, updated.

corgi/web/templates/component_manifest.json
jasinner: updated.

corgi/web/templates/component_manifest.json
jasinner: Sorry, neglected to push my changes.

corgi/tasks/manifest.py
jasinner: I filed an new issue for this: CORGI-981

corgi/tasks/manifest.py
jasinner: Updated this to read from last successful task result.

corgi/tasks/manifest.py
jasinner: Done.

PR #740, opened 2024-01-24
Add load perf tests

JimFuller-RedHat:
We need to ensure we have a good baseline defined for all performance across the service. 

This PR introduces standalone usage of [locust](https://github.com/locustio/locust) and a minimal set of tests that we can use against local, stage and prod environments.


jasinner:
Looks interesting. I think we should definitely add some load tests for common griffon queries as you have done. I don't think our problem is really high load is it? Rather just slow queries even with minimal load. Therefore I'm not sure spending the time to integrate something like Locus into CI/CD (which is valuable to measure performance change over time) is worth the effort. Did you consider just adding the Griffon tests to the existing testsuite?

JimFuller-RedHat:
Completely agree with your thoughts and actually said the same exact thing to OSIDB team when they added it there ;) ... then I had a play with locust and based on pro/cons felt it worth adding - setup was minimal and opted to not add to dev requirements on the basis we could keep for a period of time/informational. The following drove thinking:

1) we need to characterise a good perf baseline for future planed work 
2) some issues have arisen that are intermittent in nature and having a more structured load approach will help identify
3) we need to generate visible perf reports that will be consumed/analyzed by others outside time

JimFuller-RedHat:
note we rotated pw after this

PR #741, opened 2024-01-25
remove memory intensive spdx schema validation

jasinner:
This does result in unicode escape sequences such as "\u000A" getting into the SBOM due to useage of 'escapejs' in the Django template, but that doesn't seem to affect JSON, or SPDX schema validation.


JimFuller-RedHat (approved):
LGTM

PR #743, opened 2024-01-26
add locust perf tests

JimFuller-RedHat:
add standalone locust perf load tests (https://github.com/RedHatProductSecurity/component-registry/commit/dc1da10723bde0f62788004c201d1c1ed84f0525 )


JimFuller-RedHat:
merging as discussed previously

PR #745, opened 2024-01-28
Perf

JimFuller-RedHat:
* Add FasterPageNumberPagination(https://github.com/RedHatProductSecurity/component-registry/pull/745/commits/2e96e5e178c9c700533dd0d299d7e00583a2e8bc)  and made it configurable with **OPTIMISE_REST_API_COUNT**
* Add faster include filterset, enabled only on Component for now (https://github.com/RedHatProductSecurity/component-registry/pull/745/commits/345217cc5d6a11eac96c0393179feee07976498f) inspired by https://github.com/RedHatProductSecurity/osidb/pull/423
* refactor latest filter and provide new **get_latest_components** function which return multiples

**Note**- Faster page counting using estimates needs to ensure underlying database is 'primed' - eg. usually with an explicit ANALYZE or SELECT COUNT(*) FROM ... ... in testing we set **OPTIMISE_REST_API_COUNT** to False.

**Note** - both **get_latest_component** and **get_latest_components** will be deprecated in future graphdb refactor

### DIFF
--- a/corgi/core/constants.py
+++ b/corgi/core/constants.py
@@ -84,7 +84,7 @@ SBOMER_PRODUCT_MAP = {
 # So the SQL syntax will be correct in the final stored procedure code
 LATEST_FILTER_DEFINITION = (
     "get_latest_component( "
-    "model_type text, ps_ofuris text[], component_type text, component_ns text, "
+    "model_type text, ps_ofuri text, component_type text, component_ns text, "
     "component_name text, component_arch text, include_inactive_streams boolean) "
     "RETURNS uuid AS $$"
 )
@@ -125,7 +125,7 @@ CREATE OR REPLACE FUNCTION {LATEST_FILTER_DEFINITION}
             INNER JOIN "core_product"
             ON ("core_component_products"."product_id" = "core_product"."uuid")
             {LATEST_FILTER_WHERE}
-            AND (ps_ofuris IS NOT NULL AND core_product.ofuri = ANY(ps_ofuris));
+            AND (ps_ofuri IS NOT NULL AND core_product.ofuri = ps_ofuri);
 
     product_version_component_cursor CURSOR FOR
         SELECT {LATEST_FILTER_FIELDS}
@@ -134,7 +134,7 @@ CREATE OR REPLACE FUNCTION {LATEST_FILTER_DEFINITION}
             INNER JOIN "core_productversion"
             ON ("core_component_productversions"."productversion_id" = "core_productversion"."uuid")
             {LATEST_FILTER_WHERE}
-            AND (ps_ofuris IS NOT NULL AND core_productversion.ofuri = ANY(ps_ofuris));
+            AND (ps_ofuri IS NOT NULL AND core_productversion.ofuri = ps_ofuri);
 
     product_stream_component_cursor CURSOR FOR
         SELECT {LATEST_FILTER_FIELDS}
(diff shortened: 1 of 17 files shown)


corgi/core/constants.py
```diff
@@ -144,7 +144,7 @@
             ON ("core_component_productstreams"."productstream_id" = "core_productstream"."uuid")
             {LATEST_FILTER_WHERE}
             AND (include_inactive_streams OR core_productstream.active)
-            AND (ps_ofuri IS NOT NULL AND core_productstream.ofuri = ps_ofuri);
+            AND (ps_ofuris IS NOT NULL AND core_productstream.ofuri = ANY(ps_ofuris));
```
jasinner: I guess the test_latest_components_by_streams_filter_with_multiple_products test is now failing because it returns the latest package in any of the streams. In the test I assume it would now return the package with release '11' which is only part of the rhel-8 stream. 

Unfortunately I don't think that is acceptable because there might be a problem with releases prior to '11', which does not affect release '11'. Users need to know that rhel-7 ships release '10' as it's latest in that case. I think the only way to do this properly is to loop on all the ps_ofuris and calculate latest packages for each of them.

jasinner:
The rest of the changes, apart from the incorrect version returned in multiple_products test look good. Perhaps we can merge this the latest filter changes in [this commit](https://github.com/RedHatProductSecurity/component-registry/pull/745/commits/f9bcf70b7cf49b4dac767cd274c2b8f8272afc59).

corgi/core/constants.py
JimFuller-RedHat: correct, I am mid stream pushing this loop into db .... if we can wait will be done by tomorrow

jvaught-rh: I think that given it is Thursday (or Friday) already we could wait for this. Since we try not to release right before a weekend?

corgi/core/models.py
```diff
@@ -1041,7 +1041,7 @@ def _latest_components_func(
                     F("name"),
                     F("arch"),
                     Value(include_inactive_streams),
-                    function="get_latest_component",
+                    function="get_latest_components",
```
jasinner, line 1044: Should this still use the old function name without the trailing 's'? I think in that case we should also pass the ofuri argument into the database function, instead of adding it to the ofuris list.

It's probably also a good idea to rename the new database function so it's significantly different to the existing one, maybe something like 'get_latest_components_in_multiple_streams'

corgi/core/constants.py
```diff
@@ -193,3 +193,94 @@
...
+    component_uuid text;
+    component_epoch int;
+    component_version text;
+    component_release text;
+    latest_uuid text;
+    latest_epoch int;
+    latest_version text;
+    latest_release text;
+
+    product_component_cursor CURSOR FOR
```
jasinner, line 212: Did you consider calling the get_latest_component function instead of redefining it here? I think it would help reduce the burden of maintaining this code into the future. Can't say or not if this is actually exactly redefining that function, but it does look similar. If we can't call the database function here perhaps it is a good idea to make use of the Python constants so we don't have update both database function if the logic changes.

corgi/core/constants.py
```diff
@@ -84,7 +84,7 @@
 # So the SQL syntax will be correct in the final stored procedure code
 LATEST_FILTER_DEFINITION = (
     "get_latest_component( "
-    "model_type text, ps_ofuri text, component_type text, component_ns text, "
+    "model_type text, ps_ofuris text[], component_type text, component_ns text, "
```
jasinner: I think we can drop all the changes to constants.py in this commit right? We are now adding a new database function.

jasinner (approved):

corgi/core/models.py
JimFuller-RedHat: we are replacing with **get_latest_components**

corgi/core/constants.py
JimFuller-RedHat: the idea is that **get_latest_components** should 'go away' when we do graphdb refactor ...

corgi/core/constants.py
JimFuller-RedHat: yes of course you are right ! thx

PR #746, opened 2024-01-31
Add RHCOS build collector

dmonzonis:
Before this commit, Brew builds for RHCOS images are not collected since they are different than other image builds in several ways:

- They have less info compared to a usual OCI build, which contains detailed info about each of the arch-specific images.
- A Brew build for an RHCOS image is already arch-specific, instead of having the images for each arch inside the archives.
- The RPM components are found in the _commitmeta.json_ archive.

Therefore, a new collector for RHCOS builds has been implemented, which takes into account these differences.


jasinner:
Looks pretty good. Can you also add a test with some RHCOS sample data from Brew, and creating a new RHCOS component from it? I want to see what the purl will look like for example. 

There are also a couple of refactoring suggestions from Sonarcube.

jasinner:
Looks good to me, thank you.

corgi/collectors/brew.py
```diff
@@ -486,23 +485,23 @@ def get_container_build_data(self, build_id: int, build_info: dict[str, Any]) ->
 
         go_stdlib_version = ""
         remote_sources: dict[str, tuple[str, str]] = {}
+        is_rhcos_build = build_info.get("cg_name") == "coreos-assembler"
```
JimFuller-RedHat, line 488: you might consider defining a class variable (RHCOS="coreos-assembler") at the beginning of brew class and refer to that here - to give hints to future visitors ...

JimFuller-RedHat:
apart from previous suggested and failing CI please add an entry into **CHANGELOG.MD** for this work - LGTM, nice work

corgi/collectors/brew.py
dmonzonis: done

dmonzonis: done

JimFuller-RedHat (approved):
LGTM

jasinner (approved):
Looks good to me. Thanks for adding the the tests.

PR #748, opened 2024-02-04
CORGI-998: Add CI job to regularly run locust perf against stage

JimFuller-RedHat:
Add locust perf ci job that runs on regular schedule. 

**Note**: we did not add to tox intentionally


.gitlab-ci.yml
```diff
@@ -208,6 +212,23 @@ secrets:
     - *common_ci_setup
   script:
     - tox -e secrets
+  only:
```
jasinner, line 215: I think this is not part of your PR, but does this mean that mypy stage only runs during the scheduled job? Should we remove this?

jasinner (approved):
LGTM

.gitlab-ci.yml
jasinner: Whoops, I see now it's the secrets stage. Nevermind.

PR #749, opened 2024-02-05
CORGI-830 license validation

jasinner:
This allows SPDX Validation to pass. Validation is now introduced as part of the *ManifestFile.render_content() function which used a Django Template, but now uses the spdx_tools library. Using the spdx_tools library makes validation easier, and also allows to build up the extracted_licensing_info list while building the packages.

I filed a couple of follow up issues, CORGI-999, and CORGI-1000 to improve the license parsing to use less licenseRefs in a future iteration. The [SPDX Specification says](https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/#102-extracted-text-field) that we should include the entire extracted license text when using licenseRefs, which I was not able to do here because I don't have that information. So instead I linked to the [Fedora Allowed License list](https://docs.fedoraproject.org/en-US/legal/) in that field of the SPDX Document.


jasinner:

JimFuller-RedHat (approved):
if it passes CI, LGTM

PR #756, opened 2024-02-09
CORGI-954 fix container nvr

jasinner:
This updates all existing Red Hat Container images to have the original NVR, and NEVRA values from Brew. It also ensures any new Container Images use the SoftwareBuild.name value instead of the Component.name value in building the NVR, and NEVRA because the Component.name value is often an public one obtained from Pyxis.


JimFuller-RedHat (approved):
nice work, LGTM

PR #757, opened 2024-02-09
CORGI-1015: fix size overflow on nvr & nevra migration

JimFuller-RedHat:
amends 0117_update_container_nvrs.py django migration defined in #756


JimFuller-RedHat:
we want this to process over the weekend on stage, tested locally ... pulling the trigger

PR #758, opened 2024-02-12
CORGI-954 Fix Software build names for container builds, and upate container nvrs to correct values from Brew

jasinner:
This is the 3nd attempt at this issue. The first attempt had a bug where the first loop in the Container migration left out a distinct. That caused some container NVRs to be updated more than once. This PR fixes that issue by using distinct on the first loop in the 0118 migration. The 2nd attempt adjusted the length of the Component, nvr and nevra fields in a migration.  I undid the changes made in those migrations manually, see CORGI-954 for details.

This PR also fixes softwarebuild names to use the correct Brew build names instead of the component name, which could be fetched from the container name_label for container builds. Now they use the original value from Brew in all cases.


JimFuller-RedHat (approved):
third time is a charm, LGTM

PR #762, opened 2024-02-13
CORGI-1011: upgrade local dev env postgres to pg15

JimFuller-RedHat:
@RedHatProductSecurity/corgi-devs please try this out in your local dev env ... once we confirm stability, then we will upgrade stage pg.

**Note** - one may have to explicitly `>podman pull registry.redhat.io/rhel8/postgresql-15:1-44
`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       resources:
         limits:
           memory: 2G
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat:
@jasinner mentioned - if one has a local dev db with data, best to pg_dump it first (and restore after upgrade)... going to merge this.

PR #764, opened 2024-02-14
CORGI-651 add restore licenses script

jasinner:
Restore OpenLCS license data

### DIFF
--- a/scripts/restore_licenses.sh
+++ b/scripts/restore_licenses.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+set -e
+
+export PGPASSWORD=test
+
+if [[ -z "$1" ]]; then
+    licenses_file='./licenses.tar.gz'
+else
+    licenses_file="$1"
+fi
+
+echo "Decompressing ${licenses_file} ..."
+tar xvf "${licenses_file}"
+
+echo "Creating a temporary table to populate with values from licenses_file"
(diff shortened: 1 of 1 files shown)


scripts/restore_licenses.sh
```diff
@@ -0,0 +1,29 @@
...
+tar xvf "${licenses_file}"
+
+echo "Creating a temporary table to populate with values from licenses_file"
+psql -h localhost -p 5433 -U corgi-db-user --dbname corgi-db -c 'CREATE TABLE license_temp(purl varchar(1024), copyright_text text, license_concluded_raw text, openlcs_scan_url text, openlcs_scan_version text)'
+
+echo "Populating the temporary table"
+psql -h localhost -p 5433 -U corgi-db-user --dbname corgi-db -c '\copy license_temp(purl, copyright_text, license_concluded_raw, openlcs_scan_url, openlcs_scan_version) from licenses.txt'
+
+echo "Updating the core_component table with values from temporary table"
+psql -h localhost -p 5433 -U corgi-db-user --dbname corgi-db -c 'UPDATE core_component o SET copyright_text=t.copyright_text, license_concluded_raw=t.license_concluded_raw, openlcs_scan_url=t.openlcs_scan_url, openlcs_scan_version=t.openlcs_scan_version FROM license_temp t WHERE o.purl=t.purl'
```
jasinner, line 28: I've been running this for several hours in dev and it hasn't complete yet. I did seem to block the 0118 migration (which also tries to update core_component table), not sure if it blocks reads of not.

jasinner: There are over 7 million rows in the license_temp table. I don't think it's a good idea to a where clause to the 'copy to' command which dumps the data though, as we want to clear these values from the stage dump after we restore it to prod.

jasinner:

PR #772, opened 2024-02-27
CORGI-1025 return 204 http response for non root component manifest

jasinner:
We don't save component taxonomy for non-root components so attempting to generate a manifest for them results in a populated relations field, but nothing in the packages list. 
One way to fix this would be adjust provides relationship of component to traverse the graph instead of populating a many-to-many relationship. However since we don't need non-root component manifest for any specific business requirement at the moment I opted to just return 204 response content here indicating there is no content for the request.


jasinner:
@component-registry please review

PR #780, opened 2024-03-26
CORGI-1044 Duplicate components in SBOMs causing validation failure

jasinner:
I'm not sure what change cause this issue to surface now. It was happening in stage just before the upgrade to PG15
. However it does seem to be resolved by adding a distinct on manifest_components filter.


jasinner:
When looking at the latest components via the API I don't see the same duplicate components issue, so only adding this change to the manifest_components endpoint.

JimFuller-RedHat (approved):
LGTM

PR #781, opened 2024-04-02
CORGI-1046 only update cnode purl where component purl is updated

jasinner:
This is part of the fix for CORGI-1046, Postgres High Disk IO. The other part is an adjustment to the index to add purl.


JimFuller-RedHat (approved):
LGTM!

PR #786, opened 2024-04-04
CORGI-562 upgrade django

jasinner:


jasinner:
The test failure has me baffled. From some debugging it seems that we do update the namespace to REDHAT for the RPMMOD with name '389-ds', but that is not reflected in the obj.purl. If you call obj.get_purl() it shows the expected value indicating the generated value is correct.

PR #791, opened 2024-04-15
save container leaf nodes in a transaction to delay mptt updates

jasinner:
Similar to [pull/789](https://github.com/RedHatProductSecurity/component-registry/pull/789) except this time recurse the leaf node components twice. Once without a transaction to save the components, and a 2nd time in a transaction with delayed MPTT tree updates.


JimFuller-RedHat (approved):
as far as I can tell LGTM

PR #800, opened 2024-07-16
CORGI-1058 rhbq manifest

jasinner:
Don't fail the slow_fetch_pnc_sbom task on missing ProductVariant.


juspence (approved):
I was a little confused at first, but looking at the code I see we really aren't using the variant name. I guess this code, and the test for it, was just added to sanity-check the messages we're receiving.

Since it caused issues before (due to stale prod-defs data), this change LGTM and will prevent this problem from happening again. Thanks for looking into this!

